PR #74, opened 2021-09-20
add external data and appveyor yaml

IndrajeetPatil:


PavelBal:
@IndrajeetPatil Are all of these external files are actually used somewhere (vignettes or so)? Let's only keep those that are required.

PR #84, opened 2021-09-28
- Reverted usage of grepl instead of isCharInString

PavelBal:
- Check for molecule name being NA in addition to NULL when reading molecular weight in readOSPSTimeValues
- Using runSimulations instead of (missing) runSimulationsConcurrently


PavelBal:
@IndrajeetPatil FYI

Issue #60, opened 2021-09-02
Function visualizer: axis limits shod go beyond maximal function value


PavelBal:
Setting axis limits only works when the assigned value is less than the maximal resp. greater than  the minimal value given by the function (e.g., in the case example the argument is limited to 100, and the limits of x axis cannot be set to >100). It should be possible to extend them.

![image](https://user-images.githubusercontent.com/25903040/131840426-e696b157-1018-4540-964a-fa63740fead8.png)

Issue #59, opened 2021-09-02
Evaluate usage of readxl and writexl instead of openxlsx


PavelBal:
https://readxl.tidyverse.org/
https://github.com/ropensci/writexl

Comparison of performance of openxlsx and readxl.

https://lgreski.github.io/dsdepot/2020/06/13/reading-Excel-files.html#:~:text=Since%20it%20uses%20the%20Rcpp,does%20not%20rely%20on%20Java.&text=The%20readxl%20package%20is%20solely,data%20into%20another%20programming%20environment.

PR #89, opened 2021-09-30
Fixes vor issues #59 and #60

hannaei:
Replaced package openxlsx by readxl and writexl


IndrajeetPatil:
Note to self: include `.tgitconfig` (from TortoiseGit) in `.gitignore`

R/utilities-data.R
```diff
@@ -17,7 +17,7 @@ readOSPSTimeValues <- function(dataConfiguration) {
 
   observedData <- list()
   for (sheet in dataConfiguration$dataSheets) {
-    data <- readxl::read_excel(path = filePath, sheet = sheet)
+    data <- readxl::read_excel(path = filePath, sheet = sheet, .name_repair = ~ vctrs::vec_as_names(..., repair = "unique", quiet = TRUE))
```
PavelBal, line 20: @hannaei `read_excel` throws annoying messages when there are e.g. some excel cells that do not have a proper column name (values "outside" of the required excel structure). I found a workaround and added it to the wiki
https://github.com/esqLABS/esqlabsR/wiki/R-useful-information

Very annoying, but we have to think about it every time we use read_excel (or find a better solution for it).

R/utilities-data.R
```diff
@@ -63,8 +63,8 @@ readOSPSTimeValues <- function(dataConfiguration) {
       timeValues$yDimension <- yDim
       timeValues$yUnit <- yUnit
       timeValues$yErrorUnit <- yErrorUnit
-      timeValues$setMetaData(name = "StudyId", value = group$Study.Id[[1]])
-      timeValues$setMetaData(name = "PatientId", value = group$PatientId[[1]])
+      timeValues$setMetaData(name = "StudyId", value = group$`Study Id`[[1]])
```
PavelBal, line 66: Missed some renaming, so it could not find the entries in the list. @hannaei

PavelBal (approved):

PR #119, opened 2021-11-09
update website to bootstrap 5

IndrajeetPatil:
also try GHA on windows


IndrajeetPatil:
The AppVeyor failure has nothing to do with the PR since there are no changes related to PR anyway. I guess it's failing for the same reason as it was failing for ospsuite-R.

PR #121, opened 2021-11-09
use appveyor version of ospsuite

IndrajeetPatil:


.github/workflows/pkgdown.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: |
           remotes::install_deps(dependencies = TRUE)
           install.packages('https://github.com/Open-Systems-Pharmacology/rClr/releases/download/v0.9.1/rClr_0.9.1.zip')
-          install.packages('https://github.com/Open-Systems-Pharmacology/OSPSuite-R/releases/download/v10.0.72/ospsuite_10.0.72.zip')
+          install.packages('https://ci.appveyor.com/api/buildjobs/xp221iu0dabsdyl5/artifacts/ospsuite_11.0.6.zip')
```
msevestre, line 39: This will work but will alaways install a specific version of ospsuite. Better would be to manage to get the PAT stuff solve. Approving for now

msevestre (approved):

.github/workflows/pkgdown.yaml
IndrajeetPatil: Yeah, let's have this is a stopgap solution.

I will make another PR with remotes option.

PR #181, opened 2022-03-08
Replace NAs in units of readParametersFromXLS to allow the unit "" fo…

PavelBal:
…r the dimension "Dimensionless"


R/utilities-parameters.R
```diff
@@ -50,6 +50,10 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
     }
   }
 
+  # replace `NA` in units with `""` represeting the empty unit for
+  # the dimension `Dimensionless`
+  units <- tidyr::replace_na(data = units, replace = "")
```
PavelBal, line 55: The dimension "Dimensionless" has a unit `""` (empty string). When this unit is stored in excel and read in R, the value of the cell is `NA`. I am replacing here with the empty string so this becomes a valid unit.

msevestre: It can potentially be fraction FYI

PavelBal: Yepp, thanks.

IndrajeetPatil (approved):

PR #183, opened 2022-03-10
182 na individual characteristics

PavelBal:


R/utilities-individual.R
```diff
@@ -97,10 +97,23 @@ readIndividualCharacteristicsFromXLS <- function(XLSpath,
   }
 
   # Create the IndividualCharacteristics object
+  #Empty cells are read as `NA` and must be converted to `NULL` for numerical values
+  weight <- data$`Weight [kg]`[[rowIdx]]
```
PavelBal, line 101: This is so super ugly! But - weight, height, and age are optional parameters, so they can be empty in the excel file.

When reading the excel file, these empty cells are stored as `NA`. This must be converted to `NULL` before calling `createIndividualCharacteristics`.

Any idea how to implement this less ugly?

msevestre: why does it need to be converted to NULL? Maybe NA can also be supported (NA and NULL for me are really the same stuff e.g. not there)

msevestre: specifically because it relies on 
```
createSnapshotParameter <- function(value, unit) {
  if (is.null(value)) {
    return(NULL)
  }
  return(SnapshotParameter$new(value = value, unit = unit))
}
```

that could be extended to return NULL on NA as well

PavelBal:
@IndrajeetPatil Going to merge. Please review when you have time.

PR #194, opened 2022-03-21
Do not use ospsuite:: for enum

PavelBal:


R/utilities-data.R
```diff
@@ -281,4 +281,4 @@ calculateMeanDataSet <- function(dataSets, method = "arithmetic", lloqMode = LLO
 
 #' Possible entries for the `lloqMode` argument of `calculateMeans()`
 #' @export
-LLOQMode <- ospsuite::enum(list("LLOQ/2", "LLOQ", "ZERO", "ignore"))
```
PavelBal, line 284: @IndrajeetPatil For some reason, the build fails when using `ospsuite::enum`.

PR #198, opened 2022-04-01
Read params from xls hash

### LINKED ISSUE #36: readParametersFromXLS should overwrite if a parameter is already present
When reading from multiple sheets, the last sheet with a parameter path that is preset in previous paths should overwrite. Use hash instead of simply appending values to a list and by that avoid having the same path multiple times.

PavelBal:
Fixes #36


R/sensitivityCalculation.R
```diff
@@ -22,10 +22,7 @@
 #'   spreadsheet.
 #'
 #' @examples
-#'
-#' # for reproducibility
-#' set.seed(123)
-#'
+#' \dontrun{
```
PavelBal, line 25: Running those examples take too much time

R/utilities-parameters.R
```diff
@@ -21,40 +21,33 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
...
-  units <- c()
+  pathsValuesHash <- hash::hash()
+  pathsUnitsHash <- hash::hash()
 
   for (sheet in sheets) {
     data <- readExcel(path = paramsXLSpath, sheet = sheet)
 
     if (!all(columnNames %in% names(data))) {
       stop(messages$errorWrongParamsXLSStructure(paramsXLSpath))
     }
```
PavelBal, line 32: Using hash for better performance

IndrajeetPatil: Named vectors will outperform hashes because R's` vector` implementation is in `C`, while `{hash}` package is implemented in `R`.

For example, in this case, hashing is 67 times slower than the vector approach!

``` r
bench::mark(
  "hash" = {
    h <- hash::hash(letters, 1:26)
    h[["a"]]
  },
  "named vector" = {
    v <- 1:26
    names(v) <- letters
    v[["a"]]
  }
)[1:5]
#> # A tibble: 2 x 5
#>   expression        min   median `itr/sec` mem_alloc
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>
#> 1 hash          100.2us  120.6us     7842.    4.21MB
#> 2 named vector    1.4us    1.8us   504813.        0B
```

<sup>Created on 2022-04-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>


Of course, this assumes that creating a named vector is an option. But, if that's not a feasible solution, then maybe hashing is the way to go.

R/sensitivityCalculation.R
IndrajeetPatil: You can also skip the examples in plotting functions for sensitivity calculation. 

We have tests for these, so running examples is not crucial.

PavelBal:
@IndrajeetPatil Why utils version is 1.2? The develop of utils is 1.2.9000 already.

IndrajeetPatil:
No particular reason. We are currently not using anything that doesn't exist in the CRAN version, so there was no need to bump the version.

Now we need to bump it because _ospsuite_ needs the GitHub version of the utils package.

IndrajeetPatil:
The same approach can be used here:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/06a9415b31519a290826b9f498e430d5664d4dac/appveyor.yml#L45

PavelBal:
The build is failing because of this, because the latest ospsuite also uses 1.2.9000.

We need to always use the develop version. Can you make a commit for this, please?

PavelBal:
Build passing. @IndrajeetPatil Please review

IndrajeetPatil (approved):
LGTM.

PR #215, opened 2022-04-28
Add functions for simulating the standard esqlabs workflow

PavelBal:


PavelBal:
@IndrajeetPatil Sensitivity tests are failing, something with the snapshots. Could you please check it out.

IndrajeetPatil:
Probably related to some recent changes in _ospsuite_. Let me take a look.

R/ProjectConfiguration.R
```diff
@@ -0,0 +1,78 @@
...
+  inherit = ospsuite.utils::Printable,
+  cloneable = TRUE,
+  active = list(),
+  private = list(),
+  public = list(
+    #' @description
+    #' Initialize a new instance of the class
+    #' @return A new `ProjectConfiguration` object.
+    initialize = function() {
+    },
```
IndrajeetPatil, line 16: You don't need to provide an empty constructor. R6 class will provide a constructor in such cases.

``` r
library(R6)
myClass <- R6Class("myClass")
myClass$new()
#> <myClass>
#>   Public:
#>     clone: function (deep = FALSE)
```

<sup>Created on 2022-04-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/ProjectConfiguration.R
```diff
@@ -0,0 +1,78 @@
+#' @title ProjectConfiguration
+#' @docType class
+#' @description An object storing configuration used project-wide
+#' @format NULL
+ProjectConfiguration <- R6::R6Class(
+  "ProjectConfiguration",
+  inherit = ospsuite.utils::Printable,
+  cloneable = TRUE,
+  active = list(),
+  private = list(),
```
IndrajeetPatil, line 10: Why are you including empty lists here? I think this is unnecessary.

R/ProjectConfiguration.R
```diff
@@ -0,0 +1,78 @@
...
+    #' @field dataFile Name of the excel file with experimental data
+    dataFile = NULL,
+    #' @field dataImporterConfigurationFile Name of data importer configuration
+    #' file in xml format used to load the data. Must be located in the "dataFolder"
+    dataImporterConfigurationFile = NULL,
+    #' @field compoundPropertiesFile Path to the excel file containing information
+    #' about all compounds in the model. Must be located in the "dataFolder"
+    compoundPropertiesFile = NULL,
+    #' @field outputFolder Path to the folder where the results should be
+    #' saved to; relative to the "Code" folder
```
IndrajeetPatil, line 55: I think this can be moved to the top so that all `"Code"` folder related public fields are together; the same way that all fields related to `"paramsFolder"` are contiguous.

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
+#' @title ScenarioConfiguration
+#' @docType class
+#' @description An object storing configuration of a specific scenario
+#' @format NULL
+ScenarioConfiguration <- R6::R6Class(
+  "ScenarioConfiguration",
+  inherit = ospsuite.utils::Printable,
+  cloneable = TRUE,
+  active = list(
+    #' @field setTestParameters Boolean representing whether parameters defined in \code{TestParameters} are to be applied
```
IndrajeetPatil, line 10:
```suggestion
    #' @field setTestParameters Boolean representing whether parameters defined in `TestParameters` are to be applied
```

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+    #' to the simulation
+    setTestParameters = function(value) {
+      if (missing(value)) {
+        private$.setTestParameters
+      } else {
+        ospsuite.utils::validateIsLogical(value)
```
IndrajeetPatil, line 16: We had decided against namespacing `ospsuite.utils::` because they are so frequently used, and there is unlikely to be a namespace collision with any other package we are relying on.

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+      }
+    },
+    #' @field simulationTime Boolean representing whether the simulation will be brought to a steady-state first
+    simulationTime = function(value) {
+      if (missing(value)) {
+        private$.simulationTime
+      } else {
+        ospsuite.utils::validateIsNumeric(value)
+        if (value < 0) {
+          stop(paste0("simulationTime must be a positive numerical value, but the value is ", value))
```
IndrajeetPatil, line 36: This error is being used multiple times, so I am wondering if it would be better to add a new function for it (e.g. `messages$valueShouldBePositive(parameterName)` or `messages$valueShouldNotBeNegative(parameterName)`).

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+#' Default is "Individual"
+simulationType = function(value) {
+  if (missing(value)) {
+    private$.simulationType
+  } else {
+    if (value %in% c("Individual", "Population")) {
+      private$.simulationType <- value
+    } else {
+      stop("Wrong value for 'simulationType'! Accepted values are 'Individual
+                and 'Population'")
```
IndrajeetPatil, line 85: I think this should be printed to a single line:

``` r
stop("Wrong value for 'simulationType'! Accepted values are 'Individual
                and 'Population'")
#> Error: Wrong value for 'simulationType'! Accepted values are 'Individual
#>                 and 'Population'
```

<sup>Created on 2022-04-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+},
+#' @field simulationRunOptions Object of type \code{SimulationRunOptions} that will be passed
+#' to simulation runs. IF \code{NULL}, default options are used
```
IndrajeetPatil, line 90:
```suggestion
#' @field simulationRunOptions Object of type `SimulationRunOptions` that will be passed
#' to simulation runs. If `NULL`, default options are used
```

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+simulationRunOptions = function(value) {
+  if (missing(value)) {
+    private$.simulationRunOptions
+  } else {
+    ospsuite.utils::validateIsOfType(value, SimulationRunOptions, nullAllowed = T)
```
IndrajeetPatil, line 95:
```suggestion
    ospsuite.utils::validateIsOfType(value, SimulationRunOptions, nullAllowed = TRUE)
```

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+
+  #' @field scenarioName Name of the simulated scenario
+  scenarioName = NULL,
+  #' @field modelFile Name of the simulation to be loaded (must include the
+  #' extension ".pkml"). Must be located in the "modelFolder".
+  modelFile = NULL,
+  #' @field applicationProtocol Name of the application protocol to be applied. Defined
+  #' in the excel file "ApplicationParameters.xlsx"
+  applicationProtocol = NULL,
+  #' @field outputDevice Output target of the plot. If NULL (default), the figure is created in the default "plot"
```
IndrajeetPatil, line 139:
```suggestion
  #' @field outputDevice Output target of the plot. If `NULL` (default), the figure is created in the default "plot"
```

R/ScenarioConfiguration.R
```diff
@@ -0,0 +1,181 @@
...
+  #' output. Other values indicate output into a file. A list of supported outputs is provided in \code{GraphicsDevices}-enum.
```
IndrajeetPatil, line 140:
```suggestion
  #' output. Other values indicate output into a file. A list of supported outputs is provided in `GraphicsDevices`-enum.
```

R/runScenarios.R
```diff
@@ -0,0 +1,171 @@
...
+    file.path(
+      scenarioConfiguration$projectConfiguration$paramsFolder,
+      scenarioConfiguration$projectConfiguration$paramsFile
+    ),
+    scenarioConfiguration$paramSheets
+  )
+
+  individualCharacteristics <- NULL
+  if (!is.null(scenarioConfiguration$individualId)) {
+    individualCharacteristics <- readIndividualCharacteristicsFromXLS(XLSpath = file.path(scenarioConfiguration$projectConfiguration$paramsFolder, scenarioConfiguration$projectConfiguration$individualPhysiologyFile), individualId = scenarioConfiguration$individualId, nullIfNotFound = TRUE)
```
IndrajeetPatil, line 116: This is really difficult to read. Better to get each named argument on its own line in the function call.

R/runScenarios.R
```diff
@@ -0,0 +1,171 @@
...
+    if (is.null(individualCharacteristics)){
+      warning(paste0("No individual characteristics for individual id '",
+                     scenarioConfiguration$individualId, "' found."))
+    }
+
+    # Find individual-specific model parameters
+    if (scenarioConfiguration$individualId %in% readxl::excel_sheets(path = file.path(
+      scenarioConfiguration$projectConfiguration$paramsFolder,
+      scenarioConfiguration$projectConfiguration$individualParamsFile
+    ))) {
```
IndrajeetPatil, line 127: I think this will be more readable if the following is assigned to a variable

```r
readxl::excel_sheets(path = file.path(
      scenarioConfiguration$projectConfiguration$paramsFolder,
      scenarioConfiguration$projectConfiguration$individualParamsFile
    ))
```

and then that variable is used in the `if()` condition.

R/runScenarios.R
```diff
@@ -0,0 +1,171 @@
...
+      indivModelParams <- esqlabsR::readParametersFromXLS(file.path(
```
IndrajeetPatil, line 128: `esqlabsR::` not necessary. The same comment applies to other instance of this below.

R/runScenarios.R
```diff
@@ -0,0 +1,171 @@
...
+    scenarioConfiguration$projectConfiguration$modelFolder,
+    scenarioConfiguration$modelFile
+  ), loadFromCache = FALSE)
+  # Set the outputs
+  clearOutputs(simulation)
+  addOutputs(quantitiesOrPaths = enumValues(OutputPaths), simulation = simulation)
+  # Set simulation time
+  setOutputInterval(simulation = simulation, startTime = 0, endTime = scenarioConfiguration$simulationTime, resolution = scenarioConfiguration$pointsPerMinute)
+
+  esqlabsR::initializeSimulation(simulation = simulation, individualCharacteristics = individualCharacteristics, additionalParams = params, simulateSteadyState = scenarioConfiguration$simulateSteadyState, steadyStateTime = scenarioConfiguration$steadyStateTime)
```
IndrajeetPatil, line 165: Better to have one named argument per line to increase readability.

R/utilities-scenario-configuration.R
```diff
@@ -0,0 +1,115 @@
...
+#' # Set scenario name
+#' scenarioConfiguration$scenarioName <- "MyScenario"
+#' # Read scenario definition from excel
+#' scenarioConfiguration <- readScenarioConfigurationFromExcel(scenarioConfiguration)
+#' }
+readScenarioConfigurationFromExcel <- function(scenarioConfiguration) {
+  # Current scenario definition structure:
+  # "Scenario_name", "IndividualId", "ModelParameterSheets", "ApplicationProtocol",
+  # "SimulationTime", "SimulationTimeUnit", "SteadyState", "ModelFile"
+  colTypes <- c("text", "text", "text", "text", "numeric", "text", "logical", "text")
```
IndrajeetPatil, line 27: Is `"text"` type something we have internally designed?

Because I don't see it in _readr_: https://readr.tidyverse.org/articles/readr.html

R/utilities-scenario-configuration.R
```diff
@@ -0,0 +1,115 @@
...
+#' Set application protocol from excel
+#'
+#' @param scenarioConfiguration A `ScenarioConfiguration` object holding the
+#' name of the application protocol.
+#' @param simulation A `Simulation` object that will be modified.
+#' @keywords internal
+.setApplicationFromExcel <- function(scenarioConfiguration, simulation) {
+  excelFilePath <- file.path(
+    scenarioConfiguration$projectConfiguration$paramsFolder,
+    (scenarioConfiguration$projectConfiguration$scenarioApplicationsFile)
```
IndrajeetPatil, line 105: Why the additional `()` here? I don't see any precedence issues.

IndrajeetPatil (changes requested):

R/ProjectConfiguration.R
PavelBal: The idea was to group the fields by the folders and their contents. So first the folder "modelFolder", then the folder "paramsFolder" following all files located in this folder, then "dataFolder" and the files within, and finally the "outputFolder"

R/ScenarioConfiguration.R
PavelBal: Yes this is good

R/utilities-scenario-configuration.R
PavelBal: Its coming from the readxl package:
https://readxl.tidyverse.org/reference/read_excel.html

IndrajeetPatil (approved):
LGTM, thanks.

IndrajeetPatil:
@PavelBal Can you please update NEWS.md with details about the functionality which was added in this PR?

PR #222, opened 2022-05-04
Sensitivity analysis batch

### LINKED ISSUE #152: SensitivityAnalysis: Run all batches in parallel.

PavelBal:
Fixes #152

Run all batches in parallel.


PavelBal:
@IndrajeetPatil I have written the code to run all required simulations using as few simulation batches as possible, and running alll batches in parallel. You would have to re-write the following code to process the results.

All results have an ID, and mapping of parameter paths and their values is stored in a map. Let us have a quick chat about this.

IndrajeetPatil (approved):
@PavelBal Done. 

Everything should work as expected now, since the key data structure is still the same.

R/sensitivityCalculation.R
```diff
@@ -82,48 +82,131 @@ sensitivityCalculation <- function(simulation,
...
+  # Create a nested list of simulation batch results.
+  # Outer list indexes each parameter path, while inner list corresponds to each
+  # value of parameter factor.
+  simulationResultsBatch <- batchResultsIdMap
+
+  for (parameterPath in seq_along(simulationResultsBatch)) {
+    for (parameterFactor in seq_along(simulationResultsBatch[[parameterPath]])) {
+      simulationResultsBatch[[parameterPath]][[parameterFactor]] <- purrr::pluck(simulationBatchesResults, batchResultsIdMap[[parameterPath]][[parameterFactor]])
+    }
+  }
```
IndrajeetPatil, line 179: The new `simulationResultsBatch` is the same as the old one, and so everything from here on works the same way as before.

PavelBal: Nice, that was easy!

PR #224, opened 2022-05-10
Added test for loading empty sheet.

### LINKED ISSUE #223: `readParametersFromXLS` fails when a sheet does not contain entries
When trying to load a sheet with no parameters, an error is thrown:

```
Error in `vec_assign()`:
! Can't convert `replace` <character> to match type of `data` <logical>.
Run `rlang::last_error()` to see where the error occurred.
```

Here: 
https://github.com/esqLABS/esqlabsR/blob/251f8c729b417406988dd787e1b2b92357eea369/R/utilities-parameters.R#L37

PavelBal:
Fixes #223


PavelBal:
@IndrajeetPatil Any idea how this can be fixed? The problem is that when trying to load a sheet with no parameter entries, a tibble of type "logical" is created:

```
> data
# A tibble: 0 x 4
# ... with 4 variables: Container Path <lgl>, Parameter Name <lgl>, Value <lgl>, Units <lgl>
```

IndrajeetPatil:
The problem is that the default type for `NA` in R is `logical`:

``` r
mode(NA)
#> [1] "logical"
```

<sup>Created on 2022-05-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

So all empty columns in the sheet are read to be of `logical` type.

---

The solution to this is to specify `col_types` when `read_xlsx()` is called:

```r
library(esqlabsR)

paramsXLSpath <- file.path("../data/Parameters.xlsx")
sheets <- c("EmptySheet")

readxl::read_xlsx(paramsXLSpath, sheet = "EmptySheet", col_types = c("text", "text", "numeric", "text"))
#> A tibble: 0 x 4
#> ... with 4 variables: Container Path <chr>, Parameter Name <chr>, Value <dbl>, Units <chr>
```

IndrajeetPatil:
But, I guess you can't foresee how many or which types of columns will be present, so this might not be a feasible solution. 

I will see how call to `replace_na()` can be changed so that no error is produced.

PavelBal:
Ok, for this function the structure is known, so the types can be specified. thanks.

IndrajeetPatil:
I pushed something now; see if that solves the issue. Otherwise, I can revert it.

R/utilities-parameters.R
```diff
@@ -6,12 +6,20 @@
...
+#'   frame output. A list cell loads a column as a list of length 1 vectors,
+#'   which are typed using the type guessing logic from `columnTypes = NULL`, but
+#'   on a cell-by-cell basis.
 #'
 #' @return A list containing vectors 'paths' with the full paths to the
 #'   parameters, 'values' the values of the parameters, and 'units' with the
 #'   units the values are in.
 #' @export
-readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
+readParametersFromXLS <- function(paramsXLSpath, sheets = NULL, columnTypes = NULL) {
```
PavelBal, line 22: columnTypes can be hard coded here - character, character, numeric, character.

IndrajeetPatil: The problem with hard-coding like this is that reading sheets like these won't work:

> Error: Sheet 3 has 11 columns, but `col_types` has length 4.

image

IndrajeetPatil (approved):

R/utilities-parameters.R
```diff
@@ -22,19 +30,22 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
...
-    )
+    fullPaths <- paste(data[["Container Path"]], data[["Parameter Name"]], sep = "|")
     pathsValuesVector[fullPaths] <- as.numeric(data[["Value"]])
-    # replace `NA` in units with `""` represeting the empty unit for
-    # the dimension `Dimensionless`
-    pathsUnitsVector[fullPaths] <- tidyr::replace_na(data = data[["Units"]], replace = "")
+
+    # If any data is present, replace `NA` in units with `""` representing the
+    # empty unit for the dimension `Dimensionless`.
+    if (nrow(data) > 0L) {
```
PavelBal, line 46: This will not work if all units  are `NA`, I think... damn.

R/utilities-parameters.R
```diff
@@ -22,19 +22,18 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
...
-    fullPaths <- paste(data[["Container Path"]], data[["Parameter Name"]],
-      sep = "|"
-    )
+    fullPaths <- paste(data[["Container Path"]], data[["Parameter Name"]], sep = "|")
     pathsValuesVector[fullPaths] <- as.numeric(data[["Value"]])
-    # replace `NA` in units with `""` represeting the empty unit for
-    # the dimension `Dimensionless`
-    pathsUnitsVector[fullPaths] <- tidyr::replace_na(data = data[["Units"]], replace = "")
+
+    pathsUnitsVector[fullPaths] <- tidyr::replace_na(data = as.character(data[["Units"]]), replace = "")
```
PavelBal, line 36: @IndrajeetPatil Like this.

PR #266, opened 2022-06-15
Following guidelines for internal function names

IndrajeetPatil:


IndrajeetPatil:
Why do we even have `getNetHashCode()` function? It's not used anywhere.

PavelBal (approved):

PavelBal:
> Why do we even have getNetHashCode() function? It's not used anywhere.

It has been used in the past, not exactly sure what for. But let's keep it there, might be required and we would not have to re-invent the wheel.

IndrajeetPatil:
I mean, it's a single line of code (`rClr::clrGet(netWrapper$ref, "HashCode")`), so not much of a reinvention, but okay.

PR #268, opened 2022-06-16
Added simulationRunOptions parameter

PavelBal:


DESCRIPTION
```diff
@@ -32,7 +31,9 @@ Imports:
     tools,
     tlf (>= 1.4.0),
     vctrs,
-    writexl
+    writexl,
+    stringr,
+    labeling
```
IndrajeetPatil, line 36: Why do we need these two dependencies? 

I see nothing in the code from the current PR that depends on them.

IndrajeetPatil: Oh, I see. We are using them; they just weren't included as hard deps.

IndrajeetPatil (approved):

PR #283, opened 2022-07-27
Don't skip tests on CI anymore

IndrajeetPatil:


PavelBal:
I think this should be done by setting this value to False:

https://github.com/esqLABS/esqlabsR/blob/59f57e75cbdcede374c898992f1a7ef30774b23b/appveyor.yml#L22

IndrajeetPatil:
Okay, tests seem to pass now with the updated PK-Sim.

PavelBal (approved):

PR #287, opened 2022-07-28
Additional tests for utilities

IndrajeetPatil:


tests/testthat/test-utilities-individual.R
```diff
@@ -50,3 +50,30 @@ test_that("It create IndividualCharacteristics when numerical values are empty",
   expect_equal(individualCharacteristics$population, "European_ICRP_2002")
   expect_equal(individualCharacteristics$gender, "MALE")
 })
+
+## context("writeIndividualToXLS")
+
+test_that("`writeIndividualToXLS()` writes correct data to a spreadsheet", {
+  skip_if_not_installed("withr")
```
PavelBal, line 57: What would happen if `withr` is not installed and `skip_if_not_installed` is not defined? Would the test fail? In this case, I would prefer it to fail. If someone is developing the package, he should also run the tests.

tests/testthat/test-utilities-individual.R
```diff
@@ -50,3 +50,30 @@ test_that("It create IndividualCharacteristics when numerical values are empty",
...
+
+  withr::with_tempdir(
+    code = {
+      simulation <- loadSimulation(system.file("extdata", "simple.pkml", package = "ospsuite"))
```
PavelBal, line 61: Do you need the simulation here?

tests/testthat/test-utilities-simulation.R
```diff
@@ -0,0 +1,22 @@
+## context("initializeSimulation")
+
+test_that("`initializeSimulation()` loads a simulation at the minimum", {
+  simulation <- loadSimulation(system.file("extdata", "simple.pkml", package = "ospsuite"))
+  initializeSimulation(simulation, steadyStateTime = TRUE)
+  simulationResults <- runSimulation(simulation)
+  expect_s3_class(simulationResults, "SimulationResults")
```
PavelBal, line 7: What about 

`expect_true(isOfType(simulationResults, "SimulationResults"))`

It is more straightforward because simulationResults is not actually an S3 class (or not designed to be)

IndrajeetPatil: If it makes it more readable for you, I can change it, but all R6 classes also come with S3 hierarchy. 

cf. https://adv-r.hadley.nz/r6.html#introspection

image

PavelBal (approved):

PR #288, opened 2022-07-28
Extract applications

PavelBal:
- Add a function `getAllApplicationParameters` that returns all parameters of applications in a simulation
- Add a function `exportParametersToXLS` to write parameter information into an excel file that can be loaded in MoBi or R
- Add a function `writeExcel` that is a wrapper for creating a directory (if not present) and writing to excel file


R/utilities-simulation.R
```diff
@@ -150,3 +150,59 @@ compareSimulationParameters <- function(simulation1, simulation2) {
...
+getAllApplicationParameters <- function(simulation, moleculeNames = NULL) {
+  validateIsOfType(simulation, "Simulation")
+  validateIsCharacter(moleculeNames, nullAllowed = TRUE)
+
+  # If no molecules have been specified, get application parameters for all
+  # molecules in the simulation
+  moleculeNames <- moleculeNames %||% simulation$allFloatingMoleculeNames()
+
+  # Returns an object of class `Application` for each administration event
+  applications <- unlist(lapply(moleculeNames, \(x)simulation$allApplicationsFor(x)), use.names = FALSE)
```
IndrajeetPatil, line 188: We support R versions > 3.6, and so the anonymous function syntax (`\()`) can't be used with these older versions of R.

R/utilities-simulation.R
```diff
@@ -150,3 +150,59 @@ compareSimulationParameters <- function(simulation1, simulation2) {
 
   return(list(In1NotIn2 = pathsIn1NotIn2, In2NotIn1 = pathsIn2NotIn1, Different = pathsDiff))
 }
+
+#' Get paramters of applications in the simulation
+#'
+#' @param simulation A `Simulation` object
+#' @param moleculeNames Names of the molecules for which administrations the
+#' parameters will be returned. If `NUll`(default), applications for all
```
IndrajeetPatil, line 158: Something is missing in this first sentence. 

> administrations **of** the parameters?

R/utilities-simulation.R
```diff
@@ -150,3 +150,59 @@ compareSimulationParameters <- function(simulation1, simulation2) {
...
+#' molecules are returned.
+#'
+#' @details Every application event has a `ProtocolSchemaItem` container that
+#' holds parametes describing the dose, start time, infusion time etc.. This
```
IndrajeetPatil, line 162:
```suggestion
#' holds parametes describing the dose, start time, infusion time etc. This
```

R/utilities-parameters.R
```diff
@@ -39,6 +39,59 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
   return(.parametersVectorToList(pathsValuesVector, pathsUnitsVector))
 }
 
+#' Export simulation parameters to excel
+#'
+#' @description Creates an excel file with information from the passed
+#' parameters. The excel sheet will contain columns "Container Path",
+#' "Parameter Name", "Value", and "Units". The resulting file can be loaded in
+#' `MoBi` or in `R` with the function `readParametersFromXLS`.
```
IndrajeetPatil, line 47:
```suggestion
#' `MoBi` or in `R` with the function `readParametersFromXLS()`.
```

R/utilities-parameters.R
```diff
@@ -39,6 +39,59 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
...
+  if (length(output) > 0) {
+    colnames(output) <- c("Container Path", "Parameter Name", "Value", "Units")
+  }
+
+  # Write the results into an excel file.
+  # Wrap the output data frame into a list and name the list if sheet name
+  # has been provided
+  data <- list(output)
+  if (!is.null(sheet)) {
+    validateIsOfLength(sheet, 1)
```
IndrajeetPatil, line 89: All validations should happen early on to avoid failing after expensive computations. 
You can move this to the top with the addition of `, nullAllowed = TRUE`.

IndrajeetPatil (changes requested):

R/utilities-simulation.R
IndrajeetPatil: Fine by me. 

But then you should also bump the needed R version in the DESCRIPTION file.

R/utilities-parameters.R
PavelBal: Moved

IndrajeetPatil:
Looking good.

One last thing:

Can you please also make a note of all these changes in NEWS.md? Including the change in needed R version.

PavelBal:
NEWS updated

IndrajeetPatil (approved):

PR #290, opened 2022-07-28
Tests for shiny apps

IndrajeetPatil:
These looks like massive changes, but it's actually quite simple:

- To test an app, the `{shinytest2}` testing framework expects that the entire app is in a single file (`app.R`), but we were storing the components of apps in two separate files (`ui.R` and `server.R`). So I just moved their contents to `app.R` and deleted the individual files.

- Each app gets its own `tests` folder in the `{shinytest2}` testing framework. But because the apps are stored in the installation folder (`\inst`), these tests won't be run by `{testthat}`. You will need to run them manually if you ever change the apps. 

I have added only rudimentary tests to make sure that the app behavior is as expected. I doubt we actively change the look and feel of these apps much.


PavelBal:
@IndrajeetPatil Thanks. Could you give a brief intro to shiny app testing to me and @hannaei ? We could combine it with an (also brief) intro in snapshot testing.

PavelBal:
@hannaei Also take a look at this PR to get an idea.

PavelBal (approved):

hannaei (approved):

PR #291, opened 2022-07-29
Get rid of a few lints

IndrajeetPatil:


inst/FunctionVisualizer/app.R
```diff
@@ -246,8 +246,6 @@ server <- function(input, output, session) {
 
       ys <- unlist(plotdata[, -1])
       ys <- ys[is.finite(ys)]
-      y_min <- min(ys)
-      y_max <- max(ys)
```
IndrajeetPatil, line 250: @hannaei IINM, these variables are not being used anywhere in the app.

hannaei: You're right, same goes for the previous two lines - must have missed it when I made changes to the app.

hannaei (approved):

PavelBal (approved):

PR #294, opened 2022-08-01
More tests for utilities

IndrajeetPatil:


IndrajeetPatil:
@PavelBal The test is failing because there is no function `ospsuite:::getContainerTask()`. 
It should instead be `ospsuite:::.getContainerTask()`.

This is coming from [here](https://github.com/Open-Systems-Pharmacology/OSPSuite.ParameterIdentification/blob/ce91b1cb9e208f6c7441b3916dcd8501a8d0b3e4/R/utilities-simulation.R#L80).

PavelBal:
Ok this is because we install the PI package from the OSP fork, which is outdated. We should use the parameter identification from esqlabs fork. So either install from GitHub source or enable AppVeyor.

IndrajeetPatil:
> either install from GitHub source

Installation process can't access a private repository:

```r
Error: HTTP error 401.
  Bad credentials
```

> or enable AppVeyor

Doesn't work either:

Screenshot 2022-08-01 at 15 09 11

IndrajeetPatil:
I will just comment it out for now. You can have a look at it later.

PavelBal (approved):

PR #298, opened 2022-08-04
Update documentation

hannaei:
Minor changes in the code documentation


inst/FunctionVisualizer/app.R
```diff
@@ -68,7 +68,7 @@ server <- function(input, output, session) {
     if (!is.null(input$argument)) {
       currentArg <<- input$argument
     }
-    if (v$equation != input$equation) {
+    if (v$equation != input$equation & !is.null(v$snapshots)) {
```
hannaei, line 71: The app would crash when changing the equation without having saved any snapshots before.

IndrajeetPatil (approved):
LGTM, thanks!

PR #299, opened 2022-08-10
Add logo for the package

IndrajeetPatil:
Stickers are big in the R community, and a good way to market ourselves at conferences and meetings. Print them, and distribute them, so that people can put them on their laptops.

cc @StephanSchaller

esqlabsR-logo


maalbrec89:
I have no comparison to other stickers in this area. But I think it is in itself a very fine design. Logo rings could be a bit sharper but I guess the sticker will be small enough for people to notice.

IndrajeetPatil:
Here is an image of my old laptop for a reference:

image

Just to highlight that it will indeed be hard to see how sharp the circles are in the size in which the stickers are typically printed.

maalbrec89:
I think it is great. Fits well.

PavelBal (approved):

StephanSchaller:
Our name associates with this guy... maybe we can add the snake-staff to the sticker?
https://de.wiktionary.org/wiki/%C3%84skulap

PR #302, opened 2022-08-10
Example scenario

### LINKED ISSUE #295: Add a realistic example of `Scenarios.xlsx` file to installation folder
This will make it easy to include examples and tests that use this file.

PavelBal:
Fixes #295

Adds the example project folder to the tests folder.
Removes unused test files.


PavelBal:
I have added the default project folder including the R "Code" folder and all configuration excel files. This should be sufficient to write tests for all functions/classes.

svavil:
@PavelBal do you think the modeling environment script belongs here? Maybe we should drop it from the DoNotSource folder for this PR.

svavil (approved):
I've added a commit to the example-scenario branch that removes the modeling_environment.R file.

PavelBal:
@svavil Yes makes sense.

PR #315, opened 2022-09-08
Suggestions for vignettes

hannaei:
Suggestions for some small changes in the vignettes


vignettes/data-handling.Rmd
```diff
@@ -39,7 +39,7 @@ observedData <- loadObservedData(projectConfiguration = projectConfiguration, sh
 print(names(observedData))
 ```
 
-The `loadObservedData()` function loads the data from a path stored in the `projectConfiguration$dataFile`. This function returns a list of `DataSet` class instances. The resulting object will be used later to [run simulations](example-workflow.html), [plot results](example-visualization.html) and [run a sensitivity analysis](sensitivity-analysis.html).
+The `loadObservedData()` function loads the data from a path stored in the `projectConfiguration$dataFile`. This function returns a list of `DataSet` class instances. The resulting object will be used later to [plot results](example-visualization.html) to compare simulated and observed data.
```
hannaei, line 42: I think observed data is not used when running simulations and sensitivity analysis, so I deleted this. Correct me if I'm wrong

PavelBal (approved):

PR #346, opened 2022-10-25
LLOQ und ULOQ for "stringToNum"

### LINKED ISSUE #345: Extend `stringToNum` to use`LLOQMode`
`stringToNum` should consider different LLOQ modes and also how to treat values above limit of quantification

PavelBal:
Fixes #345


PavelBal:
Build is failing because `{ospsuite}` has not been updated to the latest Core.

PavelBal:
> Do we need to handle entries such as >=10?

Not really, this function is for general string conversion... For observed data, any kind of numbers can be expected. In the specific application (e.g. when using the function to read logP) we should perform the checks..

PavelBal:
Let's wait until PK-Sim is updated with the latest Core.

svavil (approved):

Issue #354, opened 2022-11-15
`createPlotsFromExcel`: support additional parameters for all configurations


PavelBal:
For the excel sheets: `plotConfiguration`, `plotGrids`, and `exportConfiguration`. The user can add columns with any property that is supporter by the configuration (e.g., in `plotGrids` sheet a new column `subtitleColor`), and the values should be applied to the according configuration. If nothing entered, default value is used ( == do not overwrite).

PavelBal:
@hannaei I merged the PR but re-open this issue, as this is not yet implemented for `plotGrids`, correct?

PR #366, opened 2023-01-16
Fix #353. Support additional parameters for plotConfiguration and …

### LINKED ISSUE #353: `createPlotsFromExcel`: sheet `exportConfiguration`
Add a sheet `exportConfiguration` to `Plots.xlsx` with the following columns (not limited to):

- `plotGridName`: expects a name of a plot grid defined in the `plotGrids`.
- `outputName`: name of the output file.

`createPlotsFromExcel` should plot the defined plot grids to PNG. It should be possible to define some but not all plot grids for export.

Once implemented, remove `outputDevice` from `ProjectConfiguration`.

hannaei:
…exportConfiguration, not yet plotGrids (#354)


R/utilities-figures.R
```diff
@@ -343,26 +344,44 @@ createPlotsFromExcel <- function(simulatedScenarios, observedData, projectConfig
...
+
+  lapply(seq_along(columns), function(i) {
+    value <- columns[[i]]
+    colName <- names(columns)[[i]]
     if (!is.na(value)) {
-      # numeric columns
-      if (colName %in% c("xAxisLimits", "yAxisLimits")) {
-        plotConfiguration[[colName]] <- eval(parse(text = value))
+      # Check if the field name is supported by the configuration class
+      if (!.validateClassHasField(object = newConfiguration, field = colName)){
```
PavelBal, line 370: @hannaei I added a check if the property that the user tries to set is supported by the configuration.

R/utilities-figures.R
```diff
@@ -343,26 +344,44 @@ createPlotsFromExcel <- function(simulatedScenarios, observedData, projectConfig
   })
   names(multiPanelPlots) <- dfPlotGrids$name
 
+  if (nrow(dfExportConfigurations) > 0) {
+    # create a list of ExportConfiguration objects from dfExportConfigurations
+    exportConfiguration <- createEsqlabsExportConfiguration(projectConfiguration)
```
PavelBal, line 349: Please add a validation for export configurations, i.e., if all required fields are populated (output name) and the referenced plot grids are defined.

R/utilities-figures.R
```diff
@@ -482,7 +482,7 @@ createPlotsFromExcel <- function(simulatedScenarios, observedData, projectConfig
   }
   # Remove white spaces
   dfPlotGrids$plotIDs <- gsub(dfPlotGrids$plotIDs, pattern = " ", replacement = "", fixed = TRUE)
-  dfPlotGrids$plotIDs <- strsplit(dfPlotGrids$plotIDs, split = "[[:punct:]]")
+  dfPlotGrids$plotIDs <- strsplit(dfPlotGrids$plotIDs, split = ",")
```
PavelBal, line 485: @hannaei We should only use `,`  as separator. Otherwise, plot ids like `P_1` are split into `P` and `1`.

PavelBal (approved):

PR #379, opened 2023-01-31
Output paths in excel

PavelBal:


PavelBal:
Fixes #362

VanessaBa (commented):
the language is a bit mied between passive voice an "yes we can!" tone.

I hope it becomes clear that the scenarios have to be predefined for the use of the `customFunction` . Probably only my Denkfehler ;-)

PR #386, opened 2023-02-10
Update description of required packages

PavelBal:


PavelBal:
@svavil Please extend such that `installPackagesLocally()` will download PK-Sim minimal from the location specified in `.developPaths$pksimMinimal` and unzips it into folder "PKSim" in the project folder:

image

PR #397, opened 2023-03-07
Add dark grey frame around legends by default.

Felixmil:
- increase osps version dependency
- update NEWS

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -18,7 +18,7 @@ URL: https://github.com/esqLABS/esqlabsR
 BugReports: https://github.com/esqLABS/esqlabsR/issues
 Depends:
     R (>= 4.1),
-    ospsuite (>= 11.1),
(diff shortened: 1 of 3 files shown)


DESCRIPTION
```diff
@@ -18,7 +18,7 @@ URL: https://github.com/esqLABS/esqlabsR
 BugReports: https://github.com/esqLABS/esqlabsR/issues
 Depends:
     R (>= 4.1),
-    ospsuite (>= 11.1),
+    ospsuite (>= 11.2.224),
```
PavelBal, line 21: Ok let's not do this... It will not work any more if using `ospsuite` built locally, because the version there is `11.1` or `11.2` but is not synchronized with the AppVeyor version.

Felixmil: Ok, i'll try to think of something to make dependency on specific more secure.

PavelBal (approved):

PR #398, opened 2023-03-07
- New function `compareSimulation()` to compare two simulations for d…

PavelBal:
…ifferences.

- Function `compareSimulationParameters()` has been removed and replaced by `compareSimulations()`

- When defining an individual of other species than human in `ScenarioConfiguration` and applying it to a human model, missing species-specific parameters are applied and the scaling works properly. Supported scalingsa are: Human to rat, human to monkey, human to rabbit.


tests/testthat/test-utilities-simulation.R
```diff
@@ -22,15 +22,17 @@ test_that("`initializeSimulation()` does not fail when additionalParams is empty
 
 ## context("compareSimulationParameters")
 
-test_that("`compareSimulationParameters()` produces no differences with identical simulations", {
+test_that("`compareSimulations()` produces no differences with identical simulations", {
```
Felixmil, line 25: Maybe add a test case for when simulations contains different paths and/or paramter values ?

PavelBal: @Felixmil Added.

Felixmil (approved):

PR #404, opened 2023-03-25
Fixed error when missing simulation results were not ignored in creat…

PavelBal:
…eDataCombinedFromExcel

### DIFF
--- a/R/utilities-data-combined.R
+++ b/R/utilities-data-combined.R
@@ -156,7 +156,7 @@ createDataCombinedFromExcel <- function(file, sheet = NULL, dataCombinedNames = 
       stop(messages$stopInvalidScenarioName(missingScenarios))
     }
     warning(messages$warningInvalidScenarioName(missingScenarios))
-    dfDataCombined <- dfDataCombined[!(missingScenarios == dfDataCombined$scenario), ]
+    dfDataCombined <- filter(dfDataCombined, dataType == "observed" | !(scenario %in% missingScenarios))
   }
   # data set name not present in observedData
   missingDataSets <- setdiff(setdiff(dfDataCombined$dataSet, names(observedData)), NA)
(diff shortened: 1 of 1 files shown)


R/utilities-data-combined.R
```diff
@@ -156,7 +156,7 @@ createDataCombinedFromExcel <- function(file, sheet = NULL, dataCombinedNames =
       stop(messages$stopInvalidScenarioName(missingScenarios))
     }
     warning(messages$warningInvalidScenarioName(missingScenarios))
-    dfDataCombined <- dfDataCombined[!(missingScenarios == dfDataCombined$scenario), ]
+    dfDataCombined <- filter(dfDataCombined, dataType == "observed" | !(scenario %in% missingScenarios))
```
PavelBal, line 159: This should remove all rows where `dataType` is `simulated` (== not observed) and the name of the scenario is one of the `missingScenarios`.

Felixmil: > and
you meant "or" right ?. 

Anyway, I would suggest to put the first condition in `( )`, it could prevent some weird things happening in case these condition evolve.

Felixmil (approved):

Felixmil (changes requested):
( after ,

R/utilities-data-combined.R
```diff
@@ -156,7 +156,7 @@ createDataCombinedFromExcel <- function(file, sheet = NULL, dataCombinedNames =
       stop(messages$stopInvalidScenarioName(missingScenarios))
     }
     warning(messages$warningInvalidScenarioName(missingScenarios))
-    dfDataCombined <- dfDataCombined[!(missingScenarios == dfDataCombined$scenario), ]
+    dfDataCombined <- filter((dfDataCombined, dataType == "observed") | !(scenario %in% missingScenarios))
```
Felixmil, line 159: `(` after `,` :
`filter(dfDataCombined, (dataType == "observed") | !(scenario %in% missingScenarios))`

Felixmil (changes requested):

R/utilities-data-combined.R
PavelBal: Damn..

PR #410, opened 2023-04-05
Fix reading custom field plot config

PavelBal:


R/utilities-figures.R
```diff
@@ -301,7 +301,7 @@ createPlotsFromExcel <- function(plotGridNames = NULL, simulatedScenarios, obser
       individual = plotIndividualTimeProfile(dataCombined, plotConfigurationList[[plotId]]),
       population = {
         quantiles <- dfPlotConfigurations[dfPlotConfigurations$plotID == plotId, ]$quantiles
-        if (is.na(quantiles)) {
+        if (is.null(quantiles) || is.na(quantiles)) {
```
PavelBal, line 304: If no column `quantiles` is found, `quantiles` is `NULL` and must be captured here.

R/utilities-figures.R
```diff
@@ -392,7 +392,7 @@ createPlotsFromExcel <- function(plotGridNames = NULL, simulatedScenarios, obser
       # For fields that require multiple values (e.g., axis limits require the
       # upper and the lower limit value), values are separated by a ','.
       # Split the input string first
-      value <- unlist(strsplit(value, split = ","))
+      value <- unlist(strsplit(as.character(value), split = ","))
```
PavelBal, line 395: When the cell entries are numerical, this failed before.

appveyor.yml
```diff
@@ -57,7 +57,7 @@ build_script:
   - Rscript -e "install.packages('https://github.com/Open-Systems-Pharmacology/rClr/releases/download/v0.9.2/rClr_0.9.2.zip', repos = NULL, type = 'binary')"
   - Rscript -e "download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/ospsuite-r/artifacts/ospsuite.zip?pr=false', destfile = 'ospsuite.zip', mode='wb'); install.packages('ospsuite.zip', repos = NULL, type = 'binary')"
   # hard dependencies (Imports)
-  - Rscript -e "install.packages(c('FME', 'readr', 'tidyr', 'ggplot2', 'patchwork'), repos = 'http://cran.us.r-project.org')"
+  - Rscript -e "install.packages(c('FME', 'readr', 'tidyr', 'ggplot2', 'patchwork', 'GenSA'), repos = 'http://cran.us.r-project.org')"
```
Felixmil, line 60: Why do we need to install GenSA ? It is not listed in package dependencies.

PavelBal: It is required by the `ospsuite.parameteridentification`.

Felixmil: then shouldn't it be an Import in `ospsuite.parameteridentification` ?

PavelBal: But it will not be automatically installed, right?

PavelBal: >I think that if parametridentification has GenSA as import then it will be installed when appVeyor runs this: 

Hmm doesn't seem to work like this, `GenSA` is part of Imports of the PI package, but Appveoyr fails if it is not installed explicitely.

Felixmil: Yes, it is just like this: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1234

Installing from zip file will not install package dependencies. 

But if Appveyors generates a `.tar.gz` for `ospsuite.parameteridentification`, then we could use `remotes::install_local()` just like [here](
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/59d42c2666a0f3023c48c3a6d2ff2c2f5d47d7a7/install_dependencies.R#L6-L7):
```r
download.file('https://ci.appveyor.com/api/projects/open-systems-pharmacology-ci/tlf-library/artifacts/tlf.tar.gz?pr=false', destfile = 'tlf.tar.gz', mode='wb')
remotes::install_local('tlf.tar.gz', dependencies = c('Depends', 'Imports'), upgrade = 'never')
```

Felixmil: We could also use `remotes::install_github("esqLABS/esqlabs.ParameterIdentification")` which would install dependencies.

PavelBal: Ok I have created a separate issue for this.

Felixmil (approved):

PR #427, opened 2023-04-21
Better error excel

### LINKED ISSUE #426: Better error when Scenarios structure is not compatible
When the expected sturcture is different (e.g. some new columns are required), an error is thrown that some data types are not matching... Throw a better error telling which columns must be presetn

PavelBal:
Fixes #426

@Felixmil @svavil I tried to implement this but failed.

The problem is that I only found a way to check for correct column names *after* the file has been loaded, but now I am stuck with converting column to required types.


R/utilities-scenario-configuration.R
```diff
@@ -28,18 +28,41 @@ readScenarioConfigurationFromExcel <- function(scenarioNames = NULL, projectConf
...
+
+  # Check if the structure is correct
+  if (length(expectedColumns) != length(names(wholeData)) && any((expectedColumns != names(wholeData)))) {
+    stop(messages$errorWrongXLSStructure(filePath = file.path(
+      projectConfiguration$paramsFolder,
+      projectConfiguration$scenarioDefinitionFile
+    ), expectedColNames = expectedColumns))
+  }
+
+  # 2DO - Cast column types
```
PavelBal, line 61: This is where I was trying to convert column types.

PavelBal:
Now the error is more informative:

` Loading from XLS failed, the file 'C:\Temp\Rtmpiy6VGB/Scenarios.xlsx' has wrong structure!
    The file should contain columns 'Scenario_name, IndividualId, PopulationId, ModelParameterSheets, ApplicationProtocol, SimulationTime, SimulationTimeUnit, SteadyState, SteadyStateTime, SteadyStateTimeUnit, ModelFile, OutputPathsIds'`

Felixmil:
@PavelBal  Is it possible to have more columns that the expectedColumns defined here:
```r
expectedColumns <- c(
    "Scenario_name", "IndividualId", "PopulationId", "ReadPopulationFromCSV", "ModelParameterSheets", "ApplicationProtocol",
    "SimulationTime", "SimulationTimeUnit", "SteadyState", "SteadyStateTime", "SteadyStateTimeUnit", "ModelFile",
    "OutputPathsIds"
  )
```

PavelBal:
@Felixmil Not for this file, no.

Felixmil:
In latest commit I implement the process in three steps:
1) read the excel file with `n_max=0` to read only the header.
2) Structure is verified like before (but using `identical()`)
3) If Structure is ok: read the whole excel file, otherwise raise error.

More about `identical()`:

```r
> x <- c("A", "B")
> y <- c("B", "A")
> identical(x, y)
[1] FALSE
> identical(x, sort(y))
[1] TRUE
```
So here, we expect that ALL the expected columns are in the header and that they are in the exact same order.

PavelBal:
I think this is a good solution, I wasn't aware of `n_max` (or was too lazy reading the docs, my bad).

PR #429, opened 2023-04-21
Lintr suggestions

PavelBal:


PavelBal:
While playing around with Visual Studio Code, I used `lintr` and implemented some of the suggestions. I think it is quite useful as at one place it revealed a typo in variable name, and at other place pointed to a not used variable.

R/project-configuration.R
```diff
@@ -3,7 +3,7 @@
 #' @description An object storing configuration used project-wide
 #' @format NULL
 #' @export
-ProjectConfiguration <- R6::R6Class(
+ProjectConfiguration <- R6::R6Class( # nolint: object_name_linter.
```
PavelBal, line 6: Bit annoying that it complains about our class names. Maybe there is an option to define a separate style for R6 class names?

R/sensitivity-calculation.R
```diff
@@ -208,17 +211,17 @@ sensitivityCalculation <- function(simulation,
       warning(messages$noPKDataToWrite())
     } else {
       # Convert tidy data to wide format for each output path
-      pkData_wide_list <- purrr::map(
+      pkDataWideList <- purrr::map(
```
PavelBal, line 214: Renamed the variable.

R/project-configuration.R
Felixmil: You can use these settings I setup for test in TLF package: https://github.com/Felixmil/TLF-Library/blob/gh-actions-test/.lintr

.lintr
```diff
@@ -0,0 +1,4 @@
+linters: linters_with_defaults(
```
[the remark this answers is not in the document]
Felixmil: Yes it is ! Lintr rules should be harmonized by project/company, not user so it makes perfect sens.

R/project-configuration.R
PavelBal: nice

PavelBal:
@Felixmil it's up to you if you want to add anything else to the default linter config. I would prefer to keep `object_length_linter` and `cyclocomp_linter` as the rules do make sense and should be reminders for us not to make the code too complex.

Felixmil:
> @Felixmil it's up to you if you want to add anything else to the default linter config. I would prefer to keep `object_length_linter` and `cyclocomp_linter` as the rules do make sense and should be reminders for us not to make the code too complex.

Should be enough for now, I think better code test coverage is more important (reach >90%).

R/utilities-population.R
```diff
@@ -57,7 +57,7 @@ readPopulationCharacteristicsFromXLS <- function(XLSpath, populationName, sheet
 #' Possible gender entries as integer values
 #'
 #' @export
-GenderInt <- enum(list(
+GenderInt <- enum(list( # nolint: object_name_linter.
```
Felixmil, line 60:
```suggestion
GenderInt <- enum(list(
```

R/utilities-population.R
```diff
@@ -164,7 +164,7 @@ extendPopulationFromXLS <- function(population, XLSpath, sheet = NULL) {
 
 #' Supported distributions for sampling
 #' @export
-Distributions <- enum(list(
+Distributions <- enum(list( # nolint: object_name_linter.
```
Felixmil, line 167:
```suggestion
Distributions <- enum(list(
```

Felixmil (approved):

PR #436, opened 2023-04-24
Check if the specified output has been simulated

### LINKED ISSUE #406: `createDataCombinedFromExcel()`: better error when simulation results not found
Currently, when trying to create `DataCombined` and results for a certain path in the simulated results are not present, following error is thrown:

```
Error in simulationResults$getValuesByPath(path, individualIds, stopIfNotFound) : 
  rmarkdown::render: No results exists for path 'Organism|blablabla' for individual IDs '0'!
```

The function should check it in advance and throw a better warning, stating the name of the `DataCombined` and `scenario`.

PavelBal:
Fixes #406

### DIFF
--- a/R/messages.R
+++ b/R/messages.R
@@ -153,6 +153,12 @@ messages$stopNoPathProvided <- function(dataCombinedName) {
   )
 }
 
+messages$stopWrongOutputPath <- function(dataCombinedName, scenarioName, path) {
+  paste0(
+    "Output path '", path, "' is defined in the DataCombined '", paste0(dataCombinedName, collapse = ", "), "' for scenario '", scenarioName, "' but has not been simulated. Please check that the output path is specified for this scenario."
+  )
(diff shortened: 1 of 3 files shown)


R/messages.R
```diff
@@ -153,6 +153,12 @@ messages$stopNoPathProvided <- function(dataCombinedName) {
   )
 }
 
+messages$stopWrongOutputPath <- function(dataCombinedName, scenarioName, path) {
+  paste0(
```
Felixmil, line 157: Maybe we could use {glue} in the future for long strings like this.

Felixmil (approved):

R/messages.R
PavelBal: @Felixmil feel free to make a suggestion, I would be glad to learn about it and I think in the `messages` there are many use cases for this.

PR #437, opened 2023-04-24
Rework install-packages.R file

Felixmil:
```
installEsqLabsR()

How do you want to install the packages? 

1: In local environment
2: Globally
> 1

Installing installation dependencies.
✔ Getting minimal version of PKSim. [332ms]
✔ Installing esqlabsR and dependencies [6m 35s]
✔ Loading esqlabsR [1.5s]
✔ Verifying PKSim installation [1.7s]
! Test simulation files could not be found.
✖ Running test simulation. [123ms]
✔ Installation successful
ℹ Restarting R session.

```


PavelBal (changes requested):

inst/extdata/install-package.R
```diff
@@ -0,0 +1,126 @@
...
+  cli::cli_progress_step("Loading esqlabsR")
+  # test if installed packages can be loaded
+  library("esqlabsR")
+
+  cli::cli_progress_step("Verifying PKSim installation")
+  # test if PK-Sim.R dll can be loaded
+  ospsuite::initPKSim()
+
+
+  cli::cli_progress_step("Running test simulation.")
```
PavelBal, line 104: Also only for local installation.

PavelBal:
OK, it works and it works well, the only unsolved thing is the version problem... I guess we have to live with it.

Could you please rename the file to `install-packages`? Otherwise we'll have to adjust the PowerShell script which copies the `install-packages` and maybe something else.

Issue #428, opened 2023-04-21
Make test parameters independent on external function


PavelBal:
In `initializeScenario()`,

```
  if (scenarioConfiguration$setTestParameters) {
    warning("INFO: 'scenarioConfiguration$setTestParameters' is set to TRUE,
            parameter values defined in 'InputCode/TestParameters.R' will be applied!")
    params <- extendParameterStructure(
      parameters = params,
      newParameters = getTestParameters()
    )
  }
```

`getTestParameters()` must be defined in the project and is not passed to the function. This is very ugly, have to find other way for applying test parameters.

Suggestion - test parameters should become part of `customParams` and be population in the project code and not in the package.

Issue #419, opened 2023-04-20
Return final parametrization of a scenario


PavelBal:
Listing only the parameters that have been set through excel or by custom functions.

> Another nice output to have in R is the final value of all the parameters that occur in the excel sheets per simulation

Issue #330, opened 2022-09-27
`runScenarios()` should perform steady-state simulation during initialization in parallel


PavelBal:
Currently, if `ScenarioConfiguration$steadyState = TRUE`, the steady-state is calculated in applied in `initializeScenario()` which is called sequentially.

Find a way to run steady-state simulations for the different Scenarios in parallel (gather ss-scenarios in a list an perform SS-simulation in `runScenarios()`?)

PR #441, opened 2023-04-27
Separate scenario initialization from actual running

### LINKED ISSUE #342: `runScenarios` - separate simulation initialization from actual running
Return the simulations and have a separate function to run them?

PavelBal:
Workflow changed to:

 1. Create `ScenarioConfiguration`
 2. Create `Scenario`
 3. Run `Scenario`

- New class `Scenario`
- New function `createScenarios()` that creates `Scenario` objects from `ScenarioConfiguration`
- `runScenarios()` requires `Scenario` objects
- `setTestParameters` has been removed from `ScenarioConfiguration`. All additional parameters must be supplied by `customParams` argument to `createScenarios()`
- `simulationRunOptions` is not a part of `ScenarioConfiguration` but an argument of `runScenarios()`
- Custom functions to simulations are not supported any more, as direct access to the `Simulation` object is possible through `Scenario` objects

- Function `initializeScenario()` removed.
- Function `setApplications()` is deprecated.

Fixes #342, #419, #428, #330


R/utilities-scenarios.R
```diff
@@ -2,311 +2,149 @@
...
-      population = populations[[scenarioName]],
-      simulationRunOptions = scenarioConfiguration$simulationRunOptions
+      simulations = scenario$simulation,
+      population = scenario$population,
+      simulationRunOptions = simulationRunOptions
     )
     simulationResults <- c(simulationResults, populationResults)
   }
 
+  # Create output list with simulation results, simulation objects, and population objects
```
Felixmil, line 70: Maybe this step can be done together with the other steps (there's already a for loop on `scenarios` earlier in code).

R/utilities-scenarios.R
```diff
@@ -2,311 +2,149 @@
...
-  # Call the custom function.
-  if (!is.null(scenarioConfiguration$customFunction)) {
-    # Set the environment of the custom function to the current environment so it
-    # has access to the `simulation` object
-    environment(scenarioConfiguration$customFunction) <- environment()
-    # Call the custom function with the arguments as defined in the scenario
-    # configuration
-    do.call(what = scenarioConfiguration$customFunction, args = scenarioConfiguration$customFunctionArgs)
+  # For each scenario configuration, create a `Scenario` object
+  for (i in seq_along(scenarioConfigurations)) {
```
Felixmil, line 135: use `{purrr}` for efficiency.

Something like this should work.

```suggestion
  scenarios <- purrr::map(scenarioConfigurations, ~Scenario$new(.x, customParams = customParams) %>% 
  purrr::set_names(purrr::map(scenarioConfigurations, ~.x$scenarioName)
```

R/utilities-scenarios.R
PavelBal: I don't think so, we don't have `simulationResults` up to this step.

R/utilities-scenarios.R
```diff
@@ -2,311 +2,149 @@
...
-          XLSpath = file.path(
-            scenarioConfiguration$projectConfiguration$paramsFolder,
-            scenarioConfiguration$projectConfiguration$populationParamsFile
-          ),
-          populationName = scenarioConfiguration$populationId,
-          sheet = "Demographics"
-        )
-        population <- createPopulation(populationCharacteristics = popCharacteristics)
-      }
+runScenarios <- function(scenarios, simulationRunOptions = NULL) {
```
Felixmil, line 15:
```suggestion
runScenarios <- function(scenarios, simulationRunOptions = NULL) {
  # List of simulation with steady-state
  steadyStateSimulations <- list()
  # Have to store steady-state times separately, because they are not part of the simulation object
  steadyStateTimes <- list()
  for (scenario in scenarios) {
    if (scenario$scenarioConfiguration$simulateSteadyState) {
      steadyStateSimulations <- c(steadyStateSimulations, scenario$simulation)
      steadyStateTimes <- c(steadyStateTimes, scenario$scenarioConfiguration$steadyStateTime)
    }
  }

  # Simulate steady-state concurrently
  if (length(steadyStateSimulations) > 0) {
    initialValues <- ospsuite.parameteridentification::getSteadyState(
      simulations = steadyStateSimulations,
      steadyStateTime = steadyStateTimes, ignoreIfFormula = TRUE,
      simulationRunOptions = simulationRunOptions
    )

    # Set initial values for steady-state simulations
    for (simulation in steadyStateSimulations) {
      ospsuite::setQuantityValuesByPath(
        quantityPaths = initialValues[[simulation$id]]$paths,
        values = initialValues[[simulation$id]]$values, simulation = simulation
      )
    }
  }

  # Create output list with simulation results, simulation objects, and population objects
  returnList <- list()
  for (scenario in scenarios) {
    scenarioName <- scenario$scenarioConfiguration$scenarioName
    simulation <- scenario$simulation
    id <- simulation$id
    population <- scenario$population
    results <- switch(scenario$scenarioType,
      "Individual" = runSimulations(
        simulations = simulation,
        simulationRunOptions = simulationRunOptions
      )[[1]],
      "Population" = runSimulations(
        simulations = simulation,
        population = population,
        simulationRunOptions = simulationRunOptions
      )[[1]]
    )

    # Retrieving quantities from paths to support pattern matching with '*'
    outputQuantities <- NULL
    if (!is.null(scenario$scenarioConfiguration$outputPaths)) {
      outputQuantities <- getAllQuantitiesMatching(
        scenario$scenarioConfiguration$outputPaths,
        simulation
      )
    }
    outputValues <- getOutputValues(results,
      quantitiesOrPaths = outputQuantities
    )

    returnList[[scenarioName]] <- list(
      simulation = simulation,
      results = results,
      outputValues = outputValues,
      population = population
    )
  }

  # Call gc() on .NET
  ospsuite::clearMemory()
  return(returnList)
}

```

PavelBal: In this case, every simulation will be executed sequentially, as opposed to calling `runSimulations()` on multiple simulation objects. The latter runs simulations in parallel on .NET side, which is much more performant than simulating simulation by simulation.

In case of population simulations, we can only run single simulation per time because it already parallelizes the different individuals.

Felixmil: Oh ok. good to know. thanks.

Felixmil (approved):

PR #448, opened 2023-05-22
Intro presentation

Felixmil:
Add a presentation equivalent to the "Get started" vignette + a vignette for listing presentation materials.

image


PavelBal:
@Felixmil some contents of the vignettes have changed after this presentation, this should be probably adjusted?

PR #458, opened 2023-06-19
proper handling of 'na'  in individual and administration protocol de…

PavelBal:
…finition


PavelBal:
What's with this `UnitConverter-001.new.png` snapshot, it is failing after each commit?

Felixmil (approved):

PR #468, opened 2023-06-26
Fix appveyor's build

Felixmil:


Felixmil:
Problem was cause by osps, solved in https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1261

PR #472, opened 2023-06-30
Fix plot title parsing

### LINKED ISSUE #471: Correctly read title of plotGrid
Following fields in Plots.xlsx should be split by `,`.

- Sheet `plotGrids`, title
- Sheet `exportConfiguration`, outputName

PavelBal:
Fixes #471


R/utilities-figures.R
```diff
@@ -342,8 +342,9 @@ createPlotsFromExcel <- function(
   plotGrids <- apply(dfPlotGrids, 1, \(row) {
     plotGridConfiguration <- .createConfigurationFromRow(
       defaultConfiguration = defaultPlotGridConfig,
-      row[!(names(row) %in% c("name", "plotIDs"))]
+      row[!(names(row) %in% c("name", "plotIDs", "title"))]
```
PavelBal, line 345: `title` column was processed as other columns, where we follow the logic that "list" of values are separated by a comma. In case of a title (and file name in the next changed code chunk), there is always only one value and should not be split.

PavelBal:
@Felixmil I am merging now. Please review when you are back.

PR #481, opened 2023-07-06
Remove entries for parameters with value == NaN

### LINKED ISSUE #480: exportParametersToXLS - ignore parameters with NaN
Some parameters might have a `NaN´ as value. These parameters should be ignored when exporting to excel.

PavelBal:
Fixes #480

### DIFF
--- a/R/utilities-parameters.R
+++ b/R/utilities-parameters.R
@@ -67,10 +67,19 @@ exportParametersToXLS <- function(parameters, paramsXLSpath, sheet = NULL) {
 
   for (paramIdx in seq_along(parameters)) {
     param <- parameters[[paramIdx]]
-    parameterContainerPath[[paramIdx]] <- param$parentContainer$path
-    parameterName[[paramIdx]] <- param$name
-    parameterUnits[[paramIdx]] <- param$unit
-    parameterValue[[paramIdx]] <- param$value
+    value <- param$value
+    if (!is.nan(value)){
+      parameterContainerPath[[paramIdx]] <- param$parentContainer$path
+      parameterName[[paramIdx]] <- param$name
+      parameterUnits[[paramIdx]] <- param$unit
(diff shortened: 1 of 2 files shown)


R/utilities-parameters.R
```diff
@@ -67,10 +67,19 @@ exportParametersToXLS <- function(parameters, paramsXLSpath, sheet = NULL) {
...
-    parameterName[[paramIdx]] <- param$name
-    parameterUnits[[paramIdx]] <- param$unit
-    parameterValue[[paramIdx]] <- param$value
+    value <- param$value
+    if (!is.nan(value)){
+      parameterContainerPath[[paramIdx]] <- param$parentContainer$path
+      parameterName[[paramIdx]] <- param$name
+      parameterUnits[[paramIdx]] <- param$unit
+      parameterValue[[paramIdx]] <- param$value
+    } else {
```
PavelBal, line 76: With this implementation, the function creates an excel file with empty rows. Is there a good way to remove all `NAs` before exporting?

Felixmil: @PavelBal  Should be. 

Which unit test / reproducible example I can use to work on this ?

PavelBal:
```
sim <- loadSimulation(system.file("extdata", "Aciclovir.pkml", package = "ospsuite"))

param1 <- getParameter("Organism|SA proportionality factor", sim)
param2 <- getParameter("Organism|Specific clearance (endosome)", sim)
param2$value <- NaN
param3 <- getParameter("Organism|Surface area scaling exponent", sim)

exportParametersToXLS(c(param1, param2, param3), paramsXLSpath = "../test.xlsx")
```

Felixmil: Thanks, done in 9e4b492

Felixmil (approved):

PR #505, opened 2023-07-31
Drafted initial content for vignette on using Plots.xlsx

PavelBal:


Felixmil:
@PavelBal  I completed your draft but I think it still need some finishing touches. Can you take a look ?

vignettes/Figures/plot-workflow.png
PavelBal, line 1: I think it makes sense to include the sources of such figures into the package. Could you please add it?

PR #509, opened 2023-08-08
Title of plot configuration is not split any more

PavelBal:
Fixing issue described here https://github.com/esqLABS/esqlabsR.reports/issues/6

### DIFF
--- a/R/utilities-figures.R
+++ b/R/utilities-figures.R
@@ -323,8 +323,8 @@ createPlotsFromExcel <- function(
                               "plotType",
                               "title",
                               "subtitle",
-                              "xLabel",
-                              "yLabel",
+                              "xlabel",
+                              "ylabel",
                               "aggregation",
                               "quantiles",
                               "foldDistance"))]
--- a/R/utilities-figures.R
+++ b/R/utilities-figures.R
@@ -318,7 +318,16 @@ createPlotsFromExcel <- function(
     plotConfiguration <- .createConfigurationFromRow(
       defaultConfiguration = defaultPlotConfiguration,
       # Have to exclude all columns that should not be vectorized
-      row[!(names(row) %in% c("plotID", "DataCombinedName", "plotType", "title", "aggregation", "quantiles", "foldDistance"))]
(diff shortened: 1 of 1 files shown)


R/utilities-figures.R
```diff
@@ -313,13 +313,18 @@ createPlotsFromExcel <- function(
...
-      c("plotID", "DataCombinedName", "plotType", "aggregation", "quantiles", "foldDistance"))],
-    1, .createConfigurationFromRow,
-    defaultConfiguration = plotConfiguration
-  )
+  defaultPlotConfiguration <- createEsqlabsPlotConfiguration()
+  plotConfigurationList <- apply(dfPlotConfigurations, 1, \(row){
+    plotConfiguration <- .createConfigurationFromRow(
+      defaultConfiguration = defaultPlotConfiguration,
+      # Have to exclude all columns that should not be vectorized
+      row[!(names(row) %in% c("plotID", "DataCombinedName", "plotType", "title", "aggregation", "quantiles", "foldDistance"))]
```
Felixmil, line 321: shouldn't subtitle, caption, xLabel and yLabel be also excluded here ?

PavelBal: You are right, this just never have been used so far.. Adding it.

R/utilities-figures.R
```diff
@@ -313,13 +313,27 @@ createPlotsFromExcel <- function(
...
+    plotConfiguration <- .createConfigurationFromRow(
+      defaultConfiguration = defaultPlotConfiguration,
+      # Have to exclude all columns that should not be vectorized
+      row[!(names(row) %in% c("plotID",
+                              "DataCombinedName",
+                              "plotType",
+                              "title",
+                              "subtitle",
+                              "xLabel",
+                              "yLabel",
```
Felixmil, line 327:
```suggestion
                              "ylabel",
```

PavelBal:
@Felixmil no, the field in the `DefaultPlotConfiguration` is called `xLabel` (capital L), so the column name would be the same.

PR #512, opened 2023-08-10
Do not fail when a scenario cannot be simulated, but show a warning

### LINKED ISSUE #493: Show informative error when a scenario fails
When a scenario simulation fails, show which did fail.

PavelBal:
Fixes #493

### DIFF
--- a/R/messages.R
+++ b/R/messages.R
@@ -1,49 +1,12 @@
 messages <- ospsuite.utils::messages
 
+# Parameters structure####
 messages$errorWrongXLSStructure <- function(filePath, expectedColNames, optionalMessage = NULL) {
   paste0(
     "Loading from XLS failed, the file '", filePath, "' has wrong structure!
     The file should contain columns '", paste0(expectedColNames, collapse = ", "), "'",
     optionalMessage
   )
-}
-
-messages$errorCouldNotCompareParameters <- function(parameter1, parameter2, optionalMessage = NULL) {
-  paste0(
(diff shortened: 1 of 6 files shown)


PavelBal:
I also tried to bring some structure to the messages file.

R/messages.R
```diff
@@ -27,52 +30,50 @@ messages$errorWrongPopulationName <- function(populationName) {
...
 }
 
+# utilities-individual####
 messages$errorWrongIndividualId <- function(individualId) {
   paste0("Individual with id ", individualId, " is not specified in the individual characteristics file!")
 }
 
+# utilities####
 messages$warningValueWithinThresholdNotExisting <- function(value, threshold, optionalMessage = NULL) {
   paste(
```
Felixmil, line 40:
```suggestion
  paste0(
```

Felixmil (changes requested):
That's way better like this.

I spoted a `paste()` instead of `paste0()` and also maybe it should be harmonized in terms of indentations:
```r
paste0("....
        .....')

# or 

paste0(
  ".....
   ......"
)
```

Another improvement suggestion would be to use `glue::glue()` for easier readability. It is not in DESCRIPTION dependency list right now but we already have it indirectly since it's a ggplot2 dependency (and many other packages we use).

PavelBal:
> Another improvement suggestion would be to use `glue::glue()` for easier readability.

Let's do it separately (feel free to implement, I have not used `glue` so far ;) )

Felixmil (approved):

PR #517, opened 2023-08-17
Some changes to default plot configurations

PavelBal:


Felixmil:
I wonder why these changes did not produce new snapshot to update 🤔

PavelBal:
Because we do not have any snapshots using the configurations :D

R/esqlabs-plot-export-configuration.R
```diff
@@ -49,15 +49,10 @@ ExportConfiguration <- R6::R6Class(
...
       if (!is.null(self$heightPerRow)) {
         # Get the number of rows as defined in the plot layout
-        nrOfRows <- plotObject$patches$layout$nrow
-        # If the value was not explicitely defined, it is `NULL`. In this case,
-        # use the the same method to calculate the number of rows for a given
-        # number of figures that is used in ggplot2::facet_wrap()
-        if (is.null(nrOfRows)) {
-          # number of plots + 1, as somehow the lenght of plotObject$patches$plots
-          # is one too short?
-          nrOfRows <- grDevices::n2mfrow(length(plotObject$patches$plots) + 1)[2]
```
Felixmil, line 59: what does it do ?

PavelBal: Instead of defining a fixed height for a figure, a height per row is defined. So for a plot with a single figure, the size is 12 times 9 cm. When having a plot with multiple panels, having a fixed makes single panels too flat. Instead, number of rows in the grid plot is calculated and multiplied by the defined height per row. So if the plot has 4 panels arranged in 2x2 grid, the height will be 2*9=18 cm.

Felixmil (approved):

PR #518, opened 2023-08-18
Change PI location from esqlabs to OSP

Felixmil:


release-process.R
```diff
@@ -48,7 +48,7 @@ desc::desc_set_remotes(
     "Open-Systems-Pharmacology/OSPSuite.RUtils@88152af44e32e714a1d5c80237c239105ed85120",
     "Open-Systems-Pharmacology/TLF-Library@03e1acfc38c283536e1841b681a04f2eb42e08f9",
     "Open-Systems-Pharmacology/OSPSuite-R@7e08e5603982f11ae72d5c8998f3d998e8a37650",
-    "esqLABS/esqlabs.ParameterIdentification@8559c93d4589c5195298c859b51f90f53d916e11"
+    "Open-Systems-Pharmacology/ospsuite.parameteridentification"
```
PavelBal, line 51: Shouldn't here a specific commit be references, as for the other packages?

Felixmil: It should be updated with latest commit reference when we release the next version yes. Same for other packages.

PavelBal (approved):

PR #520, opened 2023-08-23
No warnings for `strToNum()`

PavelBal:
- `stringToNum()` does not show a warning `NAs introduced by coercion…` when a value
cannot be converted to a numeric any more.


svavil:
I would appreciate if we could suppress only the specific warning ("NAs introduced by coercion"), but this does not seem to be possible with `suppressWarnings()`. Approving as it is.

svavil (approved):

PR #522, opened 2023-08-28
Re-structure NEWS.md

PavelBal:
Remove "develop" branch from appveyor.yml
Do not enforce old observed data naming in the test


Felixmil (approved):
I have concernes about removing `- develop` in the appveyor.yml because it caused appveyor to fail reading the file after it is processed by the `realese-process.R` script. But i'll find a fix.

PR #568, opened 2023-11-06
Throw an error if plotID (used in a plotGrid) is not unique

### LINKED ISSUE #566: `createPlotsFromExcel()`- show an error when duplicated entries for `plotGrids` or `plotConfiguration` exist
If there are multiple rows with the same `name` in the `plotGrids` sheet, `createPlotsFromExcel()` fails here: https://github.com/esqLABS/esqlabsR/blob/72b79e547eb3d632473842e85b1617350d28427c/R/utilities-figures.R#L359

To prevent this, check for duplicates early and exist with an informative message (should include the name of duplicated plot grid).

Felixmil:
fixes #566


PavelBal (changes requested):
Also field `name` in the sheet `plotGrids`

PavelBal (approved):

PR #571, opened 2023-11-07
Update pkgdown.yaml

Felixmil:
fixes https://github.com/esqLABS/esqlabsR/actions/runs/6785325756/job/18443740020

deployment action looked for `docs/` folder while everything is stored in `pkgdown/` folder:

image


svavil (approved):
We still need to test how this works in the main branch, so there could be later fixes.

PR #584, opened 2023-12-05
Improve install packages

Felixmil:
Make sure there is no prompt at the end of renv initialization
+ make sure the local PKSIM is activated when project is loaded


inst/extdata/install-packages.R
```diff
@@ -87,6 +87,11 @@ get_PKSim_Minimal <- function(install_option) {
     # test if PK-Sim.R dll can be loaded
     ospsuite::initPKSim(pksimFolderPath = "PKSim")
 
+    cat('ospsuite::initPKSim(pksimFolderPath = "PKSim")',
```
PavelBal, line 90: What was the reason for this, again?

I need more convicing, I don't like the idea of loading some libraries at project startup. This would also prevent ospsuite package from updating, because it cannot update the dlls once loaded.

Felixmil: You are right, looked convenient but this will confuse users (I'm used to commenting/uncomment Rprofile files but certainly not typical users.) I think we can close

inst/extdata/install-packages.R
```diff
@@ -252,7 +257,7 @@ setup_esqlabsR <- function() {
   initialize_project()
 
   if (install_option == 1) {
-    renv::snapshot(prompt = FALSE) # snapshot environment
+    renv::snapshot(prompt = FALSE, force = TRUE) # snapshot environment
```
PavelBal, line 260: What about this? Should we merge this change only, or do you want to integrate it into another PR?

Issue #497, opened 2023-07-20
Script for updating packages


PavelBal:
A script that updates all osp packages and esqlabsR and updates the PKSim minimal in the project folder.

Felixmil:
Would this script be for dev or users to use ? 
For user we already have: https://github.com/esqLABS/esqlabsR/blob/main/inst/extdata/install-packages.R

for dev I setup this gist that I use regularly: https://gist.github.com/Felixmil/9afed5656a4dec4979267fab7e5bfff6.

PavelBal:
For users. Will have to test if it works as expected.

Felixmil:
To update all osps packages, users just need to run: `remotes::install_github("esqLABS/esqlabsR")`. It will retrieve all latest versions of all.

So what's remaining is a script to download the portable PKSIM to a specific location and overwrite folder/files already there. Would that work for you @PavelBal

PavelBal:
Obsolete

PR #600, opened 2024-02-26
Isolate changes from #584

Felixmil:
Originaly proposed in  #584

Have to force the snapshot if the user has to run the script again after an error.


Felixmil:
Also implements #497

PavelBal (approved):

PR #616, opened 2024-03-28
Use steady-state functions for ospsuite-r

PavelBal:
- Function `exportSteadyStateToXLS` has been removed in favor of `ospsuite::exportSteadyStateToXLS` (#598, @PavelBal)

- Use `getSteadyState` from ospsuite


PavelBal:
The only checks that are failing are for code coverage, which basically did not change..

PR #629, opened 2024-04-09
 `NULL` for arguments `simulatedScenarios` and `observedData` in `createPlotsFromExcel`

### LINKED ISSUE #599: `createPlotsFromExcel()` - `observedData` can be NULL
Allow `observedData` be null, if only sim results are plotted.

PavelBal:
- Fixes #599

- Function `createPlotsFromExcel` supports `NULL` for arguments `simulatedScenarios` and `observedData` if no simulation results or observed data are required for plot generation.


PavelBal:
@Felixmil the build fails because it cannot find function `runSimulation()`, but the function was deprecated and not removed??

Felixmil:
It is also not exported anymore

Felixmil (approved):

Issue #625, opened 2024-04-09
Function signature for `sensitivitySpiderPlot`


PavelBal:
E.g. 

```
sensitivitySpiderPlot(
  sensitivityCalculation,
  outputPaths = NULL,
  parameterPaths = NULL,
  pkParameters = NULL,
DefaultPlotConfiguration
)
```

PavelBal:
https://github.com/esqLABS/esqlabsR/issues/597

rengelke:
sensitivitySpiderPlot(
  sensitivityCalculation,
  outputPaths = NULL,
  parameterPaths = NULL,
  pkParameters = NULL,
  yAxisType = "percent", "absolute"
  xAxisTYpe = "percent", "absolute"   BUT split by ParametrePath
  DefaultPlotConfiguration
)

Issue #619, opened 2024-04-09
SA spider plot: `DefaultPlotConfiguration` properties that must be supported


PavelBal:
List all properties of DefaultPlotConfiguration, that will be supported by the SA spider plot.

Laura-Villain:
Selected list:


linesColor:
 pointsColor:
caption: NULL
xLabel         = "Input parameter value [% of reference]",
  yLabel         = NULL,
  title          = unique(data$OutputPath),
legendPosition = "bottom",
  linesSize      = 1.4,
  pointsSize     = 2,
  xAxisScale     = "log",
  yAxisScale     = "lin",
  yAxisTicks
 
legendTitle: NULL
linesAlpha: 0.75
linesLinetype: NULL
pointsAlpha: 0.75
pointsShape: circle diamond triangle square invertedTriangle cross th ...
subtitle: NULL
xAxisLimits: NULL
yAxisLimits: NULL
xValuesLimits: NULL
yValuesLimits: NULL
xLabel: NULL
yLabel: NULL

Issue #597, opened 2024-02-02
Request for sensitivitySpiderPlot() to allow absolute values - y axis


wilbertdew:
sensitivitySpiderPlot() normalizes the output (y-axis) for the references value and it uses the column "PercentChangePK" from the sensitivity calculation. I would like to be able to choose for the alternative where I can also plot the absolute values as available by the column PKParameterValue in the sensitivity calculation output dataframe.

Along the same lines, it would be helpful (but less critical for me at the moment) to do the same for the x-axis and make the normalization optional. the absolute values are here also already available in the sensitivity calculation output data frame in the column "ParameterValue"

These options can give weird-looking and unhelpful plots if you put multiple parameters at the same time, but they can be helpful to plot the same parameter in different scenarios

PavelBal:
Wilber has a code for this in place.

PavelBal:
How-to in signature?

PR #632, opened 2024-04-17
Refactor spider plot

rengelke:
`sensitivitySpiderPlot` function overhaul with new parameters #625 . Now returns a `patchwork` object with multiple `ggplot2` plots.
 - `yAxisType` to toggle between `percent` and `absolute` values #597 . 
 - `xAxisScale` and `yAxisScale` to set axis scales to `log` or `lin` to be in line with `DefaultPlotConfiguration` options. 
 -  `yAxisFacetScales` to choose between `free` and `fixed` scales for y-axis as requested and to account for different y-axis types. 
  - Plot customization options through `defaultPlotConfiguration` as mentioned in #619


Felixmil:
What are all these forced-push that you do @rengelke  ?

R/sensitivity-spider-plot.R
```diff
@@ -8,91 +8,111 @@
...
-#' @param outputFolder A character string describing path folder where plots
-#'   need to be saved. The path must be relative to the current working
-#'   directory. By default (`""`), the plots will be saved in the current
-#'   working directory (which can be found using `getwd()` function). This
-#'   parameter is relevant only if `savePlots` is set to `TRUE`.
-#' @param width,height Plot size in inches. If not supplied, uses the size of
-#'   current graphics device.
-#' @inheritParams ggplot2::ggsave
+#' @param yAxisType Character string, either "percent" (percentage change) or
+#'   "absolute" (absolute values), for y-axis data normalization. Default is "percent".
```
PavelBal, line 12:
```suggestion
#'   "absolute" (absolute parameter values), for y-axis data normalization. Default is "percent".
```

R/sensitivity-spider-plot.R
```diff
@@ -8,91 +8,111 @@
...
+#'   - `legendTitle`: Title displayed for the legend.
+#'   - `linesAlpha`: Alpha transparency for line elements.
+#'   - `linesColor`: Color of the line elements.
+#'   - `linesSize`: Thickness of the line elements.
+#'   - `pointsShape`: Shape of the point elements.
+#'   - `pointsSize`: Size of the point elements.
+#'   - `subtitle`: Subtitle text for the plot.
+#'   - `title`: Main title text for the plot.
+#'   - `titleSize`: Font size of the plot title.
+#'   - `xAxisScale`: Scale type for the x-axis (`"log"` or `"lin"`).
```
PavelBal, line 33: What happens if `xAxisScale` is defined in in functions signature? Which value wins?

rengelke: Currently, defaultPlotConfiguration$xAxisScale, if modified and therefore not `NULL`, would override the settings in the function signature in `.updatePlotConfiguration`. I see that this might be confusing. 
We should document that any modifications to `defaultPlotConfiguration` will take precedence over the plot specific settings. Or discuss a better implementation.

R/sensitivity-spider-plot.R
```diff
@@ -8,91 +8,111 @@
...
+                                  yAxisScale = "lin",
+                                  yAxisFacetScales = "fixed",
+                                  defaultPlotConfiguration = NULL) {
   # input validation ------------------------
 
   # fail early if the object is of wrong type
   validateIsOfType(sensitivityCalculation, "SensitivityCalculation")
-  validateIsCharacter(outputFolder, nullAllowed = FALSE)
+  if (is.null(defaultPlotConfiguration)) {
+    defaultPlotConfiguration <- esqlabsR::createEsqlabsPlotConfiguration()
```
PavelBal, line 115: No need to specify `esqlabsR::` when used within the package.

R/sensitivity-spider-plot.R
```diff
@@ -107,104 +144,240 @@ sensitivitySpiderPlot <- function(sensitivityCalculation,
...
+
+  plotPatchwork <- patchwork::wrap_plots(plotList) +
+    patchwork::plot_annotation(
+      title = plotConfiguration$title,
+      subtitle = defaultPlotConfiguration$subtitle,
+      theme = theme(
+        plot.title = element_text(size = plotConfiguration$titleSize)
       )
+    ) +
+    patchwork::plot_layout(
```
PavelBal, line 334: `guides`, not `guide`

tests/testthat/test-sensitivity-calculation.R
```diff
@@ -389,22 +388,47 @@ test_that("sensitivitySpiderPlot plots are as expected", {
...
+  p <- sensitivitySpiderPlot(results, yAxisType = "absolute")
+  pbs <- purrr::map(seq_along(p[[n]]), ~ ggplot2::ggplot_build(p[[n]][[.x]]))
+  plotParams <- purrr::map(pbs, ~ .x$layout$panel_params[[1]]$y.range)
+
+  expect_snapshot(unlist(plotParams))
 })
 
-test_that("sensitivitySpiderPlot saves plot file", {
-  path <- "Spider_OutputPath1.png"
+test_that("sensitivitySpiderPlot correctly applies free scaling with absolute y-values", {
```
PavelBal, line 413: Ah... please create plot snapshots for new test cases.

R/sensitivity-spider-plot.R
```diff
@@ -107,104 +146,240 @@ sensitivitySpiderPlot <- function(sensitivityCalculation,
...
+          m = plotConfiguration$yAxisTicks
+        )
+        pLimits <- NULL
+      }
+
+      if (!is.null(plotConfiguration$yAxisLimits)) {
+        pLimits <- plotConfiguration$yAxisLimits
+      }
+
+      plot <- ggplot(
```
PavelBal, line 246: Please add comments (where it makes sense) what is being done in the next step. It will make debugging and adjusting easier.

R/zzz.R
```diff
@@ -1,44 +1,11 @@
 # nocov start
-.onLoad <- function(...) {
-  options(warnPartialMatchDollar = TRUE)
-  Sys.setenv("_R_CHECK_LENGTH_1_CONDITION_" = "true")
+esqlabsEnv$colorPalette <- .getEsqlabsColors()
```
PavelBal, line 2: @Felixmil Is it safe to do any assignments outside of `.onLoad`?

R/sensitivity-spider-plot.R
PavelBal: @rengelke Last to-do ;)

rengelke: Yes! :) Now the comments are quite detailed.

R/zzz.R
Felixmil: Yes it's safe. Two reasons why it's here:
1. We need it to set this in `zzz.R` so .getEsqlabsColors() has all internal function available when this is available.
2. We cannot define it **inside** .onLoad() because that function do not have access to `esqalabsEnv`

So it is in `zzz.R` but not in `.onLoad()`

PavelBal (approved):

Issue #634, opened 2024-04-18
rename `createDefaultProjectConfiguration` to `createProjectConfiguration`


Felixmil:
I don't think `Default` is necessary.

- [x] create `createProjectConfiguration()`
- [x] deprecate `createDefaultProjectConfiguration()` with warning and pass to `createProjectConfiguration`

PavelBal:
Yes

PavelBal:
https://github.com/esqLABS/esqlabsR/pull/635

Issue #633, opened 2024-04-18
Harmonize `Configuration` excel file names


Felixmil:
Some ends with `Parameters`, others don't:
![image](https://github.com/esqLABS/esqlabsR/assets/34234913/dec32729-8b80-4d0c-a905-eafd28848430)

Some are singular, other plural also.

Shall we harmonize to:
```sh
Configurations/
- Applications
- Individuals
- ModelParameters
- Plots
- Populations
- PopulationsCSV/
- Scenarios
```

PavelBal:
https://github.com/esqLABS/esqlabsR/pull/635

Issue #631, opened 2024-04-15
Rename `Parameters` folder into `Configurations`?


PavelBal:
As by now the files within this folder contain much more than just parameters?

PavelBal:
@Felixmil @dlefaudeux @Laura-Villain What do you think?

Felixmil:
Either that or `SimulationSettings` or something similar

PavelBal:
No, `SimulationSettings` has a well-defined meaning in OSP software and does not include parametrization.

PavelBal:
https://github.com/esqLABS/esqlabsR/pull/635

Issue #558, opened 2023-10-25
Rename `projectConfiguration$paramsFile` by `projectConfiguration$modelParamsFile`


Felixmil:

Felixmil:
No direct complaints. But since we will already break the `PopulationParams.xlsx` file  path in the next release, better harmonize now.

PavelBal:
Okay then.

PavelBal:
https://github.com/esqLABS/esqlabsR/pull/635

PR #635, opened 2024-04-18
Improve Project Structure

Felixmil:
Fixes:
- #558 
- #631 
- #633 
- #634 
- #637


PavelBal:
Make the PR to another branch pls

PavelBal (changes requested):
- Remove `compoundPropertiesFile` from the project configuration excel (is not used anyways)
- Add required steps to https://github.com/esqLABS/esqlabsR/issues/627

R/esqlabs-plot-export-configuration.R
```diff
@@ -1,14 +1,5 @@
 #' @title ExportConfiguration
 #' @description R6 class defining properties for saving a `ggplot` object
-#' @field name character defining the name of the file to be saved (without extension)
```
PavelBal, line 3: Why remove?

R/project-configuration.R
```diff
@@ -38,73 +38,82 @@ ProjectConfiguration <- R6::R6Class(
...
         value <- private$.populationsFile
       }
       private$.populationsFile <- value
-      private$.clean_path(value, self$paramsFolder)
+      private$.clean_path(value, self$configurationsFolder)
     },
-    #' @field scenarioDefinitionFile Name of the excel file with scenario
+    #' @field populationsCSV Name of the folder containing population defined through csv files
+    #' Must be located in the "configurationsFolder".
+    populationsCSV = function(value) {
```
PavelBal, line 82: Name it `populationsCSVfolder`?

PavelBal (changes requested):

R/esqlabs-plot-export-configuration.R
Felixmil: Because these are not arguments for initialize method.

PavelBal: They are, the `initialize` method is just defined in `TLF`. I copied the documentation here because otherwise Roxygen complained. But if it works without, fine with me (even better so, I didn't like this duplication)

R/project-configuration.R
PavelBal: Yes

R/esqlabs-plot-export-configuration.R
```diff
@@ -1,14 +1,5 @@
...
-#' @field path Path of the directory to save plot to: path and filename are
-#' combined to create the fully qualified file name. Defaults to the working directory.
-#' @field format character defining the format of the file to be saved
-#' @field width numeric values defining the width in `units` of the plot dimensions after saving
-#' @field height numeric values defining the height in `units` of the plot dimensions after saving.
```
PavelBal, line 8: Please explicitly document `height` as this is the change as compared with the default implementation @Felixmil

Felixmil: Reworked that part a bit.

PavelBal (approved):

Issue #643, opened 2024-04-22
`sensitivitySpiderPlot` - axis scaling settings


PavelBal:
Right now, axis scaling can be defined in the signature AND in the plot configuration. Discuss with users what would be the best approach:

1) value in signature overrides configuration
2) value in configuration overrides signature
3) remove axis scaling setting from the signature
4) else

rengelke:
@Laura-Villain @dlefaudeux 
Thank you for the constructive discussion regarding the implementation of axis scaling. 

As discussed, we will keep the ability to define axis scaling directly in the function signature, as it is an important parameter which should be accessed easily without the need to set it up in `defaultPlotConfiguration`. 

In summary, `xAxisScale`and `yAxisScale` parameter settings will follow this hierarchy:
1.  Plotting function signature
2. `defaultPlotConfiguration`
3. specific plot configuration set up inside the plotting function

Issue #642, opened 2024-04-19
`sensitivitySpiderPlot` - settings from `defaultPlotConfiguration` not applied


PavelBal:
```
  simPath <- system.file("extdata", "Aciclovir.pkml", package = "ospsuite")
  simulation <- loadSimulation(simPath)
  outputPaths <- "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)"
  parameterPaths <- c(
    "Aciclovir|Lipophilicity",
    "Applications|IV 250mg 10min|Application_1|ProtocolSchemaItem|Dose",
    "Neighborhoods|Kidney_pls_Kidney_ur|Aciclovir|Glomerular Filtration-GFR|GFR fraction"
  )

  results <- sensitivityCalculation(
    simulation = simulation,
    outputPaths = outputPaths,
    parameterPaths = parameterPaths
  )

myPlotConfiguration <- createEsqlabsPlotConfiguration()
myPlotConfiguration$legendPosition <- tlf::LegendPositions$outsideTop

sensitivitySpiderPlot(results, defaultPlotConfiguration = myPlotConfiguration)
```

Not applied:

![image](https://github.com/esqLABS/esqlabsR/assets/25903040/7a311a2b-5880-4eef-bf64-98b1a0f931a7)

PavelBal:
Also:

- `legendTitle`
- `title`

I think all the entries that are specified here:

```
  spiderPlotConfiguration <- list(
    legendPosition = "bottom",
    legendTitle    = "Parameter",
    linesSize      = 1.4,
    pointsShape    = 21,
    pointsSize     = 2,
    title          = unique(data$OutputPath),
    titleSize      = 14L,
    xAxisScale     = xAxisScale,
    xLabel         = "Input parameter value [% of reference]",
    yAxisScale     = yAxisScale,
    yAxisTicks     = 10L,
    yLabel         = NULL
  )
```

PR #648, opened 2024-04-29
Set and validate plot configuration

rengelke:
- Fix `.updatePlotConfiguration` function to ensure that settings from `defaultPlotConfiguration` are correctly applied and that parameters passed directly through the function signature take precedence (#642 #643 ). 
-  introduces validation of plot configuration settings against a predefined list of valid options to prevent configuration errors


PavelBal:
Remove console output for the path:

```
> sensitivitySpiderPlot(results, xAxisScale = "log")
$`Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)`
```

PavelBal (changes requested):
Remove console output for the path:

Issue #609, opened 2024-03-07
SA: Tornado plot for multiple parameters


PavelBal:
Only makes sense if multiple parameters are varied.

![image](https://github.com/esqLABS/esqlabsR/assets/25903040/a41feebb-f8c4-4114-8954-0b457fc9ee98)

PavelBal:
Or like this:



![Image](https://github.com/esqLABS/esqlabsR/assets/25903040/62ebdecc-e501-4cc9-b37a-e849dea3903c)

PavelBal:
Define signature.

Laura-Villain:
TornadoPlot(simulation,
outputPaths,
parameterPaths = list(), #more than 1 
variationAmplitude = 0.1
pkParameters = list("C_max", "t_max", "AUC_inf", "my function" = function), (warning: must be a named list)
saOutputFilePath = NULL,
simulationRunOptions = NULL,
DefaultPlotConfiguration)

PavelBal:
Does it not make sense to create these plots from the `sensitivityCalculation` as the other two plot functions do, instead running a separate SA?

rengelke:
Both is possible and we have discussed it in detail -  the main argument for running a separate SA was that the `variationAmplitude` might not be present in the `senstitivityCalculation`.
However, there are drawbacks to this too: many parameters, relatively slow, not consistent signature compared to other plotting functions.

Run it from `sensitivityAnalysis` with `variationAmplitude` limited to data in the results?

PavelBal:
What is `variationAmplitude`?

dlefaudeux:
Tornado plots are done with a specified scaling factor value, something similar to "variationRange" from the `senstitivityCalculation` but for a single value, so we suggested "variationAmplitude". We could also choose a different argument name, if this one is not explicit enough. 

I don't think there is a specific name for it, for example the [tornado ](https://cran.r-project.org/web/packages/tornado/vignettes/tornadoVignette.html) package (tornado plots on linear regression models) has different types (percentChange, ranges, percentiles), and the amount of changes or chosen percentiles is called "alpha".

PR #652, opened 2024-05-06
Add tornado plot

rengelke:
New sensitivityTornadoPlot function #609  to visualize the results of sensitivity analysis.
Each plot shows the effect of modifying parameters by a specific scaling factor
(`parameterFactor`) and its reciprocal on specific model outputs. This
visualization helps to assess the impact of parameter changes on the results,
highlighting the model's sensitivity to these parameters.

@param `sensitivityCalculation` The `SensitivityCalculation` object returned by
`sensitivityCalculation()`.
@param `outputPaths`, `parameterPaths`, `pkParameters` A single or a vector of the
output path(s), parameter path(s), and PK parameters to be displayed,
respectively. If `NULL`, all included paths and parameters present in the
supplied `SensitivityCalculation` object will be displayed in the
visualization.
@param `parameterFactor` Numeric; the scaling factor used to adjust parameters
during sensitivity analysis used in the tornado plot. Both the `parameterFactor`
and its reciprocal (`1/parameterFactor`) must be included in the
`variationRange` specified in the `sensitivityCalculation`. Default is 0.1.
@param `defaultPlotConfiguration` An object of class `DefaultPlotConfiguration`
used to customize plot aesthetics.


R/sensitivity-tornado-plot.R
```diff
@@ -0,0 +1,304 @@
...
+        )
+      }
+
+      return(plot)
+    }
+  )
+
+  # compile individual plots -----------------
+
+  plotPatchwork <- patchwork::wrap_plots(plotList, ncol = 1) +
```
Felixmil, line 276: Why use patchwork if you already use facet_wrap() ?

R/sensitivity-tornado-plot.R
```diff
@@ -0,0 +1,304 @@
...
+      plot <- ggplot(
+        dataSubset,
+        aes(
+          x = ParameterPath,
+          y = PercentChangePK,
+          fill = as.factor(ParameterFactor)
+        )
+      ) +
+        geom_col(
+          color = "grey",
```
Felixmil, line 226: I suggest color = "white" or map it to ParameterFactor

R/sensitivity-tornado-plot.R
```diff
@@ -0,0 +1,304 @@
...
+          na.rm = TRUE
+        ) +
```
Felixmil, line 228: add fixed `width` in geom_col

R/utilities-figures.R
```diff
@@ -651,6 +651,10 @@ createPlotsFromExcel <- function(
       type = "numeric",
       valueRange = c(0.1, 10)
     ),
+    parameterFactor = list(
+      type = "numeric",
+      valueRange = c(0.01, 0.99)
+    ),
```
dlefaudeux, line 657: Am I understanding that correctly? This will ensure that the parameterFactor is within 0.01 and 0.99. 

I would tend to think that it should not matter if the user give for example 0.1 or 10, it should just lead to the same results and plots.

Also do we really need to force a range for the input parameterFactor? I don't think there is a limit for the sensitivityCalculation

R/sensitivity-tornado-plot.R
```diff
@@ -0,0 +1,304 @@
...
+  # default tornado plot configuration setup ----
+
+  tornadoPlotConfiguration <- list(
+    legendPosition = "right",
+    legendTitle = "Parameter Factor",
+    subtitle = NULL,
+    title = NULL,
+    titleSize = 14,
+    xLabel = "Parameter",
+    yLabel = "Input parameter value [% of reference]"
```
dlefaudeux, line 124: I think this will be confusing, because of the cord_flip, the yLabel ends up in the final x-axis. Can we also flip the label such that what the user give as xAxis is the final xAxis. I know this is the default but it would also apply to the given plot configuration.

dlefaudeux, line 124: Also the yLabel (final x-label) should be the "PK-Parameter value [% of reference]"

R/sensitivity-tornado-plot.R
rengelke: I switched to using patchwork in `sensitivitySpiderPlot()` as required parameters couldn't be implemented with previous code using `facet_wrap()`. Now I apply facets to a single factor to keep a consistent look and preserve facet headers. This approach is now uniform across all plotting functions.

R/sensitivity-tornado-plot.R
rengelke: I switched from "white" to "grey" for bar border color for the case where the values to better visualize input parameter around 0. In this case, if border color is "white" the plot appears empty, while "grey" looks like there is actually a bar which is close to zero.

R/sensitivity-tornado-plot.R
rengelke: Set width = 0.9. However, this will not control the absolute widths of bars across different plots, as these are influenced by the overall dimensions of the graphics device and the plot layout settings.

R/utilities-figures.R
rengelke: Extended to valueRange = c(1e-16, 1e16)

R/sensitivity-tornado-plot.R
rengelke: Indeed, this is confusing. Swapped labels as suggested

PavelBal:
@Laura-Villain @Felixmil @dlefaudeux Please review until Monday so @rengelke can proceed. Two reviews will be sufficient.

Felixmil (approved):

R/sensitivity-tornado-plot.R
```diff
@@ -0,0 +1,304 @@
...
+          y = PercentChangePK,
+          fill = as.factor(ParameterFactor)
+        )
+      ) +
+        geom_col(
+          color = "grey",
+          na.rm = TRUE
+        ) +
+        coord_flip()
+
```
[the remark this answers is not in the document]
PavelBal: Yws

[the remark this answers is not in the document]
rengelke: Mentioning this, you reminded me to make it more symmetric as well

PR #672, opened 2024-05-27
- Add new function `writeParameterStructureToXLS`

### LINKED ISSUE #670: New function `writeParametersToXLS`
`{esqlabsR}` has a function `readParametersFromXLS()` that reads parameter sets from an Excel file as supported by MoBi and returns a list containing vectors 'paths' with the full paths to the parameters, 'values' the values of the parameters, and 'units' with the units the values are in. This is the default "Parameter structure" used across the package.

There is currently no function that would do the opposite - write such a "parameter structure" into excel. The function `exportParametersToXLS` write `Parameter` objects, but not the list structure.

Proposed name for the new function `writeParameterStructureToXLS`? And then rename `readParametersFromXLS` to `readParameterStructureFromXLS` and deprecate `readParametersFromXLS` for a while?

@Felixmil @dlefaudeux @Laura-Villain what do you think? Need your input on usability.

PavelBal:
Fixes #670

It still lacks the implementation of appending to a file instead of overwriting, and maybe some tests. I would ask you to do this.


R/utilities-parameters.R
```diff
@@ -39,6 +39,55 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
...
 }
 
+#' Write parameter structure to excel that can be loaded in MoBi
+#'
+#' @param parameterStructure A list containing vectors 'paths' with the full paths to the parameters,
+#' 'values' the values of the parameters, and 'units' with the units the values are in.
+#'
+#' @param paramsXLSpath Path to the excel file
+#' @param sheet (Optional) name of the excel sheet
+#' @param extend If TRUE, the existing excel file will be extended with the new
```
PavelBal, line 49: Not implemented yet!

R/utilities-parameters.R
```diff
@@ -39,6 +39,55 @@ readParametersFromXLS <- function(paramsXLSpath, sheets = NULL) {
...
+  )
+
+  # Write the results into an excel file.
+  # Wrap the output data frame into a list and name the list if sheet name
+  # has been provided
+  data <- list(output)
+  if (!is.null(sheet)) {
+    names(data) <- sheet
+  }
+  writeExcel(data = data, path = paramsXLSpath)
```
PavelBal, line 88: Make `writeExcel` internal? I think (and hope) that no one uses it directly.

Felixmil: done in a9e347a

R/utilities-parameters.R
Felixmil: done in 964f738

Felixmil (approved):

NAMESPACE
```diff
@@ -67,19 +67,15 @@ export(sourceAll)
 export(startFunctionVisualizer)
 export(startUnitConverter)
 export(stringToNum)
-export(writeExcel)
```
PavelBal, line 70: Add note to NEWS (breaking change?)

R/utilities-file.R
```diff
@@ -66,7 +66,6 @@ readExcel <- function(path, sheet = NULL, ...) {
 #' @param data Data frame or named list of data frames that will be sheets in
 #' the xlsx
 #' @inheritParams writexl::write_xlsx
-#' @export
 writeExcel <- function(data, path, col_names = TRUE) {
```
PavelBal, line 69: prepend `.` to the name to stay consistent with naming conventions.

Felixmil: done in a2f2994

NAMESPACE
Felixmil: done in a2f2994

PR #676, opened 2024-06-06
Fix pksim installation path

Felixmil:
Something was wrong in the documentation 😅

### DIFF
--- a/vignettes/install-pksim.Rmd
+++ b/vignettes/install-pksim.Rmd
@@ -49,8 +49,8 @@ version of PK-Sim.
 
 ## Extract PK-Sim
 
-Extract the downloaded portable PK-Sim to `C:/Program Files/Open Systems
-Pharmacology/PK-Sim 12.0`.  
+Extract the downloaded portable PK-Sim to this location `C:/Open Systems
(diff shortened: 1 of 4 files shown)


rengelke (approved):
Nothing to add

vignettes/install-pksim.Rmd
```diff
@@ -69,17 +69,12 @@ unzip("pksim_minimal.zip", exdir = "PKSim")
 file.remove("pksim_minimal.zip")
 ```
 
-Then, include the following code lines at the begining of the projects scripts:
+Then, include the following code lines at the beginning of the projects scripts:
 
 ```r
 ospsuite::initPKSim(pksimFolderPath = "PKSim")
```
PavelBal, line 75: This only applies for esqlabs-projects, when minimal PK-Sim is located in the `Code` folder.

Felixmil: Well, it can also apply for self-contained project that are not esqlabs project.
this is why it's under the section "# Install and use PK-Sim in a project"

PavelBal: Ah yes, I am always cofused when GitHub collapses parts of the code..

Issue #623, opened 2024-04-09
Function signature for `sensitivityCalculation`


PavelBal:
Define inputs and outputs.

- https://github.com/esqLABS/esqlabsR/issues/621
- https://github.com/esqLABS/esqlabsR/issues/622

Laura-Villain:
sensitivityCalculation(
  simulation,
  outputPaths,
  parameterPaths,
  variationRange = list(range1,range2),  (same size than parameterPaths)
variationType = list("absolute","relative"), (same size than parameterPaths)
  pkParameters = list("C_max", "t_max", "AUC_inf", "my function" = function), (warning: must be a named list)
  saOutputFilePath = NULL,
  simulationRunOptions = NULL
)

PavelBal:
Is `variationRange` a list of numeric vectors, e.g., `list(c(0.5, 1, 2), c(1,2,3))`? Does this mean that different variation parameters might have different variation ranges? Yes, doable, we can support this. We should pay attention to how this will be represented in the plots.

Then it should be possible to provide *one* vector that will be applied to all parameters. So `variationRange = c(1,2,3)` means that this varation range (or absolute values) will be applied for all parameters. Otherwise, `variationRange` must be of the same length as `parameterPaths`.

Same applies to `variationType`.

Use `ospsuite.utils::toList` to ensure that `variationRange` and `variationType` are lists (and check for their length).

I propose to have the custom functions not as part of the `pkParameters` argument (because the function is not a PK-parameter), but as a separate argument `customOutputFunctions = NULL`. Must be a named list (check for names), each group entry must be a function with arguments `SimulationResults, outputPaths`.

PR #685, opened 2024-07-23
Allow user defined pk parameters

### LINKED ISSUE #205: Sensitivity analysis with user defined pk parameters
"Following non-standard PK parameters will not be calculated:
olol"

Check what exactly happens, this should be supported.

rengelke:
sensitivityCalculation with new parameter `customOutputFunctions`

@param customOutputFunctions A named list with custom function(s) for PK parameter calculation. User-defined functions should have either 'x', 'y', or both 'x' and 'y' as parameters which correspond to x-Dimension (time) or y-Dimension values from simulation results.

Related to #623 
Closes #205


R/utilities-sensitivity-calculation.R
```diff
@@ -2,49 +2,221 @@
...
+#' This function calculates standard PK analyses from simulation results and
+#' converts them into a dataframe. If custom output functions are provided,
+#' it calculates user-defined PK analyses and integrates them with the standard
+#' PK data.
 #'
-#' @inheritParams .simulationResultsToTimeSeriesDataFrame
+#' @param simulationResults List of simulation results from which PK analyses
+#' are to be calculated.
+#' @param parameterPath Path to the parameter in the simulation results.
+#' @param customOutputFunctions Optional list of custom output functions for
```
Felixmil, line 38: Can it work with anonymous functions ? And with `~` syntax  (e.g ~x+2)

rengelke: Yes, we already can pass it as a list of anonymous functions
`list("minmax" = function(y) min(y[y != 0]) / max(y))`

Purrr specific `~` syntax for anonymous functions could be implemented and translated using `purrr::as_mapper` or `rlang::as_function`

Felixmil: If other function syntax are supported, corresponding tests should be added.

There also the shorthand `\(x)`: `\(x) x + 1` ➡️ `function(x) x + 1` (cf [this](https://stat.ethz.ch/R-manual/R-devel/library/base/html/function.html))

rengelke: So far only classical function syntax is supported

R/utilities-sensitivity-calculation.R
```diff
@@ -2,49 +2,217 @@
...
+#'
+#' @return A dataframe containing the custom PK values.
+#'
+#' @keywords internal
+#' @noRd
+.calculateCustomPK <- function(simulationResults, customOutputFunctions) {
+  # validate customOutputFunctions
+  validateIsOfType(customOutputFunctions, "list", nullAllowed = TRUE)
+  validateIsOfType(customOutputFunctions, "function", nullAllowed = TRUE)
+  validateHasOnlyNonEmptyStrings(names(customOutputFunctions))
```
Felixmil, line 126: I would maybe create a new validate... function in {esqlabsR} for checking if lists are named. Because the message returned by this validation will not be clear to the user: 
```
> customOutputFunctions <- list(1, 2, 3)
> ospsuite.utils::validateHasOnlyNonEmptyStrings(names(customOutputFunctions))
Error in ospsuite.utils::validateHasOnlyNonEmptyStrings(names(customOutputFunctions)) : 
  ospsuite.utils::validateHasOnlyNonEmptyStrings: argument 'names(customOutputFunctions)' has empty strings!
```

We could return a more explicit message: `customOutputFunctions is not a named list` for example

Felixmil (commented):
More explicit error when providing unnamed list ?

Felixmil (approved):

NEWS.md
```diff
@@ -26,6 +26,9 @@
     multiple `DataSet` objects) can now be added to the plots and will be
     included automatically in the `outputPath` if the dimensions are convertible
     (\#674).
+    
+  - Function `sensitivityCalculation()` supports new parameter `customOutputFunctions`, 
+    allowing user-defined functions for PK parameter calculation (\#685).
```
PavelBal, line 31: > allowing user-defined functions for PK parameter calculation

"allowing user-defined functions for outputs to be evaluated.

R/sensitivity-calculation.R
```diff
@@ -16,11 +16,17 @@
 #'   the following parameters will be considered: `"C_max"`, `"t_max"`,
 #'   `"AUC_inf"`. If `NULL`, all available PK-parameters will be calculated. You
 #'   can also specify custom PK parameters.
+#' @param customOutputFunctions A named list with
+#' custom function(s) for PK parameter calculation. User-defined functions should
```
PavelBal, line 20: "A named list with custom function(s) for output calculation".

R/sensitivity-calculation.R
```diff
@@ -16,11 +16,17 @@
...
+#' have either 'x', 'y', or both 'x' and 'y' as parameters which correspond to
+#' x-Dimension (time) or y-Dimension values from simulation results.
```
PavelBal, line 22: Add a comment on what should be the output of the function (a single numerical value, I guess?).

R/sensitivity-calculation.R
```diff
@@ -16,11 +16,17 @@
...
+#' @param ... Additional parameters passed to the function.
```
PavelBal, line 23: The ellipsis (`...`) is part of the `customOutputFunctions` signature? The way it is written here, it is part of the `sensitivityCalculation`  function, which is not correct. Am I right?

PavelBal (commented):
Also please add an example in the function documentation and additionally an example in the vignette.

Or create a branch and I will add an example in the vignette.

PavelBal:
@rengelke I added some comments after the PR was merged.

rengelke:
> @rengelke I added some comments after the PR was merged.

Addressed in PR #720

PR #688, opened 2024-07-26
Change units in SA time profiles

### LINKED ISSUE #610: SA: Time profiles - change units
Best support can be achieved when using the plotIndividualTimeProfile() from ospsuite-r

rengelke:
Feature to change units in sensitivityTimeProfiles()
Closes #610 

@param xUnits, yUnits  Lists of units for the x-axis and y-axis, respectively. If a list of length one is provided, it will be applied to all `outputPaths` if conversion is possible. If a list of multiple units is provided, the units list should correspond to the `outputPaths`, and units conversion will be applied accordingly. If `NULL`, default units from the simulation results will be used.


Felixmil (commented):
Do you thinkg it possible to add tests to improve coverage on the new code ? There are 3 warnings from codecov.

rengelke:
> Do you thinkg it possible to add tests to improve coverage on the new code ? There are 3 warnings from codecov.

Added tests; 1 codecov warning should be gone. The remaining 2 will be fixed in a separate PR.

R/sensivitity-time-profiles.R
```diff
@@ -499,3 +515,82 @@ sensitivityTimeProfiles <- function(sensitivityCalculation,
...
+  unitsConvertable <- TRUE
+
+  tryCatch(
+    {
+      convertUnits(dataCombined, xUnit = xUnit, yUnit = yUnit)
+    },
+    error = function(e) {
+      unitsConvertable <<- FALSE
+    }
+  )
```
Felixmil, line 539: Another approach, FYI:

```r
ans <- try(
  {
    convertUnits(dataCombined, xUnit = xUnit, yUnit = yUnit)
  },
  silent = TRUE
)

if (inherits(ans, "try-error")) unitsConvertable <- FALSE
```

Felixmil (approved):

R/sensivitity-time-profiles.R
rengelke: Yes, it's an alternative

PR #691, opened 2024-07-30
Improved dynamic axis breaks, limits and labeling

rengelke:


R/utilities-figures.R
```diff
@@ -699,11 +703,18 @@ createPlotsFromExcel <- function(
 #'
 #' @keywords internal
 #' @noRd
-.calculateLimits <- function(x) {
-  limits <- c(
-    (if (min(x, na.rm = TRUE) <= 0) 1.01 else 0.99) * min(x, na.rm = TRUE),
-    (if (max(x, na.rm = TRUE) > 0) 1.01 else 0.99) * max(x, na.rm = TRUE)
-  )
+.calculateLimits <- function(x, scaling = NULL) {
```
Felixmil, line 706: Is this function used to add "blank" area below and above data points ?

R/utilities-figures.R
```diff
@@ -699,11 +703,18 @@ createPlotsFromExcel <- function(
...
+    limits <- c(
+      min(x[x > 0], na.rm = TRUE) * 0.9,
+      max(x[x > 0], na.rm = TRUE) * 1.1
+    )
+  } else {
+    limits <- c(
+      (if (min(x, na.rm = TRUE) <= 0) 1.01 else 0.99) * min(x, na.rm = TRUE),
+      (if (max(x, na.rm = TRUE) > 0) 1.01 else 0.99) * max(x, na.rm = TRUE)
+    )
+  }
```
Felixmil, line 717: Is this function used to add "blank" area below and above data points ?
If yes, I would suggest to use [the expand argument](https://ggplot2.tidyverse.org/reference/expansion.html)
instead of the limit one.

R/utilities-figures.R
```diff
@@ -674,8 +674,12 @@ createPlotsFromExcel <- function(
 .calculateBreaks <- function(x, scaling = NULL, ...) {
```
Felixmil, line 674: I would have used {scales} that contains functions that already do this kind of stuff.
cf:
- https://scales.r-lib.org/reference/label_log.html
- https://scales.r-lib.org/reference/breaks_log.html

Felixmil (commented):
Plots have really improved ! However, I would evaluate the possibility to reuse some of the functions from {scales}. Also, I think it's a good opportunity to  take a look at scales argument in ggplot to make our code a bit less "hacky"

R/utilities-figures.R
rengelke: Good suggestion - didn't think about that. I have now replaced the existing functionality with label_log() and breaks_log() from the scales package.

R/utilities-figures.R
rengelke: The function is used to pre-calculate limits, which are necessary for assembling plots into facets. Limits also help with filtering data for log-transformed axis plotting to avoid the ggplot log transform warning message. Tried `scales::expand_range`, but it did not produce the desired result.

Felixmil (approved):

PR #692, opened 2024-07-30
New project structure

### LINKED ISSUE #721: Mention the get started vignette earlier in the README

### LINKED ISSUE #735: install-packages.R - remove "option 2" (global installation)
I think by now "global installation" via the script is not required. Remove this option (which currently leads to confusion by some users).

- Update WI on BlueSpice

Felixmil:
Generally improved, harmonized and easier to use project structure
closes #721 
closes #735


Felixmil:
@PavelBal maybe include this in the intermediate release before new project workflow ?

PavelBal:
@Felixmil No, the intermediate release should be backward-compatible and work with the current projects, so the users can get the benefits of the new fixes/features without having to change the project structure (e.g., renaming Excel files) in the middle of their projects.

R/project-configuration.R
```diff
@@ -32,154 +32,228 @@ ProjectConfiguration <- R6::R6Class(
...
+      }
+
+      # If one of the excel configuration is not expected, return an error.
+      for (property in names(data)) {
+        if (!(property %in% names(self))) {
+          cli::cli_abort(c("x" = "Property {property} is not a valid configuration property for {self$projectConfigurationFilePath}"))
+        }
+      }
+      private$.projectConfigurationData <- data
+    },
```
Felixmil, line 234: Ensure the new project structure is backward compatible.

Felixmil:
@PavelBal This PR also supports previous Excel configuration and provide a function to update from "old" structure to the new one. (see comment above)

PavelBal:
Okay I will test it with a project first ;)

PavelBal:
> provide a function to update from "old" structure to the new one. (see comment above)

Where? could not find it, and there is also no mentioning of it in the NEWS (or did I miss it?)

Maybe provide a description of what was changed in the README?

Also the vignettes must be updated accordingly!

Felixmil:
> Where? could not find it, and there is also no mentioning of it in the NEWS (or did I miss it?)

The way it works is that you can still import previous project file structure (and display a warning that it is old but still supported). It will internally be converted to the new one. To convert to new structure, use just has to then export using the `$save()` method

All of this is explained when user import old project configuration:
https://github.com/esqLABS/esqlabsR/pull/692/files#diff-df7e0ee8879e22d10aea0afb4a44c8e145f3ad6e6d0534fcc8016fa4336ae5c1R205-R209


https://github.com/esqLABS/esqlabsR/blob/521077f992e0db6957632108a596221b54fbc21e/R/project-configuration.R#L205-L208C27

> Maybe provide a description of what was changed in the README?

It's described in the NEWS:

https://github.com/esqLABS/esqlabsR/pull/692/files#diff-51920e95310ebfbc1ae31709f3b95f89afffbf4f1a6e38e8b2b406e2fb6197eaR11-R23

> Also the vignettes must be updated accordingly!

All vignettes were updated:
image

PavelBal:
> All of this is explained when user import old project configuration:

Ah okay, it wasn't shown because I was running it from Quarto.

R/project-configuration.R
```diff
@@ -222,25 +296,27 @@ ProjectConfiguration <- R6::R6Class(
...
+      private$printLine("Plots", fs::path_rel(as.character(self$plotsFile)))
+      private$printLine("Data folder", fs::path_rel(as.character(self$dataFolder)))
+      private$printLine("Data file", fs::path_rel(as.character(self$dataFile)))
       private$printLine("Data importer configuration", fs::path_rel(as.character(self$dataImporterConfigurationFile)))
-      private$printLine("Compound Properties File", fs::path_rel(as.character(self$compoundPropertiesFile)))
       private$printLine("Output folder", fs::path_rel(as.character(self$outputFolder)))
       invisible(self)
     },
     #' @description Export ProjectConfiguration object to ProjectConfiguration.xlsx
+    #' @param path A string representing the path to the exported file.
```
PavelBal, line 314: Add that the path is relative to the current wd.

PavelBal (changes requested):

PavelBal:
> It's described in the NEWS:

Hmm this is not sufficient. Describe a workflow of how to convert an existing project structure to the new one.

- Which folders and files to rename, which files to move to where...

DESCRIPTION
```diff
@@ -69,5 +69,4 @@ RoxygenNote: 7.3.2
 Remotes: 
     ospsuite.utils=Open-Systems-Pharmacology/OSPSuite.RUtils,
     tlf=Open-Systems-Pharmacology/TLF-Library,
-    ospsuite=Open-Systems-Pharmacology/OSPSuite-R,
-    Open-Systems-Pharmacology/ospsuite.parameteridentification
```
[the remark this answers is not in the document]
Felixmil: We are not using it anywhere in the package so it is not a dependency.

PavelBal:
> To export to new structure, user can use the $save() method after importing their "old" project structure.

Not sure it worked: (on the right - the old configuration, on the left - the new after "save"):

image

Many entries are missing, is this right?

DESCRIPTION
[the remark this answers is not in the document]
Felixmil: Yes, or in renv specification for new project setup workflow. Will do.

R/project-configuration.R
Felixmil: If an absolute path is provided, then it will be written at absolute location. Otherwise it's relative. As this is the case for all path arguments in base R this is to be expected, however, I will add a mention to this in the docstring.

PavelBal: Yeah maybe I just didn't use R that much recently :D but it took me a while to understand where the new file is..

Felixmil:
> > It's described in the NEWS:
> 
> Hmm this is not sufficient. Describe a workflow of how to convert an existing project structure to the new one.
> 
> * Which folders and files to rename, which files to move to where...

This is not necessary because:
- If user has old project structure -> It still works with newer ESQlabs
- If user starts a new project -> He gets the new structure when using `init_project()`

Felixmil:
> > To export to new structure, user can use the $save() method after importing their "old" project structure.
> 
> Not sure it worked: (on the right - the old configuration, on the left - the new after "save"):
> 
> image
> 
> Many entries are missing, is this right?

Fixed in latest pushes. + now there is a dedicated test for this.

PavelBal:
> Do you think it is necessary ? Users can keep using their current project structure without changing anything:

Yes I think it is. First of all, saving the old configuration as the new structure does not make any sense without adapting the structure of the project.

And the typical use case would be - re-using an "old" project in a new one. You would like to convert it to the new structure to be compatible with other projects, take the advantages from the new developments and so on. So all in all - yes I as a user would like to know how can I upgrade my project to the new structure without having to create a new one.

PavelBal:
@Felixmil When doing check locally (also with the main branch), I get this error:

```
   Quitting from lines 83-95 [unnamed-chunk-6] (esqlabsR-project-structure.Rmd)
   Error: processing vignette 'esqlabsR-project-structure.Rmd' failed with diagnostics:
   there is no package called 'reactable'
   --- failed re-building 'esqlabsR-project-structure.Rmd'
```

Should the package be in the "suggest" section?

PavelBal (approved):

PR #695, opened 2024-08-02
Implement absolute x-axis values in sensitivitySpiderPlot()

### LINKED ISSUE #641: `sensitivitySpiderPlot` - absolute values in x axis
> Along the same lines, it would be helpful (but less critical for me at the moment) to do the same for the x-axis and make the normalization optional. the absolute values are here also already available in the sensitivity calculation output data frame in the column "ParameterValue"

These options can give weird-looking and unhelpful plots if you put multiple parameters at the same time, but they can be helpful to plot the same parameter in different scenarios

rengelke:
Closes #641


R/sensitivity-spider-plot.R
```diff
@@ -270,7 +290,7 @@ sensitivitySpiderPlot <- function(sensitivityCalculation,
 
       plot <- ggplot(
         dataSubset,
-        aes(x = ParameterFactor, y = !!sym(yColumn), group = ParameterPath)
+        aes(x = !!sym(xColumn), y = !!sym(yColumn), group = ParameterPath)
```
Felixmil, line 293: I would not use bang bang and sym syntax here. Please use the recommended `.data` pronoun (cf [here](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html#using-aes-and-vars-in-a-package-function)).
I would take the opportunity to update the rest of the code.

R/utilities-sensitivity-calculation.R
```diff
@@ -89,17 +89,16 @@
...
   pkData <- dplyr::group_by(pkData, ParameterPath, PKParameter) %>%
     dplyr::group_modify(.f = ~ .computePercentChange(.)) %>%
-    dplyr::ungroup()
+    dplyr::ungroup() %>%
+    dplyr::select(
+      "OutputPath",
+      dplyr::starts_with("Parameter"),
+      dplyr::starts_with("PK"),
+      Unit,
+      dplyr::everything(),
```
Felixmil, line 98: why use conditionnals and then `everything()` and a `-` right after ? Is it for reordering the columns ? Is it necessary to reorganize them ? I know this is not really part of this PR but I would clean this or add a comment explaining why.

At first sight, only saying to the select fonction "all variable but `individualId`" should be enough. Translated to dplyr language it would be:
```r
dplyr::select(-any_of("IndividualId"))
```

And then use `dplyr::relocate()` to reorganize columns:
```r
dplyr::relocate(
 "OutputPath",
  dplyr::starts_with("Parameter"),
  dplyr::starts_with("PK"),
  Unit
)

rengelke: I guess the intention was to reorder the columns, but I agree that this looks a bit convoluted. Yes, also I try not to touch code which is not related to the PR.

R/sensitivity-spider-plot.R
rengelke: Thanks, it's more explicit that way

R/utilities-sensitivity-calculation.R
rengelke: To be honest, I ould leave it like that:

``
dplyr::select(
      "OutputPath",
      dplyr::starts_with("Parameter"),
      dplyr::starts_with("PK"),
      Unit,
      dplyr::everything(),
      -dplyr::any_of("IndividualId")
    )
``

Use `any_of()` seems reasonable and is recommended. But reordering columns using `select()`  is an acceptable approach. For me the benefit of not using an additional function `dplyr::relocate` overweights.

Felixmil: Question is how are `dplyr::select()` performance on big dataframes compared to `dplyr::relocate()` ?

Felixmil (approved):

PR #705, opened 2024-09-02
Update and simplify installation/project setup script

Felixmil:
Remove PKSIM installation steps
+ remove rClr installation steps


rengelke (approved):
Code: Removed PK-Sim installation, now using cli explicitly. 
Tested locally, works fine.

PR #710, opened 2024-09-17
Validation object & individual modification validation

AKostiv8:
Closes #661 

```r

 devtools::load_all()
#ℹ Loading esqlabsR

# Create a project
p <- testProject()

# Check available individuals in configuration
p$configurations$individuals
# $Indiv1
# • Individual ID: Indiv1
# • Characteristics:
#   • Specy: Human
# • Population: European_ICRP_2002
# • Gender: MALE
# • Weight: 73
# • Height: 176
# • Age: 30
# • Protein: NA
# • Ontogeny: NA
# • Parameters:
#   • GFR

# Modify to non declared individual
 p$configurations$scenarios$TestScenario$individual <- "Non-declared individual"

# Run validation
p$validate()
# ℹ Warnings:
#   ! Scenario: TestScenario - The individual 'Non-declared individual' does not exist in the individuals configuration for scenario 'TestScenario'

# Modify `TestScenario2` and add one more unknown individual
 p$configurations$scenarios$TestScenario2$individual <- "Non-declared individual 2"
# Run validation
p$validate()
# ℹ Warnings:
#   ! Scenario: TestScenario - The individual 'Non-declared individual' does not exist in the individuals configuration for scenario 'TestScenario'
#   ! Scenario: TestScenario2 - The individual 'Non-declared individual 2' does not exist in the individuals configuration for scenario 'TestScenario2'

# Switch back to the defined individual in the `TestScenario2`
p$configurations$scenarios$TestScenario2$individual <- "Indiv1"
# Run validation
p$validate()
# ℹ Warnings:
#   ! Scenario: TestScenario - The individual 'Non-declared individual' does not exist in the individuals configuration for scenario 'TestScenario'

# Get validation object with the warnings 
p$get_warning_manager()$get_warnings()
# $TestScenario
# $TestScenario$INDIVIDUAL_NOT_FOUND
# [1] "The individual 'Non-declared individual' does not exist in the individuals configuration for scenario 'TestScenario'"
```

PR #728, opened 2024-10-14
Add Object validation in Project

AKostiv8:
Originally: https://github.com/esqLABS/esqlabsR/pull/710


Felixmil:
I think something went wrong during some merge or PR creation: some files are missing (WarningManager is one of them I think)

tests/testthat/test-project.R
```diff
@@ -126,3 +126,44 @@ test_that("Model is not defined", {
...
+  result <- p$exportToJSON(interactiveInput = FALSE)
+
+  expect_null(result)
+})
+
+
+test_that("Export aborts with invalid directory in non-interactive mode", {
+  p <- testProject()
+  # Capture cli output using testthat::expect_message()
+  expect_message(
```
Felixmil, line 143: use `expect_error` and fill the regex argument to check for error message

R/project.R
```diff
@@ -185,19 +185,33 @@ Project <-
       #' @description Export the entire project to a JSON file
       #' @param filePath The file path where the JSON will be saved
       #' @return NULL
-      exportToJSON = function() {
+      exportToJSON = function(interactiveInput = TRUE, dirPath = NULL, fileName = NULL) {
```
Felixmil, line 188: Also provide an `export()` **function** defined like this:

`export <- function(project, json = TRUE, xlsx = TRUE){}`

This function would call internally `project$exportToJSON` and in the future `project$exportToExcel` once we have it.

R/project.R
```diff
@@ -185,19 +185,33 @@ Project <-
...
+          dirPath <- readline(prompt = "Please enter the directory path where to save the file: ")
 
-        # Validate directory path
-        if (!dir.exists(dirPath)) {
-          cli::cli_alert_danger("Invalid directory path. Export aborted.")
-          return(invisible(NULL))
-        }
+          # Validate directory path
+          if (!dir.exists(dirPath)) {
+            cli::cli_alert_danger("Invalid directory path. Export aborted.")
```
Felixmil, line 196: use `cli_abort` instead and do not return anything ?

tests/testthat/test-project.R
```diff
@@ -126,3 +126,44 @@ test_that("Model is not defined", {
     "MODEL_NOT_FOUND"
   )
 })
+
+
+# Project export test ----------------------------------------------------------
```
Felixmil, line 131: Also add tests for import.

R/project.R
```diff
@@ -185,19 +185,33 @@ Project <-
...
+            return(invisible(NULL))
+          }
+
+          # Prompt for the file name
+          fileName <- readline(prompt = "Please enter the file name (without extension): ")
+        } else {
+          # Ensure both parameters are provided if interactiveInput is FALSE
+          if (is.null(dirPath) || is.null(fileName)) {
+            cli::cli_alert_danger("Both 'dirPath' and 'fileName' must be provided when 'interactiveInput' is FALSE.")
+            return(invisible(NULL))
```
Felixmil, line 206: It's not the first time I see functions returning NULL when everything's ok. But could you elaborate why we want to do so ?

We don't usually do it in the rest of the package now for homogeneity concerns I would either: not do it or implement it everywhere (but it requires a very good reason to spend time on this).

PR #732, opened 2024-10-18
Remove purrr::map

### LINKED ISSUE #678: Get rid of `purrr::map_dfr` etc
`purrr::map_dfr`  is superseded, replace it by the more up-to-date approaches.

Or better just get rid of all map purr etc

rengelke:
Closes #678

### DIFF
--- a/R/sensitivity-spider-plot.R
+++ b/R/sensitivity-spider-plot.R
@@ -111,19 +111,16 @@ sensitivitySpiderPlot <- function(sensitivityCalculation,
                                   xAxisType = "percent",
                                   yAxisType = "percent",
                                   defaultPlotConfiguration = NULL) {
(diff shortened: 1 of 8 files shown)


Felixmil:
What is the rationale behind replacing functionals by for loops here ?

Felixmil (commented):
Why replace functionals with for loops ? Even if we don't want to use purrr, there is still `lapply()` and co.

rengelke:
> Why replace functionals with for loops ? Even if we don't want to use purrr, there is still `lapply()` and co.

The switch to for loops was made to simplify debugging, as requested. Since performance was the same in my tests comparing `lapply()`, `purrr::map()`, and `for` loops, I didn’t open a discussion.

Felixmil (approved):

PR #768, opened 2025-02-17
Remove all non essential project files

### LINKED ISSUE #766: Remove ESQlabs projects' specific files from example project
Example project in `inst/extdata/examples/TestProject/` (which is replicated when using `init_project()` contains files that are specific to ESQlabs internal projects (e.g. `Code/` folder) in addition to `{esqlabsR}` files like `projectConfiguration.xlsx` or other excel files.

Actions:
- [ ] Copy ESQlabs project files to the template repository
- [ ] Delete them from this  repo

Felixmil:
Closes #766


PavelBal:
Wait. And how is anyone outside of esq supposed to use the package then? I think the default project should be kept here, along with the "init" script. Otherwise the package is not usable.

Felixmil:
The package still contains what's required to initialize a project:
- init_project()
- ProjectCOnfiguration.xlsx
- COnfigurations/ folder with all default excel files

These also serve as testing material.

I Only removed the following folders:
image

PavelBal:
@Felixmil Why did you remove the R project from the repo?

I now have to manually create a new project, set the install args properly etc... Why do you think it is better not to synchronize it?

PavelBal:
E.g. the `tidyverse` packages all have the .rpoj file in the repos.

Felixmil:
> @Felixmil Why did you remove the R project from the repo?
> 
> I now have to manually create a new project, set the install args properly etc... Why do you think it is better not to synchronize it?

You should not have to create it manually because it is created when a repo a cloned using RStudio Project Creation.
But if the tidyverse team sync them, there is probably a good reason. I will try to find information on that.

PavelBal:
Maybe because not everyone is cloning with RStudio ;) I would prefer to have it back.

PavelBal:
I brought back the RStudio project and the required folders (e.g. `Results` folder is specified in the project configuration and is used by the `saveResults` function.

PavelBal (approved):

Felixmil:
@PavelBal It looks like you also reverted the Code/ folder that is now supposed to be in our project template and not here.

PavelBal:
Why would you prefer to remove them? There is no proprietary information included. I think it is nice to have a working example distributed with the package.

Felixmil:
> Why would you prefer to remove them? There is no proprietary information included. I think it is nice to have a working example distributed with the package.

They are not documented nor mentionned anywhere.
When a user just wants to initialize a project he still gets these extra files that can be confusing for a new user or unnecessary for an experienced one.
They are not necessary and I feel like initializing a project should only create necessary files.
So I would either add an option to create these when using init_project or move these to project template.

PR #783, opened 2025-03-03
780-create-plots-return-data-combined

### LINKED ISSUE #780: `createPlotsFromExcel` should return the created data combined

PavelBal:
Fixes #780

- `createPlotsFromExcel` now returns a list with the items `plots` and `dataCombined`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: esqlabsR
 Title: esqLABS utilities package
-Version: 5.3.0.9007
+Version: 5.3.0.9010
 Authors@R: c(
     person("esqLABS GmbH", role = c("cph", "fnd")),
     person("Pavel", "Balazki", , "pavel.balazki@esqlabs.com", role = c("cre", "aut"),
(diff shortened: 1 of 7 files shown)


PavelBal:
Okay I see your point.

> we can extend createPlotsFromExcel() to be able use dataCombined provided by the user.

Yes, this would allow to pre-process DataCombined before feeding them into createPlots.

The "problem" is that the user has to manually select the DataCombined that are used by the plots. So I might want to create three plots, but they use 10 different DataCombined. `createPlotsFromExcel` takes care of creating the required DataCombined internally.

Felixmil:
I see, then `createPlotsFromExcel` can check if provided `dataCombined` are enough to build plots, if not, return an error or propose to the user to build all necessary `dataCombined` ?

PavelBal:
> I see, then createPlotsFromExcel can check if provided dataCombined are enough to build plots, if not, return an error (listing missing ones) or propose to the user to build all necessary dataCombined ?

Yes I like this (returning the list of DC names that must be included).

PR #786, opened 2025-03-05
Remove "Code" folder from the example project

PavelBal:
Follow up on https://github.com/esqLABS/esqlabsR/pull/768#issuecomment-2698255823

- Integrate the default workflow scirpt into README
- Update generated figure
- Styler


PavelBal:
I have extended the example in the README by the content of the "default scenario" that was in the Code folder.

PR #794, opened 2025-03-16
Plots from excel data combined argument

### LINKED ISSUE #787: `createPlotsFromExcel` should accept `DataCombined` as argument
I see, then `createPlotsFromExcel` can check if provided `dataCombined` are enough to build plots, if not, return an error (listing missing ones) or propose to the user to build all necessary `dataCombined` ?

_Originally posted by @Felixmil in https://github.com/esqLABS/esqlabsR/issues/783#issuecomment-2696878827_

PavelBal:
- Fixes #787


R/enum.R
```diff
@@ -7,7 +7,7 @@
 #' @param overwrite if TRUE and a `key` exists, it will be overwritten with the
 #'   new value. Otherwise, an error is thrown. Default is `FALSE`.
 #'
-#' @return Enum with added key-value pair.
+#' @returns Enum with added key-value pair.
```
Felixmil, line 10: What's the reason behind using `returns` alias for `return` ?
Mention this new standard in NEWS ?

R/utilities-data-combined.R
```diff
@@ -130,6 +127,63 @@ createDataCombinedFromExcel <- function(
...
+#' Each data combined used by the specified plot grids will be created.
+#' @param projectConfiguration Object of class `ProjectConfiguration`
+#' that contains information about the output paths and the excel file
+#' where plots are defined.
+#' @inheritParams createDataCombinedFromExcel
+#' @returns A list of `DataCombined` objects that are required to create plots
+#' specified in `plotGridNames`
+#'
+#' @export
+createDataCombinedForPlots <- function(
```
Felixmil, line 145: The name is confusing in regards of `createDataCombinedFromExcel`
I think we should stick to `createDataCombinedFromExcel` only and either:
- Provide a function that list dataCombinedNames necessary for a set of plotGrids. Then, the user can pass it to createDataCombinedFromExcel (like it is done right now but with 2 different functions). Or:
- Add an optional `plotGridNames` in `createDataCombinedFromExcel` and integrate `createDataCombinedForPlots`'s logic there.

Felixmil (commented):
- [ ] Question about return**s**
- [ ] Concerns about having both `createDataCombinedForPlots` and `createDataCombinedFromExcel`

R/enum.R
PavelBal: Suddenly, when using the "Code -> Insert Roxygen Skeleton" in RStudio, it creates `@returns` instead of `@return`. Also:

image

So I migrated everything to "returns" so the new skeletons are consistent with the rest of the code.

I am not sure that this is worth mentioning in the NEWS? This is purely internal...

R/utilities-data-combined.R
PavelBal: Yeah I see the point. The problem I see is that `createDataCombinedFromExcel` is not that user friendly in terms of that the user has to manually provide the path to the file and the sheet name. The new `createDataCombinedForPlots` follows the same signature as `createPlotsFromExcel` where the user passes the project configuration, which is more convenient for the majority, I think. I think that the standard workflows should rely on the project configuration without the users specifying any paths.

PavelBal:
btw - why does the check say

```
Warning: Undocumented arguments in Rd file 'createDataCombinedForPlots.Rd'
  'simulatedScenarios' 'observedData' 'stopIfNotFound'
```

though I have specified 

`#' @inheritParams createDataCombinedFromExcel`

?

R/utilities-data-combined.R
Felixmil: I understand, maybe we should improve createPlotsFromExcel  to be able to handle ProjectCOnfiguration ?

NEWS.md
```diff
@@ -1,11 +1,34 @@
 # esqlabsR (development version)
 
+## Breaking changes
+
+- The function `createDataCombinedFromExcel()` gets a new signature. The arguments
```
PavelBal, line 5: @VanessaBa Please review from the user point of view

R/utilities-data-combined.R
```diff
@@ -1,35 +1,51 @@
...
+  if (is.null(dataCombinedNames) && is.null(plotGridNames)) {
+    return(list())
   }
 
+  # If plotGridNames are provided, extract the names of required data combined
+  # and add them to the passed data combined names
+  if (!is.null(plotGridNames)) {
+    # Combine the passed data combined names with the names required for
+    # the passed plots
+    dataCombinedNames <- union(dataCombinedNames, .extractDataCombinedNamesForPlots(projectConfiguration = projectConfiguration, plotGridNames = plotGridNames))
```
PavelBal, line 43: Extracting of data combined names required for plots is moved to an internal function.

R/utilities-data-combined.R
```diff
@@ -196,13 +211,55 @@ createDataCombinedFromExcel <- function(
...
+#' @noRd
+.extractDataCombinedNamesForPlots <- function(projectConfiguration, plotGridNames) {
+  # read sheet "plotGrids" with info for plotGridConfigurations
+  dfPlotGrids <- readExcel(projectConfiguration$plotsFile,
+    sheet = "plotGrids"
+  )
+
+  # Filter for only specified plot grids
+  # Throw an error if a plot grid name that is passed is not defined in the excel file
+  missingPlotGrids <- setdiff(plotGridNames, unique(dfPlotGrids$name))
```
PavelBal, line 242: Here is some code duplication with `createPlotsFromExcel`

R/utilities-figures.R
```diff
@@ -270,8 +270,12 @@ createEsqlabsExportConfiguration <- function(outputFolder) { # nolint: object_le
 #' @param outputFolder Optional - path to the folder where the results will be
 #' stored. If `NULL` (default), `projectConfiguration$outputFolder` is used. Only
 #' relevant for plots specified for export in the `exportConfiguration` sheet.
+#' @param dataCombinedList A (named) list of `DataCombined` objects as input
```
PavelBal, line 273: New argument that allows to pass DataCombined to the plot function

R/utilities-figures.R
```diff
@@ -280,62 +284,36 @@ createPlotsFromExcel <- function(
...
     stopIfNotFound = TRUE) {
   validateIsOfType(observedData, "DataSet", nullAllowed = TRUE)
   validateIsOfType(projectConfiguration, "ProjectConfiguration")
   validateIsString(plotGridNames, nullAllowed = TRUE)
+  validateIsOfType(dataCombinedList, "DataCombined", nullAllowed = TRUE)
 
-  # read sheet "plotGrids" with info for plotGridConfigurations
-  dfPlotGrids <- readExcel(projectConfiguration$plotsFile,
-    sheet = "plotGrids"
+  plotConfigurations <- .readPlotConfigurations(
```
PavelBal, line 296: Reading of excel sheet exctracted to an internal function to avoid code duplication (used by createPlotsFromExcel and getDataCombinedNames

tests/testthat/test-create-plots-from-excel.R
```diff
@@ -531,6 +530,83 @@ test_that("It creates plots for all plot grids when plotGridNames is NULL", {
...
+
+plots <- createPlotsFromExcel(
+  simulatedScenarios = simulatedScenarios,
+  observedData = observedData,
+  projectConfiguration = projectConfiguration,
+  dataCombinedList = dataCombinedList,
+  stopIfNotFound = TRUE
+)
+
+vdiffr::expect_doppelganger(title = "firstPlot", print(plots[[1]]))
```
PavelBal, line 551: This test ALWAYS fails on my machine due to non-visible changes! Any idea what is happening?

PavelBal: @Felixmil @rengelke I am lost here... If it does not work I will just remove the test.

PavelBal: @Felixmil @rengelke can you try running the tests on this branch locally? Do you also get snapshot changes every time?

PavelBal: This happens already at simulation results level:

``` 
  yval <- simulatedScenarios$TestScenario$outputValues$data$`Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)`

  expect_snapshot_value(yval, style = "deparse", tolerance = 1e-6)
```

The solution is to save simulation results as csv and load them for tests.

rengelke: Yes, it's the simulated data. glad that workaround with csv works.

PavelBal: @rengelke could you review, please?

PR #812, opened 2025-04-15
- The function `escapeForRegex()` has been removed from the package.

### LINKED ISSUE #809: Review of `utilities` functions
We have some functions in the `utilities` file that are of historical nature, and I would like to review their validity.

- https://github.com/esqLABS/esqlabsR/blob/f8c59720414061e2664501ff9966b6c9bf4fd77f/R/utilities.R#L63
#' Calculate geometric mean of a numeric vector

@wilbertdew @VanessaBa @Felixmil @rengelke First, please review if the implementation is correct, and second, is this maybe already offered by some "standard" package? I don't want to implement something that has already be done by others and is well maintained

- https://github.com/esqLABS/esqlabsR/blob/f8c59720414061e2664501ff9966b6c9bf4fd77f/R/utilities.R#L75
@wilbertdew @VanessaBa @Felixmil @rengelke same here

- https://github.com/esqLABS/esqlabsR/blob/f8c59720414061e2664501ff9966b6c9bf4fd77f/R/utilities.R#L129
Not sure this is helpful, usually we can just set "fixed = TRUE" when matching a string, right?

PavelBal:
Fixes #809


rengelke (approved):

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Type: Package
 Package: esqlabsR
 Title: esqLABS utilities package
-Version: 5.3.0.9026
+Version: 5.3.0.9027
```
Felixmil, line 4:
```suggestion
Version: 5.3.0.9027
```
dev version should not be bumped manually, this is taken care by GHA.

Felixmil (approved):

wilbertdew:
late comment on #809 , geomean and geosd are useful and not directly available in standard packages as far as i know

PavelBal:
@wilbertdew Yes they will remain ;)

PR #815, opened 2025-04-17
ignore-rows-without-scenario-name

PavelBal:
`readScenarioConfigurationFromExcel()` ignores rows where `Scenario…_name` is empty.


Felixmil:
We should add tests with faulty data.
However, since we use the same data for tests and for initProject(), the users would end up with weird projects on initialization (using initProject()).
Unless we split ESQlabsR project template and test data. Will open a new issue for discussion.

PR #651, opened 2024-05-06
Project-based Workflow

Felixmil:
## New Workflow Preview

``` r
devtools::load_all()
#> ℹ Loading esqlabsR
#> Loading required package: ospsuite
#> 
#> Loading required package: rClr
#> 
#> Loading the dynamic library for Microsoft .NET runtime...
#> Loaded Common Language Runtime version 4.0.30319.42000

# Create a project
p <- testProject()

# Project Overview
p
#> 
#> ── Project Configuration ───────────────────────────────────────────────────────
#>   • Project Configuration:
#>   'inst/extdata/examples/TestProject/ProjectConfiguration.xlsx'
#>   • Model folder: 'inst/extdata/examples/TestProject/Models/Simulations'
#>   • Configurations folder: 'inst/extdata/examples/TestProject/Configurations'
#>   • Model Parameters:
#>   'inst/extdata/examples/TestProject/Configurations/ModelParameters.xlsx'
#>   • Individuals:
#>   'inst/extdata/examples/TestProject/Configurations/Individuals.xlsx'
#>   • Populations:
#>   'inst/extdata/examples/TestProject/Configurations/Populations.xlsx'
#>   • PopulationsCSV:
#>   'inst/extdata/examples/TestProject/Configurations/PopulationsCSV'
#>   • Scenarios:
#>   'inst/extdata/examples/TestProject/Configurations/Scenarios.xlsx'
#>   • Applications:
#>   'inst/extdata/examples/TestProject/Configurations/Applications.xlsx'
#>   • Plots: 'inst/extdata/examples/TestProject/Configurations/Plots.xlsx'
#>   • Data folder: 'inst/extdata/examples/TestProject/Data'
#>   • Data file:
#>   'inst/extdata/examples/TestProject/Data/TestProject_TimeValuesData.xlsx'
#>   • Data importer configuration:
#>   'inst/extdata/examples/TestProject/Data/esqlabs_dataImporter_configuration.xml'
#>   • Compound Properties File:
#>   • Output folder: 'inst/extdata/examples/TestProject/Results'
#> 
#> ── Configurations ──────────────────────────────────────────────────────────────
#>   • Scenarios: 5
#>   • Models: 1
#>   • Model Parameters: 3
#>   • Individuals: 1
#> 
#> ── Scenarios ───────────────────────────────────────────────────────────────────
#>   • ✅ TestScenario
#>   • ✅ TestScenario2
#>   • ✅ PopulationScenario
[...]

PR #821, opened 2025-04-25
feat: add snapshot and restore project configuration functions

Felixmil:
Features to implement:

- [x] Excel <-> Json interoperability (snapshot/restore workflows)
- [x] Synchronisation status report function

- Introduced `snapshotProjectConfiguration()` to export project configurations from Excel to JSON for version control.
- Added `restoreProjectConfiguration()` to recreate Excel files from JSON snapshots.
- Updated documentation for both functions and added examples in the project structure vignette.
- Included new tests to ensure functionality of the snapshot and restore processes.


R/utilities-config-json.R
```diff
@@ -0,0 +1,456 @@
...
+  project_config = "ProjectConfiguration.xlsx",
+  output_dir = "."
+) {
+  # Convert to ProjectConfiguration object if path is provided
+  if (is.character(project_config)) {
+    project_config <- createProjectConfiguration(project_config)
+  }
+
+  # Validate project_config
+  if (!inherits(project_config, "ProjectConfiguration")) {
```
PavelBal, line 22: How can this ever happen? The name of the file is never passed by the used, and is created from the name of the excel file... no?

R/utilities-config-json.R
```diff
@@ -0,0 +1,456 @@
+#' Export project configuration Excel files to JSON
+#'
+#' @description
+#' Exports all Excel configuration files in the Configurations folder of an esqlabsR project
+#' to a single JSON file. This allows for easier version control and programmatic manipulation.
+#'
+#' @param project_config A ProjectConfiguration object or path to ProjectConfiguration excel file. Defaults to "ProjectConfiguration.xlsx".
+#' @param output_dir where the JSON file will be saved. Defaults to the current working directory.
+#'
+#' @return Invisibly returns the exported configuration data structure
```
PavelBal, line 10: What do you think about returning the `json` structure?

R/utilities-config-json.R
```diff
@@ -0,0 +1,456 @@
...
+#' This allows for recreating the project configuration from version-controlled JSON.
+#'
+#' @param json_path Path to the JSON configuration file. Defaults to "ProjectConfiguration.json".
+#' @param output_dir Directory where the Excel files will be created. Defaults to the current working directory.
+#'
+#' @return A ProjectConfiguration object initialized with the regenerated ProjectConfiguration.xlsx
+#' @export
+restoreProjectConfiguration <- function(
+  json_path = "ProjectConfiguration.json",
+  output_dir = "."
```
PavelBal, line 214: Same as above - should default to the directory of input json file.

PavelBal:
My suggestions to the output json structure, to make it more convenient for humans:

- Remove the level `files` (strong opinion about this):
image

- Remove the level `projectConfiguration`, leave only `ProjectConfiguration`:
image

- Rename `modelParameters` to `ModelParameterSets`
- `individuals` to `Individuals`
- Same for the following sections, start capitalized


Can we add allowed values for certain columns? e.g., the column `dataType` in file `Plots`, sheet `DataCombined` can only accept `simulated` or `observed`. But this is for a different PR.

R/utilities-config-json.R
Felixmil: Input can be a path to `ProjectConfiguration.xslx` OR an `ProjectConfiguration` object.
So if it is not a character string (a path) that was provided, we need to check the object type.

R/utilities-config-json.R
Felixmil: It is returned but invisibly.
So you could do:
`json_congig <- snapshotProjectConfiguration(...)`
to retrieve the output json (while the json file is also written).

If running `snapshotProjectConfiguration(...)`, only the json file is created and the json structure is not printed (it would be too long in the console and difficult to read).

Felixmil:
> Can we add allowed values for certain columns? e.g., the column dataType in file Plots, sheet DataCombined can only accept simulated or observed. But this is for a different PR.

This kind of validation will be implemented in #651.

R/utilities-config-json.R
PavelBal: I meant that the returned data is a nested list and not a json string.

R/utilities-config-json.R
Felixmil: Indeed, remnant from when the user was able to pass a different file name for output. But now that the user can only provide outputDir, not required anymore. Will delete.

Felixmil:
Added `projectConfigurationStatus()` function as a convenient way for the user to scan differences between Excel files and project snapshot. Here's a preview of the output:

```r
! Excel configuration files are NOT in sync with JSON snapshot.                                                                                                                               

── File Sync Status: ──

✔ ProjectConfiguration.xlsx.xlsx: in-sync
✔ projectConfiguration.xlsx: in-sync
✔ modelParameterSets.xlsx: in-sync
✔ Individuals.xlsx: in-sync
✔ Populations.xlsx: in-sync
✖ Scenarios.xlsx: out-of-sync
✔ Applications.xlsx: in-sync
✖ Plots.xlsx: out-of-sync

── Details: ──

• Scenarios.xlsx
  • Different data in sheets: Scenarios
• Plots.xlsx
  • New sheets: CombinedTestSheet

── Suggested Actions: ──

To resolve these differences, you can:
  • `snapshotProjectConfiguration()` - Save the changes from Excel files to the project snapshot.
  • `restoreProjectConfiguration()` - Recreate the Excel files according to the configuration snapshot.
```

PavelBal:
pkdown failing @Felixmil

PavelBal:
> ✔ ProjectConfiguration.xlsx.xlsx: in-sync

`*.xlsx.xlsx`? :D

PavelBal:
@Felixmil could you change the description of the package?

Description: The {esqlabsR} package facilitates and standardizes the modeling and simulation of physiologically based kinetic (PBK) and quantitative systems pharmacology/toxicology (QSP/T) models implemented in the [Open Systems Pharmacology Software](https://www.open-systems-pharmacology.org/) (OSPS).

Felixmil:
Shall we include this in next release @PavelBal ? It is ready if we want to.

Felixmil:
Except for the last point. All your suggestions were implemented

PavelBal (approved):

PR #824, opened 2025-04-25
sensitivityTimeProfiles y label dimension fix

### LINKED ISSUE #823: `sensitivityTimeProfiles` axis labels not always correct in respect to dimension
When changing the unit of y axis to `ng/ml`, the label still says `Concentration (molar)`, though it should be `Concentration (mass)`.

```
p <- sensitivityTimeProfiles(results)
print(p)
```

Image

Changing units: 

`sensitivityTimeProfiles(results, yUnits = list("ng/ml"))`

Image

PavelBal:
- Fixes #823 

Fixed a bug when the dimension in the y-axis label of `sensitivityimeProfiles()` did not match the unit

### DIFF
--- a/R/sensivitity-time-profiles.R
+++ b/R/sensivitity-time-profiles.R
@@ -216,10 +216,12 @@ sensitivityTimeProfiles <- function(sensitivityCalculation,
       unique(data$xUnit), "]"
     )
   }
+  yUnitForPlot <- unique(data$yUnit)
+  yDimensionForPlot <- ospsuite::getDimensionForUnit(yUnitForPlot)
   if (is.null(plotConfiguration$yLabel)) {
     plotConfiguration$yLabel <- paste0(
-      unique(data$yDimension), " [",
-      unique(data$yUnit), "]"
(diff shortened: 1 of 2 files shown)


R/sensivitity-time-profiles.R
```diff
@@ -216,10 +216,12 @@ sensitivityTimeProfiles <- function(sensitivityCalculation,
       unique(data$xUnit), "]"
     )
   }
+  yUnitForPlot <- unique(data$yUnit)
+  yDimensionForPlot <- ospsuite::getDimensionForUnit(yUnitForPlot)
```
Felixmil, line 220: Is it possible that more than one result is returned here ? If yes, how to handle this ? We should at least add a [1] for now to make sure to print only the first element.

rengelke (approved):

R/sensivitity-time-profiles.R
PavelBal: No this should never happen. `getDimensionForUnit()` always returns one value (or there is a bug).

PR #834, opened 2025-05-13
add snapshot utilities

PavelBal:
- New function `addPlotsToExcelFromSnapshot()` to add entries for plots defined for all simulations of a PK-Sim snapshot to the `Plots.xlsx` file.
- New function `createScenarioConfigurationsFromSnapshot()` to create scenario configurations for all simulations of a PK-Sim snapshot.


R/utilities-scenario-configuration.R
```diff
@@ -253,3 +253,106 @@ setApplications <- function(simulation, scenarioConfiguration) {
...
+#' Existing scenarios will not be overwriten. If the scenario configuration
+#' file already contains a scenario with the same name, the resulting file will
+#' have a duplicate entry. It is the resposibility of the user to ensure that no duplicate
+#' scenarios are present in the excel file.
+#'
+#' @param scenarioConfigurations A list of `ScenarioConfiguration` objects
+#' @param projectConfiguration An object of type `ProjectConfiguration`
+#'
+#' @export
+addScenariosToExcel <- function(scenarioConfigurations, projectConfiguration) {
```
PavelBal, line 268: @wilbertdew you also have something similar, right? To write scenario confirugaration into the excel? Do you want to share you code so maybe we can integrate it here?

wilbertdew: Of course! We did not use any scenarioConfiguration object, I believe, but defined it more or less as you would in Excel. We developed a bit more elaborate script to create and modify (e.g., copy-paste a selection of scenarios, change a selection of columns) all of the necessary excel files. All the defaults are also in there so in principle this script also worked without filling in anything, in which case it would just have one scenario from a .pkml file. Here is an example from a project, where you can see how it was used:
[https://esqlabs.sharepoint.com/:u:/s/SysPharm/EbDOQ7GPe51Ikdh_j503fb4BkMsilgYchUtL2Ozms3IFXg?e=GhgTZI](url)

NEWS.md
```diff
@@ -14,6 +14,14 @@
 in the `customParams` argument are applied last. Up to this version, they 
 were overwritten by the administration protocol (\#817)
 
+- New function `addScenariosToExcel()` to add scenarios configurations to the `Scenarios.xlsx` file.
```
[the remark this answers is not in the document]
wilbertdew: makes sense to me

R/utilities-scenario-configuration.R
PavelBal: @wilbertdew  ping :)

wilbertdew: Argh, think I sent the wrong version sorrysorrysorry. this is the right one without that function: https://esqlabs.sharepoint.com/:u:/s/SysPharm/EcqNYGpkMAtPjk3JQ0PmNEgB6jJ9PYNiu3urv4Vzem_6kw?e=7v8Xg6

PR #840, opened 2025-05-23
Tornado plot axis mapping and axis limits

rengelke:


Felixmil (approved):

PavelBal:
So there is still no way to change the limits? I am surprized that this cannot be implemented... It is a common case that the user would like to zoom in (e.g., the highest sensitivity is 100% and other parameters have sensitivity <1, it is impossible to get some useful information from the plot without zooming in)

PavelBal:
> I’ve added a new parameter xAxisZoomRange directly to sensitivityTornadoPlot(),

Is this a separate PR? Or didn't you push yet?

PR #862, opened 2025-07-31
Add support for saving and loading sensitivity calculations

### LINKED ISSUE #829: Save & load sensitivity results
Could we have a function to save and load a sensitivity analysis to avoid having to rerun it to adapt a plot?

rengelke:
Adds two new functions: `saveSensitivityCalculation()` and `loadSensitivityCalculation()`.

### Comments
- `Simulation` is optional when loading; it is automatically restored from the internally saved file path if not provided.
- The `.meta` suffix is used when saving the `sensitivityCalculation` container using `saveRDS()` to prevent users from opening it as a `.rds` file without simulation results.
- Simulation results are saved as individual uncompressed `.csv` files, as per user request.

Closes #829


R/utilities-sensitivity-calculation.R
```diff
@@ -526,3 +526,171 @@
...
+      # Clear `SimulationResults` while preserving named list structure
+      simulationResults[[i]][[j]] <- list(NULL)
+    }
+  }
+
+  # save sensitivityCalculation w/o `SimulationResults`
+  sensitivityCalculation$simulationResults <- simulationResults
+  saveRDS(
+    sensitivityCalculation,
+    file.path(outputDir, "sensitivityCalculation.meta")
```
Felixmil, line 606: nice trick

R/utilities-sensitivity-calculation.R
```diff
@@ -526,3 +526,171 @@
...
+                                       overwrite = FALSE) {
+  validateIsOfType(sensitivityCalculation, "SensitivityCalculation")
+  validateIsString(outputDir)
+  validateIsLogical(overwrite)
+
+  if (dir.exists(outputDir)) {
+    if (!overwrite) {
+      stop(messages$errorOutputDirExists(outputDir))
+    }
+    unlink(outputDir, recursive = TRUE)
```
Felixmil, line 577: hopefully the user doesn't store other files in the output directory before using `overwrite=T`. Maybe ask user for confirmation ?

vignettes/sensitivity.Rmd
Felixmil, line 1: Add an example section for how to save/load SA results ? Or at least mention the feature exists.

Felixmil (commented):
Very nice, our users will be happy with this !
Only comments and questions.

PR #900, opened 2025-10-28
use glue

### LINKED ISSUE #514: Use glue() instead of paste0() in messages
Another improvement suggestion would be to use `glue::glue()` for easier readability. It is not in DESCRIPTION dependency list right now but we already have it indirectly since it's a ggplot2 dependency (and many other packages we use).

_Originally posted by @Felixmil in https://github.com/esqLABS/esqlabsR/pull/512#pullrequestreview-1572064394_

sorinvoicu:
closes #514

### DIFF
--- a/R/messages.R
+++ b/R/messages.R
@@ -22,7 +22,7 @@ messages$wrongParametersStructure <- function(argumentName) {
 
 # Enum####
 messages$errorEnumPutListMultipleKeys <- function() {
-  "Trying to put multiple keys, but only one key is allowed!"
+  cliFormat("Trying to put multiple keys, but only one key is allowed!")
 }
 
 # utilities-population####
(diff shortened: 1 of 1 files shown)


tests/testthat/test-sensitivity-tornado-plot.R
```diff
@@ -48,7 +48,7 @@ test_that("sensitivityTornadoPlot fails with incorrect input", {
 test_that("sensitivityTornadoPlot fails with invalid parameterFactor", {
   expect_error(
```
rengelke: Let's take the chance to improve tests to using actual message functions for expected regex. then we avoid having this type of failures

Same in tests/testthat/test-sensitivity-tornado-plot.R:67

rengelke (changes requested):
Clear and to the point.
Minor issues:
- remove unneccesary as.character() conversion
- dont use pipes if doesnt improve readabillity significantly
- update NEWS

UPDATE as discussed:
- convert to `ospsuite.utils::cliFormat`

rengelke (approved):

PR #912, opened 2025-11-20
Extending validation

nautilus69:
in order to make this  https://github.com/esqLABS/ESQapp/issues/193 
we had to extend the validations of the excel files. so that we can have these requirement on the esqlabsR side: 

    Perform ALL validation logic
    Return structured ValidationResult objects (or similar)
    Include critical errors, warnings, and validated data
    Handle all business rules (required fields, uniqueness, references, etc.)

### DIFF
--- a/tests/testthat/test-validation.R
+++ b/tests/testthat/test-validation.R
@@ -446,3 +446,502 @@ test_that(".validateApplicationsFile handles valid applications file", {
 
   unlink(temp_file)
 })
+
+# Tests for .validateIndividualsFile ----
+
+test_that(".validateIndividualsFile detects missing file", {
+  result <- esqlabsR:::.validateIndividualsFile("nonexistent_file.xlsx")
+  expect_false(result$is_valid())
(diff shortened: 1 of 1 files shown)


R/validation-utils.R
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  return(result)
+}
+
+#' Categorize validation messages
+#' @keywords internal
+.categorize_message <- function(message) {
+  if (grepl("missing|empty|not found", message, ignore.case = TRUE)) {
+    return("Missing Fields")
```
sorinvoicu: perfect case of case_when usage?

PavelBal:
Yeah agree with Felix suggestion.

R/validation-utils.R
PavelBal: @nautilus69

PavelBal: @nautilus69 please rewrite using `case_when` here

PR #956, opened 2026-02-24
- Replace .validateUnitInOspUnits  by ospsuite.utils::validateEnumVal…

PavelBal:
…ue()

### DIFF
--- a/R/sensivitity-time-profiles.R
+++ b/R/sensivitity-time-profiles.R
@@ -606,7 +606,7 @@ sensitivityTimeProfiles <- function(
     return(rep(list(NULL), length(outputPaths)))
   }
   # validate if units are valid in any dimension
-  allowedUnits <- unlist(ospUnits, use.names = TRUE)
+  allowedUnits <- enum(unlist(ospUnits, use.names = FALSE))
   lapply(units, \(unit) {
     ospsuite.utils::validateEnumValue(unit, allowedUnits, nullAllowed = TRUE)
   })


sorinvoicu:
The PR replaces the old .validateUnitInOspUnits() helper with a direct call to ospsuite.utils::validateEnumValue():                                                                                                                                                   
                                                                                                                                                                                                                                                                        
  OLD (worked correctly)                                                                                                                                                                                                                                              
  lapply(units, .validateUnitInOspUnits, TRUE)                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                        
  NEW (broken)                                                                                                                                                                                                                                                        
  allowedUnits <- unlist(ospUnits, use.names = TRUE)                                                                                                                                                                                                                    
  lapply(units, \(unit) {                                                                                                                                                                                                                                               
    ospsuite.utils::validateEnumValue(unit, allowedUnits, nullAllowed = TRUE)                                                                                                                                                                                           
  })

  The bug

  unlist(ospUnits, use.names = TRUE) mangles the enum names. ospUnits is a named list of named vectors, one per dimension (e.g. $Time, $Mass). After unlist(..., use.names = TRUE), the names of the resulting vector become composite strings like "Time.month(s)",
  "Mass.month(s)", etc. instead of just "month(s)".

  ospsuite.utils::validateEnumValue() validates by looking up the input against the names of the enum vector. Since "month(s)" is not a name in allowedUnits (only "Time.month(s)" etc. are), it fails:

  Error: month(s) is not a valid value in `allowedUnits`.
  Did you mean one of these: month(s), month(s), and month(s)?

  The three identical suggestions are the VALUES (not names) of the 3 entries in allowedUnits that fuzzy-match — they all happen to have the value "month(s)" from different dimensions.

  The old .validateUnitInOspUnits() worked correctly because it tried validateEnumValue against each dimension's sub-enum separately (where names were just "month(s)"), succeeding on the first dimension that contained the unit.

  Why macOS passes

  The test counts differ:
  - Ubuntu: FAIL 1 | SKIP 1 | PASS 1132
  - macOS: FAIL 0 | SKIP 2 | PASS 1130

  macOS has 2 skips vs Ubuntu's 1. The failing test (sensitivityTimeProfiles applies y-unit conversion for multiple paths, test-sensitivity-time-profiles.R:237) is skipped on macOS — most likely due to a skip_on_os("mac") guard or a condition like
  skip_if(!isTRUE(getOption("ospsuite.useSimulator"))) that prevents it from running on macOS ARM64 in CI where the .NET-based simulator may not be fully available.

  Fix

  Either revert to the old per-dimension loop approach, or strip the composite names after unlist:

  Option: strip composite names so names remain just the unit strings
  allowedUnits <- unlist(ospUnits, use.names = FALSE)
  names(allowedUnits) <- unlist(lapply(ospUnits, names))

PR #993, opened 2026-04-09
696 osp plots

pchelle:
WIP on OSPSuite.Plots migration

## Summary by CodeRabbit

* **New Features**
  * New plot-grid API for assembling multi-plot layouts and styling; new Predicted-vs-Observed default shows LLOQ on both axes; axis scale names use "linear"/"log".

* **Breaking Changes**
  * Export-configuration/export-based workflow removed; outputFolder/export config no longer used; plotting now uses updated plot/config/grid model and requires newer plotting deps.

* **Documentation**
  * Vignettes, examples and reference docs updated to reflect the new plotting workflow and configuration APIs.

* **Tests**
  * Updated/added tests and snapshots to match new behavior and rendering stability.


pchelle:
Some snapshot tests are now failing due to ospsuite.plots update from yesterday (https://github.com/Open-Systems-Pharmacology/OSPSuite.Plots/pull/106)

R/esqlabs-plot-export-configuration.R
```diff
@@ -32,13 +32,69 @@ ExportConfiguration <- R6::R6Class(
...
+    },
+
+    #' @description Print properties of export configuration
+    #' @return Export configuration properties
+    print = function() {
+      ospsuite.utils::ospPrintItems(
+        list(
+          "Format/Device" = self$format,
+          "Width" = paste(self$width, self$units),
+          "Height Per Row" = paste(self$width, self$units),
```
Felixmil: same as Width ?

R/sensivitity-time-profiles.R
```diff
@@ -338,37 +309,38 @@ sensitivityTimeProfiles <- function(
...
         x = plotConfiguration$xLabel,
         y = plotConfiguration$yLabel,
         title = NULL,
         color = plotConfiguration$legendTitle
       )
 
     plot <- plot +
-      theme_bw(base_size = 11) +
+      ggplot2::theme_bw(base_size = 11) +
       theme(
```
Felixmil: missing `ggplot2::`

R/sensivitity-time-profiles.R
```diff
@@ -412,16 +384,17 @@ sensitivityTimeProfiles <- function(
     patchwork::plot_annotation(
       title = plotConfiguration$title,
       subtitle = plotConfiguration$subtitle,
-      theme = theme(
-        plot.title = element_text(size = plotConfiguration$titleSize)
-      )
+      theme = ggplot2::theme(plot.title = element_text(size = plotConfiguration$titleSize))
```
Felixmil: missing `ggplot2::`

R/sensitivity-spider-plot.R
```diff
@@ -305,34 +280,34 @@ sensitivitySpiderPlot <- function(
...
-      plot <- plot + scale_x_log10()
+      plot <- plot + ggplot2::scale_x_log10()
     }
 
     if (isTRUE(plotConfiguration$yAxisScale == "log")) {
       plot <- plot +
-        scale_y_log10(
+        ggplot2::scale_y_log10(
           limits = pLimits,
           expand = expansion(mult = c(0.01, 0.1)),
```
Felixmil: Missing `ggplot2::`

R/utilities-figures.R
```diff
@@ -1158,3 +1134,139 @@ createPlotsFromExcel <- function(
     plotConfigurations = dfPlotConfigurations
   ))
 }
+
+
+#' Set ospsuite.plots defaults settings for ESQLabs workflows
+#'
+setESQTheme <- function() {
```
Felixmil: Should it be exported ?

R/utilities-figures.R
```diff
@@ -244,17 +250,8 @@ createEsqlabsExportConfiguration <- function(outputFolder) {
...
-  exportConfiguration$dpi <- 300
-  # NULL is not supported by ExportConfiguration, so we should assign here
-  # something useful. NULL in the ProjectConfiguration currently means "do not
-  # export".
-  exportConfiguration$format <- "png"
-  exportConfiguration$width <- 18
-  exportConfiguration$heightPerRow <- 12
-  exportConfiguration$units <- "cm"
+  # TODO: when TLF is removed from dependency, remove "esqlabsR::"
+  exportConfiguration <- esqlabsR::ExportConfiguration$new(path = outputFolder)
```
Felixmil: Isn't TLF removed here ?

R/utilities-figures.R
```diff
@@ -1158,3 +1134,139 @@ createPlotsFromExcel <- function(
...
+
+
+#' Return list of properties from a plot type
+#'
+#' @param plotType One of `"timeProfiles"`, `"spiderPlot"`, `"tornadoPlot`
+#' @returns A list
+#' @keywords internal
+.plotConfigurationFromType <- function(plotType = NULL) {
+  if(is.null(plotType)){
+    return(NULL)
```
Felixmil, line 1310: Print a warning ?

R/esqlabs-plot-export-configuration.R
```diff
@@ -52,7 +108,14 @@ ExportConfiguration <- R6::R6Class(
...
       }
-      super$savePlot(plotObject, fileName)
+      ospsuite.plots::exportPlot(
+        plotObject,
+        filename = fileName %||% self$name,
+        filepath = self$path,
+        width = self$width,
+        height = self$height,
+        device = self$format
+      )
```
Felixmil: where dpi and units are used ?

R/utilities-figures.R
```diff
@@ -344,111 +341,78 @@ createPlotsFromExcel <- function(
...
+            quantiles = .fieldFromExcel("quantiles", plotConfigurationRow, defaultConfiguration),
+            nsd = .fieldFromExcel("nsd", plotConfigurationRow, defaultConfiguration)
+          )
+        },
+        observedVsSimulated = {
+          defaultConfiguration <- formals(ospsuite::plotPredictedVsObserved)
+          ospsuite::plotPredictedVsObserved(
+            plotData = dataCombined,
+            yUnit = .fieldFromExcel("yUnit", plotConfigurationRow, defaultConfiguration),
+            # xyScale = .fieldFromExcel("xAxisScale", plotConfigurationRow, defaultConfiguration),
```
Felixmil: commented code: Necessary to keep ?

R/sensitivity-tornado-plot.R
```diff
@@ -265,21 +248,21 @@ sensitivityTornadoPlot <- function(
...
       ) +
-      theme(
+      ggplot2::theme(
         legend.position = plotConfiguration$legendPosition,
-        panel.grid.minor = element_blank(),
-        text = element_text(size = 11),
-        axis.text.y = element_text(margin = margin(l = 20, unit = "pt"))
+        panel.grid.minor = ggplot2::element_blank(),
+        text = ggplot2::element_text(size = 11),
+        axis.text.y = ggplot2::element_text(margin = margin(l = 20, unit = "pt"))
```
Felixmil: missing `ggplot2::`

R/sensivitity-time-profiles.R
```diff
@@ -316,19 +287,19 @@ sensitivityTimeProfiles <- function(
...
     if (isTRUE(plotConfiguration$xAxisScale == "log")) {
-      plot <- plot + scale_x_log10()
+      plot <- plot + ggplot2::scale_x_log10()
     }
     if (isTRUE(plotConfiguration$yAxisScale == "log")) {
       plot <- plot +
-        scale_y_log10(
+        ggplot2::scale_y_log10(
           limits = pLimits,
           expand = expansion(mult = c(0.01, 0.1)),
```
Felixmil: Missing `ggplot2::`

R/utilities-figures.R
```diff
@@ -126,28 +126,26 @@ col2hsv <- function(color) {
...
 
-#' @title Create an instance of `DefaultPlotConfiguration` R6 class
+#' @title Create a `DefaultPlotConfiguration` list of properties
 #' @rdname createEsqlabsPlotConfiguration
 #'
 #' @description
 #'
-#' An instance of `DefaultPlotConfiguration` R6 class from `{tlf}` package is
-#' needed for creating visualizations with the `{ospsuite}` package.
+#' An list of plot properties needed for creating visualizations with the `{ospsuite}` package.
```
Felixmil: "**A** list ..."

PavelBal:
@pchelle I decided to remove export configuration completely.
Can you please create a PR into your branch (and accept/merge after review) from this: https://github.com/esqLABS/esqlabsR/tree/remove-export-config

R/esqlabs-plot-export-configuration.R
PavelBal: @Felixmil ignore export config... I decided we don't need it.

R/utilities-figures.R
rengelke: Chck also zzz.R L8 `tlf::setDefaultMaxCharacterWidth`

R/utilities-figures.R
```diff
@@ -344,111 +341,78 @@ createPlotsFromExcel <- function(
...
-        ]$foldDistance
-        if (is.na(foldDist)) {
-          plotObservedVsSimulated(dataCombined, plotConfigurationList[[plotId]])
-        } else {
-          plotObservedVsSimulated(
-            dataCombined,
-            plotConfigurationList[[plotId]],
-            foldDistance = as.numeric(unlist(strsplit(foldDist, split = ",")))
+  # Set default plot settings prior updating based on  sheet "plotConfiguration"
+  setESQTheme()
```
rengelke: To avoid global side-effects, this should be called during setup and not from inside a function. 
or allow control via parameter e.g. `applyTheme = FALSE`

rengelke (changes requested):

R/esqlabs-plot-export-configuration.R
pchelle: This is a typo, I will fix it anyways.
But indeed with `ospsuite.plots`, you can also define plot export configuration (height is not in there though...):

``` r
grep(pattern = "export", names(ospsuite.plots::getDefaultOptions()), value = TRUE)
#> [1] "ospsuite.plots.exportWidth"  "ospsuite.plots.exportUnits" 
#> [3] "ospsuite.plots.exportDevice" "ospsuite.plots.exportDpi"
```

<sup>Created on 2026-04-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/utilities-figures.R
pchelle: That's a good question.
Since this was a draft, I did not export and hard coded the content based on what was in `{esqlabsR}` but it could certainly be exported and some options be exposed for the user. Quid of the options to be exposed

R/utilities-figures.R
pchelle: @PavelBal 
What would be the best logic here ?

The default argument for `xyScale` is `NULL` and will create a log-log plot. 
Users can set their own scales in Excel PlotConfiguration however they have 2 columns `xAxisScale` and `yAxisScale` which could lead to conflicting inputs. 
I defined 3 cases below and suggested how R should handle it, let me know if this seems okay:

* 1)  `xAxisScale` and `yAxisScale` undefined in Excel $\rightarrow$ R uses default (log-log plot)
* 2)  `xAxisScale` or `yAxisScale` defined in Excel $\rightarrow$ R uses the one defined as axis scale (checking first if within lin/log axes)
* 3)  `xAxisScale` and `yAxisScale` defined in Excel $\rightarrow$ R checks if the scales are equal and within lin/log, if not throw an error about conflicting scales, if okay use the scale

PavelBal: yes, all make sense

R/utilities-figures.R
pchelle: The function is currently only a draft, I am not sure what would be best here:
Initially we had a whole family of R6 classes for plot configurations in tlf whose structure was re-used here.

Should we create a simpler plot configuration R6 class to store all the relevant properties and tune default as a function of `plotType` knowing that currently there is nothing in `{ospsuite.plots}` ?
Or is a simple named list enough for the few properties that could be defined by users ?

R/utilities-figures.R
```diff
@@ -344,111 +314,161 @@ createPlotsFromExcel <- function(
...
-            foldDistance = as.numeric(unlist(strsplit(foldDist, split = ",")))
+  # Set default plot settings prior updating based on  sheet "plotConfiguration"
+  if(applyTheme){
+    setESQTheme()
+  }
+  plotList <- lapply(
+    seq_along(dfPlotConfigurations$plotID),
+    function(rowIndex) {
+      plotConfigurationRow <- dfPlotConfigurations[rowIndex, ]
+      if (is.null(plotConfigurationRow$DataCombinedName)) {
```
Felixmil: I think this guard is dead code. `plotConfigurationRow` is a 1-row data.frame, so `$DataCombinedName` resolves to a length-1 vector — possibly `NA`, but never `NULL`. So this `return()` never fires, and a row with an empty `DataCombinedName` falls through to the `dataCombinedList[[NA]]` lookup below.

Did you mean `is.na(plotConfigurationRow$DataCombinedName)`? If so, also worth checking that `names(plotList) <- dfPlotConfigurations$plotID` on line 472 still lines up: the lapply will return NULL for skipped rows and the names need to match either way.

R/utilities-figures.R
```diff
@@ -344,111 +314,161 @@ createPlotsFromExcel <- function(
...
+        return()
+      }
+      .validateLogScaleAxisLimits(
+        plotConfigurationRow,
+        plotID = plotConfigurationRow$plotID
+      )
+      dataCombined <- dataCombinedList[[plotConfigurationRow$DataCombinedName]]
+      switch(
```
Felixmil, line 354: There's no default branch on this `switch()`. If a user has a typo in `plotType` (or an old value that's been removed), `switch()` returns `NULL` and then the `+ ggplot2::labs(...)` on line 464 throws `non-numeric argument to binary operator`, which is going to be confusing to debug from the user's side.


Would be nicer to fail fast with a clear message listing the valid plot types.

R/utilities-figures.R
```diff
@@ -344,111 +314,161 @@ createPlotsFromExcel <- function(
...
-        ]$foldDistance
-        if (is.na(foldDist)) {
-          plotObservedVsSimulated(dataCombined, plotConfigurationList[[plotId]])
-        } else {
-          plotObservedVsSimulated(
-            dataCombined,
-            plotConfigurationList[[plotId]],
-            foldDistance = as.numeric(unlist(strsplit(foldDist, split = ",")))
+  # Set default plot settings prior updating based on  sheet "plotConfiguration"
+  if(applyTheme){
```
Felixmil: This is the line I'm most nervous about. `setESQTheme()` does `ggplot2::theme_update()`, `options(ospsuite.plots.watermarkEnabled = ...)`, and `ospsuite.plots::setDefaults(...)` — all of which persist for the rest of the R session. So a user calling `createPlotsFromExcel()` once gets every subsequent ggplot in their analysis silently restyled.

A couple of options:
1. Default `applyTheme = FALSE` and let users opt in.
2. Save the current theme/options on entry and restore on exit (`withr::defer(ggplot2::theme_set(old))`).
3. At minimum, make the side effect very loud in the function's docs.

Minor: `if(applyTheme){` is missing the spaces, `air` will fix it.

R/utilities-figures.R
```diff
@@ -652,30 +633,46 @@ createPlotsFromExcel <- function(
...
-            axisLimitsField = limitsField,
-            axis = check$axis
-          ))
-        }
+    for (limitsField in check$limits) {
+      limitsValue <- plotConfiguration[[limitsField]]
+      if (is.null(limitsValue)) {
+        next
+      }
+      if (is.na(limitsValue)) {
```
Felixmil: If `limitsValue` is already a length-2 vector here, `is.na(limitsValue)` returns a length-2 logical and the `if` warns about `the condition has length > 1`. Either swap to `anyNA(limitsValue)`, or guard with `length(limitsValue) == 1 && is.na(limitsValue)` if the intent really is "single missing cell".

R/utilities-figures.R
```diff
@@ -1154,7 +1102,205 @@ createPlotsFromExcel <- function(
...
+
+#' If field from PlotConfiguration is undefined (`NA`),
+#' use default of `{ospsuite.plots}` function
+#'
+#' @param fieldName Name of field to query from Excel table
+#' @param plotConfigurationRow Plot configuration table row
+#' @param defaultConfiguration Default ESQLabs values for the configuration
+#' @importFrom stats na.exclude
+#' @keywords internal
+.fieldFromExcel <- function(
```
Felixmil, line 1224: I'd push back on the design here a bit: `.fieldFromExcel()` reaches into `plotConfigurationRow[[fieldName]]` for fields the user may not have in their sheet (`xAxisScale`, `lloqOnBothAxes`, `xyScale`, `xValuesLimits`, `yValuesLimits`, `foldDistance`, etc.). For a missing column the extract is `NULL` and `is.na(NULL)` returns `logical(0)`, which then errors as `argument is of length zero` inside the `if`.

The old `.createConfigurationFromRow` iterated `names(row)` and only acted on columns that existed, which is gentler on users with older Excel templates. Could we add a `if (!fieldName %in% names(plotConfigurationRow)) return(defaultConfiguration[[...]])` at the top, or convert `plotConfigurationRow` to a list-with-defaults earlier?

R/utilities-figures.R
```diff
@@ -1154,7 +1102,205 @@ createPlotsFromExcel <- function(
...
+    }
+    foldDistance <- .parseExcelMultiValueField(plotConfigurationRow[[
+      "foldDistance"
+    ]])
+    return(ospsuite.plots::getFoldDistanceList(folds = foldDistance))
+  }
+  if (fieldName %in% c("xAxisScale", "yAxisScale")) {
+    # Default value for a axis scale is linear
+    if (is.na(plotConfigurationRow[[fieldName]])) {
+      return("linear")
```
Felixmil: Returning `"linear"` is fine for `ospsuite.plots`, but elsewhere in this file (e.g. `.plotConfigurationFromType()` line 1264 sets `xAxisScale = "lin"`, and the sensitivity plot code uses `"lin"` too). So depending on which code path constructed the config, the same logical value is sometimes `"lin"` and sometimes `"linear"`. Worth picking one and converting at the boundar, otherwise downstream `==` comparisons will be subtly inconsistent.

R/utilities-figures.R
```diff
@@ -1154,7 +1102,205 @@ createPlotsFromExcel <- function(
...
+      ) |>
+      na.exclude() |>
+      unique()
+    # Default value for xyScale is log creating log-log obs vs simulated
+    if(isEmpty(xyScale)){
+      return(defaultConfiguration[[fieldName]])
+    }
+    # Note that at this stage .validateLogScaleAxisLimits was already assessed
+    tryCatch(
+      ospsuite.utils::validateIsOfLength(xyScale, 1),
```
Felixmil: Style nit: the `tryCatch` around a length check feels heavy. `if (length(xyScale) > 1) stop(messages$conflictingAxesScales(plotConfigurationRow$plotID), call. = FALSE)` reads more directly. Also the indentation on the `xyScale = .fieldFromExcel(...)` call (around lines 414-417 at the call site) is off — `air` will catch that.

R/utilities-figures.R
```diff
@@ -1154,7 +1102,205 @@ createPlotsFromExcel <- function(
...
+#' @param legendPosition Default position of legends as defined in `ggplot2::theme`
+#' One of "none", "left", "right", "bottom", "top"
+#' 
+#' @param legendJustification Default alignment/justification of legends as defined in `ggplot2::theme`
+#' One of "left", "right", "center"
+#' 
+#' @param watermarkEnabled Logical allowing whether ospsuite.plots watermark is enabled
+#' 
+#' @export
+setESQTheme <- function(legendPosition = "top",
```
Felixmil: Since this is exported and globally side-effecting, it'd be worth a `@section Side effects:` or a prominent `@description` note saying it modifies `ggplot2`'s global theme, sets `options(ospsuite.plots.watermarkEnabled)`, and calls `ospsuite.plots::setDefaults()` — and that those effects persist until the session ends. Right now a user reading the docs wouldn't know.

R/plot-grid.R
```diff
@@ -0,0 +1,411 @@
...
+      tag_sep = plotGridConfiguration$tagSeparator,
+      theme = ggplot2::theme(
+        plot.title = ggtext::element_textbox_simple(
+          color = plotGridConfiguration$titleColor,
+          size = plotGridConfiguration$titleSize,
+          face = plotGridConfiguration$titleFontFace,
+          family = plotGridConfiguration$titleFontFamily,
+          halign = plotGridConfiguration$titleHorizontalJustification,
+          valign = plotGridConfiguration$titleVerticalJustification,
+          hjust = plotGridConfiguration$captionHorizontalJustification,
```
Felixmil: I think this is a copy-paste bug: inside the `plot.title = ...` block, `hjust` and `vjust` (lines 87-88) point at `captionHorizontalJustification` / `captionVerticalJustification` instead of `titleHorizontalJustification` / `titleVerticalJustification`. So title justification is currently being driven by caption fields, which would explain any "setting `titleHorizontalJustification` does nothing" reports.

R/plot-grid.R
```diff
@@ -0,0 +1,411 @@
...
+          )
+        ),
+        plot.subtitle = ggtext::element_textbox_simple(
+          color = plotGridConfiguration$subtitleColor,
+          size = plotGridConfiguration$subtitleSize,
+          face = plotGridConfiguration$subtitleFontFace,
+          family = plotGridConfiguration$subtitleFontFamily,
+          halign = plotGridConfiguration$subtitleHorizontalJustification,
+          valign = plotGridConfiguration$subtitleVerticalJustification,
+          hjust = plotGridConfiguration$captionHorizontalJustification,
```
Felixmil: Same copy-paste as in the title block: `hjust` / `vjust` for `plot.subtitle` reference `captionHorizontalJustification` / `captionVerticalJustification`. Should be `subtitleHorizontalJustification` / `subtitleVerticalJustification`.

R/plot-grid.R
```diff
@@ -0,0 +1,411 @@
...
+      tag_prefix = plotGridConfiguration$tagPrefix,
+      tag_suffix = plotGridConfiguration$tagSuffix,
+      tag_sep = plotGridConfiguration$tagSeparator,
+      theme = ggplot2::theme(
+        plot.title = ggtext::element_textbox_simple(
+          color = plotGridConfiguration$titleColor,
+          size = plotGridConfiguration$titleSize,
+          face = plotGridConfiguration$titleFontFace,
+          family = plotGridConfiguration$titleFontFamily,
+          halign = plotGridConfiguration$titleHorizontalJustification,
```
Felixmil, line 86: Minor design question: passing both `halign`/`valign` *and* `hjust`/`vjust` to `ggtext::element_textbox_simple` is redundant — `halign`/`valign` control alignment within the textbox, `hjust`/`vjust` control placement of the box itself. Worth picking which axis of control you want to expose to users and dropping the other, otherwise the public field surface is confusing ("why are there two of these?").

Felixmil (commented):
Couple of concerns inline.

R/messages.R
```diff
@@ -900,3 +886,10 @@ messages$warningPIOptimizationFailed <- function(piTaskName, errorMessage) {
 messages$messageRunningPITask <- function(piTaskName) {
   cliFormat("Running PI task: {.val {piTaskName}}")
 }
+
+messages$conflictingAxesScales <- function(plotID) {
+  cli::format_message(c(
+    "x" = "{.strong observedVsSimulated} plot with plotID {.val {plotID}} has conflicting {.field xAxisScale} and {.field yAxisScale}",
+    "i" = "Use either {.val linear} or {.val log} in both columns"
+  ))
+}
```
Felixmil: Tiny nit: file ends without a trailing newline (`\ No newline at end of file` in the diff). `air` should fix it on next run.

tests/testthat/test-plot-grid.R
```diff
@@ -0,0 +1,136 @@
+test_that("plots grid produces error with wrong input type", {
+  expect_error(plotGrid(DataSet$new(name = "DS")))
+})
+
+setESQTheme()
+ls_plots <- list(
```
Felixmil, line 5: Calling `setESQTheme()` at the top level of the test file means the global theme is mutated for every test file that runs after this one in the suite. Same goes for the plot-grid setup at lines 22-39, top-level state in test files makes failures hard to localize.

Could we move both into the relevant `test_that()` blocks (or `local({...})`) and use `withr::defer(ggplot2::theme_set(ggplot2::theme_grey()))` to restore? I suspect this is also why `Plot Grid can have very long texts` needs `skip_on_os("mac")` further down, leaked theme state from earlier tests can cause SVG divergence.

tests/testthat/test-utilities-figures.R
```diff
@@ -64,15 +64,8 @@ test_that("createEsqlabsPlotGridConfiguration() creates object with chosen defau
...
-test_that("createEsqlabsExportConfiguration() creates object with chosen defaults", {
-  myProjConfig <- ProjectConfiguration$new()
-  myEC <- createEsqlabsExportConfiguration(myProjConfig$outputFolder)
-  expect_true(isOfType(myEC, "ExportConfiguration"))
-  expect_equal(myEC$units, "cm")
-})
-
-
 test_that("esqlabsPlotConfiguration fields match DefaultPlotConfiguration", {
+  skip("Skipping due to migration towards ospsuite.plots")
```
Felixmil: This test is now permanently skipped post-migration. Could we either delete it (since `createEsqlabsPlotConfiguration()` returns a list now and the comparison no longer makes sense) or rewrite it to assert the list-based contract? Skipped-but-kept tests tend to rot.

tests/testthat/test-create-plots-from-excel.R
```diff
@@ -886,90 +843,8 @@ test_that("It trows an error if a plot grid requires a plot id that is not defin
...
-      # Get the most recently created/modified folder in the Figures directory
-      latestDir <- .getLatestDirectory(file.path(tempDir, "Figures"))
-      expect_true(file.exists(file.path(latestDir, "Aciclovir1.png")))
-      expect_true(file.exists(file.path(latestDir, "Aciclovir2.png")))
-    }
-  )
-})
-
 test_that("It throws an error when trying to set a property that is not supported by the configuration", {
+  skip("Currently properties that are not supported are simply discarded")
```
Felixmil: The skip message admits this is a behavior regression — the old code errored on unknown columns; the new code silently drops them. Silent drops are exactly the failure mode the original test was guarding against (someone misspells `xLabel` and gets no warning).

Could we either restore the validation in `.fieldFromExcel` (warn or error on unknown columns), or call this out as a deliberate breaking change in NEWS.md?

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # esqlabsR (development version)
 
+## Breaking changes
+
+This version updates the figure creation workflow to use the new approach implemented in `{ospsuite}` version 12.4.2. See ["Plotting with ospsuite.plots"](https://www.open-systems-pharmacology.org/OSPSuite-R/articles/plotting-with-ospsuite-plots.html) for more details.
+
+- Removed sheet "exportConfiguration" from the "Plots.xlsx" configuration. Automated export of figures generated from excel sheets to files is not supported any more. To export figures, use `ospsuite.plots::exportPlot()` or the standard `ggplot2` export functions
+- Removed class `ExportConfiguration` that was overriding `ExportConfiguration` from the `tlf` package.
```
Felixmil, line 14: The breaking-changes section is good. Could we add a couple more notes for users with existing projects?

- `createPlotsFromExcel()` now applies a global ggplot theme by default (via `setESQTheme()`); pass `applyTheme = FALSE` to opt out.
- Excel `Plots.xlsx` files without the new columns (`lloqOnBothAxes`, `xyScale`, `xValuesLimits`, etc.) may need to be regenerated.
- Unknown columns in `plotConfiguration` are now silently ignored rather than rejected (was an error, see related comment on `test-create-plots-from-excel.R`).

R/utilities-figures.R
pchelle: Indeed meant as `is.na`, I'll double check the plotID

R/utilities-figures.R
pchelle: I am also not confident about this approach and tried to match with the previous hardcoded settings.

@Felixmil , @PavelBal and @rengelke  Which approach would make more sense for replacing this: 

- [ ] Using `setESQTheme()` when loading the package (`.onLoad`) ?
- [ ] No plot settings and leaving `setESQTheme()` for users who want to apply the styling ?
- [ ] Hard coding a plot configuration (currently as draft for sensitivity time profile, spider plot and tornado plot)
- [ ] Save the current theme/options on entry and restore on exit

PavelBal: I would prefer to first try it out with the default `ospsuite` settings, so not overriding anything in `esqlabsR`. Only apply additional defaults when necessary. For this, https://github.com/Open-Systems-Pharmacology/OSPSuite-R/pull/1912 must be solved first, I think.

Then add somewhere in the warnings, that `ospsuite` applies some deafults to the theme, and thath the original theme can be restored from ospsuiteEnv.

tests/testthat/test-utilities-figures.R
pchelle: Agree, I removed it

tests/testthat/test-create-plots-from-excel.R
pchelle: I restored the validation and included the new `lloqOnBothAxes` in expected column names

tests/testthat/test-plot-grid.R
pchelle: Because of the many side effects of `setDefaults()`, I now only define as little as possible: 

- 2 of `ospsuite.plots` options are set on load to prevent unwanted warnings (watermark and esqlabs color palette)
- `theme()` is simply added as a ggplot layer, no more `theme_update()` with potential unwanted side effects

pchelle:
@copilot resolve the merge conflicts in this pull request

Felixmil (changes requested):
need merge conflict to be solved manually (or rerun tests before commiting merge conflict resolution)

R/plot-grid.R
```diff
@@ -0,0 +1,411 @@
...
+.convertAngleToOrientation <- function(angle) {
+  # use modulo 360 to in case minus angles were provided
+  return(
+    switch(
+      as.character(angle %% 360),
+      "0" = "upright",
+      "90" = "left-rotated",
+      "180" = "right-rotated",
+      "270" = "inverted"
+    )
```
rengelke, line 16: no default for non defined angles

rengelke (changes requested):
Great, only minor issue which should be fixed

Felixmil:
@pchelle only failing snapshots on macos are the sensitivity ones (+ the ones already skipped on macos)
**Cause:** `ospsuite` imports `tlf`, whose `.onLoad` (`TLF-Library/R/zzz.R`) calls `showtext::showtext_auto()`. That hook rasterizes every glyph to filled SVG paths using the system "sans" font, which resolves to different files across OSes (Helvetica on macOS, DejaVu/Liberation on Linux). The committed snapshots were generated on Windows, so a single glyph in the title string `"Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)"` lands 0.01 pt off on macOS — same diff in all 17 files.

It was the same issue we had before for plots with long titles.

I will push a temporary fix in which we disable explicitely showtext in `tests/testthat/setup.R`.
The "permanent" fix will be to remove showtext as direct dependency from ospsuite-R (I don't know why it's actually there, showtext was tlf dependency, not ospsuite...)

rengelke (approved):

Felixmil (approved):

PavelBal (changes requested):
If `plotConfiguration` sheet contains the columns `pointsShapre` or `legendPosition`, the following error is thrown:


```
Error in `.validatePlotConfigurationFromExcel()`:
! Trying to apply property `pointsShape` and `legendPosition` that is not supported by the configuration `DefaultPlotConfiguration`!
```

pchelle:
@rengelke 

The following test is failing which seems to be related to the parameter identification package.
Should I create a specific issue for that ?

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Error ('test-utilities-pi.R:560:3'): createPITasks applies objectiveFunctionOptions from PIConfiguration columns ──
Error in `ospsuite.utils::validateIsOption(value, ObjectiveFunctionSpecs[names(value)])`: Option validation failed:

residualWeightingMethod : "std" not allowed.
Allowed values: "none" and "error".
Backtrace:
    ▆
 1. ├─esqlabsR::createPITasks(readPITaskConfigurationFromExcel(projectConfiguration = projectConfigurationLocal)) at test-utilities-pi.R:560:3
 2. │ └─esqlabsR:::.createSinglePITask(...) at esqlabsR/R/utilities-pi.R:52:5
 3. │   └─esqlabsR:::.createPIConfigurationFromConfig(configurations = piTaskConfiguration$piConfiguration) at esqlabsR/R/utilities-pi.R:158:3
 4. └─ospsuite.parameteridentification (local) `<fn>`(base::quote(`<named list>`)) at esqlabsR/R/utilities-pi.R:568:3
 5.   └─ospsuite.utils::validateIsOption(value, ObjectiveFunctionSpecs[names(value)]) at ospsuite.parameteridentification/R/PIConfiguration.R:82:9
 ```

**Comment by PavelBal** [comment]
Should this be added to package load?

```
if (requireNamespace("showtext", quietly = TRUE)) {
  showtext::showtext_auto(enable = FALSE)
}
```

I am tired of the inconsistencies originating from showtex... I think we already disable this for the tests - but then we should also disable it globally.

pchelle:
> Should this be added to package load?
> 
> ```
> if (requireNamespace("showtext", quietly = TRUE)) {
>   showtext::showtext_auto(enable = FALSE)
> }
> ```
> 
> I am tired of the inconsistencies originating from showtex... I think we already disable this for the tests - but then we should also disable it globally.

Thanks to @Felixmil implementation of `geom_point_osp()`, `showtext` should not be in a dependency of `ospsuite.plots` or `esqlabsR` anymore. 
So, by migrating to `ospsuite.plots`, we would not need to run this showtext code.

Felixmil:
> > Should this be added to package load?
> > ```
> > if (requireNamespace("showtext", quietly = TRUE)) {
> >   showtext::showtext_auto(enable = FALSE)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am tired of the inconsistencies originating from showtex... I think we already disable this for the tests - but then we should also disable it globally.
> 
> Thanks to @Felixmil implementation of `geom_point_osp()`, `showtext` should not be in a dependency of `ospsuite.plots` or `esqlabsR` anymore. So, by migrating to `ospsuite.plots`, we would not need to run this showtext code.

The problem is that we import `{ospsuite}`, which still relies on `{tlf}` and automatically loads it.

pchelle:
> > > Should this be added to package load?
> > > ```
> > > if (requireNamespace("showtext", quietly = TRUE)) {
> > >   showtext::showtext_auto(enable = FALSE)
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I am tired of the inconsistencies originating from showtex... I think we already disable this for the tests - but then we should also disable it globally.
> > 
> > 
> > Thanks to @Felixmil implementation of `geom_point_osp()`, `showtext` should not be in a dependency of `ospsuite.plots` or `esqlabsR` anymore. So, by migrating to `ospsuite.plots`, we would not need to run this showtext code.
> 
> The problem is that we import `{ospsuite}`, which still relies on `{tlf}` and automatically loads it.

Got it ! 
The deprecation of the `tlf` dependency may still take time.
Indeed, better preventing showtext to randomly be enabled.

PR #1004, opened 2026-04-19
Rewrite tests for better performance

PavelBal:
Test runtime reduced from 14 to 7 minutes. No snapshot updates -> tests contents did not change

## Release Notes

* **Documentation**
  * Updated example parameter paths across sensitivity analysis functions (sensitivity calculation, spider plot, time profiles, and tornado plot) to reflect current data structure.

* **Chores**
  * Enabled parallel test execution for improved test performance.
  * Refactored internal test infrastructure to use shared fixtures for consistency and maintainability.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -62,6 +62,7 @@ Suggests:
 VignetteBuilder:
     knitr
 Config/testthat/edition: 3
+Config/testthat/parallel: true
 Config/rcmdcheck/ignore-inconsequential-notes: true
 Encoding: UTF-8
 Language: en-US
--- a/R/sensivitity-time-profiles.R
+++ b/R/sensivitity-time-profiles.R
@@ -101,7 +101,7 @@
 #'   parameterPaths = namedParameterPaths
 #' )
 #'
-#' sensitivitySpiderPlot(resultsNamed)
+#' sensitivityTimeProfiles(resultsNamed)
 #' }
 #' @export
 sensitivityTimeProfiles <- function(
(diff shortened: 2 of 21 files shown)


PavelBal:
@coderabbit full freview

nautilus69 (changes requested):
I suggest test-enum.R may need to be excluded from parallel execution, because of the nature of .NET runtime. it may fail for no code related reasons on windows machine. 

I've found out that Two different Aciclovir.pkml files are used:
  - ospsuite::system.file("extdata", "Aciclovir.pkml") : uses Applications|...
  - TestProject/Models/Simulations/Aciclovir.pkml : uses Events|...

I am not sure if these should been one or not. I had to accept a lot of snapshots when to make the tests passed. 
In addition a minor change for the sake of clean code and readability: 

tests/testthat/create-data-combined-objects.R (lines 7-9)

 

  Each DC is stored as `.testFixtures$fooDC` and exposed via an accessor that
  returns a deep clone, so tests can safely mutate what they receive without
  poisoning other tests (same pattern as OSPSuite-R PR #1859).

 should change to: 

  Each DC is stored as `.testFixtures$fooDC` and exposed via an accessor.
  For data.frames, R's copy-on-write semantics provide automatic isolation
  when tests mutate their copy.


"Deep Clone" is misleading.

PavelBal:
Probably it does not make sense to do it now. Lets tackle this again, when the new project structure and plotting with ospsuite.plots is implemented.

Issue #1003, opened 2026-04-17
Run tests in parallel


PavelBal:
And implement other GHA improvements we had recently in ospsuite.r

PR #1005, opened 2026-04-27
Parallel tests

pchelle:
Not sure which other GHA improvements are suggested in #1003

## Summary by CodeRabbit

* **Chores**
  * Enabled parallel execution for the test suite.
  * Excluded renv artifacts from package builds.
  * Cleaned up print method registration metadata.
* **Bug Fixes**
  * Adjusted package initialization to keep SVG text portable when showtext is available.
* **Tests**
  * Removed macOS skips so plot and sensitivity snapshot tests run on macOS.
  * Added a safety check to ensure plot generation succeeds before snapshot comparisons.
  * Tuned test-session options for stricter condition-length checks and quieter lifecycle messaging.


tests/testthat.R
```diff
@@ -5,4 +5,9 @@ library(ospsuite.utils)
 # Disable these warnings for the tests
 options(warnPartialMatchDollar = FALSE)
 
+numCores <- parallel::detectCores()
+numCores <- if (is.na(numCores)) 1L else numCores
+Sys.setenv(TESTTHAT_PARALLEL = "true")
```
PavelBal: You don't have to set it here, if `Config/testthat/parallel: true` is already defined in the DESCRIPTION.

tests/testthat.R
```diff
@@ -5,4 +5,9 @@ library(ospsuite.utils)
...
+Sys.setenv(TESTTHAT_CPUS = as.character(numCores))
+
```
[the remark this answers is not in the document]
PavelBal: I think the "hardcoded" (which is not really hardcoded - it gets the number of cores for the respective machine) setting of number of CPUs is okay. `testthat` by default uses only two cores, so if tests are run on a machine with more cores, we are wasting computation power

DESCRIPTION
```diff
@@ -65,6 +65,7 @@ VignetteBuilder:
     knitr,
     quarto
 Config/testthat/edition: 3
+Config/testthat/parallel: true
 Config/rcmdcheck/ignore-inconsequential-notes: true
```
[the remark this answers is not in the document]
Felixmil: this is not a user facing change

tests/testthat/setup.R
```diff
@@ -1 +1,31 @@
...
+#
+# `tlf` (a transitive dep of `ospsuite`) calls `showtext::showtext_auto()` on
+# load, which makes every graphics device rasterize glyphs to filled SVG paths
+# using FreeType + the system "sans" font. That font resolves to different
+# files on different OSes (Helvetica on macOS, DejaVu/Liberation on Linux),
+# producing sub-pixel differences in glyph outlines and breaking byte-exact
+# vdiffr snapshot comparisons across platforms. Disabling showtext makes
+# svglite emit portable `<text>` elements with the actual string content
+# instead of platform-dependent glyph paths.
+if (requireNamespace("showtext", quietly = TRUE)) {
```
PavelBal: Should this maybe be `onLoad`, so not only the tests, but the actuall work with the package is not affected? Otherwise we might muss any ugly behabvior.

Felixmil: I agree

Felixmil: done in [351187d](https://github.com/esqLABS/esqlabsR/pull/1005/commits/351187d33e47b9c17f687d26479f83b173b929e3)

PavelBal (approved):

Issue #700, opened 2024-08-14
All messages must be stored in the `messages.R`


PavelBal:
https://github.com/esqLABS/esqlabsR/blame/7ccdedf7f2c6ce1aee7a0bcf955161a4224f8b60/R/utilities-data-combined.R#L85

PR #1017, opened 2026-05-04
JSON Driven ESQlabsR Projects

### LINKED ISSUE #1048: Project core: section fields bypass invalidation, clone safety, serialization and lifecycle edge cases

## Decision needed: mutation contract for section fields

`scenarios`, `outputPaths`, `individuals`, `plots`, `parameterIdentification`, and the other section fields are plain public R6 fields (R/project.R:241-278), so direct writes bypass `private$.invalidate()` entirely. Verified consequences:

- `project$scenarios[["X"]] <- sc` flips neither `modified` nor `validatedSinceMutation`, so after one `validateProject()` call, `runScenarios()` / `createPlots()` / `runPI()` skip re-validation of changed data (the exact hazard the flag exists for).
- `sync()` partially recovers via content diff but misattributes the drift as `json_modified` ("JSON file has been modified externally") instead of `unsaved_changes`.
- The hazard is documented as the recommended usage: the `createScenariosFromPKML()` example tells users to run `project$scenarios <- c(project$scenarios, scenarios)`.

Since #1047 dropped the R6 delegate methods, the standalone `add*`/`remove*` functions are the only safe surface, but there is no object-accepting mutator (`addScenario()` only builds from scalar arguments). Either:

1. make section fields active bindings that call `private$.invalidate()` on set, or
2. declare direct field writes unsupported, and add the missing object-accepting mutators so the docs never need to recommend the bypass.

This decision gates what the chapter 9 vignettes may document.

## Serialization and field-shape bugs

- [x] R/project.R:237-239: `outputPaths` is documented as a "Named character vector" but is a named **list** (parser at project.R:521, mutators, and test-project.R:219 all agree). Following the doc corrupts saves: jsonlite serializes a named atomic vector as a JSON **array**, silently dropping every id (verified; the length-1 case degrades to a bare string). Fix the doc and add a serializer guard in `.outputPathsToJson()`.
- [x] R/project-to-json.R:38: saving an empty `Project$new()` writes `"schemaVersion": null`, which `loadProject()` then rejects ("Unsupported schemaVersion"). Default it like the Excel bridge does (`%||% "2.0"`, project-excel.R:51).
- [x] R/scenarios.R:50-66 + R/project-to-json.R:192-207: `steadyState: false` with a declared `steadyStateTime`/`steadyStateTimeUnit` parses fine but serializes both fields to `null`, so a load/save cycle silently loses data, violating the serializer's own structural-identity contract (project-to-json.R:6-7).
- [x] R/individuals.R:17-28 vs R/populations.R:26-39: individuals round-trip through a field whitelist (unknown JSON fields silently deleted on load/save) while populations pass unknown fields through. Pick one forward-compatibility behavior.
- [x] R/scenarios.R:41-46: a declared `simulationTimeUnit` is silently dropped when `simulationTime` is null.
- [x] R/project-to-json.R:172-176: `populationId` emission is keyed off the derived `simulationType`, so drifted objects (both are plain public fields on `Scenario`) lose `populationId` silently on save.

## R6 hygiene

- [x] ~~`Project$clone()` is unsafe: scenarios live inside a list field, so neither `clone()` nor `clone(deep = TRUE)` copies the `Scenario` R6 objects; mutating a clone's scenario mutates the source while both `modified` flags stay FALSE (verified empirically). Add a `deep_clone` hook or set `cloneable = FALSE`. There are zero clone tests package-wide.~~ Resolved differently from the proposed `deep_clone` / `cloneable = FALSE`: `Scenario` is now a plain-data record (copy semantics), so `Project$clone()` produces fully independent section data with no hook needed and `cloneable = TRUE` kept. The one remaining shared-by-reference case (programmatic `DataSet` objects added via `addObservedData()`, which wrap external handles) is documented on the class. Clone-semantics regression tests added.
- [x] R/project.R:307-311: `.markSaved()` clears `validatedSinceMutation`, contradicting the field's own docs ("cleared by any mutation") and forcing a redundant full re-validation after every `saveProject()`. Pin the intended behavior either way with a test.
- [x] R/project.R:421-431: `print()` calls `length()` on the `plotConfiguration` / `plotGrids` data frames, reporting **column** counts as plot counts.

## Lifecycle paths

- [x] R/project-lifecycle.R:144: `isProjectInitialized()` globs `*Project*.xlsx` against the full path returned by `fs::dir_ls()`, so any directory whose own path contains "Project" (most project folders) false-positives. Glob on the basename.
- [x] R/project-lifecycle.R:198-207: `initProject(overwrite = FALSE)` over an existing project calls `utils::menu()` with no `interactive()` guard; scripts and CI die with a raw "menu() cannot be used non-interactively" instead of a cli abort suggesting `overwrite = TRUE`.
- [x] R/project-lifecycle.R:77-99: `saveProject(project, newPath)` clears `modified` but leaves `jsonPath` / `projectFilePath` / `projectDirPath` pointing at the old file: `sync()` then reports against a file that was not written, the next bare `saveProject(project)` writes to the old location, and relative paths keep resolving against the old directory. Rebind on save-as, or document the asymmetry.
- [x] R/project-lifecycle.R:80-96: `saveProject()` duplicates `.saveProjectJson()` instead of delegating, and loses its parent-directory guard; bad paths surface as raw connection errors.
- [x] R/project-lifecycle.R:214-219: the `file.copy()` result is discarded, so partial template copies pass silently.
- [x] R/project.R:503-508: `loadProject()` on a corrupt non-JSON file leaks a raw jsonlite lexical error with no file context.

## Required tests

The full suite passes today (1014 tests) while all of the above is broken or unpinned. The fixes must land with regression tests, in particular:

- empty `Project$new()` save then reload,
- clone semantics (whatever the decision is),
- `steadyState: false` plus non-null `steadyStateTime` round trip,
- a regression test for the section-field mutation decision (direct write must either invalidate or be impossible),
- save-as path rebinding, corrupt-file load, and the `initProject()` minimal / `createExcel = FALSE` / decline-overwrite paths (all currently untested),
- `.warnOnCrossReferenceErrors()` (currently zero tests; no fixture with a dangling reference is ever loaded).

### LINKED ISSUE #1051: Validation gaps: parameter-set validator never runs, missing cross-reference checks, cli glue injection
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. The two headline findings were reproduced by execution.

`validateProject()` is the gate that `runScenarios()` / `createPlots()` / `runPI()` rely on, but it has holes exactly where the chapter seams are: things it waves through then fail later with worse errors, or never fail at all.

## Findings

- [x] **Critical**, R/validation.R:435-453: `.validateParameterSets()` reads `set$paths` / `set$values` / `set$units`, a shape neither the parser nor the mutators produce (sets are lists of `{containerPath, parameterName, value, units}` records), so `set$paths` is always NULL and every set short-circuits at `length(paths) == 0`. **All three parameter-set sections are effectively unvalidated.** Reproduced: a set containing `list(containerPath = "", parameterName = "", value = "abc", units = NULL)` passes with zero critical errors and zero warnings. The only existing test (test-validation.R:292-299) feeds the legacy shape, which is why this never fired. Related to #1026 (validators should operate on the JSON shapes directly); fixing this validator is the concrete first step.
- [x] R/validation.R:255-273: `.abortValidationErrors()` passes raw validation messages (which embed user-controlled ids) into `cli::cli_abort()` bullets where they are glue-interpolated. A scenario named `Dose {mg}` makes the abort itself crash ("Could not evaluate cli `{}` expression"); `S{1}` silently renders as `S1`. Escape braces before interpolation.
- [x] Scenario to Output Path is the one inbound reference with **no cross-reference check**: it is in `.warnIfReferenced()`'s entity list (R/validation.R:342-344) and the docstring (280-282) promises the next `validateProject()` will surface dangling references, but no check exists. Reproduced end to end: `removeOutputPath()` of a path a scenario uses, then `validateProject()` reports zero critical errors, then `saveProject()` aborts in `.scenariosToJson()` (R/project-to-json.R:140-145).
- [x] dataCombined `observed[*]$dataSet` references are never checked (R/validation.R:662-682 checks only `simulated`), and `removeObservedData()` is the only remover in the package without a referenced-elsewhere warning, so a dangling dataset reference is invisible until `createDataCombined()` fails at run time. A fully static name check is impossible for excel/script sources, but the cheap structural check (non-empty `dataSet` per observed row) plus the removal warning are both feasible.
- [x] R/plots.R:115-209: `.validatePlots()` never checks `plotType` membership in `.validPlotTypes` (only `addPlot()` enforces the enum), so an unknown type coming from JSON passes `validateProject()` and then crashes `createPlots()` with `attempt to set an attribute on NULL` (reproduced), or silently drops the plot from a mixed grid.
- [x] R/plots.R:196-204 vs R/create-plots.R:410-412: severity mismatch. The validator demotes unknown plot-grid `plotIDs` to a warning with the rationale "unknown grid ids fail softly during plot creation", but `createPlots()` hard-errors on exactly that condition via `.validatePlotGridsFromExcel()`. Align the two (and fix the docstring, which is factually wrong).
- [x] R/plots.R:183-206: the plotGrid reference check is skipped entirely whenever `plotConfiguration` is empty, which is exactly the state after removing the last plot (a flow `removePlot()` explicitly warns about).
- [x] R/parameter-identification.R:364-416, 466-483: `.validatePI` does not check that parameter/mapping `scenarios` are a subset of the task's `scenarios`; such configs pass `validateProject()` then hard-fail the runtime build. The runtime is also internally inconsistent about it: step 1 silently skips unknown scenarios while steps 2 and 3 abort.
- [x] R/validation.R:71-73, 98-105: `validationResult$set_data()` and the `data` field are never populated by any adapter (`get_summary()$has_data` is always FALSE) although the documented contract says the result carries the validated data; `validationSummary()` (779-804) still reports `files_with_errors` / `files_with_warnings` though the keys are sections, not files. Implement or trim.
- [x] Scenario `modelFile` existence is never checked by `validateProject()`: the scenarios validator (R/scenarios.R, `.validateScenarios()`) only verifies `modelFile` is a non-empty string, never that the referenced `.pkml` resolves on disk. A scenario pointing at a missing or renamed model file passes `validateProject()` with zero critical errors and no warning, then fails late inside `ospsuite::loadSimulation()` at `runScenarios()` time (R/scenario-execution.R, the `loadSimulation()` call after `modelFilePath` resolution) with an upstream OSPS message that carries no scenario or project context. This is the surviving half of the `must_work` removal tracked in #1031: `.clean_path()` (R/project.R) was deliberately made side-effect-free (no `mustWork` stat) to stop accessor/`print()` reads from emitting `fileNotFound` warnings, which is correct, but nothing took over the existence check it used to do. The fix belongs here, not in `.clean_path()`: add a filesystem-existence check to the scenarios validator that emits a `add_warning("File Not Found", ...)` (or critical error) per scenario whose resolved `modelFile` is absent, mirroring the pattern `.validateObservedData()` already uses for `excel` / `pkml` / `script` entries (R/observed-data.R). Reuse the same path-resolution rule as the runtime (absolute `modelFile` used as-is, relative resolved against `project$modelFolder`). Folder fields (`modelFolder` / `dataFolder` / `configurationsFolder`) lost the same lazy stat; decide whether a missing-but-set folder is worth its own validator warning too.

## Required tests

The suite passes today (1014 tests) while the parameter-set validator validates nothing; that is the headline lesson. **Validation tests must build their fixtures through the parser or the public mutators, never hand-rolled shapes**, otherwise the validator and the data can drift apart again unnoticed. The fix PR must include:

- an invalid parameter set in the real parsed shape that `validateProject()` must flag (critical error), for all three parameter-set sections,
- cross-reference tests for scenario to outputPath and dataCombined to observed dataSet,
- a scenario name containing glue metacharacters (`{`, `}`) driven through `.abortValidationErrors()`,
- an unknown `plotType` fixture that `validateProject()` must flag,
- the currently untested cross-reference branches (grep-verified zero coverage): scenario to `populationId`, scenario to `modelParameterSets`, scenario to `applicationProtocol`, application to `parameterSets`, dataCombined to `scenarios`, and PI outputMapping to `outputPathId`,
- a scenario whose `modelFile` does not resolve on disk (relative against `modelFolder`, and absolute), asserting `validateProject()` surfaces it up front rather than deferring to the `loadSimulation()` abort at run time.

### LINKED ISSUE #1052: createPlots()/createDataCombined() break on plots outside grids, grid subsets, and round-tripped projects
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. The first four findings were reproduced by execution.

`createPlots()` and `createDataCombined()` work for the Example fixture but break on perfectly valid projects: plots outside grids, grid subsets, projects round-tripped through JSON, and plots built via `addPlot()`.

## Findings

- [x] **Critical**, R/create-plots.R:62-74 and 443-446: `createPlots()` validates the **full** `plotConfiguration` against a DataCombined list built only for the requested grids (`cfgDf` is never filtered, although the comment at line 62 claims it is). Reproduced consequences: (1) a Plot that belongs to no Plot Grid makes even `createPlots(project)` with all grids abort; (2) requesting a `plotGridNames` subset aborts whenever other plots use other DataCombined. The error is doubly misleading: it claims a DataCombined that exists in the project "is not present", and says "sheet" (Excel vocabulary) on the JSON path.
- [x] R/plots.R:292-303 and 387: `.namedDotsAsRow()`'s `list(v)` wrapping is undone by `as.data.frame(newRow)`, so `addPlot(p, "P1", "DC1", "population", quantiles = c(0.05, 0.5, 0.95))` (its documented usage) writes 3 duplicate rows with a mangled column name; `createPlots()` then aborts on duplicate plotIDs. Needs `I(list(v))`.
- [x] R/create-plots.R:473 and 555: unguarded `title` access errors with "subscript out of bounds" when the column is absent. This is reachable from persisted projects: `.dataFrameToListOfLists()` drops NA cells on save, so a project where no plot sets a title round-trips to JSON without the field and reloads without the column. The `subtitle` access right below shows the correct `%in% names(row)` guard.
- [x] R/create-plots.R:521-524: same class of bug for `foldDistance` in the observedVsSimulated branch (`if (is.na(foldDist))` on NULL gives "argument is of length zero"); the population branch (509-516) shows the correct `!is.null(x) && !is.na(x)` pattern.
- [x] R/create-plots.R:59-64: unknown `plotGridNames` are silently dropped; a typo returns `list()` with no diagnostic even with `stopIfNotFound = TRUE`, while `createDataCombined()` aborts on unknown `dataCombinedNames` (R/data-combined.R:54-56). Align.
- [x] R/create-plots.R:566-580: the `warningInvalidPlotID` soft path is unreachable (`.validatePlotGridsFromExcel()` already aborted on the same condition), so it is dead code and there is no soft path for invalid grid references at all.
- [x] R/data-combined.R:196-209 and 237-264: with `stopIfNotFound = FALSE`, a simulated entry skipped for a wrong output path is not removed from `dfDataCombined`; if it declares offsets, the transform block then operates on an all-NA row and surfaces an opaque .NET `System.MissingMethodException` from `toUnit(NA, ...)`. The missing-scenario and missing-dataset paths do remove their rows (368-371, 383-385); this branch should too.
- [x] R/data-combined.R:180-209: a Scenario Result whose run failed (`results = NULL`, which `.collectScenarioResult()` produces with only a warning) is misdiagnosed as a wrong output path (`messages$stopWrongOutputPath`). Detect the failed-run case and report it as such.
- [x] R/data-combined.R:33-34: `dataCombinedNames` is not validated while `plotGridNames` is.
- [ ] Type-specific plot fields (`quantiles`, `aggregation`, `nsd` on non-population plots; `foldDistance` on non-observedVsSimulated plots) are silently ignored, with no warning from either `.validatePlots()` or `createPlots()`. Decide whether to warn.

Unknown `plotType` passing `validateProject()` and the validator/runtime severity mismatch for grid references are tracked with the validation gaps in #1051. The legacy `*FromExcel` validator naming and DataFrame reshape are tracked in #1026.

## Required tests

The suite passes today while all of the above is broken, because only the `individual` Plot Type is ever dispatched end to end: `population`, `observedVsSimulated`, `residualsVsSimulated`, and `residualsVsTime` are never built in any test, and the entire transformation block (offsets and scale factors, R/data-combined.R:230-290) is untested (every fixture value is null). The fix PR must include:

- each plot type built end to end,
- a project with a plot outside any grid, and `createPlots(plotGridNames = subset)`,
- a save, load, `createPlots()` round trip (catches the dropped-column crashes),
- offsets/scale factors actually applied,
- `addPlot()` with vector-valued fields,
- `createDataCombined(stopIfNotFound = FALSE)` at the public level with a wrong path plus declared offsets.

### LINKED ISSUE #1053: Parameter identification: builder bounds ordering, dead units and stopIfParameterNotFound, weight round trip
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. Findings marked "reproduced" were verified by execution against the TestProject fixture.

The runPI tryCatch narrowing itself (f52efcbe) checked out as correct: build errors propagate, only the optimisation run is wrapped, and no cleanup was orphaned. The remaining issues are in the runtime builder and the record round trip.

## Findings

- [x] R/parameter-identification.R:494-496: the builder assigns `minValue` and `maxValue` **before** `startValue`, but the upstream `PIParameters` setters validate bounds against the current start value, which at that point is still the model default. Any record whose declared bounds do not bracket the model default hard-fails the build with a misleading message (reproduced: bounds 0.2/0.8 vs model default 1 gives "The maximal value cannot be smaller than the start value!"). Fix: assign `startValue` first.
- [x] R/parameter-identification.R:476-498: `PIParameter$units` is parsed, stored, documented, printed, and serialized, but never applied to the runtime (`runtime$unit <- p$units` is missing), so bounds and start values are silently interpreted in the display unit.
- [x] R/parameter-identification.R:649-653: the `stopIfParameterNotFound` argument of `runPI()` is dead: documented as "Forwarded to `.prepareScenario()`", never referenced in the body, and `.prepareScenario()` does not accept it (R/scenario-execution.R:244 hardcodes TRUE). `runPI(project, stopIfParameterNotFound = FALSE)` is a silent no-op.
- [x] R/parameter-identification.R:338 + R/project-to-json.R:366: `weight` does not survive a JSON round trip (reproduced): a numeric vector reparses as a bare list, a scalar double as integer; the constructor coerces the other numeric fields with `as.double()` but not `weight`. Downstream, `setDataWeights(setNames(list(m$weight), ...))` then nests a list inside a list.
- [x] R/parameter-identification.R:702-718: each task is built **inside** the run loop, after previous tasks were already optimised; a late build error (path typo, missing dataset) propagates out of `runPI()` and discards all completed optimisation results, potentially hours of compute. Build every runtime first (fail fast on configuration), then optimise.
- [x] R/parameter-identification.R:884-886, 979-981: auto-generated ids use `length(list) + 1L`, which collides after a removal (reproduced: add, auto-add, remove the first, auto-add again aborts with "already exists" and no hint that an explicit `id` is the workaround). Scan for a free id instead.
- [x] R/parameter-identification.R:793-804: `addPITask()` dereferences `m$outputPathId` over the raw `outputMappings` argument before any type check, so malformed input dies with raw base errors ("$ operator is invalid for atomic vectors", "argument is of length zero") instead of `errorPIWrongElementType` (reproduced with both shapes).
- [x] R/parameter-identification.R:75-141: `PIOutputMapping()` validates the id/reference fields but not `scaling`, the offsets/factors, or `weight`; `as.double("a")` yields NA with a warning, and an NA offset later kills the build with "missing value where TRUE/FALSE needed" (from `hasNonDefaultTransform`).
- [x] R/validation.R:336-359: `.warnIfReferenced()` does not cover PI inbound references: removing an output path referenced by a `PIOutputMapping`, or a scenario referenced by a `PITask`, warns nothing (the dangling reference only surfaces at the next `validateProject()`).
- [x] R/parameter-identification.R:184: bare `stop(messages$errorPIWrongElementType(...))` missed by the cli sweep c2d6089d (the parallel `parameters` branch at line 170 was converted).

## Required tests

test-parameter-identification.R is 1275 lines, yet none of the bugs above are covered (the suite passes today). Grep-verified gaps the fix PR must close:

- `runPI(piTaskNames = ...)`: zero occurrences anywhere, neither subset selection nor the unknown-name abort,
- the explicit `observedData` argument at the `runPI()` level,
- all of `.buildPIConfiguration()` (the `type` to `objectiveFunctionType` remap and the `simulationRunOptions` block),
- the non-default `setDataTransformations` / `setDataWeights` builder branches, and the `weight` round trip,
- a bounds-not-bracketing-model-default regression test,
- duplicate mapping ids in `.validatePI`, removal with an unknown `taskId`, `addPIParameter` explicit duplicate-id abort, and `runPI()` on a project with zero PI tasks.

### LINKED ISSUE #1054: Scenario Results save/load round trip loses population and re-derives outputValues

The save/load round trip for Scenario Results violates the documented four-field record contract (`simulation`, `results`, `outputValues`, `population`).

## Findings

- [x] R/scenario-results.R:148-153: `loadScenarioResults()` reconstructs a **three-field** record with no `population`, even though `saveScenarioResults()` writes `<scenario>_population.csv` for population scenarios (lines 69-77) and `runScenarios()` always produces the four-field record. The population CSV that was written is never re-read; every consumer relying on the `population` key breaks after a reload.
- [x] R/scenario-results.R:143-147: reload extracts `outputValues` with `quantitiesOrPaths = results$allQuantityPaths` instead of the scenario's recorded `outputPaths` (compare R/scenario-execution.R:336-339), so the reloaded `outputValues` are not guaranteed to equal what `runScenarios()` produced.
- [x] R/scenario-results.R:28-92: the per-scenario `tryCatch` in `saveScenarioResults()` downgrades any error to a misleading base `warning("Cannot save to path '...'")` plus `message(cond)` (the real error may have nothing to do with the path) and silently continues the loop. Convert to cli with the actual error attached, and decide whether a failed save should be loud.
- [x] R/scenario-results.R:30: the parameter is still named `projectConfiguration` for what its own roxygen calls "A `Project` object" (a retired term per the project vocabulary), and the `@param` doc still describes the old three-field record shape (with a `SimulatioResults` typo). Renaming the argument is breaking, so this needs a deliberate decision plus a deprecation shim.
- [x] Naming decision: the developer vocabulary doc says Scenario Results are "persisted by `exportScenarioResults()`, reloaded by `importScenarioResults()`" while the exported names are `saveScenarioResults()` / `loadScenarioResults()`. Reconcile one way or the other.

## Required tests

There is **no save/load round-trip test at all** today: `saveScenarioResults()` has zero tests and `loadScenarioResults()` has a single negative test (error on missing folder). That is why both round-trip bugs are invisible while the suite passes. The fix PR must include round trips for an individual scenario and a population scenario, asserting the four-field record shape and `outputValues` equality with the original run.

Felixmil:
> **This PR body is the live status ledger for the 6.0.0 project-model rework.** The design and decision record (why the format changed, the rationale trail, the superseded decisions) lives in issue **#1088**. The former umbrella #1030 is wound down; its open items were re-homed to #1088.

## What this PR delivers

This is the consolidation branch for the full 6.0.0 project-model rework of esqlabsR. The JSON-primary chapters and wave-1 fixes were assembled here first (folding reviewed, test-pinned work onto this branch); the definitions-tree layer that builds on that foundation then lands on top of it as a stacked, human-reviewable PR series. The branch merges to `v6` once, after a final review pass.

The branch carries chapters 1–8d (the foundational schema through the Excel-deprecation polish) on top of `v6`, plus all of wave 1, plus the whole migration stack (#1089–#1101 and the review-round-1 layer #1104, all merged onto `json-based-project`). Remaining before the final review pass: the open review/lifecycle PRs (#1117, #1119, #1121) and the JSON Schema disposition (#1068 / #1069). The bug items of #1055 also landed via #1060; its remaining cross-section asymmetry sweep and PI inbound-reference warnings stay open for 6.0.x. The Excel bridge round-trip (#1049) and the initial-conditions port (#973, via #1083) have since landed in the migration stack and are no longer separate pending items.

## The definitions-tree layer

Building on the JSON-primary foundation, the definitions-tree layer makes a tree of one-per-definition JSON files under `definitions/` the source of truth, with Excel demoted to import/export. The design rationale is in #1088. Two decisions from the original pitch were superseded during the build:

- **Explicit-save, not write-through.** The layer originally landed as write-through (every `add*`/`set*`/`remove*` wrote to disk immediately, no `saveProject()`). #1117 then replaced that with an explicit-save lifecycle: memory is the source of truth for a loaded project, and `saveProject(project)` reconciles the on-disk tree. See #1117 for the lifecycle verbs and the dirty-state / undo surface.
- **"definition", not "entity".** #1104 (via #1114) renamed the code-level "entity" vocabulary (entity-files, entity-id, entity-list) to "definition" throughout. The on-disk `definitions/` tree name is unchanged.

**#1104 is the effort's canonical reference** (based directly on `json-based-project`; now merged).

## Status ledger

Everything that makes up the 6.0.0 delta, in the order it was assembled: the JSON-primary chapter stack, the wave-1 correctness fixes, the migration stack that swaps in the definitions tree, and the open post-migration review/lifecycle PRs, then the JSON Schema disposition.

| PR | What it adds | Status |
|---|---|---|
| #1017 | Chapter 1 — v2.0 schema, internal `Project` R6 class, parser and serializer | ✅ Merged |
| #1018 | Chapter 2 — typed `Scenario` data class; `Project$.parseScenarios()` | ✅ Merged |
| #1020 | Chapter 3 — `runScenarios()` and `loadProject()` drive off the parsed `Project` | ✅ Merged |
| #1023 | Chapter 4 — `validateProject()` with per-section adapter dispatch | ✅ Merged |
| #1022 | Chapter 5 — `loadObservedData`, `createDataCombined`, `createPlots` accept `Project` | ✅ Merged |
| #1032 | Chapter 6 — full `add*` / `remove*` mutation surface across all sections | ✅ Merged |
| #1037 | Chapter 7 — `runPI()` drives off the parsed `Project`; legacy `PITaskConfiguration` removed | ✅ Merged |
| #1044 | Chapter 8a — `ProjectConfiguration` and `Project` merged into one exported class | ✅ Merged |
| #1045 | Chapter 8b — `saveProject()`, `initProject()`, `createScenariosFromPKML()` rewrite | ✅ Merged |
| #1046 | Chapter 8c — per-domain file reorganisation; test fixture rationalisation | ✅ Merged |
| #1047 | Chapter 8d — `runPI()` failure semantics, cli convergence, snapshot refresh | ✅ Merged |
| #1060 | Wave-1 (closes #1048, #1050) — Project core / mutation surface: active-binding invalidation on section fields, clone safety, serialization and lifecycle fixes; `createScenariosFromPKML` load and seed (via #1061) | ✅ Merged |
| #1062 | Wave-1 (closes #1054) — Scenario Results save/load round trip: restores population and recorded `outputValues` | ✅ Merged |
| #1063 | Wave-1 (closes #1051) — validation gaps: parameter-set validator shape, cross-references, cli glue injection, `modelFile` existence | ✅ Merged |
| #1064 | Wave-1 (closes #1053, #1075) — PI runtime: parameter/mapping scenario-subset check, `PIParameter$units`, transform and weight; per-algorithm option defaults merge | ✅ Merged |
| #1065 | Wave-1 (closes #1052, #1078) — plots robustness: `createPlots()` / `createDataCombined()` survive plots outside grids, grid subsets, round-tripped specs, and `addPlot()` vector fields; JSON identifier casing unified (`plotID` → `plotId`, `DataCombinedName` → `dataCombinedName`) | ✅ Merged |
| #1089 | Migration PR1 — persistence engine: definition-file tree + id canonicalization | ✅ Merged |
| #1090 | Migration PR2 — Project class core: write-through bindings, syncStatus, snapshot verbs | ✅ Merged |
| #1091 | Migration PR3 — scenarios: tree persistence + authoring + PKML + run wiring | ✅ Merged |
| #1092 | Migration PR4 — parameter sets: unify the three kinds + vectorized authoring | ✅ Merged |
| #1093 | Migration PR5 — definition authoring: individuals/populations/applications/output-paths/observed-data | ✅ Merged |
| #1094 | Migration PR6 — plots restructure: three per-definition kinds + keyed lists + create-plots | ✅ Merged |
| #1095 | Migration PR7 — parameter identification reshape | ✅ Merged |
| #1096 | Migration PR8 — validation and prints/UX | ✅ Merged |
| #1097 | Migration PR9 — Excel bridge (import/export) | ✅ Merged |
| #1098 | Migration PR10 — docs: vignettes + README + NEWS + pkgdown (Chapter 9) | ✅ Merged |
| #1099 | Migration PR11 — fixtures: bundled example + test-data project trees | ✅ Merged |
| #1100 | Migration PR12 — tail: sensitivity + settings + utils + test hardening | ✅ Merged |
| #1101 | Migration PR13 — regenerate `man/*.Rd` + `NAMESPACE` (migration boundary) | ✅ Merged |
| #1102 | Migration PR14 — post-migration polish (lower-tier review + issue-ledger fixes) | ❌ Closed — fully contained in #1104, superseded |
| #1104 | Migration PR15 — review round 1: cross-stack review fixes (**canonical reference**) | ✅ Merged |
| #1117 | Replace write-through with the explicit-save lifecycle | 🔄 Open — base `json-based-project` |
| #1119 | Modeler-friendly docs for the project lifecycle | 🔄 Open — stacked on #1117 |
| #1121 | Make `Project` method-first; free functions forward to public methods | 🔄 Open — base `json-based-project` |
| #1068 | v6 project-snapshot JSON Schema (faithful) + v5→v6 diff | ❓ Disposition pending |
| #1069 | v6 project-snapshot JSON Schema (proposed clean variant) | 📝 Draft / RFC |


R/project.R
```diff
@@ -0,0 +1,213 @@
...
+#
+# `Project` is therefore intentionally kept internal:
+#
+#   * Not added to `NAMESPACE` (no `@export`).
+#   * No mutation API. Public fields are read-only; mutations would belong to a
+#     later milestone where the class actually owns project state.
+#   * No auto-validation, no cross-reference resolution, no path resolution
+#     beyond what's required to print sensibly.
+#
+# Construction goes through `.loadProjectJson()` (`R/project-parse.R`).
```
Felixmil: This is the part of the PR I'd most like reviewers to internalize — the existence of a parallel `Project` class alongside `ProjectConfiguration`. They look like they could collide but they don't: `ProjectConfiguration` is the Excel-paths-only class that drives the runtime today; `Project` is the parsed in-memory shape that will eventually own everything. They coexist on purpose for the next several chapters.

**Future:** Chapter 8 (Excel deprecation) is where `ProjectConfiguration` actually gets renamed/replaced. Until then, both live side by side and each handles its own lifecycle.

R/project.R
```diff
@@ -0,0 +1,213 @@
...
+          "  plots:           ",
+          length(private$.plots$dataCombined %||% list()), " dataCombined / ",
+          length(private$.plots$plotConfiguration %||% list()), " plot(s) / ",
+          length(private$.plots$plotGrids %||% list()), " grid(s)\n",
+          sep = ""
+        )
+      }
+      invisible(self)
+    }
+  ),
```
Felixmil, line 716: Every active field is `stop("<field> is read-only")` on assignment. Deliberate — the class has no mutation API yet, and I don't want callers reaching past the parser's output and tweaking fields on a parsed `Project`.

**Future:** Chapter 6 introduces the full programmatic surface (`addScenario` / `removeScenario`, `addIndividual` / `removeIndividual`, etc., for every section, plus the generic S3 `addParameter` / `removeParameter`). At that point these read-only guards get replaced by setters that route through a `private$.invalidate()` lifecycle hook — sets `modified = TRUE`, clears `validatedSinceMutation`.

R/project-parse.R
```diff
@@ -0,0 +1,64 @@
...
+#' @noRd
+.loadProjectJson <- function(path) {
+  if (!is.character(path) || length(path) != 1L || is.na(path)) {
+    stop("`path` must be a single non-NA string.", call. = FALSE)
+  }
+  if (!file.exists(path)) {
+    stop("Project file does not exist: ", path, call. = FALSE)
+  }
+
+  raw <- jsonlite::fromJSON(path, simplifyVector = FALSE)
```
Felixmil: `simplifyVector = FALSE` is load-bearing. Without it, `jsonlite` would happily collapse `outputPathIds: ["Aciclovir_PVB"]` into a length-1 character vector, and the round-trip serializer would then emit it back as a JSON scalar string — schema drift on a value that was unambiguous in the input.

Keeping it `FALSE` means every section comes through as a list of lists, exactly mirroring the JSON tree. The `Project` class stores that shape verbatim.

The serializer (`R/project-to-json.R`) makes the same promise on the way out — see the round-trip tests in `tests/testthat/test-project-to-json.R` for the contract.

R/project-parse.R
```diff
@@ -0,0 +1,64 @@
...
+
+  schemaVersion <- raw$schemaVersion
+  if (!identical(schemaVersion, "2.0")) {
```
Felixmil: Only one validation in this chapter — the schema version. Anything else (typoed field names, missing required keys, cross-references between sections) gets caught later or not at all in this PR. Intentional — keeping the parser as a pure shape transform.

**Future:** Chapter 4 lands `validateProject()` and a per-section validator dispatcher (each section file owns a `.<section>ValidatorAdapter`). That's where structural and cross-reference validation goes — kept separate from parsing so the seam stays clean: parser does shape, validator does semantics.

R/project-to-json.R
```diff
@@ -0,0 +1,153 @@
...
+    observedData = .observedDataToJson(project),
+    outputPaths = .outputPathsToJson(project),
+    scenarios = .scenariosToJson(project),
+    modelParameters = .modelParametersToJson(project),
+    individuals = .individualsToJson(project),
+    populations = .populationsToJson(project),
+    applications = .applicationsToJson(project),
+    plots = .plotsToJson(project)
+  )
+
```
Felixmil, line 122: The per-section delegation here is the architectural shape that matters more than what the helpers actually do today. Right now every helper (`.filePathsToJson`, `.scenariosToJson`, …) is a one-line passthrough — the parser is JSON-faithful, so there's nothing to transform on the way out.

**Future:**
- `.filePathsToJson` will convert absolute paths back to relative against `project$projectDirPath` (Chapter 3, when the runtime starts caring about path resolution).
- `.scenariosToJson` will rewrite literal `outputPaths` back to `outputPathIds` referencing `project$outputPaths`, and convert `steadyStateTime` back to its declared unit (Chapter 3).
- Per-section files (`R/scenarios.R`, `R/individuals.R`, …) will eventually own the parser/serializer pair for their section as a co-located unit (Chapter 7's file reorg).

Landing the call shape now means future chapters can add the transformations one section at a time without rearranging the top-level.

R/project-to-json.R
```diff
@@ -0,0 +1,153 @@
...
+  project$plots
+}
+
+# Shape-coercion helper -------------------------------------------------------
+
+# `list()` is ambiguous in JSON: jsonlite renders an empty list as `[]`, but
+# the schema requires `{}` for the map-shaped sections (`filePaths`,
+# `outputPaths`, `applications`, `modelParameters`). Setting a zero-length
+# names attribute triggers jsonlite's named-list serialization path.
+.asJsonObject <- function(x) {
```
Felixmil, line 464: Subtle one — `list()` is ambiguous in JSON. `jsonlite` defaults an empty list to `[]`, but the schema requires `{}` for the map-shaped sections (`filePaths`, `outputPaths`, `applications`, `modelParameters`). `setNames(list(), character(0))` flips jsonlite into named-list serialization mode, which emits `{}`.

The round-trip test `"empty map sections survive a round-trip as empty named lists"` pins this — once the serializer has emitted `{}` once, every subsequent reload + save is stable. Without this helper, an empty project would round-trip from `{}` to `[]` to `[]`.

Felixmil (commented):
Posting a few inline notes to flag the load-bearing parts of this chapter and what later chapters will change.

R/project-to-json.R
```diff
@@ -0,0 +1,153 @@
...
+  if (!is.character(path) || length(path) != 1L || is.na(path)) {
+    stop("`path` must be a single non-NA string.", call. = FALSE)
+  }
+  parent <- dirname(path)
+  if (!dir.exists(parent)) {
+    stop("Parent directory does not exist: ", parent, call. = FALSE)
+  }
+
+  json <- .projectToJson(project)
+  writeLines(json, path)
```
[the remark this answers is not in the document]
Felixmil: Applied — but reshaped to match the end-state form on `json-as-primary-input-v2`: `.saveProjectJson()` now calls `jsonlite::write_json()` directly (instead of `writeLines()` on a pre-built string). `jsonlite::write_json` writes UTF-8 to disk regardless of locale, which addresses the same hazard `useBytes = TRUE` would. Side effect: `.projectToJson()` now returns the JSON-shaped list rather than a string, mirroring end-state's `.projectToJson()` shape.

R/project-parse.R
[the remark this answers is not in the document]
Felixmil: Skipping. End-state on `json-as-primary-input-v2` (which this PR is the first chapter toward) calls `jsonlite::fromJSON()` bare too, after a separate `fs::file_exists()` check — same shape as here. Adding a `tryCatch` wrapper would deviate from end state, so I'll leave the lexer error verbatim. The default jsonlite message includes a column pointer; the file path comes from the calling context.

R/project-to-json.R
```diff
@@ -0,0 +1,153 @@
...
+# JSON array of observed-data source entries.
+.observedDataToJson <- function(project) {
+  project$observedData
+}
+
+# JSON object with `dataCombined` / `plotConfiguration` / `plotGrids` arrays;
+# `null` when the project carries no plots section.
+.plotsToJson <- function(project) {
+  project$plots
+}
```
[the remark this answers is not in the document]
Felixmil: Skipping. Verified the round-trip behavior: when `plots` is parsed from `{}` in the input, `jsonlite::fromJSON` produces a named empty list, and the serializer emits `{}` correctly without `.asJsonObject()`. The only path that would hit the suggested guard is a programmatic `p$plots <- list()`, but `plots` is read-only in this PR (mutation API arrives in Chapter 6). End state's `.plotsToJson` doesn't use `.asJsonObject` either — it transforms the sub-sections through `.dataCombinedToNestedJson` and `.dataFrameToListOfLists`, which always produce well-shaped lists. That richer body lands in Chapter 5; the current passthrough is the right placeholder until then.

R/project.R
```diff
@@ -0,0 +1,213 @@
...
+        cat(
+          "  plots:           ",
+          length(private$.plots$dataCombined %||% list()), " dataCombined / ",
+          length(private$.plots$plotConfiguration %||% list()), " plot(s) / ",
+          length(private$.plots$plotGrids %||% list()), " grid(s)\n",
+          sep = ""
+        )
+      }
+      invisible(self)
+    }
```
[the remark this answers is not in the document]
Felixmil: Skipping for this chapter. End state replaces the print method more thoroughly than this — it uses `ospsuite.utils::ospPrintItems` and shows the *resolved* `modelFolder` / `dataFolder` / `outputFolder` paths (the meaningful subset of `filePaths`), plus the working directory and a per-sub-section breakdown of `plots`. Adding a "filePaths: N path(s)" line now would land a different design that the print-method overhaul will then re-replace. The lean Chapter 1 `print()` is intentional.

nautilus69 (changes requested):
Here are my two cents, by what I understand from the instruction Felix posted on teams: 

1. call. = FALSE missing from read-only guards (R/project.R:293 and all 11 siblings)

  Every error in the parser and serializer uses call. = FALSE, but the active binding guards don't:

  if (!missing(value)) stop("schemaVersion is read-only")
  > should be:
  if (!missing(value)) stop("schemaVersion is read-only", call. = FALSE)

  This is inconsistent within the PR itself — not a future-chapter concern.

  2. Redundant mustWork = TRUE in normalizePath (R/project-parse.R:73)

  if (!file.exists(path)) stop("Project file does not exist: ", path, call. = FALSE)
  jsonPath <- normalizePath(path, winslash = "/", mustWork = TRUE)

  mustWork = TRUE throws if the file doesn't exist — but that's already caught two lines above. If a race condition hits between the
  two calls, the error message from normalizePath is less readable than yours. Drop mustWork = TRUE or remove the file.exists() guard
   and let normalizePath do the work.

  3. plotGrids.plotIDs is a scalar string, not an array (Project.json:665)

  "plotGrids": [{ "plotIDs": "P1", ... }]

  Every other multi-value field in the fixture is an array: "outputPathIds": ["Aciclovir_PVB"], "modelParameters": ["Global",
  "Aciclovir"]. This looks like an inconsistency that will surface as a special case in Ch 5 (plots-from-project) when plotIDs needs
  to be iterated. If it's intentionally a comma-separated string (like simulationTime), a short comment saying so would prevent the
  next chapter from normalizing it silently.

Felixmil:
> Here are my two cents, by what I understand from the instruction Felix posted on teams:
> ...

Thanks, addressed all three in `ac3eee2b`.

1. **`call. = FALSE` on read-only guards** — switched the 12 active-binding guards to `cli::cli_abort()` with `{.field}` markup. Gives the same clean output without an `Error in (function (value) ...)` prefix.

2. **Redundant `mustWork = TRUE`**, flipped to `mustWork = FALSE`. The `file.exists()` guard above already produces the readable error

3. **`plotGrids.plotIDs` shape** — kept as a comma-separated string by design (matches the existing Excel `Plots` sheet convention and the end-state schema in `json-as-primary-input-v2`, where the same shape is reused). Added a note on the `@field plots` roxygen so the next PR doesn't silently normalise it into an array.

nautilus69 (approved):
Approved.

R/project-parse.R
```diff
@@ -0,0 +1,64 @@
...
+#' @param path Path to a `Project.json` file. Must exist and declare
+#'   `schemaVersion == "2.0"`.
+#'
+#' @return A `Project` (R6) holding the parsed sections.
+#'
+#' @keywords internal
+#' @noRd
+.loadProjectJson <- function(path) {
+  if (!is.character(path) || length(path) != 1L || is.na(path)) {
+    stop("`path` must be a single non-NA string.", call. = FALSE)
```
rengelke: Several stop() calls use direct strings instead of entries in messages.R with ospsuite.utils::cliFormat()

rengelke (commented):
Solid parser foundation!

R/project-parse.R
Felixmil: Thanks. Applied for the one path-validation message that's reused across `R/project-parse.R` and `R/project-to-json.R` — added `messages$invalidPathArgument()` and routed both sites through it (c16e1ae5).

Left the other three `stop()` calls inline (file-not-found, unsupported schemaVersion, parent-dir-missing). They're each used in a single location, and per the project rule of thumb I'm trying to keep the `messages.R` namespace for entries that are genuinely shared.

PavelBal:
@coderabbit full review

inst/extdata/projects/Example/Project.json
```diff
@@ -0,0 +1,195 @@
+{
+  "schemaVersion": "2.0",
+  "esqlabsRVersion": "6.0.0",
+  "filePaths": {
+    "modelFolder": "Models/Simulations/",
```
PavelBal: Should this be renamed to `simulationsFolder`?

inst/extdata/projects/Example/Project.json
```diff
@@ -0,0 +1,195 @@
...
+    "configurationsFolder": "Configurations/",
+    "modelParamsFile": "ModelParameters.xlsx",
+    "individualsFile": "Individuals.xlsx",
+    "populationsFile": "Populations.xlsx",
+    "populationsFolder": "Populations",
```
PavelBal: Missing backslash (to stay consistent with other folder names)

PavelBal: Missing backslash (to stay consistent with other folder names)

PavelBal: Missing backslash (to stay consistent with other folder names)

inst/extdata/projects/Example/Project.json
```diff
@@ -0,0 +1,195 @@
...
+      "simulationTimeUnit": "h",
+      "steadyState": false,
+      "steadyStateTime": null,
+      "steadyStateTimeUnit": null,
+      "overwriteFormulasInSS": false,
+      "modelFile": "Aciclovir.pkml",
+      "outputPathIds": ["Aciclovir_PVB"]
+    }
+  ],
+  "modelParameters": {
```
PavelBal: Suggest renaming to `ParameterSets`

inst/extdata/projects/Example/Project.json
PavelBal: Missing backslash (to stay consistent with other folder names)

inst/extdata/projects/Example/Project.json
```diff
@@ -0,0 +1,195 @@
...
+    {
+      "individualId": "Adult_male",
+      "species": "Human",
+      "population": "European_ICRP_2002",
+      "gender": "MALE",
+      "weight": 73,
+      "height": 176,
+      "age": 30,
+      "proteinOntogenies": "CYP3A4:CYP3A4",
+      "parameters": [
```
PavelBal: This is not up-to-date with the new (V6) way to define individal parameters, as introduced in this PR https://github.com/esqLABS/esqlabsR/pull/970

Multiple individual parameter sets (sheets) can be assigned to an individual.

PavelBal (changes requested):

PavelBal (changes requested):
Changes as requested.

PavelBal (changes requested):
Changes as requested.

inst/extdata/projects/Example/Project.json
Felixmil: Why not, do we still need to nest it unders `Models/` ? Is it used for something else ?

PavelBal: The idea was to store all kind of "models" in the `Models/` folder - PK-Sim/MoBi projects, other models (if we have them, e.g. RxODE). And the pkml simulations live in the `Simulations` subfolder

inst/extdata/projects/Example/Project.json
Felixmil: This will be implemented in future chapter.

inst/extdata/projects/Example/Project.json
Felixmil: Original idea was to store all parameterSets in there (individuals, applications, modelParameters). Each with a unique key that can be referenced in other entities.
This way, we don't end up with individualParameters, applicationParameters, sections. what do you think ?

inst/extdata/projects/Example/Project.json
Felixmil: Then `Models/PKML/` and `pkmlFolder` ?

inst/extdata/projects/Example/Project.json
PavelBal: I think for the user it is important to have a clear separation between these three parameter set types - parameter sets applied to the global model, then only if an individual is defined, and separate from applications

inst/extdata/projects/Example/Project.json
PavelBal: :D pkml can be everything - a whole simulation, a separate building block, observed data... We have to make clear that we are loading a simulation, and only pkmls that represent a simulation should go there

inst/extdata/projects/Example/Project.json
Felixmil: Ok but `parameterSets` wording works for the 3 of them, its a generic term. Would this be ok ?

`Project$modelParameterSets`
`Project$individualParameterSets`
`Project$applicationParameterSets`

Another option:
`Project$ParameterSets$models`
`Project$ParameterSets$applications`
`Project$ParameterSets$individuals`

inst/extdata/projects/Example/Project.json
Felixmil: I assumed that since it is under `Models/` it could not be bb, observed data etc.. but you're right.
For now let's keep it as-is because all my other PR depends on this. We'll be able to change that easily once they are all merged into one.

inst/extdata/projects/Example/Project.json
PavelBal: Yes!

R/project-parse.R
rengelke: Yes, makes sense. The applied repo rule is to store each message in messages.R. It was mentioned several time e.g.  #700

Felixmil: You are right. I will align.

This rule should be documented somewhere ?

sorinvoicu:
- `test-project-parse.R` and `test-project-to-json.R` both define an identical `example_project_json_path()` function
- readLines() in text-matching tests has no encoding declaration
  test-project-to-json.R (the "empty map sections serialize as JSON objects" test) reads the output file with:
  text <- paste(readLines(out, warn = FALSE), collapse = "\n")
  On Windows, readLines() defaults to the system locale encoding, not UTF-8. Since .saveProjectJson() writes UTF-8 via jsonlite::write_json, the read-back could silently mangle non-ASCII characters in a future fixture containing e.g. accented species names. Use readLines(out, encoding = "UTF-8", warn = FALSE) to match the write side.

tests/testthat/test-project-parse.R
```diff
@@ -0,0 +1,225 @@
+# Tests for the v2.0 Project.json parser. These tests cover only the parser
+# and the internal `Project` class — the parser is not wired into runScenarios,
+# validators, or plotting on this branch. Tests use `:::` because both are
+# intentionally unexported.
+
+example_project_json_path <- function() {
```
sorinvoicu: duplicate with test-project-to-json.R

tests/testthat/test-project-to-json.R
```diff
@@ -0,0 +1,170 @@
...
+  jsonlite::write_json(
+    list(schemaVersion = "2.0", esqlabsRVersion = "6.0.0"),
+    empty,
+    auto_unbox = TRUE
+  )
+  project <- esqlabsR:::.loadProjectJson(empty)
+
+  out <- withr::local_tempfile(fileext = ".json")
+  esqlabsR:::.saveProjectJson(project, out)
+  text <- paste(readLines(out, warn = FALSE), collapse = "\n")
```
sorinvoicu: Use readLines(out, encoding = "UTF-8", warn = FALSE) to match the write side

sorinvoicu (approved):

R/project-parse.R
Felixmil: I'll address this in following chapter so I don't have to cascade rebase all next chapters.

sorinvoicu:
syncStatus() calls an undefined .projectSyncStatus (project.R:704), and the write-through path silently rounds all numerics to 4 decimals (entity-files.R:311, missing digits = NA).

sorinvoicu:
The codebase explicitly treats Project.json as untrusted for path fields. project.R:1357 routes the four working folders through .resolveWorkingFolder() → .pathEscapesRoot(), and tests/testthat/test-project.R:585 states the rationale: "an absolute (or ../-escaping) folder value from an untrusted Project.json would otherwise be an attacker-chosen root."

definitionsFolder bypasses that check entirely:

- project.R:1431 — private$.definitionsFolder <- jsonData$definitionsFolder, raw from JSON
- project.R:952-955 — getter and setter, no validation
- definition-files.R:68 — file.path(projectDirPath, definitionsFolder)

And it is the field that feeds the two destructive paths:

- project-lifecycle.R:660-668 — .clearProjectArtifacts() reads definitionsFolder back out of the container via .readContainerField() (which swallows parse errors) and calls unlink(file.path(destination, folder), recursive = TRUE, force = TRUE). A Project.json carrying "definitionsFolder": "../.." makes initProject(dest, overwrite = TRUE) recursively delete the grandparent tree.
- definition-files.R:454-488 — .writeDefinitionTree() documents that it "OWNS the <kind>/ directory" and deletes every .json not in the keep-set. With an escaping definitionsFolder, a plain saveProject() — no overwrite flag needed — deletes non-matching .json files in a directory outside the project.

The second is the worse one: shared/downloaded project directories are the normal distribution mode for this package, and it needs no destructive-sounding call to trigger.

There is no escape test for definitionsFolder; it's covered only for defaults and for legitimate relocation ("defs", "other-defs"). The asymmetry with its four siblings reads as an oversight rather than a decision.

Fix: run it through .validateFilenameSegment() (file-utils.R:325) — which already exists for exactly this, and rejects separators plus ./.. — at both the JSON read and the setter. It's a single path segment by design, so that's a tighter and simpler guard than .pathEscapesRoot(). Add the matching rejection test.

inst/extdata/update-packages.R
PavelBal: Do we need this file at all?

PR #1020, opened 2026-05-04
Chapter 3: drive runScenarios from a parsed Project

Felixmil:
Chapter 3 of the JSON-based-project rework (issue #908). Builds on Chapter 2 (PR #1018, `scenario-as-data`).

## Summary

- **`loadProject()` exported** as the JSON-first workflow's first public seam, marked `lifecycle::badge('experimental')`. Thin wrapper around `.loadProjectJson()`.
- **`Scenario` (plain-data class) exported.** The legacy R6 `Scenario` is renamed `LegacyScenario` and soft-deprecated; it stays alive for the v6 Excel-driven `createScenarios()` flow until the Excel deprecation chapter.
- **`runScenarios()` accepts a `Project` directly:** `runScenarios(project, scenarioNames = NULL, customParams = NULL, simulationRunOptions = NULL, validate = TRUE)`. The legacy `runScenarios(scenarios = ...)` form keeps working but is soft-deprecated. Internal type-sniff dispatcher routes the modern path to `.runScenariosFromProject()` and the legacy paths to `.runLegacyScenarios()`.
- **New `R/scenario-execution.R`** holds `.mergeScenarioParameters()` (five-layer merge: model → species → individual → application → custom), `.prepareScenario()`, `.collectScenarioResult()`, `.runScenariosFromProject()`, plus two transitional bridge helpers (`.parameterGroupToStructure`, `.findById`) that flatten the JSON-faithful section parses on demand. Both bridges go away once the section parsers become typed (Chapter 6/7).
- **Read-only path-resolution active fields on `Project`:** `configurationsFolder`, `modelFolder`, `populationsFolder` (rooted at `projectDirPath`). Setters and `.invalidate()` lifecycle deferred to Chapter 6.
- **`Scenario$outputPaths` is now a named character vector** (`names = ids`, `values = literal paths`). Fixes the reverse-resolution lossiness when two ids map to the same literal path; the serializer reads `names()` directly. Documented as the one deliberate deviation from end-state.
- **Test fixture `tests/testthat/data/TestProject/`** (Project.json + Aciclovir.pkml + TestPopulation.csv) plus three new helpers in `tests/testthat/helpers.R`: `testDataDirectory()`, `testProjectJSONPath()`, `testProject()`.

## Notes for review

[...]

Issue #908, opened 2025-11-17
ESQlabsR should use `json` snapshot as source of truth


Felixmil:
Excel Files are just an alternative UI (eventually replaced by ESQapp)

Felixmil:
Implemented and shipped as the 6.0.0 project model: JSON is the single source of truth (a `Project.json` container plus a one-file-per-definition `definitions/` tree), and Excel is demoted to an interchange bridge through `importProjectFromExcel()` and `exportProjectToExcel()`. The NEWS breaking-changes section credits this issue across the whole rework.

Closing as done. Design record: #1088. Consolidation PR: #1017.

PR #1022, opened 2026-05-06
Chapter 5: drive data combination and plotting from a parsed Project

Felixmil:
Chapter 5 of the JSON-based-project rework (issue #908). Builds on Chapter 3 (PR #1020, `runscenarios-from-project`). Skips Chapter 4 — `validate-project` lands in parallel and Chapter 5 doesn't depend on it.

## Summary

- **`loadObservedData(project)` exported** as the project-driven loader. Dispatches across `excel` / `pkml` / `script` source declarations in `project$observedData`. The `programmatic` source type errors fast — its dependencies (private `.programmaticDataSets` slot, `addObservedData()`) land with the mutation API in Chapter 6.
- **Legacy observed-data loaders soft-deprecated.** `loadObservedData(projectConfiguration, ...)` is renamed to `loadObservedDataFromExcel()`; together with `loadObservedDataFromPKML()`, both are wrapped in `lifecycle::deprecate_soft()` and route through internal `.legacy` implementations. Internal callers (`R/utilities-pi.R`) updated to use the renamed `.loadObservedDataFromExcelLegacy` directly.
- **`createDataCombined(project, ...)` exported.** Reads from `project$plots$dataCombined` (named list keyed by name), resolves observed sources via `loadObservedData()` internally. `createDataCombinedFromExcel()` keeps working as a soft-deprecated wrapper. Both entry points share `.createDataCombinedFromProcessedDF()` so the build/transform/validate logic lives in one place.
- **`createPlots(project, ...)` exported.** Iterates `project$plots$plotConfiguration` and `project$plots$plotGrids` (data.frames) to build per-grid plots. `createPlotsFromExcel()` is soft-deprecated. Both entry points share `.createPlotGridsFromDataFrames()`.
- **`Project$plots` parses into the asymmetric in-memory shape** future chapters consume: `dataCombined` as a **named list** keyed by name (drops the redundant `name` field on each entry), `plotConfiguration` and `plotGrids` as **data.frames** with NA-padded missing fields. New helpers `.parseNestedDataCombined` / `.dataCombinedToNestedJson` and `.listOfListsToDataFrame` / `.dataFrameToListOfLists` form the parse↔serialize pair. Wire-level round-trip is the contract; in-memory NA-vs-absent identity is intentionally lossy.
- **`Project$dataFolder` resolved active field.** Mirrors `modelFolder` / `populationsFolder` / `configurationsFolder`; resolves `filePaths$dataFolder` against `projectDirPath`. Read-only.
- **Bundled fixtures.** `inst/extdata/projects/Example/` gets the Aciclovir PKML + observed-data Excel + importer config + TestPopulation CSV, restored from `chapters-base`. `tests/testthat/data/TestProject/` gets the Excel observed-data binaries plus a populated `observedData` entry in `Project.json`.

## Notes for review

- **Validation deferred to Chapter 4.** No `.observedDataValidatorAdapter()`, no `.plotsValidatorAdapter()`, no `validate = TRUE` parameter on `createPlots()`. Lightweight inline checks (`.validateObservedDataEntry`) handle dispatch-time concerns only. Chapter 4 will add the convention-based section adapters once both branches merge.
- **No cache, no `getObservedDataNames()`.** Both land with the mutation API in Chapter 6 where they actually matter (programmatic adds invalidate the cache). `loadObservedData()` re-loads on each call; `createDataCombined()` calls it exactly once per request.
- **File naming preserved.** No rename of `R/utilities-data-combined.R` → `R/data-combined.R` or `R/utilities-figures.R` → `R/create-plots.R` — those moves stay scoped to Chapter 7 (`r-file-reorg`) so reviewers track logic changes here without rename noise. New `R/observed-data.R` is a fresh file, not a rename.
- **Lifecycle deprecation noise in legacy tests.** `tests/testthat/setup.R` sets `lifecycle_verbosity = "quiet"` so the legacy callers (`createPlotsFromExcel`, `createDataCombinedFromExcel`, `loadObservedDataFromExcel`, `loadObservedDataFromPKML`) don't drown the test reporter in deprecation warnings. Removed when the Excel bridge retires.
- **NEWS bullet placed under "Internal (work in progress)"** alongside Chapter 1 and Chapter 3, even though the entry announces newly-exported public functions and `lifecycle::deprecate_soft` warnings on legacy ones — matches the chapters-base convention of keeping the JSON-primary workflow flagged as experimental until the full sequence lands.


R/utilities-data-combined.R
```diff
@@ -1,27 +1,162 @@
...
+}
+
+# Convert the named-list `dataCombined` spec from project$plots into the
+# flat tibble the legacy Excel-driven code path expects. One row per
+# entry (simulated or observed). Caller must pre-filter to specs that
+# actually have entries; empty DCs are handled in `createDataCombined`.
+#
+# @keywords internal
+# @noRd
+.specsToDataCombinedDataFrame <- function(specs) {
```
Felixmil, line 92: **NOTE — transitional.** This helper reshapes the JSON `dataCombined` spec into the legacy flat-DF shape so the existing validator and processor can be reused without change. It goes away in a later chapter when validators move to operate on specs directly. Also: the validator it feeds (`.validateDataCombinedFromExcel`) loses its `FromExcel` suffix once the Excel path is removed.

R/utilities-data-combined.R
```diff
@@ -1,27 +1,162 @@
...
+      rows[[length(rows) + 1L]] <- .specEntryToRow(name, "simulated", entry)
+    }
+    for (entry in spec$observed %||% list()) {
+      rows[[length(rows) + 1L]] <- .specEntryToRow(name, "observed", entry)
+    }
+  }
+  dplyr::bind_rows(rows)
+}
+
+.specEntryToRow <- function(dataCombinedName, dataType, entry) {
```
Felixmil, line 106: **NOTE — transitional.** The 13 hardcoded fields mirror the legacy DF columns. This list disappears together with `.specsToDataCombinedDataFrame` once the JSON path stops flowing through the legacy DF processor.

R/utilities-figures.R
```diff
@@ -817,15 +722,18 @@ createPlotsFromExcel <- function(
...
-    unlist(trimws(scan(
-      text = as.character(plotId),
-      what = "character",
-      sep = ",",
-      quiet = TRUE
-    )))
-  })
+  # (and not lists of plot ids). Skipped when plotIDs is already a list-column
+  # (e.g. when this validator runs a second time inside the shared helper).
+  if (!is.list(dfPlotGrids$plotIDs)) {
```
Felixmil, line 727: **NOTE — transitional.** Idempotency guard so the validator can run twice without errors: once pre-call from the legacy `.readPlotConfigurations` path, once inside `.createPlotGridsFromDataFrames` for the JSON path. Removed once validators stop transforming DF shapes (later chapter).

R/utilities-figures.R
```diff
@@ -1158,3 +1066,165 @@ createPlotsFromExcel <- function(
...
+.createPlotGridsFromDataFrames <- function(
+  dfPlotConfigurations,
+  dfPlotGrids,
+  dataCombinedList
+) {
+  dfPlotConfigurations <- .validatePlotConfigurationFromExcel(
+    dfPlotConfigurations,
+    names(dataCombinedList)
+  )
+  dfPlotGrids <- .validatePlotGridsFromExcel(
```
Felixmil, line 1092: **NOTE — transitional names.** Both `.validatePlotConfigurationFromExcel` and `.validatePlotGridsFromExcel` are now shared between the JSON and the deprecated Excel paths. The `FromExcel` suffix gets dropped when Excel goes away.

Felixmil (commented):
Inline NOTEs marking transitional code so reviewers can spend their attention on the parts that survive into v2 (`json-as-primary-input-v2`). The flagged lines are deliberate bridges; cleanup is tracked for later chapters.

R/project.R
nautilus69: plotGrids is NOW a data.frame, so plotGrids[[i]]$plotIDs indexes the i-th column, not a row. The example notation is wrong for the shape this chapter introduces. Should be plotGrids$plotIDs[i] or describe it as a column.

nautilus69 (changes requested):
Minor change requested, beside that approved.

R/observed-data.R
```diff
@@ -0,0 +1,135 @@
...
+  ds <- ospsuite::loadDataSetFromPKML(filePath = filePath)
+  stats::setNames(list(ds), ds$name)
+}
+
+.loadObservedScript <- function(entry, dataFolder) {
+  filePath <- .resolveDataPath(entry$file, dataFolder)
+  cli::cli_inform(c(
+    "i" = "Sourcing observed-data script: {.path {filePath}}"
+  ))
+  result <- source(filePath, local = TRUE)$value
```
sorinvoicu, line 121: arbitrary code execution vector

R/observed-data.R
```diff
@@ -0,0 +1,135 @@
...
+    if (is.null(entry[[field]])) {
+      cli::cli_abort(
+        messages$observedDataMissingField(entryIndex, entry$type, field)
+      )
+    }
+  }
+  invisible(TRUE)
+}
+
+.resolveDataPath <- function(file, dataFolder) {
```
sorinvoicu, line 85: check paths against outside of project references

R/observed-data.R
Felixmil: True, but we already had some `source()` call in ESQlabsR (*cf* `sourceAll()`).
I will add a note in the documentation (tracked by #1033)

Felixmil: Tracked as #1033 (sub-issue of #1030, the JSON-primary follow-ups umbrella). The `script` source type is a designed feature of v2 — end-state `json-as-primary-input-v2` keeps the same `source(filePath, local = TRUE)` call (R/observed-data.R:248), and the `sourceAll()` / PI-script workflows have always sourced user-supplied R from a project. So no change in this chapter PR; #1033 covers the documentation-side mitigations (security note in `?loadObservedData`, schema docs, possible one-shot `cli_warn`).

R/observed-data.R
Felixmil: I have to admit we never had this concerns before, but it is 100% legit.
Existing design already has this vulnerability.
This is a good opportunity solve this, tracking in #1034.

R/observed-data.R
sorinvoicu: just had a flaskback about our discussion with Marco about functional risks :)

rengelke (approved):
Clean extension of ch. 3 pattern.

PR #1023, opened 2026-05-06
Chapter 4: validate Project objects through a convention-based dispatcher

Felixmil:
Chapter 4 of the JSON-based-project rework (issue #908). Builds on Chapter 5 (PR #1022, `plots-from-project`). Lands the validation framework that Chapter 5 deferred and that Chapter 6's mutation API will hook into.

## Summary

- **`validateProject(project)` exported** as the public entry point. Returns a named list of `validationResult` objects, one per section, in canonical order. Sets `project$validatedSinceMutation` on a clean run so subsequent `runScenarios()` / `createPlots()` calls can short-circuit redundant re-validation. Path inputs are not accepted here — load the project first via `loadProject()`.
- **Convention-based section dispatcher.** `.validationSections` lists the canonical order; `.lookupSectionValidatorAdapter()` resolves each name to `.<section>ValidatorAdapter` in the package namespace. Adding a new section is one entry in the list plus one adapter at the top of the section's R file; the dispatcher does not change. Section order matches the end-state branch.
- **Per-section adapters at the top of each section's file.** `R/utilities-scenarios.R` (outputPaths, scenarios, applications), `R/utilities-individual.R` (individuals, individualParameterSets), `R/utilities-population.R` (populations), `R/utilities-parameters.R` (modelParameterSets, applicationParameterSets), `R/observed-data.R` (observedData), `R/utilities-figures.R` (plots, covering dataCombined / plotConfiguration / plotGrids together). Each adapter pulls the right slice of `project` and delegates to a `.validate<Section>` function nearby.
- **Cross-references checked.** `.validateCrossReferences()` covers `scenario.individualId` / `populationId` / `modelParameterSets` / `applicationProtocol`, `individual.parameterSets`, `application.parameterSets`, and `dataCombined.simulated.scenario`. Skips itself with a single named warning when any prior section flagged a critical error so cross-reference noise does not pile on broken sections. The end-state's per-section `references()` walker is left for a later deepening PR (open question in the spec).
- **`validatedSinceMutation` lifecycle on Project.** Read-only active field; `.markValidated()` flips it after a clean full validation; private `.invalidate()` clears it (and gets a TODO marker for the `modified` flag Chapter 6 will add). No mutators ship in this chapter, so the flag only resets on a fresh `loadProject()`.
- **Pre-flight validation in `runScenarios()` and `createPlots()`.** Both gain a `validate = TRUE` default and route through `.ensureValid()`, which runs the relevant section validators and aborts with a formatted critical-error summary. Short-circuits when `validatedSinceMutation` is `TRUE`. Set `validate = FALSE` to skip when the caller has already validated.
- **Legacy Excel cross-reference validator renamed** to `.validateCrossReferencesExcel` so the canonical name is free for the JSON path. The Excel body is unchanged; only the lone caller in `validateAllConfigurations()` and the legacy tests update.

## Notes for review

- **End-state alignment.** Public surface (`validateProject`, `.validationSections`, `.lookupSectionValidatorAdapter`, `.<section>ValidatorAdapter`, `.validate<Section>`, `.ensureValid`, `.abortValidationErrors`, `.markValidated`/`.markModified`/`.invalidate` lifecycle on Project) tracks `json-as-primary-input-v2` so chapters 5+ rebase cleanly.
- **Adapter file layout.** Adapters live in the existing `R/utilities-*.R` files because Chapter 7 owns the per-domain rename. Keeps reviewers tracking logic changes here without rename noise; the adapter convention works regardless of file layout.
- **`.extractEntryIds` is transitional.** `individuals` and `populations` are stored as JSON-array (unnamed) lists in the current parser; the helper extracts their IDs for cross-reference checks. Drops out once the parser converts them into named lists keyed by id (planned alongside the Chapter 7 file reorg). Marked in the doc.
- **Scope of `validateProject()`.** Path inputs deliberately omitted — chapter 9 / vignettes can broaden the entry point if needed. Keeping the chapter-4 surface narrow.
- **No NEWS entry.** This branch is unreleased v6 work; the chapter sequence ships its public-facing announcement at the end.


R/validation.R
```diff
@@ -0,0 +1,558 @@
...
+#' These sections are stored as unnamed lists where each entry carries
+#' its own ID field (`individualId` / `populationId`). Returns a
+#' character vector of those IDs (with `NA` filtered out) for
+#' cross-reference checks. Transitional helper: drops out once the parser
+#' converts individuals/populations into named lists keyed by id (planned
+#' alongside the file reorg in Chapter 7).
+#'
+#' @keywords internal
+#' @noRd
+.extractEntryIds <- function(entries, idField) {
```
Felixmil, line 202: **NOTE — transitional.** `.extractEntryIds` exists only because the current parser stores `individuals` and `populations` as JSON-array (unnamed) lists where each entry carries its own `individualId` / `populationId` field. The helper extracts those IDs for cross-reference checks. Goes away once the parser converts both sections into named lists keyed by id (planned alongside the Chapter 7 file reorg) — at that point `names()` resolves the same thing for free.

R/validation.R
```diff
@@ -0,0 +1,558 @@
...
+          "' references undefined applicationProtocol '",
+          sc$applicationProtocol,
+          "'"
+        )
+      )
+    }
+  }
+
+  individualSetKeys <- names(project$individualParameterSets %||% list())
+  for (entry in project$individuals %||% list()) {
```
Felixmil, line 483: **NOTE — transitional iteration.** Walks individuals as a positional array because the current parser stores them as an unnamed list. Once individuals become a named list keyed by `individualId` this collapses to `for (id in names(project$individuals))` plus `project$individuals[[id]]$parameterSets`.

Felixmil (commented):
Inline NOTEs flagging the helpers/loops that exist only because the current parser stores individuals/populations as JSON arrays. Reviewers can skip these — they disappear when those sections become named lists keyed by id (planned alongside the Chapter 7 file reorg).

R/utilities-individual.R
```diff
@@ -1,3 +1,87 @@
...
+  if (is.null(individuals) || length(individuals) == 0) {
+    result$add_warning("Data", "No individuals defined")
+    return(result)
+  }
+
+  ids <- .extractEntryIds(individuals, "individualId")
+  result <- .check_no_duplicates(ids, "individualId", result)
+
+  requiredFields <- c("species", "gender")
+  for (i in seq_along(individuals)) {
```
Felixmil, line 44: **NOTE — transitional iteration.** `seq_along()` + extracting `individualId` from each entry, including the `Required field individualId is missing` critical error, is a consequence of the JSON-array storage. Once individuals are stored as a named list keyed by `individualId`, the iteration becomes `for (id in names(individuals))` and the missing-id check is structurally impossible.

R/utilities-population.R
```diff
@@ -1,3 +1,97 @@
...
+
+  if (is.null(populations) || length(populations) == 0) {
+    result$add_warning("Data", "No populations defined")
+    return(result)
+  }
+
+  ids <- .extractEntryIds(populations, "populationId")
+  result <- .check_no_duplicates(ids, "populationId", result)
+
+  for (i in seq_along(populations)) {
```
Felixmil, line 33: **NOTE — transitional iteration.** Same shape as `.validateIndividuals`: walks the JSON-array storage of `populations` and the embedded `Required field populationId is missing` check both go away when populations become a named list keyed by `populationId` (Chapter 7 alongside the file reorg).

R/observed-data.R
```diff
@@ -12,6 +13,135 @@
...
+        paste0(
+          entryLabel,
+          " has invalid type '",
+          entry$type,
+          "'. Must be one of: ",
+          paste(validTypes, collapse = ", ")
+        )
+      )
+      next
+    }
```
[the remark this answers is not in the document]
Felixmil: Skipping. The runtime aborts on `programmatic` via `messages$observedDataProgrammaticNotYetAvailable()` (R/observed-data.R:177), and the `programmatic` type is reserved for the mutation API in Chapter 6 — flagging it as a critical error here would block valid future-state JSON. The runtime error is the right place to surface "not yet implemented" until then.

R/validation.R
```diff
@@ -0,0 +1,558 @@
...
+        paste0(
+          "Set '",
+          setName,
+          "' in ",
+          sectionName,
+          ": paths and values have different lengths"
+        )
+      )
+      next
+    }
```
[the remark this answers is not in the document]
Felixmil: Skipping. `.parseParameterSets` constructs `paths`, `values`, and `units` in lockstep (R/project-parse.R), so divergent lengths are not reachable from a parsed Project. The end-state's `.validateParameterSets` (json-as-primary-input-v2:R/validation.R) doesn't check this either; adding the guard would diverge from the canonical shape without protecting any real-world input.

R/validation.R
```diff
@@ -0,0 +1,558 @@
...
+          scName,
+          "' references undefined populationId '",
+          sc$populationId,
+          "'"
+        )
+      )
+    }
+
+    if (!is.null(sc$modelParameterSets) && length(sc$modelParameterSets) > 0) {
+      invalidSets <- setdiff(sc$modelParameterSets, modelParamKeys)
```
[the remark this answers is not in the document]
Felixmil: Skipping. The Scenario class declares `modelParameterSets` as a character vector and `.parseScenarios` already `unlist()`s it at parse time (R/project-parse.R:170). The end-state's cross-reference check uses the same bare `setdiff` — the parser invariant holds, so a defensive `unlist()` here would diverge from the canonical shape without protecting any reachable input.

R/utilities-figures.R
```diff
@@ -287,9 +423,17 @@ createPlots <- function(
...
   dataCombinedList = NULL,
-  stopIfNotFound = TRUE
+  stopIfNotFound = TRUE,
+  validate = TRUE
 ) {
   ospsuite.utils::validateIsOfType(project, "Project")
+  if (isTRUE(validate)) {
+    .ensureValid(
+      project,
+      sections = c("plots", "scenarios", "crossReferences"),
```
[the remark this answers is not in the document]
Felixmil: Good catch — fixed in 57f83ad. `createPlots` → `createDataCombined` → `loadObservedData` does hit observedData entries; the pre-flight now includes `observedData` so a broken section aborts before plot generation rather than at runtime.

nautilus69 (approved):
checked locally, Approved.

sorinvoicu (approved):

R/validation.R
```diff
@@ -0,0 +1,558 @@
...
+  adapterName <- paste0(".", section, "ValidatorAdapter")
+  ns <- asNamespace("esqlabsR")
+  if (!exists(adapterName, envir = ns, mode = "function", inherits = FALSE)) {
+    cli::cli_abort(c(
+      "No validator adapter found for section {.val {section}}.",
+      "i" = "Define {.code {adapterName} <- function(project) ...} in the section's R file."
+    ))
+  }
+  get(adapterName, envir = ns, mode = "function", inherits = FALSE)
+}
```
rengelke: Not wrong, but the dispatcher feels over-engineered for a package with 11 stable sections. 
The named list of validation adapters is a more R idiomatic version of this implementation and would catch missing adapters at load time rather than at runtime.

rengelke (commented):
Layering looks good overall: section validators run independently, cross-reference run last as second pass and `runScenarios` / `createPlots` require a validation pass before running.

R/validation.R
Felixmil: adressed in [34a5171](https://github.com/esqLABS/esqlabsR/pull/1023/commits/34a51714a1620fa99c40d105cf6cc232258d7a8e) and in end state branch

rengelke (approved):

PR #1032, opened 2026-05-07
Chapter 6: programmatic add*/remove* mutation API for Project

Felixmil:
Chapter 6 of the JSON-based-project rework (issue #908). Builds on Chapter 4 (PR #1023, `validate-project`).

## Summary

- **Programmatic mutation API exported** for every section the JSON loader populates. Each section gets a standalone `addX(project, ...)` / `removeX(project, ...)`, plus an R6 delegate of the same name on `Project` so `project$addX(...)` and `addX(project, ...)` are equivalent. Function names, argument shapes, error/warning text, FK checks, and the auto-removal of empty parameter sets after the last entry is removed all match the end-state branch (`json-as-primary-input-v2`).
- **`modified` lifecycle flag on `Project`.** Read-only active field paired with the existing `validatedSinceMutation`. Both are routed through the private `.invalidate()` helper. Every successful mutator calls `project$.markModified()`, which flips `modified` to `TRUE` and clears `validatedSinceMutation` — the integration hook that prevents a stale validation result from letting `runScenarios()` / `createPlots()` short-circuit on a now-invalid project.
- **`removeX()` warn-then-remove semantics.** Removing a missing key warns and no-ops. Removing a key still referenced elsewhere emits a `cli::cli_warn()` listing the referrers and proceeds with the removal. New `.warnIfReferenced()` helper in `R/validation.R` walks scenarios for inbound references (individual / population / application / modelParameterSet / outputPath) and the holder section for parameter-set families.
- **Section active bindings made settable** so the end-state-style mutator pattern `project$x <- new` works from inside the standalone functions. Metadata bindings (`schemaVersion`, `jsonPath`, `validatedSinceMutation`, `modified`, ...) stay read-only.
- **Individuals / populations stored as named lists keyed by id.** Flips the parser to build named lists (matching the end-state shape on `json-as-primary-input-v2`) instead of JSON-array unnamed lists. Drops the transitional `.findById` and `.extractEntryIds` helpers; section validators and cross-reference checks now iterate via `names()` and look entries up by `[[id]]` directly. Applications also gain an `Application` class attribute so future S3 dispatch can hook in without re-walking the parser.
- **Observed-data programmatic source wired up.** `addObservedData()` accepts a `DataSet` or a config list; private `.programmaticDataSets` slot holds runtime DataSets. `loadObservedData()` extended to merge programmatic entries and cache `getObservedDataNames()`. Programmatic-source dispatch — errored out in Chapter 5 — now succeeds.
- **Per-section file landing:**
  - `R/utilities-scenarios.R`: `addOutputPath` / `removeOutputPath`, `addScenario` / `removeScenario`.
  - `R/utilities-individual.R`: `addIndividual` / `removeIndividual`, `setIndividualParameterSets`, `addIndividualParameterSet` / `removeIndividualParameterSet`, `addIndividualParameterSetEntry` / `removeIndividualParameterSetEntry`.
  - `R/utilities-population.R`: `addPopulation` / `removePopulation`.
  - `R/applications.R` (new): `addApplication` / `removeApplication`, `setApplicationParameterSets`, `addApplicationParameterSet` / `removeApplicationParameterSet`, `addApplicationParameterSetEntry` / `removeApplicationParameterSetEntry`.
  - `R/utilities-parameters.R`: `addModelParameterSet` / `removeModelParameterSet`, `addModelParameterEntry` / `removeModelParameterEntry`, plus shared `.addParameterEntry` / `.removeParameterEntry` / `.findParameterEntryIndex` / `.validateParameterEntryArgs` helpers used by every parameter-set family.
  - `R/observed-data.R`: `addObservedData` / `removeObservedData` / `getObservedDataNames`.
  - `R/utilities-figures.R`: `addPlot` / `removePlot`, `addPlotGrid` / `removePlotGrid`, `addDataCombined` / `removeDataCombined`.

## Notes for review

- **End-state alignment.** Every public name, argument order, error message, and FK rule tracks `json-as-primary-input-v2`. The R6 delegate methods are thin shims (`addX = function(...) addX(self, ...)`) so the standalone is the canonical implementation.
- **Adapter file layout.** Mutators live in the existing `R/utilities-*.R` files and the new `R/applications.R` (no Application section file existed before). The per-domain rename stays scoped to Chapter 7 (`r-file-reorg`) so reviewers track logic changes here without rename noise.
- **`.warnIfReferenced` placement.** Lives in `R/validation.R` because it's the only validation-adjacent helper that mutators need. Does a forward walk over `project$scenarios` plus the holder section for parameter-set families. Cross-section references (e.g. plots referencing scenarios) are not walked yet — Chapter 4's `.validateCrossReferences()` covers them at validation time, and the warn-then-remove story prefers proceeding over hard-aborting.
- **Auto-cleanup of empty parameter sets.** Removing the last entry from an `individualParameterSet` / `applicationParameterSet` / `modelParameterSet` removes the set itself. Matches the end-state behaviour; the alternative (keeping empty named sets around) would create dangling FK targets that nothing produces.
- **Observed-data programmatic flow.** `loadObservedData()` no longer errors on `source = "programmatic"`. An orphan sentinel (programmatic entry declared in JSON but never `addObservedData`-ed) silently produces an empty result; the existing test flipped from "errors" to "skips" to assert this. The cache (`.observedDataNamesCache`) is invalidated on every `addObservedData` / `removeObservedData`.
- **JSON round-trip preserved.** Individuals / populations / applications still serialize as JSON arrays — the named-list keys are re-attached as the matching id field on the way out (`.individualsToJson`, `.populationsToJson`, `.applicationsToJson`).
- **No NEWS entry.** This branch is unreleased v6 work; the chapter sequence ships its public-facing announcement at the end.


R/project.R
```diff
@@ -141,7 +147,9 @@ Project <- R6::R6Class(
     #' @field outputPaths Named list mapping output-path IDs to literal output
     #'   path strings.
     outputPaths = function(value) {
```
Felixmil, line 149: **SKIM — uniform setter pattern.** The `if (!missing(value)) private$.x <- value` branch was added to every section active binding (`outputPaths`, `scenarios`, `modelParameterSets`, `individualParameterSets`, `applicationParameterSets`, `individuals`, `populations`, `applications`, `observedData`, `plots`) so the standalone mutators can do `project$x <- new` from inside the package. The setter bodies are identical across sections — the contract worth reviewing is the read-only/settable split (metadata bindings like `schemaVersion`, `jsonPath`, `validatedSinceMutation`, `modified` stay read-only; section bindings become settable). Matches the end-state shape on `json-as-primary-input-v2`.

R/project.R
```diff
@@ -281,17 +305,343 @@ Project <- R6::R6Class(
...
-    #'   programmatic change. Clears the `validatedSinceMutation` flag so
-    #'   that any cached validation result is invalidated. No mutators ship
-    #'   with this chapter; the hook exists so future `add*` / `remove*`
-    #'   APIs can route through a single chokepoint. Not intended for
-    #'   end-user use.
+    #'   programmatic change. Sets the `modified` flag and clears the
+    #'   `validatedSinceMutation` flag so that any cached validation
+    #'   result is invalidated. Not intended for end-user use.
     #' @keywords internal
     .markModified = function() {
```
Felixmil, line 313: **REVIEW CAREFULLY — integration hook with Chapter 4.** Every successful mutator below ends with `project$.markModified()` (or `self$.markModified()` from the R6 delegates). `.markModified()` flips `modified` to `TRUE` and clears `validatedSinceMutation` via `.invalidate()` (line 705). This is the integration point that prevents a stale validation result from letting `runScenarios()` / `createPlots()` short-circuit on a now-invalid project. If a mutator forgets to call `.markModified()`, the validation cache becomes a correctness bug — worth scanning each `addX` / `removeX` / `setX` to confirm they all route through here.

Felixmil (commented):
Inline notes to help focus the review: a few sections will simplify in later chapters and can be skimmed; a few are long-lived semantics worth a careful read.

R/project.R
```diff
@@ -281,17 +305,343 @@ Project <- R6::R6Class(
...
       private$.invalidate()
       invisible(self)
     },
 
+    #' @description Add a scenario programmatically. Delegates to the
+    #'   standalone [addScenario()] function.
+    #' @param scenarioName Character. Name for the new scenario.
+    #' @param modelFile Character. Name of the `.pkml` model file.
+    #' @param ... Additional arguments passed to [addScenario()].
+    addScenario = function(scenarioName, modelFile, ...) {
```
Felixmil, line 323: **SKIM — ~30 R6 delegates are thin shims.** From here through ~line 640, every `addX` / `removeX` / `setX` method on `Project` is a one-line forwarder to the standalone of the same name (`addX(self, ...)`). The standalone is the canonical implementation; the delegate exists so `project$addX(...)` and `addX(project, ...)` are equivalent. Review the standalones (in `R/utilities-*.R` / `R/applications.R` / `R/observed-data.R`) — the delegates have nothing meaningful to review beyond "forwards to the standalone with the right argument names." The shape may collapse in a later chapter (e.g. metaprogrammed delegate generation); for now the explicit list mirrors the end-state branch.

R/validation.R
```diff
@@ -204,32 +204,89 @@ validateProject <- function(project) {
...
+#'   `"modelParameterSet"`, `"individualParameterSet"`,
+#'   `"applicationParameterSet"`, `"outputPath"`.
+#' @param id Character scalar of the id being removed.
+#' @return `invisible(NULL)`.
 #' @keywords internal
 #' @noRd
-.extractEntryIds <- function(entries, idField) {
-  if (is.null(entries) || length(entries) == 0) {
-    return(character(0))
+.warnIfReferenced <- function(project, entityType, id) {
```
Felixmil, line 206: **SKIM — transitional walker.** `.warnIfReferenced()` hand-walks `project$scenarios` for inbound references (individual / population / application / modelParameterSet / outputPath) plus the holder section for parameter-set families. The end-state branch (`json-as-primary-input-v2`) replaces this with the per-section `references()` walker mentioned in Chapter 4's PR notes — at that point `.warnIfReferenced` collapses to `references(project, entityType, id)` and the per-`entityType` switch goes away. **What's worth reviewing here is the warn-then-remove semantics**: missing key → warn + no-op, referenced key → `cli::cli_warn()` listing referrers + proceed with the removal (not abort). Those rules match the end state and won't change.

R/utilities-parameters.R
```diff
@@ -475,3 +475,262 @@ setParameterValuesByPathWithCondition <- function(
...
+  idx <- .findParameterEntryIndex(parameters, containerPath, parameterName)
+  if (length(idx) == 0L) {
+    cli::cli_warn(
+      "parameter {.val {paste(containerPath, parameterName, sep = '|')}} not found; no-op."
+    )
+    return(parameters)
+  }
+  parameters <- parameters[-idx]
+  if (length(parameters) == 0L) {
+    return(NULL)
```
Felixmil: **REVIEW CAREFULLY — auto-remove empty parameter sets.** Returning `NULL` here is the signal that the last entry of the set was just removed. Every caller (`removeModelParameterEntry`, `removeIndividualParameterSetEntry`, `removeApplicationParameterSetEntry`) checks `is.null(updated)` and removes the named set itself (e.g. `project$modelParameterSets[[id]] <- NULL`), then runs `.warnIfReferenced()` because the set's id is now gone. The alternative (keeping empty named sets around) would create dangling FK targets that nothing produces, so this matches the end-state semantics. Worth a careful read because (a) it's non-obvious that removing one entry can delete the whole set, and (b) the warn-if-referenced fires only on the auto-cleanup branch, not on every entry removal.

R/observed-data.R
```diff
@@ -174,15 +187,172 @@ loadObservedData <- function(project) {
...
-      )
+      "programmatic" = NULL
     )
-    allDataSets <- c(allDataSets, dataSets)
+    if (!is.null(dataSets)) {
+      allDataSets <- c(allDataSets, dataSets)
+    }
   }
+  # Merge runtime programmatic store, then cache names.
+  allDataSets <- c(allDataSets, state$.programmaticDataSets)
```
Felixmil, line 197: **REVIEW CAREFULLY — programmatic source now resolved at load time.** Chapter 5 errored on `source = "programmatic"`; here `loadObservedData()` merges the runtime store (`project$.programmaticDataSets`, populated by `addObservedData()`) into the result and caches the resolved names on `project$.observedDataNamesCache` for `getObservedDataNames()`. **Two semantics worth scrutinising:** (1) an orphan sentinel (a `{type: "programmatic", name: ...}` entry declared in JSON but never `addObservedData`-ed at runtime) silently produces an empty result rather than erroring — the existing test `tests/testthat/test-observed-data.R` was flipped from "errors" to "skips" to assert this. (2) The cache is invalidated on every `addObservedData` / `removeObservedData`, so name lookups never go stale across mutations. If either of those is wrong the runtime fails silently rather than loudly.

R/utilities-figures.R
```diff
@@ -1372,3 +1372,361 @@ createPlotsFromExcel <- function(
...
+
+  newRow <- c(
+    list(
+      name = name,
+      plotIDs = paste(plotIDs, collapse = ", ")
+    ),
+    .namedDotsAsRow(...)
+  )
+  newRowDf <- as.data.frame(newRow, stringsAsFactors = FALSE)
+
```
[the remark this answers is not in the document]
Felixmil: Fixed in f0eef43e via the shared `.namedDotsAsRow()` change — covers `addPlotGrid` too. Mirrored on `json-as-primary-input-v2` (763a733).

R/utilities-figures.R
```diff
@@ -1372,3 +1372,361 @@ createPlotsFromExcel <- function(
...
+    )]
+    if (length(referencingGrids) > 0) {
+      cli::cli_warn(c(
+        "Removed plot {.val {plotID}} is still referenced by {length(referencingGrids)} plot grid{?s}:",
+        "*" = "{referencingGrids}"
+      ))
+    }
+  }
+
+  project$plots$plotConfiguration <- df[df$plotID != plotID, , drop = FALSE]
```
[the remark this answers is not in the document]
Felixmil: Fixed in f0eef43e: `df[which(df$plotID != plotID), , drop = FALSE]` so NA rows in the plotID column are excluded rather than producing rows of NAs. Mirrored on `json-as-primary-input-v2` (763a733).

R/utilities-figures.R
```diff
@@ -1372,3 +1372,361 @@ createPlotsFromExcel <- function(
...
+  validateIsOfType(project, "Project")
+  .requireNonEmptyString(name, "name")
+
+  df <- project$plots$plotGrids
+  if (is.null(df) || nrow(df) == 0 || !(name %in% df$name)) {
+    cli::cli_warn("plot grid {.val {name}} not found; no-op.")
+    return(invisible(project))
+  }
+
+  project$plots$plotGrids <- df[df$name != name, , drop = FALSE]
```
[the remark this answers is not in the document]
Felixmil: Fixed in f0eef43e: same `which()` change for `removePlotGrid`. Mirrored on `json-as-primary-input-v2` (763a733).

R/observed-data.R
```diff
@@ -174,15 +187,172 @@ loadObservedData <- function(project) {
...
+      },
+      logical(1)
+    ))
+    if (length(matchIdx) == 0L) {
+      matchIdx <- which(vapply(
+        project$observedData,
+        function(e) identical(e$type, "programmatic"),
+        logical(1)
+      ))
+    }
```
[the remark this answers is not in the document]
Felixmil: Fixed in f0eef43e: dropped the fallback that removed an arbitrary programmatic sentinel when the name match failed. `addObservedData()` always stamps the name on the sentinel it writes, so a missing-name match means the entry was either hand-edited or pre-existing in the JSON — silently removing some other programmatic entry is the wrong recovery. Mirrored on `json-as-primary-input-v2` (763a733).

R/utilities-parameters.R
```diff
@@ -475,3 +475,262 @@ setParameterValuesByPathWithCondition <- function(
...
+  existingIdx <- .findParameterEntryIndex(
+    parameters,
+    containerPath,
+    parameterName
+  )
+  newEntry <- list(
+    containerPath = containerPath,
+    parameterName = parameterName,
+    value = as.double(value),
+    units = if (nchar(units) == 0L) NULL else units
```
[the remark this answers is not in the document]
Felixmil: Fixed in f0eef43e by adding `is.na(units)` to `.validateParameterEntryArgs()`, consistent with how `containerPath` / `parameterName` / `value` are validated. `units = NA` now aborts with a clear message instead of crashing on `nchar(NA)`. Mirrored on `json-as-primary-input-v2` (763a733), where end-state would otherwise silently store `NA_character_` in the units vector.

R/utilities-figures.R
```diff
@@ -1372,3 +1372,361 @@ createPlotsFromExcel <- function(
...
+
+  newRow <- c(
+    list(
+      plotID = plotID,
+      DataCombinedName = dataCombinedName,
+      plotType = plotType
+    ),
+    .namedDotsAsRow(...)
+  )
+  newRowDf <- as.data.frame(newRow, stringsAsFactors = FALSE)
```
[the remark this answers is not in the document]
Felixmil: Fixed in f0eef43e. `.namedDotsAsRow()` now wraps multi-length values (and lists) in `list()` so `addPlot(..., xValuesLimits = c(0, 100))` produces a single list-column row instead of recycling. Same fix landed on `json-as-primary-input-v2` (763a733).

R/utilities-figures.R
```diff
@@ -1372,3 +1372,377 @@ createPlotsFromExcel <- function(
...
+
+  if (is.null(project$plots)) {
+    project$plots <- list(
+      dataCombined = list(),
+      plotConfiguration = data.frame(),
+      plotGrids = data.frame()
+    )
+  }
```
nautilus69: addPlot has an unreachable plots-section initializer:

The dataCombinedName check above always fails when project$plots is NULL (names(NULL) is NULL, name %in% NULL is FALSE), so the if (is.null(project$plots)) block can never execute.

nautilus69 (changes requested):
Beside the addPlot: 

On these:
removeModelParameterEntry / 
removeApplicationParameterSetEntry / 
removeIndividualParameterSetEntry  


the .markModified() fires on no-operation

R/utilities-parameters.R:572, 
R/applications.R:209, 
R/utilities-individual.R:579. 
All three follow the same pattern:

  updated <- .removeParameterEntry(project$xParameterSets[[id]], containerPath, parameterName)
  if (is.null(updated)) {
    .warnIfReferenced(project, "xParameterSet", id)
    project$xParameterSets[[id]] <- NULL
  } else {
    project$xParameterSets[[id]] <- updated
  }
  project$.markModified()

The set-not-found case is handled with an early-return that correctly skips .markModified(). The entry-not-found-within-an-existing-set case is not. 
.removeParameterEntry() warns and returns the unchanged list, the else branch reassigns identical content, and .markModified() fires anyway. Result: a no-op user-warn flips modified to TRUE and clears validatedSinceMutation, forcing redundant re-validation on the next runScenarios() / createPlots().  

Although it is Not a correctness bug, just gratuitous cache invalidation that contradicts the PR description's "every successful mutator calls .markModified()" invariant.

R/utilities-individual.R
```diff
@@ -335,3 +321,286 @@ applyIndividualParameters <- function(individualCharacteristics, simulation) {
...
+#' @export
+#' @family individual
+addIndividual <- function(project, individualId, species, ...) {
+  validateIsOfType(project, "Project")
+  errors <- character()
+
+  if (
+    !is.character(individualId) ||
+      length(individualId) != 1L ||
+      is.na(individualId) ||
```
sorinvoicu, line 348: inconsistent NULL / NA between default parameter value and checks - see addScenario() as well. I assume consistency needs double checked across multiple parameters as well.

R/utilities-scenarios.R
```diff
@@ -576,3 +576,301 @@ loadScenarioResults <- function(scenarioNames, resultsFolder) {
...
+#'   Default `FALSE`.
+#'
+#' @returns The `project` object, invisibly.
+#' @export
+#' @family scenario
+addScenario <- function(
+  project,
+  scenarioName,
+  modelFile,
+  individualId = NULL,
```
sorinvoicu, line 714: individualId is NULL, yet checks in other functions are for is.na() ?

R/utilities-figures.R
```diff
@@ -1372,3 +1372,377 @@ createPlotsFromExcel <- function(
...
+  existingGrids <- project$plots$plotGrids
+  if (
+    !is.null(existingGrids) &&
+      nrow(existingGrids) > 0 &&
+      name %in% existingGrids$name
+  ) {
+    cli::cli_abort("plot grid {.val {name}} already exists")
+  }
+
+  unknown <- setdiff(plotIDs, project$plots$plotConfiguration$plotID)
```
sorinvoicu: silently always-fails when plotConfiguration has no plotID column

sorinvoicu (commented):
in general, check for inconsistencies between NULL, NA and "" id columns, silent failures or under-reporting.

R/utilities-figures.R
Felixmil: Addressed in 5519ef63

R/utilities-individual.R
Felixmil: Addressed in 5519ef63. Mirrored on end state branch (b6c4a4cd).

R/utilities-scenarios.R
Felixmil: Addressed in 5519ef63. Mirrored on end state branch (b6c4a4cd).

R/utilities-figures.R
Felixmil: Addressed in 5519ef63. Mirrored on end state branch (b6c4a4cd).

Felixmil:
@nautilus69 addressed in 533f351a. Mirrored on endstate branch (1ffc3bb5).

sorinvoicu:
> in general, check for inconsistencies between NULL, NA and "" id columns, silent failures or under-reporting.

should we consider similar approach for addIndividual, addPopulation, addApplication?

Felixmil:
> > in general, check for inconsistencies between NULL, NA and "" id columns, silent failures or under-reporting.
> 
> should we consider similar approach for addIndividual, addPopulation, addApplication?

All three already implement the same id-validation pattern: !is.character(id) || length(id) != 1L || is.na(id) || nchar(id) == 0. So they reject (fails is.character), NA (fails is.na), and "" (fails nchar == 0) with a clear error before doing anything else. The parameterSets arg also rejects undefined ids.

Issue #1031, opened 2026-05-07
Restore must_work existence-check in Project$.clean_path()


Felixmil:
## Context

Sub-issue of #1030 (umbrella for follow-ups from JSON-primary chapter PRs). Surfaced by @sorinvoicu during review of #1020.

## Current state on `runscenarios-from-project` (PR #1020)

`Project\$.clean_path()` declares `must_work = TRUE` in its signature but the body never references it:

```r
.clean_path = function(
  path,
  parent = NULL,
  must_work = TRUE,
  replace_env_vars = TRUE
) {
  # ... body uses fs::path_abs() but never checks file_exists() ...
}
```

## End-state (`json-as-primary-input-v2`)

The end-state body **does** use `must_work`: it warns (once per path) when the resolved path doesn't exist and `must_work = TRUE`. The signature is identical; only the body diverges.

```r
if (!fs::file_exists(abs_path) && must_work == TRUE) {
  if (!(abs_path %in% private\$.warned_paths)) {
    warning(messages\$fileNotFound(abs_path))
    private\$.warned_paths <- c(private\$.warned_paths, abs_path)
  }
}
```

So the chapter slice in #1020 stripped the existence-check body but kept the arg. The arg is currently dead, but removing it would just mean reintroducing it later when the existence check returns.

## Resolution

When the path-resolution layer fully lands (likely the chapter that introduces mutable `Project` state and the validator pipeline), restore the `must_work` body and the `.warned_paths` private slot from end-state, and add coverage that exercises both `must_work = TRUE` (warns once) and `must_work = FALSE` (silent) paths.

No action needed on #1020 — leaving the dead arg avoids a dead-code → restore-arg flip-flop across chapters.

Refs: #908, #1030, #1020

Felixmil:
Status update against `excel-deprecation-and-reorg` (chapter 8d): the premise is overtaken. The plan above was to leave the dead `must_work` argument in place, but commit b532ea01 removed it entirely; the current signature is `.clean_path(path, parent, replace_env_vars)` (R/project.R:473) and the `.warned_paths` slot is gone too. The underlying feature (warn when a resolved path does not exist) is still absent on the branch.

[...]

PR #1044, opened 2026-06-10
Chapter 8a: merge Project class and consolidate the Excel bridge

Felixmil:
First slice of chapter 8 of the JSON-primary rework, split out of #1038 for reviewability. The Excel-first surface is removed or soft-deprecated and the canonical in-memory representation collapses to a single exported `Project` R6 class, with the Excel bridge consolidated into one explicit import/export module. The diff is deletion-heavy: most of it is the removal of the legacy Excel-shape machinery.

## Project class

The internal `Project` class and the exported `ProjectConfiguration` class collapse into a single exported `Project`. The merged class owns both the parsed JSON sections (scenarios, individuals, populations, applications, observed data, plots, parameter identification) and the Excel-bridge file paths (`modelParamsFile`, `individualsFile`, `populationsFile`, `scenariosFile`, `applicationsFile`, `plotsFile`, plus the writable folder fields). `ProjectConfiguration()` survives as a soft-deprecated wrapper that forwards to `Project$new()`. `Project$new(projectFilePath)` is the single constructor; the `.loadProjectJson` shim and its piecewise constructor are gone.

## Excel bridge

`importProjectFromExcel()`, `exportProjectToExcel()`, and `projectStatus()` are new exports that consolidate the scattered Excel readers/writers into `R/project-excel.R`. The old `snapshotProjectConfiguration()`, `restoreProjectConfiguration()`, and `projectConfigurationStatus()` are now soft-deprecated wrappers that forward to the new names.

## Removals

- `ScenarioConfiguration`, `LegacyScenario`, and `createScenarios()` are deleted. `runScenarios()` accepts only a `Project`.
- `addScenarioConfigurationsToExcel()`, `readScenarioConfigurationFromExcel()`, `createScenarioConfigurationsFromPKML()`, and `setApplications()` are deleted.
- `validateAllConfigurations()` is deleted; `validateProject()` is the JSON-driven replacement.
- The Excel-shape validator files (`R/validation-{project-configuration, all-configurations, scenarios, individuals, populations, plots, models-applications, cross-references}.R`) are deleted; `isAnyCriticalErrors()` and `validationSummary()` move into `R/validation.R`.

Part of #908.

## Review stack

1. #1044 (this PR)
2. #1045 
3. #1046 
4. #1047


Felixmil:
_Review feedback assisted by the [critical-code-reviewer skill](https://github.com/posit-dev/skills/blob/main/posit-dev/critical-code-reviewer/SKILL.md)._

## Summary

**BLUF: Structurally the right move** — collapsing two parallel class hierarchies and consolidating the scattered Excel readers into one bridge is genuinely good cleanup, and the deletion-heavy diff is mostly legitimate. But the merge regressed a safety invariant the previous commits deliberately built, the new 1166-line migration bridge ships with effectively zero functional tests, and there's a layer of cargo-cult parameter threading and dead code that suggests the new file wasn't read back critically before submission. This is the load-bearing PR of a migration stack, and the migration path itself is the least-tested code in it.

## Critical Issues (Blocking)

**1. The `modified` / `validatedSinceMutation` lifecycle is now bypassable — and that silently runs unvalidated projects.**

In the old `Project`, `scenarios`, `individuals`, `populations`, `applications`, `observedData`, `plots`, `parameterIdentification`, `outputPaths`, and the three `*ParameterSets` were **read-only active bindings**. Mutation was only possible through the `add*`/`remove*` API, which calls `.markModified()` → `.invalidate()` → clears `validatedSinceMutation`.

This PR converts every one of those into a **plain public writable field** (`R/project.R:463-541`, `public = list(... scenarios = NULL, ...)`). Now `project$scenarios <- ...` and `project$scenarios[["x"]] <- ...` succeed silently and do **not** touch `modified` or `validatedSinceMutation`.

Combine that with the validation cache (`R/validation.R:151-154`):

```r
.ensureValid <- function(project, sections, opName) {
  if (isTRUE(project$validatedSinceMutation)) {
    return(invisible(NULL))   # <- skips validation entirely
  }
  ...
}
```

Failure mode: `validateProject(p)` sets the flag `TRUE` → user assigns `p$scenarios <- <invalid>` directly (now a supported path: the class is `@export`ed and the field is a documented `@field`) → `runScenarios(p)` / `createPlots(p)` see `validatedSinceMutation == TRUE` and **skip validation**, running an invalid project. The whole point of `validatedSinceMutation` was to make this impossible.

The inconsistency is visible inside the same class: the `filePaths`-backed folder fields (`modelFolder`, `dataFolder`, …) are active bindings that call `private$.invalidate()` on write (`R/project.R:219-232`), but the content fields don't. Either keep the content fields as private + active bindings that invalidate on write (assign to `private$.x` inside `.read_json`), or accept that the lifecycle flags are now decorative.

**2. The 1166-line Excel bridge — the stated purpose of this PR — has no functional tests.**

`tests/testthat/test-project-excel.R` is 40 lines containing three tests, each asserting only a `lifecycle_warning_deprecated` and wrapping the actual call in `tryCatch(..., error = function(e) NULL)` so every real error is swallowed. A grep of the whole `tests/` tree finds **zero** assertions exercising `importProjectFromExcel`, `exportProjectToExcel`, `projectStatus`, `.projectSync`, or the hand-rolled escape codec `.formatArrayToCommaList` / `.parseCommaListToArray`.

The PR deletes ~6,000 lines of Excel-machinery tests and replaces the new code's coverage with three deprecation smoke tests. For a migration bridge, "untested" means users' real projects are the test suite. At minimum: a round-trip test (`Excel → import → JSON → load → export → Excel`, assert stability), a `projectStatus` in-sync/out-of-sync pair, and a unit test for the comma-escape codec (`R/project-excel.R:1146-1162` has bespoke character-by-character escape-state logic — exactly the kind of code that's wrong in a corner you didn't think about).

## Required Changes

**3. `schemaVersion` is threaded through six parsers as a no-op, with docs that lie.** Every `.parseExcel*` function (`R/project-excel.R:714, 753, 785, 820, 855`) takes `schemaVersion` documented as *"Used to adapt parsing logic when importing Excel files generated by an older schema"* — and none branch on it. Pure speculative generality, and the roxygen misdescribes behavior that doesn't exist. Remove the parameter or implement the branching.

**4. Dead `esqlabsRVersion` read in `importProjectFromExcel`** (`R/project-excel.R:52-53`): read from the Excel with a careful fallback, then unconditionally overwritten at line 98 with `as.character(utils::packageVersion("esqlabsR"))`. The read value is never used. Honor the source file's version or delete lines 52-53.

**5. Dead `ignoreVersionCheck = TRUE` parameter** in `projectStatus` (`R/project-excel.R:461`) — declared, never referenced in the body.

**6. `must_work == TRUE`** (`R/project.R:921`). Comparing a logical to `TRUE`. Write `if (!fs::file_exists(abs_path) && must_work)`.

**7. Reads have side effects: accessors emit warnings, and `print()` triggers them.** `.clean_path` calls `warning(messages$fileNotFound(...))` when `must_work` (default `TRUE`) and the path is absent (`R/project.R:921-926`). The `modelFolder`/`configurationsFolder`/`dataFolder` active bindings call it with the default on every read, and `print()` reads all of them (`R/project.R:681-705`). So printing a freshly-loaded project whose folders don't exist yet spews `fileNotFound` warnings — from a print method. Property reads must be side-effect-free; `print()` should pass `must_work = FALSE`, and folder accessors arguably shouldn't stat the filesystem on read at all.

**8. Export/import filename asymmetry breaks the round-trip and `projectStatus`.** `exportProjectToExcel` hardcodes output filenames — `"ModelParameters.xlsx"`, `"Individuals.xlsx"`, … (`R/project-excel.R:315-396`) — and always writes into a `"Configurations/"` subdir (line 279). But it writes the `Project.xlsx` Property rows from `.getFilePathsData()`, which may hold *custom* names (e.g. `modelParamsFile = "MyParams.xlsx"`), and `importProjectFromExcel` resolves files by those custom property names. So a project with non-default file/folder names exports a `Project.xlsx` pointing at files that aren't there, and `projectStatus` then reports out-of-sync. Export must honor the stored names (and `configurationsFolder`), not the defaults.

## Suggestions

**9.** `R/project-excel.R:402-404` ships a `# TODO: verify symbol exists` comment claiming `messages$restoredProject` may be missing. The code actually calls `messages$restoredProjectConfiguration`, which exists (`R/messages.R:146`). The call is fine — delete the stale comment.

**10.** `importProjectFromExcel` writes JSON and returns without ever loading it back through `Project$new()` / `validateProject()`. The migration can silently emit malformed JSON; a cheap post-write round-trip would catch it at migration time.

**11.** `jsonPath` (public mutable field) and `projectFilePath` (read-only active binding) hold the same absolute path, set together in `.read_json` (`R/project.R:941-944`). Two sources of truth for one value — pick one.

**12.** The `ProjectConfiguration()` shim (`R/project.R:865-871`) swallows `...` silently. Old callers passing extra args get them dropped without warning.

**13.** Minor: `initialize(projectFilePath = character())` then tests `!missing(projectFilePath) && length(...) > 0L` (`R/project.R:600-607`) — the `!missing` is redundant given the `character()` default.

Credit where due: the comma-escape codec is the one spot where the corner case was clearly considered (escaping `\` and `,` so ids round-trip). Now write the test that proves it.

## Verdict

**Request Changes.** Two blockers: the bypassable validation lifecycle (#1) can run unvalidated projects, and the flagship migration bridge is untested (#2). Neither is a rewrite — #1 is "make the content fields invalidate on write like the folder fields already do," and #2 is "add the round-trip and codec tests." The required-changes list is mostly dead code and lying docs that a read-back would have caught.

---

Felixmil:
## Review triage: cross-referenced against the stack and the audit issues
I cross-referenced all 13 findings against `merge-simulation` (which stacks #1044 → #1047 plus the in-flight Wave fix PRs #1060 to #1067) and against the chapter 8d audit issues (#1048 to #1066). Most of the blocking and behavioral findings are already fixed downstream or already tracked with an open fix PR; what genuinely belongs in **this** PR is the dead-code and lying-docs cleanup.

### Status of each finding

| # | Finding | Status now | Home |
|---|---|---|---|
| 1 | Content fields bypass invalidation (blocker) | **Fixed** downstream | #1048 / #1060 (in merge-simulation) |
| 2 | Excel bridge has ~no functional tests | Open (2 narrow tests exist, the broad suite does not) | #1049 |
| 3 | `schemaVersion` threaded as a no-op, docs lie | **Open, fix here** | this PR |
| 4 | Dead `esqlabsRVersion` read | **Open, fix here** | this PR |
| 5 | Dead `ignoreVersionCheck` param | **Open, fix here** | this PR |
| 6 | `must_work == TRUE` comparison | **Fixed** downstream (`.clean_path` no longer takes `must_work`) | end-state |
| 7 | Reads have side effects (print warns) | **Fixed** downstream (`.clean_path` is side-effect-free) | end-state; sibling #1031 closed into #1051/#1063 |
| 8 | Export/import filename + folder asymmetry | Open (behavioral, entangled with the round-trip rework) | #1049 |
| 9 | Stale `# TODO: verify symbol exists` comment | **Open in this diff, fix here** | this PR |
| 10 | No post-write round-trip self-check | Open (defensive guard, depends on 8) | #1049 |
| 11 | Two sources of truth for the project path | **Fixed** downstream (`jsonPath` is now a read-only binding) | #1048 / #1060 |
| 12 | `ProjectConfiguration()` shim swallows `...` | Open (deprecation-surface decision, defer) | #1058 |
| 13 | Redundant `!missing` in `initialize()` | **Open, fix here** | this PR |

A note on finding 7: it was resolved by **removing** the existence stat from `.clean_path()` rather than restoring it, which is the right call (property reads must be side-effect-free). The existence check that disappeared is now re-homed in the validation layer (a `validateProject()`-time check that a scenario's `modelFile` resolves on disk), tracked on #1051 and implemented by #1063. The sibling issue #1031 was closed in favor of that.

R/project-excel.R
```diff
@@ -0,0 +1,1146 @@
...
+  result
+}
+
+#' Convert parameter structures to Excel sheet data frames
+#' @param parameterSets Named list of parameter structures (paths, values,
+#'   units)
+#' @returns Named list of data frames suitable for Excel sheets
+#' @keywords internal
+#' @noRd
+.parameterStructuresToExcelSheets <- function(parameterSets) {
```
nautilus69, line 864: .parameterStructuresToExcelSheets() expects the parallel-vector shape list(paths, values, units), but every call site (:309, :324, :372) feeds it project$modelParameterSets etc, whose canonical shape is a list of entry objects [{containerPath, parameterName, value, units}].

So params$paths is NULL will result in the early-return at :868 (length(params$paths) == 0) fires for every set then a 0-row sheet is written.

R/project-excel.R
```diff
@@ -0,0 +1,1146 @@
...
+    for (sheetName in names(project$plots)) {
+      df <- project$plots[[sheetName]]
+      if (is.data.frame(df) && nrow(df) > 0) {
+        plotSheets[[sheetName]] <- df
+      }
+    }
+    if (length(plotSheets) > 0) {
+      .writeExcel(plotSheets, file.path(configDir, "Plots.xlsx"))
+    }
+  }
```
nautilus69, line 393: This is about backward compatibility: 

only writes sheets where is.data.frame(df) && nrow(df) > 0. But .parsePlots() stores dataCombined as a named list (not a data.frame), so it's silently skipped. plotConfiguration/plotGrids (data.frames) survive; DataCombined definitions are lost on JSON→Excel.

R/validation.R
nautilus69, line 350: THIS IS NOT FOR THIS PR BUT AFFECTS THE OUTCOME, but this PR makes validation.R the canonical JSON-driven validator replacing validateAllConfigurations()

reads set$paths/set$values; the adapter (utilities-parameters.R:11) passes project$modelParameterSets  (entry-list shape). So length(paths) == 0 → next for every set (repro: set Global -> SKIPPED, entries never validated). validateProject()  therefore never checks parameter-set contents (empty paths, non-numeric values, duplicate paths) — for any of the three *ParameterSets  sections.

nautilus69 (commented):
These two are documentations which can be ommited at the moment. 

1- Dangling Rd link [saveProject()] (project.R:34); no such exported function/topic exists (only internal .saveProjectJson). R CMD check will warn.
2- Stale docs that this PR's own deletions invalidated: @examples at utilitiesscenarios.R:230-233 and :325-328 still call createScenarios() and readScenarioConfigurationFromExcel() (both deleted here) plus ProjectConfiguration; comment at validation.R:7references the deleted validateAllConfigurations()/file.


The rest are requested changes

R/project-excel.R
Felixmil: Already identified in tracked in #1049

R/project-excel.R
Felixmil: Already identified and tracked in #1049

R/validation.R
Felixmil: Already identified and tracked in #1051

nautilus69 (approved):

PR #1045, opened 2026-06-10
Chapter 8b: JSON-first project lifecycle API

Felixmil:
Second slice of chapter 8 of the JSON-primary rework, split out of #1038 for reviewability. With the Excel-first machinery removed in the previous slice, the remaining API gaps for a fully JSON-driven workflow land here so chapter 9 can rewrite the vignettes against a stable surface: save, init, and PKML scenario seeding. This is the slice with the densest new code.

## Lifecycle

- `saveProject(project, path)` is the exported inverse of `loadProject()`. It calls `project$.markSaved()`.
- `initProject()` is rewritten with `type = c("minimal", "example")` and `createExcel = TRUE`; the minimal template is seeded from a new `inst/extdata/projects/Blank/Project.json`. By default it produces Excel side-cars next to the JSON.
- `isProjectInitialized()` also recognizes `Project.json`.
- `exampleProjectPath()` returns the bundled example `Project.json`; `exampleProjectConfigurationPath()` is soft-deprecated.

## PKML scenario seeding

`createScenariosFromPKML(pkmlFilePaths, project, ...)` replaces the removed `createScenarioConfigurationsFromPKML()` and attaches scenarios directly to a `Project` (new `R/scenario-from-pkml.R`).

Note: this slice lands before the per-domain file reorganization (next slice in the stack), so the touched files still carry their pre-reorg names.

Part of #908.

## Review stack

1. #1044 
2. #1045 (this PR)
3. #1046 
4. #1047


Felixmil:
## Triage: every finding is resolved downstream in the stack

Cross-referenced each item against `merge-simulation` (the integration branch stacking #1044 to #1047 plus the "Wave" fix PRs #1060 to #1067) and the chapter 8d audit. **The substance of this review is the same set the audit caught, and it is already fixed, primarily by #1061 ("Fix createScenariosFromPKML load, seed, save", `Closes #1050`).** Verdicts are verified against the end-state, not this slice in isolation.

| # | Finding | Status | Resolved by |
|---|---|---|---|
| 1 | Silent scenario overwrite (duplicate-key data loss) | ✅ Resolved | #1061 |
| 2 | Vignette removal breaks R CMD check | ✅ Resolved | end-state (#908) |
| 3 | Leaked agent caller name in snapshots | ✅ Resolved | end-state |
| 4 | Unguarded unit comparison throws "length zero" | ✅ Resolved | #1061 |
| 5 | Test coverage decorative | ✅ Resolved | #1061 |
| 6 | Unchecked `file.copy` return in `initProject` | ✅ Resolved | end-state |
| 7 | PR description contradicts code (attach vs. list) | ✅ Resolved | #1061 |
| S2 | base `warning()` vs cli | ✅ Resolved | #1061 |
| S4 | `sapply` to `vapply(..., character(1))` | ✅ Resolved | #1061 |
| S5 | `.getScenarioCount` doc says Integer, returns double | ✅ Resolved | #1061 |
| S6 | Obvious comments | ✅ Mostly removed | #1061 |
| S3 | Bespoke outputPaths recycling fork | 🟡 Style only | #1061 (no longer a correctness risk) |

<details>
<summary>Per-finding detail</summary>

#1061 rewrites the seeding path so `createScenariosFromPKML()` mutates the project in place (routing output paths through `addOutputPath()` and scenarios through `addScenario()`) and returns it invisibly, instead of returning a detached list. That single rewrite resolves the data-loss bug, the unit-comparison crash, the attach contract, and the missing coverage.

**1. Silent scenario overwrite.** The same bug the audit logged in #1050 (`c("S", "S", "S_2")` returns 2 scenarios). De-dup now tracks the accumulated final keys rather than the original-name history, so that input yields three distinct scenarios (`S`, `S_2`, `S_2_2`), pinned by a regression test.

**2. Vignette removal breaks R CMD check.** On the end-state every vignette that still calls the removed Excel-first API is set to `eval = FALSE` with a `TODO(#908)` marker, so vignette re-building no longer executes the deleted API and CI is green. The vignette rewrite itself is the chapter 9 worklist (#908).

**3. Leaked agent caller name (`btw_mcp_server()`).** A snapshot-generation artifact: `ospsuite.utils::validateIsOfType()` interpolates the outermost calling frame into the message, so generating the snapshot through an MCP/agent harness captured that harness on the stack. The frame is environment-dependent (`test_file()` under a normal run, `tryCatch()` when called directly), so it is the wrong thing to pin in a snapshot regardless. On the end-state these assertions no longer snapshot a caller-sensitive message, so the leak cannot recur.

**4. Unguarded unit comparison.** Simulation-time extraction now guards the display unit with `%||%` before the comparison, with a test covering the NULL-unit case.

**5. Test coverage decorative.** `test-scenario-from-pkml.R` goes from 4 shallow tests to roughly 40, covering exactly the branchy logic this finding called out: duplicate-name expansion, vector recycling, named-vector output paths and id collisions, comma-splitting, simulation-time extraction and unit conversion, steady-state, model-file resolution, transactional rollback, and end-to-end load/seed/save round-trips. This was #1050's headline "Required tests" list.

**6. Unchecked `file.copy` return.** On the end-state the copy result is checked and a partial failure aborts with the failed source files listed, instead of proceeding to load a half-written project.

**7. PR description contradicts code.** The function now genuinely attaches: it mutates the project via `addScenario()` and returns `invisible(project)`, like the other `add*` mutators, exactly what the audit's "no supported way to attach the result" finding asked for. Description and code now agree.

**Suggestions.** S2 (cli), S4 (`vapply`), S5 (integer return), S6 (comments) are all resolved downstream. S3: the rewrite routes output paths through `addOutputPath()` and handles the named-vector case correctly; the list-vs-vector fork structure remains but is no longer a correctness risk.

</details>

**Net:** the substance here is the #1050 audit, already implemented by #1061 against the integration branch. Re-fixing it on this slice in isolation would conflict with that rewrite. The intent behind every blocking finding is addressed downstream and verified on `merge-simulation`.

rengelke (approved):
Nice, this is a big one! Project is finally the single source of truth with a clean `saveProject()`/`loadProject()` round-trip, and building scenarios straight from PKML.

All found issues "resolved downstream". Thanks. Approved

PR #1046, opened 2026-06-10
Chapter 8c: per-domain file reorganization and test fixtures

Felixmil:
Third slice of chapter 8 of the JSON-primary rework, split out of #1038 for reviewability. Surviving code lands in its final per-domain home, project loading and test fixtures are rationalized, and the remaining user-facing Excel-only helpers come out. Mostly file movement and deletion; best reviewed commit by commit (4 commits).

## Per-domain file reorganization

`R/utilities-scenarios.R` splits into `R/scenarios.R` and a new `R/output-paths.R`. `R/utilities-individual.R` becomes `R/individuals.R`, `R/utilities-population.R` becomes `R/populations.R`, `R/utilities-parameters.R` becomes `R/parameters.R`, `R/utilities-figures.R` becomes `R/plots.R`, and `R/utilities-data-combined.R` becomes `R/data-combined.R`. The final commit aligns the remaining layout with the end state (`R/create-plots.R`, `R/plots-utils.R`, `R/scenario-results.R`, `R/utils.R`, `R/validation.R` consolidation). Test files are renamed to mirror sources.

## Excel-only helper removal

User-facing Excel-only helpers are deleted; the explicit interop wrappers from the first slice (`importProjectFromExcel()`, `exportProjectToExcel()`, `projectStatus()`) are the remaining Excel surface. Removed exports: `exportParametersToXLS()`, `loadObservedDataFromExcel()`, `loadObservedDataFromPKML()`, `readIndividualCharacteristicsFromXLS()`, `readPopulationCharacteristicsFromXLS()`, `writeIndividualToXLS()`, `writeParameterStructureToXLS()`.

## Project loading and test fixtures

Project loading is rationalized: parsing consolidates into `Project$new()` and `loadProject()` warns on unresolved cross-references at load time. The legacy `inst/extdata/examples/TestProject/` Excel project moves to `tests/testthat/data/TestProjectExcel/` (test-only, not user-shipped). The `tests/data/` flat-files directory consolidates into `tests/testthat/data/`. Test helpers in `tests/testthat/helpers.R` are reduced from ten to five with clear roles: `getTestDataFilePath()`, `testProject()`, `exampleProject()`, `testProjectExcelPath()` plus its `Configurations/` sibling, and `with_temp_project()`.

Part of #908.

## Review stack

Replaces #1038. Review bottom up; each PR's base is its predecessor's branch, so each diff shows only its own slice.

1. #1044 
2. #1045 
3. #1046 (this PR)
4. #1047

PR #1047, opened 2026-06-10
Chapter 8d: runPI semantics, cli convergence, and chapter-chain polish

Felixmil:
Final slice of chapter 8 of the JSON-primary rework, split out of #1038 for reviewability. Behavior polish and convention sweeps across the chapter chain: a single mutation surface, stricter `runPI()` failure semantics, cli error and warning style, test enablement with a snapshot refresh, and small cleanups.

## Mutation surface

`Project` no longer exposes mutation methods (`project$addScenario()`, `project$removeScenario()`, etc.). The free-function form is canonical: call `addScenario(project, ...)`, `removeScenario(project, ...)`, and the matching mutators directly. The R6 wrappers were pure delegation and have been dropped to keep one mutation surface to maintain.

## runPI failure semantics

`runPI()` no longer wraps the build phase in `tryCatch`. Build errors (typos in parameter paths, unknown outputs, missing observed data) now hard-fail and surface immediately. Only `runtime$run()` is wrapped, so numerical optimization failures still degrade to `result = NULL` and the loop continues with the remaining tasks. The soft-fail entry now also carries the built runtime on `task` (was `NULL`), so callers can inspect the optimization setup.

## Error and warning style

Chapter-chain code converges on `cli::cli_abort()` / `cli::cli_warn()`. Every `stop(messages$...)` / `warning(messages$...)` site introduced or rewritten by chapters 1 to 8 (PI module, parsed-Project section files, scenario-from-pkml, project lifecycle and Excel bridge, esqlabs-env, plus inline string-concat `stop()` sites in `project.R`, `project-to-json.R`, `scenarios.R`) now goes through cli. The `messages$` registry entries are unchanged; the only visible behavioral difference is the condition class (`rlang_error` / `rlang_warning`) and cli's terminal-width wrapping. Legacy code on `v6` (sensitivity, plots-utils, data-utils, utils, create-plots Excel helpers) keeps its existing call style; converging it is tracked as a follow-up.

## Tests and snapshots

showtext is disabled for tests, previously skipped tests are enabled on macOS, and snapshots are refreshed accordingly (the bulk of the diff size).

## Cleanups

- `Scenario$asList` now iterates an explicit `.scenarioFieldNames` constant instead of `ls(self)` reflection, so adding methods to `Scenario` no longer changes the JSON-shape contract.
- Vignette code execution is disabled pending the chapter 9 rewrite.
- Dead `must_work` / `.warned_paths` machinery in `Project$.clean_path()` and the dead `.safe_validate` / `.categorize_message` helpers are dropped.
- `exportProjectToExcel()` flattens `outputPaths` with `unlist()` when building the OutputPaths sheet, with round-trip coverage.

Part of #908.

## Review stack

1. #1044 
2. #1045 
3. #1046 
4. #1047 (this PR)


R/esqlabs-env.R
```diff
@@ -25,7 +25,7 @@ esqlabsRSettingNames <- enum(names(esqlabsEnv))
 #' getEsqlabsRSetting("packageName")
 getEsqlabsRSetting <- function(settingName) {
   if (!(any(names(esqlabsEnv) == settingName))) {
-    stop(messages$errorPackageSettingNotFound(settingName, esqlabsEnv))
+    rlang::abort(messages$errorPackageSettingNotFound(settingName, esqlabsEnv))
```
Felixmil: **Required Changes (consistency)** — The PR description states the chapter-chain converges on `cli::cli_abort()` / `cli::cli_warn()`, and explicitly lists `esqlabs-env` as converted. This site uses `rlang::abort()` instead. The artifact contradicts the stated intent.

Functionally `rlang::abort()` and `cli::cli_abort()` both produce a `rlang_error`, so the condition class is the same — but `rlang::abort()` does *not* re-run cli markup/width-wrapping over the message, which is the whole point of the sweep (see the snapshot re-wrap in `parameter-identification.md`). So this lone site behaves subtly differently from every other converted site.

Either make it consistent:

```suggestion
    cli::cli_abort(messages$errorPackageSettingNotFound(settingName, esqlabsEnv))
```

or, if `rlang::abort()` was a deliberate choice (e.g. because `errorPackageSettingNotFound` lives in `ospsuite.utils` and you specifically want to avoid re-processing), drop `esqlabs-env` from the "converged to cli" list in the description.

---

R/data-combined.R
```diff
@@ -52,7 +52,7 @@ createDataCombined <- function(
     names(allSpecs)
   )
   if (length(missingNames) > 0) {
-    stop(messages$stopDataCombinedNamesNotFound(missingNames))
+    cli::cli_abort(messages$stopDataCombinedNamesNotFound(missingNames))
```
Felixmil, line 55: **Verify (latent edge case, applies to the whole `cli::cli_abort(messages$...)` sweep)** — The `messages$` registry entries return strings that are *already* resolved by `cliFormat()` / `cli::format_message()`. Wrapping a pre-resolved string in `cli::cli_abort()` means cli parses it as a template a **second** time. The intended effect (terminal-width re-wrapping) is confirmed by the `parameter-identification.md` snapshot, so the common path is fine.

The residual failure mode: any message whose *resolved* text contains a literal `{` or `}` — e.g. an interpolated path, label, or value that happens to carry a brace — will be re-interpreted by cli on the second pass and throw `unexpected '}'` *from inside the error path*, masking the real error. Domain paths use `|` separators so the probability is low, but it's a footgun for user-supplied identifiers.

Not blocking. If you want to harden it, pass raw templates to `cli_abort` (let it format once) rather than pre-formatting in the registry, or escape with `{{`/`}}`. Flagging here as representative of every converted call site, not just this line.

---

R/scenario.R
```diff
@@ -6,6 +6,27 @@
 # Population loading, no parameter merging. Construction is by direct
 # field assignment.
 
+# Explicit field list backing `Scenario$asList`. Kept in sync with the
+# v2.0 `Project.json` scenario entry shape so adding a method on
+# `Scenario` does not change the JSON-shape contract.
+.scenarioFieldNames <- c(
```
Felixmil, line 12: **Strong Suggestion** — Replacing `ls(self)` reflection with an explicit constant is the right call: it stops methods from leaking into the JSON shape. But the comment claims this is "kept in sync with the v2.0 schema" while nothing *enforces* that sync. If someone adds a new public data field to `Scenario` and forgets to append it here, `asList` will silently drop it from the JSON contract — and `self[[missingName]]` returns `NULL` rather than erroring, so a typo in this constant also fails silently.

The accompanying test (`asList exposes exactly the v2.0 schema fields`) does not catch this — it asserts the names equal the same hardcoded list, so it's tautological w.r.t. the constant. Consider deriving the field set from the object and diffing, e.g. assert that `.scenarioFieldNames` equals `setdiff(names(Scenario$public_fields), <known-non-serialized>)`, so field drift breaks the build instead of silently corrupting the round-trip.

---

tests/testthat/test-scenario.R
```diff
@@ -24,6 +24,36 @@ test_that("Scenario has the documented field defaults", {
   expect_null(sc$modelParameterSets)
 })
 
+test_that("Scenario$asList exposes exactly the v2.0 schema fields", {
```
Felixmil, line 27: **Strong Suggestion** — This test hardcodes the exact same 15 names as `.scenarioFieldNames`, so it only verifies that `asList` echoes the constant — it cannot detect the failure it's meant to guard against (a `Scenario` field added without updating the constant). To make it load-bearing, assert the constant against the object's actual public data fields, e.g. derive the expected set from `Scenario$public_fields` minus the known non-serialized members, then check both `asList` names and `.scenarioFieldNames` against it.

---

Felixmil (commented):

## Summary

This is a clean, disciplined convention-sweep PR, and the discipline shows. I went in assuming the 12k-line deletion was hiding something; it mostly isn't. The R6 mutation-method removal is genuinely pure delegation, the dead-code removal in `Project$.clean_path()` is verifiably dead (all 11 call sites passed `must_work = FALSE`, so the warning branch was unreachable), the `sapply` → `vapply`/`lapply` conversions are correct type-hardening, and the `runPI()` semantics change is matched by two focused tests that exercise both the hard-fail (build) and soft-fail (optimisation) paths. NEWS.md and NAMESPACE are kept honest with the breaking changes. The `deprecate_warn` `when =` normalisation to `"6.0.0"` is correct given `DESCRIPTION` is at `6.0.0.9001` — the prior `"7.0.0"`/`"6.1.0"` values were speculative.

The findings below are minor. None is blocking; none risks data corruption or a logic error. I'm marking this Needs Discussion only because of the one self-contradiction against the PR description and two tests/abstractions that don't enforce what they claim to.

## Required Changes

1. **`R/esqlabs-env.R:28`** — uses `rlang::abort()` while the PR description lists `esqlabs-env` among the sites converged to `cli::cli_abort()`. Same condition class, but no cli re-wrapping, so it behaves unlike every other converted site. Make it consistent or correct the description. (inline)

## Strong Suggestions

2. **`R/scenario.R:12` / `tests/testthat/test-scenario.R:27`** — `.scenarioFieldNames` is a manually-maintained mirror of the `Scenario` data fields, and both the comment and the test claim it stays "in sync" with the schema, but nothing enforces it. The test is tautological (asserts `asList` echoes the same hardcoded list), and `self[[missingName]]` returns `NULL` rather than erroring, so a forgotten field or a typo silently corrupts the JSON contract. Derive the expected set from `Scenario$public_fields` so drift breaks the build. (inline)

3. **cli double-processing across the sweep** (representative inline on `R/data-combined.R:55`) — `messages$` entries return already-resolved cli strings, so `cli::cli_abort(messages$...)` parses them a second time. The intended re-wrap is confirmed by the snapshot diff; the residual risk is a resolved message containing a literal `{`/`}` (interpolated identifier) throwing from inside the error path. Low probability with `|`-delimited domain paths, but a footgun worth a comment or escaping strategy.

## Noted

4. **`R/observed-data.R:14`** (outside the diff hunks, so noting here rather than inline) — the source-types comment still describes the `programmatic` type as "DataSets added at runtime via `project$addObservedData()`". This PR removes that R6 method; the canonical call is now `addObservedData(project, ...)`. Stale after this change.

Remaining items are minor. Good PR.

## Verdict

Needs Discussion

---

R/esqlabs-env.R
Felixmil: Fixed in 9428efee: converted this site to `cli::cli_abort(messages$errorPackageSettingNotFound(...))` so it matches every other converted site and the PR description.

Verified the message has no `{`/`}` or `\` so the cli re-processing is safe, but it does re-wrap at terminal width, which broke the raw-message regex in `test-esqlabsr-settings.R`. Switched that assertion to `expect_snapshot(error = TRUE)` (full message reviewable, robust to wrapping) and recorded the snapshot.

Also fixed the other item from your review while here: the `observed-data.R` source-types comment now reads `addObservedData(project)` instead of the removed `project$addObservedData()` method.

R/scenario.R
Felixmil: Agreed the constant + test don't enforce sync, and deriving from `Scenario$public_fields` is the right hardening. Deferring it out of this PR, though: `mutation-surface` (#1060, closes #1048) reworks this exact code as part of the "Scenario as plain-data record" decision (audit theme 8). On that branch `R/scenario.R` and `tests/testthat/test-scenario.R` are removed entirely (`scenario.R` is merged into `scenarios.R` and `Scenario` becomes a plain-data record), so any enforcement we add here gets thrown away in the rebase. Better to land the derive-from-object check against the post-#1060 shape. Tracking under #1048.

tests/testthat/test-scenario.R
Felixmil: Same disposition as the `scenario.R:12` thread: correct that this test only echoes the constant and can't catch field drift. Not fixing it on this branch because #1060 deletes this whole test file when `Scenario` becomes a plain-data record; the load-bearing version (assert `.scenarioFieldNames` against the object's actual public data fields) belongs against that end-state. Tracked under #1048.

R/data-combined.R
Felixmil: Confirmed real, and not blocking. One nuance worth recording: the second cli pass is actually load-bearing for at least one site — `messages$warningPIOptimizationFailed` carries a `\\` line-fold that only resolves on the `cli_warn()` re-process, so removing the double-processing wholesale would regress it. The residual risk you flag (a resolved message carrying a literal `{`/`}` throwing on the second parse) is genuine but low given `|`-delimited domain paths. This is cross-cutting across every converted `cli::cli_abort(messages$...)` site, so it belongs in the messages/cli convention sweep rather than this PR. Tracked under #1057.

Felixmil:
## Code review

Disciplined convention-sweep at the tip of the chapter-8 stack; the structural change is sound. The load-bearing behaviour is correct: the `runPI()` build/optimization `tryCatch` split works as described (build via `.createSinglePITask()` hard-fails outside the wrapper, only `runtime$run()` is wrapped, soft-fail entry carries `task = runtime`), the R6 mutation-method removal is pure delegation with no dangling call sites, the `globalVariables`/`colorPalette`/`_R_CHECK_LENGTH_1_CONDITION_` relocation out of `.onLoad` is a genuine fix, dead code is cleanly removed, and man/NAMESPACE/`_pkgdown.yml`/NEWS are all in sync (`document()` produces no drift). One substantive defect was found and is now fixed in this PR; the rest are minor or owned elsewhere in the stack.

Findings are ordered by criticality. The last two columns make tracking explicit: **Tracked (issue)** links the issue that owns the item, **Following PR** links a later PR that fixes/addresses it (or notes when none exists yet). Severity status: open / fixed in this PR / not a bug.

| # | Issue | Location | Severity | Status | Tracked (issue) | Following PR |
|---|-------|----------|----------|--------|-----------------|--------------|
| 1 | **cli double-processing.** `cli::cli_abort/warn(messages$…)` re-parses an already-rendered string (`cliFormat` = `cli::format_inline`) as a glue template; a literal `{`/`}` in interpolated user/external data crashes with a cryptic glue error, masking the real diagnostic. Worst case confirmed end-to-end: an uncontrolled optimizer `e$message` with braces made `runPI()` hard-crash the whole loop instead of degrading to `result = NULL`, defeating this PR's own soft-fail feature. | `R/parameter-identification.R` (runPI soft-fail handler) | High | fixed in this PR | [#1057](https://github.com/esqLABS/esqlabsR/issues/1057) (for the remaining ~68 sites) | this PR (the runPI site); remainder has no PR yet, see #1057 note below |
| 2 | `.md` snapshots refreshed (cli reflow + showtext). Legitimate, not a blessed-broken-message regression (test data uses `\|`-delimited paths with no braces, so messages render fine). | `tests/testthat/_snaps/populations.md`, `…/parameter-identification.md` | Low | not a bug | — | — |
| 3 | `exportProjectToExcel()` `unlist(outputPaths, use.names = FALSE)` paired with a separate `names()` column. Well-formed input is fine and round-trip-tested; a malformed element is fragile (`NULL`/nested → hard error; `character(0)` → silent row misalignment). | `R/project-excel.R:353` | Low | open | [#1049](https://github.com/esqLABS/esqlabsR/issues/1049) | none yet (wave-2; no open PR) |
| 4 | `.scenarioFieldNames` was a hand-maintained mirror with no drift guard. Verified to match the field set exactly at this PR's head, so no live drift. | `R/scenario.R:9-24`; `tests/testthat/test-scenario.R:27-50` | Low | open in this PR | [#1048](https://github.com/esqLABS/esqlabsR/issues/1048) | ✅ [#1060](https://github.com/esqLABS/esqlabsR/pull/1060) (commit `0270bfdb`) adds the drift guard. On that branch `Scenario` is a plain-data record, so it ties `.scenarioFieldNames` to the `Scenario()` formals (`expect_identical(.scenarioFieldNames, names(formals(Scenario)))`); a field added to only one source now breaks the build |
| 5 | Error assertions in some test files weakened from exact `messages$…` matchers to loose substrings (side effect of finding 1). Real coverage reduction, counter to the house `expect_snapshot(error = TRUE)` style. | `tests/testthat/test-data-combined.R:55,91,107,118,172`; `test-parameters.R:44,82` | Low | open | [#1057](https://github.com/esqLABS/esqlabsR/issues/1057) | none yet (rides the #1057 cli sweep) |
| 6 | Stale reference to the removed `createDefaultProjectConfiguration()` (the removal itself is correctly a NEWS breaking-change bullet; NAMESPACE/man/`_pkgdown.yml` consistent). Non-executing fence, so no build break. | `vignettes/project-migration-guide.Rmd:92,103` (pre-existing) | Low | open | [#908](https://github.com/esqLABS/esqlabsR/issues/908) (chapter-9 docs) | none yet (chapter-9 docs rewrite) |
| 7 | runPI tryCatch split; R6 mutation removal; dead-code removal (`.safe_validate`/`.categorize_message`/`must_work`); `globalVariables`/`colorPalette`/`_R_CHECK_LENGTH_1_CONDITION_` relocation; `deprecate_warn(when="6.0.0")` normalization; `sapply`→`vapply`/`lapply`; SVG showtext churn; `esqlabs-env.R` fully cli; `observed-data.R:14` comment fixed | various | n/a | not a bug | — | — |

Note on #1057 (finding 1 remainder): the ~68 other `cli_*(messages$…)` sites are the same latent pattern but interpolate low-brace-risk values (paths, names). The fix is a deliberate stack-wide convention decision tracked in [#1057](https://github.com/esqLABS/esqlabsR/issues/1057). Important caveat recorded there: the migration `cliFormat` planned in #1057 does **not** fix this class on its own (the `cliFormat` end-state crashes identically), so the issue must not be considered closed by that migration alone.

### Bottom line

The architecture and all the cleanup/feature work are good (runPI split, mutation-surface removal, dead-code purge, globals/`.onLoad` relocation, deprecation handling). The one substantive defect (finding 1) is fixed here, with the systemic remainder correctly scoped to #1057. Findings 3-6 are non-gating and owned by their respective stack issues. No open blocker remains in this PR.

rengelke (approved):
This is a clean.

Issue #1050, opened 2026-06-10
createScenariosFromPKML() output cannot be saved: load, seed, save is broken


Felixmil:
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. The headline findings were reproduced live.

The 8b/8d pair `createScenariosFromPKML()` plus `saveProject()` does not compose: **load, seed, save is broken on the default path**.

## Findings

- [ ] **Critical**, R/scenario-from-pkml.R:486-491: seeded scenarios store unnamed `outputPaths` (bare `vapply()` result; same at 481 for the comma-split user input), but `.scenariosToJson()` (R/project-to-json.R:130-147) requires the named id-to-path shape and aborts with `Scenario "..." has outputPaths without ids.` Even when the user passes named paths, the ids are never registered in `project$outputPaths`, so the serializer's `setdiff(pathIds, names(project$outputPaths))` check aborts next. The fix needs both halves: name the vector, and register (or reuse) the ids in `project$outputPaths`.
- [ ] R/scenario-from-pkml.R:576-598: with `steadyState = TRUE`, the block converts `steadyStateTime` using a local `timeUnit` but never assigns `sc$steadyStateTimeUnit`, so `saveProject()` hits the serializer's "simulateSteadyState=TRUE but steadyStateTimeUnit is NULL" stop. One-line fix.
- [ ] R/scenario-from-pkml.R:388-400: duplicate-name resolution checks only the original names; `scenarioNames = c("S", "S", "S_2")` returns 2 scenarios instead of 3 (the renamed second "S_2" is overwritten by the third), silently.
- [ ] R/scenario-from-pkml.R:406-409: `fs::path_rel(pkmlPath, start = project$modelFolder)` dies with a raw fs error ("argument is not a character vector") when `modelFolder` is NULL (empty `Project$new()`, or a Project file whose `filePaths` omits it). Guard with a proper message or fall back to the absolute path.
- [ ] R/scenario-from-pkml.R:443-453: the default `applicationProtocol` goes through `.sanitizeExcelSheetName(warn = TRUE)`, so a JSON-only user with a 32-character scenario name gets a warning about Excel sheet-name limits and a silently truncated protocol name, in the flagship JSON-first API.
[...]

Issue #1058, opened 2026-06-10
Docs and metadata: NEWS baseline, dependency floors, deprecation surface, stale references


Felixmil:
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. Single collection issue for everything docs and metadata.

## Release metadata

- [x] DESCRIPTION: `ospsuite (>= 12.2.0)` is behind the released floor; main requires `>= 12.4.2` since 5.7.0 ("earlier versions fail to load data correctly", #1000). The v6 chain forked before that bump; merging v6 over main would silently regress a released breaking requirement. *(Done on branch in wave 0: floor raised to `>= 12.4.2` on 8c `b1fcfbec`, rippled to 8d. Not yet on v6.)*
- [x] DESCRIPTION declares `R (>= 4.4)` while the repo development instructions still mandate R 4.1 compatibility (no `_$x` placeholder syntax etc.). Decide which is authoritative and update the other. *(Resolved in wave 0: `R (>= 4.4)` is authoritative; ospsuite 12.4.2 itself `Depends: R (>= 4.4)`, so R 4.1 is not satisfiable. The stale R 4.1 note in the dev instructions was corrected; no committed repo diff since those instructions live outside the package tree.)*
- [ ] NEWS.md is baselined against 5.6.0: it jumps from "(development version)" straight to "# esqlabsR 5.6.0"; the released 5.7.0 section (Excel PI workflow, ospsuite 12.4.2 bump, #1007 fix) exists only on main. Restore it and re-baseline the dev bullets, in particular: the PI rework sits under "Internal (work in progress)" although it removes and re-signatures API that 5.7.0 shipped publicly; line 88 duplicates the #1007 fix that already shipped in 5.7.0; line 45 claims `loadObservedDataFromExcel()` was "removed (soft-deprecated since 5.7.0)" though that name never shipped in any release, while the real released delta (`loadObservedData()` signature change from `(projectConfiguration, sheets, ...)` to `(project)`) has no Breaking-changes bullet; lines 71-80 still describe the superseded #970 Individuals-column contract; lines 82-84 are a duplicate `## New features` heading describing removed API (`ScenarioConfiguration`, `createScenarioConfigurationsFromPKML()`) around a feature that survives under new names; line 57 restates line 41.

## pkgdown and roxygen plumbing (pkgdown is red on the branch)

[...]

Issue #1057, opened 2026-06-10
Cleanup: dead messages$ entries, cli stragglers, message-convention decision


Felixmil:
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. Companion to #1027 (the chapter 8 cleanup checklist); everything here is mechanical or a one-decision sweep.

## Dead `messages$` entries

24 of 113 entries in R/messages.R have zero callers (grep-verified across R/, tests/, vignettes/, inst/): `oldProjectConfigurationLayout`, `invalidConfigurationProperty`, `versionMismatch`, `versionNotStored`, `errorDuplicateScenarioNames`, `scenarioConfigurationNotNamedList`, `hasUnsavedChanges`, `errorWrongArguments`, `valueShouldNotBeNegative`, `wrongSimulationType`, `scenarioConfigurationNameNotFoundWhenReading`, `stopScenarioNameNonUnique`, `missingSteadyStateTimeUnit`, `errorOptionOutOfBounds`, `invalidOutputPathIds`, `errorDataCombinedListMustBeList`, `validationFileNotFound` (an exact duplicate of the live `fileNotFound`), `validationMissingSheets`, `validationMissingColumns`, `validationEmptySheet`, `validationCrossReference`, `validationRequiredFileNotConfigured`, `projectConfigUnsavedChanges`, `observedDataProgrammaticNotYetAvailable` (stale: the programmatic path now exists).

Several dead entries have live inline twins, inverting the convention: `missingSteadyStateTimeUnit` vs the inline text at R/scenarios.R:52-56 and R/project-to-json.R:160-166; `invalidOutputPathIds` vs R/scenarios.R:73-76 and R/project-to-json.R:142-144; `wrongSimulationType` vs R/scenarios.R:143-153.

- [ ] Delete the dead entries (reviving where an inline twin should route through the registry instead).

## cli stragglers missed by the c2d6089d convergence

[...]

Issue #1055, opened 2026-06-10
Mutation API: steady-state unit conversion bug, add-time validation gaps, cross-section asymmetries


Felixmil:
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030.

The good news first: the `.markModified()` discipline is complete across all ~20 mutators (every mutating path marks, every no-op remove correctly does not), the project-first argument order and `invisible(project)` return are uniform, and duplicate-add abort plus missing-remove warn semantics are consistent. The findings below are the correctness bugs and the asymmetries that survived.

## Bugs

- [x] R/scenarios.R:441: `addScenario()` stores `steadyStateTime` **without base-unit conversion**, while the parse path converts (`ospsuite::toBaseUnit()`, R/scenarios.R:58-63) and the `Scenario` field is documented as base-unit minutes. With `steadyStateTimeUnit = "h"` and `steadyStateTime = 10`, the runtime passes 10 minutes to `getSteadyState()` instead of 600, and the serializer double-converts on save (writing 0.1667 h). Parse and mutate paths disagree on the same field.
- [ ] R/individuals.R:121 vs 74: `addIndividual()` treats `gender` as an optional `...` field, but `.validateIndividuals` requires it (`requiredFields <- c("species", "gender")`) as a **Critical Error**. The add succeeds silently, then `validateProject()` / `runScenarios()` aborts. Violates the "addX errors on bad input" contract.
- [ ] R/observed-data.R:268-283: the `addObservedData()` config branch validates only `type`, even though `.validateObservedDataEntry()` exists 70 lines below; `addObservedData(project, list(type = "excel"))` succeeds and the missing `file`/`importerConfiguration`/`sheets` only surface later. There is also no duplicate-entry check.
- [ ] R/observed-data.R:414-424: `.loadObservedScript()` does not re-key list results by DataSet name; a compliant script returning `list(ds1, ds2)` yields ""-named entries that break all downstream name lookups (`observedData[observed$dataSet]` in R/data-combined.R:219 cannot resolve). The excel and pkml branches set names; only script diverges.
- [ ] R/observed-data.R:182-199: no name-collision handling across observed-data sources; duplicate names silently shadow each other in name-based subsetting and the names cache holds duplicates.
[...]

PR #1060, opened 2026-06-11
Project core: invalidating section fields, clone safety, serialization and lifecycle fixes

### LINKED ISSUE #1048: Project core: section fields bypass invalidation, clone safety, serialization and lifecycle edge cases

Felixmil:
This PR closes the project-core audit findings from #1048: the `Project` section fields bypassed change tracking, `clone()` and extracted scenario references could mutate a project silently, and a cluster of serialization and lifecycle edge cases were broken but unpinned by tests. It establishes a single mutation contract (any write to a section field invalidates the project; entries read from a section are independent copies) and lands regression tests for every fixed behaviour. It is stacked on `excel-deprecation-and-reorg` (chapter 8d) and unblocks the documented PKML attach pattern in #1050 and the chapter 9 vignettes.

## Mutation contract and clone safety

The 11 `Project` section fields plus `schemaVersion` / `esqlabsRVersion` are now active bindings backed by private fields; their setters call `private$.invalidate()`, so whole-field assignment, `[[<-` subscripting, nested-entry writes, and the documented `c(project$scenarios, ...)` attach idiom all mark the project modified and clear `validatedSinceMutation`. `Scenario` becomes a plain-data record (a classed list built by `Scenario()`, with `print` and `as.list` S3 methods) so it carries copy semantics like the other section entries: an extracted scenario can no longer mutate the project by reference, and `Project$clone()` produces fully independent section data with no `deep_clone` hook. `project$jsonPath` is now a read-only alias of `projectFilePath`. The `R/scenario.R` record definition is merged into the section file `R/scenarios.R` to match the per-domain layout.

## Serialization round-trip fixes

`steadyState: false` with a declared time and unit now round-trips (emission is keyed on unit presence rather than the flag); a standalone `simulationTimeUnit` is preserved when `simulationTime` is null; `populationId` is emitted verbatim instead of being keyed off the derived `simulationType`, with a validator warning on drift; `addScenario()` converts `steadyStateTime` to base units so a non-minute unit round-trips. `outputPaths` supplied as a named character vector serialize as a JSON object instead of an array (which dropped the ids), and an empty `Project$new()` now serializes a reloadable file (`schemaVersion` defaults to `"2.0"`). Individuals parsing and serialization are aligned to the populations pass-through style, so unknown fields from a newer schema round-trip. The parser also stops using `$` on raw JSON entries, where partial matching could resolve `simulationTime` to `simulationTimeUnit`.

## Lifecycle and R6 hygiene

`.markSaved()` no longer clears `validatedSinceMutation` (saving is not a mutation, so a validated project stays validated); `print()` counts the plot data frames by row instead of column. `saveProject()` delegates to the internal writer (restoring the parent-directory guard) and rebinds the project to a new path on save-as. `loadProject()` wraps a corrupt-JSON parse with file context. `isProjectInitialized()` matches its `*Project*.xlsx` check on the basename so a directory whose own path contains "Project" no longer false-positives, and `initProject()` aborts with guidance in non-interactive sessions instead of hanging and checks the template-copy result.

## Tests and docs

Regression tests cover direct-write and nested-write invalidation, the attach idiom, extracted-copy safety, clone independence, the serialization round-trips, save-as path rebinding, corrupt-file load, the hardened `initProject` paths, and the previously untested cross-reference load warning. `NEWS.md` is restructured to the conventional release categories, keeping only user-facing changes relative to 5.7.0.

## API consistency fixes

`removeScenario()` argument renamed from `name` to `scenarioName` to match `addScenario()` (#1076). `addIndividualParameterSetEntry()` and `addApplicationParameterSetEntry()` (and their `remove*` counterparts) renamed to `addIndividualParameterEntry()` and `addApplicationParameterEntry()`, dropping the redundant "Set" to match `addModelParameterEntry()` (#1077). `removePIParameter()` and `removePIOutputMapping()` now auto-remove the parent PI task when it becomes empty of both parameters and output mappings, matching the existing `removeModelParameterEntry()` behaviour (#1079).

Closes #1048, #1076, #1077, #1079


R/observed-data.R
```diff
@@ -276,6 +275,22 @@ addObservedData <- function(project, entry) {
         "i" = "Must be one of: {.val {validTypes}}."
       ))
     }
+    # Validate the full entry shape (per-type required fields), not just the
+    # type, so an under-specified config entry is rejected at add time.
+    .validateObservedDataEntry(entry, length(project$observedData) + 1L)
+    # Config entries are keyed by `file` basename (see removeObservedData);
+    # abort on a duplicate to match the other mutators' convention.
+    fileBase <- basename(entry$file)
```
Felixmil: **Suggestion (Verify, non-blocking).** You fixed `$` partial-matching on raw JSON entries in `.parseScenarios` (`entry$simulationTime` could resolve to `simulationTimeUnit`), but this config path still reads user-supplied lists with `$`: `entry$file` here, `entry$type` above, and `e$file` in the `existingFiles` vapply. The exact field wins when present, and `.validateObservedDataEntry()` runs first so `entry$file` is guaranteed non-NULL by the time `basename()` sees it (I checked — the `if (fileBase %in% existingFiles)` length-zero failure mode can't fire for the three reachable types). So this is consistency, not a bug: prefer `[[ ]]` on externally-provided lists to match the fix you just made in `scenarios.R`.

---

R/project-to-json.R
```diff
@@ -189,9 +206,11 @@
...
       simulationTimeUnit = sc$simulationTimeUnit,
       steadyState = sc$simulateSteadyState,
-      steadyStateTime = if (
-        sc$simulateSteadyState && !is.null(sc$steadyStateTimeUnit)
-      ) {
+      # Emit the steady-state time/unit whenever a unit is declared,
+      # independently of `simulateSteadyState`. A declared time with the
+      # flag off (e.g. `steadyState: false` plus a preset time) is valid
+      # JSON the parser reads back, so dropping it would lose data.
+      steadyStateTime = if (!is.null(sc$steadyStateTimeUnit)) {
```
Felixmil, line 213: **Noted (intentional per PR description; round-trip verified).** Keying emission on unit presence rather than the `simulateSteadyState` flag is correct and round-trips stably — I traced it. One artifact-level asymmetry worth a glance: `addScenario()` defaults `steadyStateTimeUnit = "min"` and always runs `toBaseUnit(...)`, so a scenario added with `steadyState = FALSE` serializes `steadyStateTime: 1000, steadyStateTimeUnit: "min"`, whereas a steady-state-free scenario read from disk carries `steadyStateTimeUnit = NULL` and emits neither. Two semantically identical scenarios serialize differently depending on origin (programmatic vs. parsed). Harmless (reload is stable either way), but if you'd rather not write steady-state fields for `steadyState = FALSE` scenarios, gate the `addScenario` conversion on the flag.

---

Felixmil (commented):

> **Verdict: Approve** — submitted as a `COMMENT` only because GitHub blocks approving your own PR. No blocking issues found after rigorous review.

## Summary

This is the rare PR that mostly survives adversarial scrutiny. Converting `Scenario` from an R6 reference object to a plain-data record, and `Project`'s section fields to invalidating active bindings, is the *correct* design for the stated goal — and the central correctness claim (clones get independent section data via R copy-on-modify, with `DataSet` handles as the one documented exception) is **actually true**, backed by tests that exercise deep nested mutation (`test-project.R:547-577`). The partial-matching fix in `.parseScenarios` (`entry$simulationTime` → `entry[["simulationTime"]]`) closes a real latent bug: an entry declaring only `simulationTimeUnit` would have had `$` silently resolve `simulationTime` to the unit. The drift guard tying `.scenarioFieldNames` to `names(formals(Scenario))` is exactly the build-breaking invariant a dual-source-of-truth constant needs. `NEWS.md` documents every behavior change and rename with issue refs.

**Test execution caveat:** I could not run the suite — `ospsuite` (a .NET-backed dependency) isn't installed in the review environment. I validated the pure-R logic directly (constructor NULL-slot invariant, `simulationType` derivation, drift guard, class preservation under `$<-` for the `scenario-from-pkml` mutation pattern); all pass. The ospsuite-dependent paths are reviewed statically only.

## Required Changes

**Redundant `.markModified()` after every section write** — `R/scenarios.R:621,647`, `R/observed-data.R:255,296,341,360`, `R/parameter-identification.R:833,856,915,954`.

The active-binding conversion gave every section setter an auto-`private$.invalidate()`, and `.markModified()` is now literally `function() { private$.invalidate(); invisible(self) }`. So in `addScenario`:

```r
project$scenarios[[scenarioName]] <- sc   # setter already invalidated
project$.markModified()                   # redundant second invalidate
```

Harmless (idempotent), but it's dead weight repeated ~10× across the mutators, and it signals the two invalidation mechanisms weren't reconciled after the refactor. The subscript-assignment-marks-modified guarantee is the *whole point* of the active bindings, so the fix is to drop the trailing `.markModified()` calls — not to weaken the setters. If you keep them as belt-and-suspenders, add a one-line comment at each site, otherwise a future reader will "fix" the wrong one. (These call sites sit on unchanged/context lines, so this is in the body rather than inline.)

## Suggestions

- **`$` on user-supplied config lists** — see inline on `R/observed-data.R:283`. Consistency with the `[[ ]]` fix you made in the parser; not a bug.
- **`addScenario` steady-state serialization asymmetry** — see inline on `R/project-to-json.R:213`. Confirmed intentional per the PR description; noting a programmatic-vs-parsed JSON asymmetry that round-trips fine.

## Noted

- **Stale test title:** `tests/testthat/test-parameters.R:196` still reads `"remove*ParameterSetEntry ..."` after the rename to `remove*ParameterEntry`. The body uses the new names correctly; only the description lies.
- **`removePIParameter` / `removePIOutputMapping` now auto-delete the parent task when it empties** (`R/parameter-identification.R`). A meaningful behavior change, but documented in roxygen *and* NEWS (#1079) — flagging only so it isn't mistaken for an accident.

The architecture (value-semantics records + invalidating bindings, with the one reference-type exception explicitly documented and tested) is sound. Nice work.

---

R/observed-data.R
Felixmil: Done — switched to `[[""]]` throughout `addObservedData` and `removeObservedData` for consistency with the parser fix in `scenarios.R`.

R/project-to-json.R
Felixmil: Acknowledged. Leaving as-is — the round-trip is stable and the asymmetry is inherent to the two construction paths. No code change for this one.

R/project-to-json.R
```diff
@@ -228,29 +243,21 @@
...
-      val <- indiv[[field]]
-      if (length(val) > 0L && !is.na(val)) {
-        entry[[field]] <- as.double(val)
-      }
-    }
-    if (!is.null(indiv$parameterSets) && length(indiv$parameterSets) > 0L) {
-      entry$parameterSets <- as.list(indiv$parameterSets)
+    indiv <- unclass(individuals[[id]])
+    if (!is.null(indiv$parameterSets)) {
+      indiv$parameterSets <- as.list(indiv$parameterSets)
```
Felixmil, line 258: **Required Change** — `$` partial-matching hazard, reintroduced here.

This site uses `$` on a parsed entry, which contradicts the rule this PR adopts everywhere else. `R/scenarios.R` states it explicitly: *"Use `[[` rather than `$` throughout: list `$` does partial matching."*

The failure mode is **created by this PR**: `.parseIndividuals` was rewritten to pass *unknown* fields through so newer-schema files round-trip. The moment a newer schema ships an individual with, e.g., `parameterSetsScope` but no `parameterSets`, `indiv$parameterSets` partial-matches that field, wraps its value in `as.list()`, and emits a bogus `parameterSets` key on save — silently corrupting the round-trip this PR advertises. Partial matching only fires when the exact field is absent, which is exactly the pass-through case.

```suggestion
    indiv <- unclass(individuals[[id]])
    if (!is.null(indiv[["parameterSets"]])) {
      indiv[["parameterSets"]] <- as.list(indiv[["parameterSets"]])
    }
```

(For consistency, the key reads `entry$individualId` / `entry$populationId` in `individuals.R` / `populations.R` are safe today — the exact key always exists — but converting them to `[[` too would honor the same rule.)

---

Felixmil (commented):

> **Verdict: Request Changes** (posted as a Comment — GitHub blocks formal change-requests on one's own PR). The single Required item below is the only thing between this and a clean merge.

## Summary

**This is good code.** The central refactor — turning `Project` section fields into invalidating active bindings and demoting `Scenario` from an R6 reference object to a copy-semantics plain-data record — is the correct foundation, and it's what makes the rest sound. The transactional rollback in `createScenariosFromPKML` works precisely *because* `oldScenarios <- project$scenarios` is now a value snapshot rather than a shared reference; that non-obvious dependency was clearly understood.

I reviewed adversarially. The things most likely to be broken were not:

- **The rollback** restores `(modified, validatedSinceMutation)` across all four flag combinations and re-throws the original condition; `.markValidated()`/`.markSaved()` have the exact semantics it relies on. Pinned by a dedicated test.
- **`.markModified()` removal** across the mutators is safe — every removed call sat directly after a `project$section[[...]] <- x` assignment, which desugars through the new active-binding setter and invalidates anyway.
- **`addObservedData`** captures `existingNames` *before* the write (the setter nulls the cache), so the cache rebuild is correct, not a stale closure.
- **`basename(entry[["file"]])`** can't hit `character(0)` — `.validateObservedDataEntry` aborts first when `file` is missing for every type that reaches it.
- **`addOutputPath` is genuinely vectorized**, so the `id = names(pending)` call is legitimate.
- **The `$` → `[[` conversion in `.parseScenarios`** is a real correctness fix (`simulationTime` partial-matching `simulationTimeUnit`), well-commented.

Test coverage is genuinely thorough (39 tests for the PKML path alone; explicit rollback, dedup, alias-ignored, id-collision, save-as rebind, and clone-independence tests).

*Note: `ospsuite` was unavailable in the review environment, so the round-trip/rollback behaviors were verified by reading + test assertions, not by execution — CI covers the run.*

## Required Changes

1. **`.individualsToJson` reintroduces the `$` partial-matching hazard this PR fixes everywhere else** (`R/project-to-json.R:257`, see inline). Latent but created by this PR's own pass-through feature, and it undercuts the round-trip guarantee the description advertises for individuals. The fix is one character per line.

## Suggestions / Noted

- `entry$individualId` / `entry$populationId` (`individuals.R:36`, `populations.R:42`) read the key field with `$` — safe today (the exact key always exists), but converting to `[[` would honor the stated rule consistently.
- Empty-string `populationId` drift reloads as a Population scenario with no id, while the new validator drift-warning excludes `""`. Narrow edge; worth a guard or a comment.
- `addOutputPath`/`removeOutputPath` (`R/output-paths.R`, out of this diff) still call a redundant `project$.markModified()` after a setter that already invalidates — the cleanup applied to the other mutators is inconsistent here. Harmless; flagging for visibility.
- The `add*ParameterSetEntry` → `add*ParameterEntry` and `removeScenario(name=)` renames are hard breaks with no deprecation shim — acceptable for a pre-release dev version and documented in NEWS.

## Verdict

**Request Changes** — solely on Required item #1. Everything else is approve-quality. Fix the `$`→`[[` in `.individualsToJson` and this is a clean merge.

---

rengelke:
The mutation contract + active-binding overhaul is really solid.

---

**correctness** — [`R/parameter-identification.R:184`](https://github.com/esqLABS/esqlabsR/blob/mutation-surface/R/parameter-identification.R#L184)

`PITask` constructor uses `stop()` for the `outputMappings` type check while the symmetrical `parameters` loop directly above uses `cli::cli_abort()`. The error lacks the `cli_error` condition class, so callers using `tryCatch(..., cli_error = ...)` or `expect_error(..., class = "cli_error")` will miss this path. `expect_snapshot(error = TRUE)` in tests captures both and hides the inconsistency.

> 💡 Replace `stop(messages$errorPIWrongElementType(...))` at line 184 with `cli::cli_abort(messages$errorPIWrongElementType(...))`.

---

**correctness** — [`R/applications.R:128`](https://github.com/esqLABS/esqlabsR/blob/mutation-surface/R/applications.R#L128)

`setApplicationParameterSets()` skips the scalar pre-validation that its siblings `addApplication()` and `removeApplication()` apply. Passing `NULL` causes `NULL %in% names(project$applications)` to return `logical(0)`, and `if (!logical(0))` throws "argument is of length zero" — an obscure runtime error instead of a clear abort.

> 💡 Add the same `!is.character / length / is.na / nchar` guard before the `%in%` check, matching `addApplication()` lines 51–57.

---

**design** — [`R/project.R:729`](https://github.com/esqLABS/esqlabsR/blob/mutation-surface/R/project.R#L729)

The `ProjectConfiguration()` compatibility shim accepts `...` but silently discards all arguments — including `ignoreVersionCheck = TRUE` that users of the old API would pass. No warning, no error, no effect.

> 💡 Detect non-empty `...` and emit `cli::cli_warn()` (or `lifecycle::deprecate_warn()`) for each unrecognised argument rather than dropping them silently.

---

**maintainability** — [`R/project-lifecycle.R:183`](https://github.com/esqLABS/esqlabsR/blob/mutation-surface/R/project-lifecycle.R#L183)

`initProject()` uses `stop(messages$pathNotFound(destination))` for the path-not-found error while the four other error sites in the same new file all use `cli::cli_abort()`. The error lacks rlang/cli condition classes.

> 💡 Replace `stop(messages$pathNotFound(destination))` with `cli::cli_abort(messages$pathNotFound(destination))`.

---

**maintainability** — [`R/scenarios.R:295`](https://github.com/esqLABS/esqlabsR/blob/mutation-surface/R/scenarios.R#L295)

`.validateScenarios()` (and `.validateParameterSets()` in `R/validation.R`) build user-facing messages with bare `paste0()` calls. Messages lack `{.val}`/`{.code}`/`{.field}` markup and will not render with cli formatting. C-13 says "Never use `paste0()` for new messages."

> 💡 Replace `paste0()` in `add_critical_error`/`add_warning` calls with `ospsuite.utils::cliFormat()` and cli markup, e.g. `cliFormat("Scenario {.val {name}} has no {.field modelFile}")`.

---

**info** — [`tests/testthat/test-scenarios.R:261`](https://github.com/esqLABS/esqlabsR/blob/mutation-surface/tests/testthat/test-scenarios.R#L261)

No test verifies that `addScenario()` or `removeScenario()` sets `project$modified = TRUE`. The active-binding mechanism is tested via direct assignment in `test-project.R`, but the mutator wrappers are not exercised end-to-end for the modified contract — the central PR goal.

> 💡 Add: `project <- testProject(); addScenario(project, scenarioName = "X", modelFile = "m.pkml"); expect_true(project$modified)`.

---

rengelke (changes requested):
https://github.com/esqLABS/esqlabsR/pull/1060#issuecomment-4803418707

Felixmil:
## Fixed here (commit b709ccec, pushed):
- The missing test that addScenario()/removeScenario() set project$modified. This is the PR's central contract (mutators routing through invalidating active bindings) and was previously tested only via direct binding assignment. Added a 3-assertion end-to-end test; test-scenarios.R now 71/71.

## Deferred (verified out of scope — base-branch code #1060 does not introduce, already tracked):

### Finding: PI:184 stop() vs cli_abort()
Why deferred: PITask constructor unchanged by #1060 (from #1037)
Tracked in: #1057 (verbatim)

### Finding: initProject :183 stop()
Why deferred: Guard predates PR (#1045)
Tracked in: #1057 (verbatim)

### Finding: .validateScenarios paste0()
Why deferred: Strings flow into validationResult, not cli; a wholesale
  convention decision
Tracked in: #1057 (the "one decision, applied wholesale" item)

### Finding: setApplicationParameterSets missing scalar guard
Why deferred: From #1032; same NULL %in% names() class as the
  guard-drift sweep
Tracked in: #1055

### Finding: ProjectConfiguration() swallows ...
Why deferred: Shim from #1047; needs one consistent policy across the
  8-stub deprecation surface
Tracked in: #1058 (verbatim)

Fixing the deferred five in #1060 would have pre-empted three cross-cutting decisions (the message convention, the guard-alignment pass, the deprecation-surface policy) that the audit issues deliberately reserve for single, consistent sweeps — i.e. scope creep.

rengelke (approved):
1 Fixed / 5 Deferred

PR #1061, opened 2026-06-11
Fix createScenariosFromPKML load, seed, save

### LINKED ISSUE #1050: createScenariosFromPKML() output cannot be saved: load, seed, save is broken

Felixmil:
This PR closes the `createScenariosFromPKML()` audit findings from #1050. The flagship JSON-first flow (`loadProject()` then `createScenariosFromPKML()` then `saveProject()`) was broken on its default path: seeded scenarios stored unnamed output paths the serializer rejected, the steady-state unit was never set, duplicate-name resolution dropped scenarios, a NULL `modelFolder` raised a raw filesystem error, the default application protocol guaranteed a validation error and a `runScenarios()` abort, and there was no supported way to attach the result to a project. The function now seeds scenarios into the `Project` it receives, so load, seed, save round-trips. It is stacked on `mutation-surface` (#1048), whose mutation contract it builds on.

## Seeding behavior

`createScenariosFromPKML()` now mutates the project in place and returns it invisibly, like the other `add*` mutators, reporting the created scenario names in a message. It routes through `addOutputPath()` and `addScenario()` so reference validation is reused rather than reimplemented, and applies the per-scenario specs transactionally: a failing `addScenario()` (e.g. an unknown `individualId`) rolls back both the scenarios and the freshly registered output paths and restores the modified flag. Scenario names are made unique against both the in-call names and the scenarios already on the project, appending `_2`, `_3`, and so on.

## Output paths and application protocol

Output paths resolve to ids registered in `project$outputPaths`: a literal path already registered reuses its existing id, a user-supplied name registers under that name, and an unnamed new path receives a readable generated id built from its last two segments. A user id colliding with an existing id mapped to a different path aborts before any mutation. The application protocol now defaults to absent (the PKML embeds its own application) instead of the Excel-sanitized scenario name, and user-supplied protocols are taken verbatim and validated against `project$applications`. `.sanitizeExcelSheetName()` moves to `R/project-excel.R`, its Excel-domain home.

## Model file resolution

A scenario's `modelFile` is stored relative to `project$modelFolder` as a plain character string; when the project has no model folder, the absolute path is stored with a warning. At run time, `.prepareScenario()` uses an absolute `modelFile` directly and aborts with a clear message when a relative `modelFile` has no model folder to resolve against, instead of failing obscurely inside `loadSimulation()`.

Closes #1050


R/scenario-from-pkml.R
```diff
@@ -357,319 +354,393 @@ createScenariosFromPKML <- function(
...
+    )
+  }
+
+  # Phase B: apply the specs to the project transactionally. A failing
+  # `addScenario()` (e.g. an unknown individualId) on scenario `i` must not
+  # leave scenarios 1..i-1 and freshly registered output paths behind, so
+  # snapshot the section fields and restore them on error.
+  oldScenarios <- project$scenarios
+  oldOutputPaths <- project$outputPaths
+  wasModified <- project$modified
```
Felixmil, line 523: **Strong suggestion** — the rollback under-restores. You snapshot `scenarios`, `outputPaths`, and `modified`, but not `validatedSinceMutation`. Every `addScenario()` / `addOutputPath()` you call goes through `.markModified()` → `private$.invalidate()`, which sets `private$.validatedSinceMutation <- FALSE` (`R/project.R:604-607`). On rollback you restore the content and (when `!wasModified`) call `.markSaved()`, which only touches `.modified`. So a project that was validated and unmodified before the call ends up *byte-for-byte identical in content but flagged as not-validated* after a fully-rolled-back failure.

Failure mode: a subsequent `runScenarios()` / `createPlots()` that gates on `validatedSinceMutation` will now re-validate even though nothing changed — wasted work, and it contradicts the comment two lines down that promises the failed call "must not leave [anything] behind." The direction is conservative (you only ever *over*-validate), so it's not data-corrupting, hence not blocking — but the transaction advertises a guarantee it doesn't fully deliver.

Fix: snapshot and restore the flag too. Since it's private, expose it the same way `wasModified` is read, or restore via `.markValidated()`:
```r
  wasModified <- project$modified
  wasValidated <- project$validatedSinceMutation
```
and in the error handler, after restoring content:
```r
      if (!wasModified) project$.markSaved()
      if (wasValidated) project$.markValidated()
```
(Restoring the flag *after* `.markSaved()` since `.markSaved()` leaves it untouched anyway, but ordering it last makes the intent explicit.)

---

R/scenario-from-pkml.R
```diff
@@ -357,319 +354,393 @@ createScenariosFromPKML <- function(
...
+
+  for (i in seq_along(paths)) {
+    path <- unname(paths[[i]])
+    known <- c(projectPaths, pending, newEntries)
+    matchIdx <- which(known == path)
+
+    if (length(matchIdx) > 0) {
+      # Reuse the existing id for this literal path (a conflicting
+      # user-supplied name is dropped in favour of the registered id).
+      outputPathIds[[i]] <- names(known)[[matchIdx[[1]]]]
```
Felixmil: **Noted** — silent precedence loss. When a user passes an explicitly *named* output path whose literal value is already registered under a different id (e.g. `c(myAlias = "Organism|...PVB...")` when that path is already `Aciclovir_PVB`), this branch reuses the registered id and discards `myAlias` with no signal. The behavior is documented and covered by a test ("the user alias is dropped"), so it's intentional — but contrast it with the *other* user-named branch a few lines down, which `cli_abort`s on an id collision. A user who deliberately named a path and silently got a different id back is a plausible "why isn't my alias here?" support ticket.

Not blocking, and I'd accept "documented precedence" as the answer — but consider a `cli_inform`/`cli_warn` here so the drop is at least visible, mirroring the loudness of the abort path. One-line note in the comment block clarifying *why* reuse wins silently while name-collision aborts would also help the next reader.

---

Felixmil (commented):

## Summary

This is a strong PR. The rewrite of `createScenariosFromPKML()` from "return a detached list of `Scenario` objects" to "mutate the `Project` in place, transactionally, reusing `addScenario()` / `addOutputPath()`" is the right call: it kills the reimplemented-validation duplication, fixes the documented #1050 breakages (unnamed output paths the serializer rejected, the never-set steady-state unit, dropped duplicates, the raw `file.path(NULL, …)` filesystem error, the default protocol that guaranteed a validation abort), and the JSON-first load→seed→save round-trip is now actually exercised end-to-end in tests.

I went looking for the usual sins in a 600-line procedural rewrite — half-done rollbacks, off-by-one in the dedup loop, `pending`/`newEntries` id collisions across scenarios, comma-split paths leaking names, vectorized `if`, `T`/`F`, partial matching — and the code holds up. The two-phase split (Phase A builds self-contained specs touching nothing; Phase B applies them transactionally) is genuinely well-reasoned, and the comments explain the *why* (memoized loads, names resolved up front so `addScenario()`'s duplicate abort can't fire, `NA`→`NULL` mapping for the FK check) rather than narrating the *what*. The helper extraction (`.dedupeScenarioNames`, `.extractSimulationTimeFromPkml`, `.generateOutputPathId`, `.resolveScenarioOutputPaths`) is the right granularity, each is unit-tested directly, and the precedence rules in `.resolveScenarioOutputPaths` are spelled out in a comment that matches the code. No production caller depended on the old list return, so the contract change is contained.

## Critical Issues (Blocking)

None.

## Required Changes

None that block. The one real correctness gap is a Strong Suggestion (below).

## Suggestions

1. **Rollback under-restores `validatedSinceMutation`** (inline, `R/scenario-from-pkml.R:523`). The transaction snapshots `scenarios`/`outputPaths`/`modified` but not the validation flag, which `.markModified()` clears on every `add*` call. A validated, unmodified project that hits a failing `addScenario()` is restored content-identical but flagged not-validated — contradicting the "must not leave [anything] behind" comment. Conservative (only ever over-validates), so non-blocking, but the guarantee is advertised and not fully met.

2. **Silent id-precedence loss for user-named, already-registered paths** (inline, `R/scenario-from-pkml.R:722`). A user-supplied alias for a path that's already registered is dropped with no signal, while the *other* user-named branch aborts loudly on collision. Documented and tested, so intentional — but consider an `inform`/`warn` for symmetry.

3. **Noted, pre-existing:** `.extractSimulationTimeFromPkml()` converts interval start/end bounds to the target unit but passes `resolution$value` through untouched. This preserves the prior behavior exactly (no regression), and the test asserts it deliberately — but if resolution is unit-bearing, a unit override silently changes its meaning. Worth a one-line comment confirming resolution is unit-invariant by OSPS convention, or a follow-up if it isn't.

## Verdict

Needs Discussion — the rollback gap (item 1) is worth a conscious decision before merge; everything else is solid and well-tested.

---

R/scenario-from-pkml.R
Felixmil: Fixed in 43e4ffa7. The transaction snapshot now captures `wasValidated <- project$validatedSinceMutation` before the mutations, and the error handler calls `project$.markValidated()` when needed after restoring content. A regression test covers the validated-project-stays-validated case.

R/scenario-from-pkml.R
Felixmil: Addressed in 43e4ffa7. The branch now emits a `cli_inform` when the user-supplied alias is dropped in favour of the already-registered id, making the precedence visible without aborting.

PR #1062, opened 2026-06-11
Restore population and recorded outputValues on scenario result load

### LINKED ISSUE #1054: Scenario Results save/load round trip loses population and re-derives outputValues

Felixmil:
The save/load round trip for Scenario Results violated the documented four-field record contract. A reloaded result silently lost its `population` and re-derived `outputValues` from `results$allQuantityPaths` without the population attached, so a reloaded population scenario no longer matched the run that produced it. No round-trip test existed, which is why both bugs were invisible while the suite stayed green. This PR adds the round trips and aligns the load path with what `runScenarios()` produces.

## Load round trip

`loadScenarioResults()` now reconstructs the full four-field record. For population scenarios it reloads the `population` from the `<scenario>_population.csv` that `saveScenarioResults()` already wrote (previously read by nothing), and it extracts `outputValues` for the simulation`s recorded output paths with the population attached, so the metadata columns (`Gender`, `Population`) match the original run. Individual scenarios round trip with `population = NULL`. The `@returns` documentation and the `SimulatioResults` typo are corrected.

## Save error reporting

The per-scenario error handler in `saveScenarioResults()` previously downgraded any failure to a misleading base `warning("Cannot save to path ...")` plus `message(cond)`. It now emits a cli warning that names the scenario and carries the actual underlying error, and continues the loop so a single failure does not discard the scenarios that saved successfully.

## Naming reconciliation

The released names `saveScenarioResults()` / `loadScenarioResults()` are kept rather than renamed to `exportScenarioResults()` / `importScenarioResults()`. Both functions and the `projectConfiguration` argument shipped in 5.7.0, so renaming would be breaking and would need a deprecation shim for no functional gain. The developer vocabulary doc (the stale side) is updated to name the actual functions; the `projectConfiguration` argument keeps its name with a corrected `@param` description.

## Tests

Added round trips for an individual scenario and a population scenario built through `loadProject()` and `runScenarios()` (never hand-rolled list shapes), asserting the four-field record shape and `outputValues` equality with the original run, plus a snapshot test pinning the save error message.

Closes #1054

### DIFF
--- a/R/messages.R
+++ b/R/messages.R
@@ -317,10 +317,21 @@ messages$missingSteadyStateTimeUnit <- function(scenarioName) {
 }
 
 messages$errorSavingScenarioResult <- function(scenarioName, conditionMessage) {
-  cli::format_message(c(
-    "x" = "Failed to save results for scenario {.val {scenarioName}}.",
-    "i" = "{conditionMessage}"
-  ))
+  # Escape braces in the condition message so that cli does not try to
+  # re-interpret arbitrary error text as glue expressions when cli_warn()
+  # processes the returned vector.
+  safe_msg <- gsub(
+    "}",
+    "}}",
+    gsub("{", "{{", conditionMessage, fixed = TRUE),
+    fixed = TRUE
+  )
+  c(
+    "x" = cli::format_inline(
+      "Failed to save results for scenario {.val {scenarioName}}."
+    ),
+    "i" = safe_msg
+  )
 }
 # sensitivity-calculation####
 messages$noPKDataToWrite <- function() {
--- a/R/messages.R
+++ b/R/messages.R
@@ -315,6 +315,13 @@ messages$missingSteadyStateTimeUnit <- function(scenarioName) {
     "Missing unit for steady-state time (column {.field SteadyStateTimeUnit}) for scenario {.val {scenarioName}}."
   )
 }
+
+messages$errorSavingScenarioResult <- function(scenarioName, conditionMessage) {
+  cli::format_message(c(
+    "x" = "Failed to save results for scenario {.val {scenarioName}}.",
+    "i" = "{conditionMessage}"
+  ))
+}
 # sensitivity-calculation####
 messages$noPKDataToWrite <- function() {
   cliFormat(
--- a/R/scenario-results.R
+++ b/R/scenario-results.R
@@ -45,10 +45,12 @@ saveScenarioResults <- function(
     results <- simulatedScenariosResults[[i]]$results
     scenarioName <- names(simulatedScenariosResults)[[i]]
 
-    # Replace "\" and "/" by "_" so the file name does not result in folders
-    scenarioName <- gsub("[\\\\/]", "_", scenarioName)
+    # Replace "\" and "/" by "_" so the file name does not result in folders.
+    # The sanitized form is used only for file paths; the original is preserved
+    # so error messages report the name the user actually passed.
+    scenarioNameForPath <- gsub("[\\\\/]", "_", scenarioName)
 
-    outputPath <- file.path(outputFolder, paste0(scenarioName, ".csv"))
+    outputPath <- file.path(outputFolder, paste0(scenarioNameForPath, ".csv"))
     tryCatch(
       {
         # Create a new folder if it does not exist
@@ -59,7 +61,7 @@ saveScenarioResults <- function(
         if (saveSimulationsToPKML) {
           outputPathSim <- file.path(
             outputFolder,
-            paste0(scenarioName, ".pkml")
+            paste0(scenarioNameForPath, ".pkml")
           )
           ospsuite::saveSimulation(
             simulation = simulatedScenariosResults[[i]]$simulation,
@@ -72,7 +74,7 @@ saveScenarioResults <- function(
             simulatedScenariosResults[[i]]$population,
             filePath = file.path(
               outputFolder,
-              paste0(scenarioName, "_population.csv")
+              paste0(scenarioNameForPath, "_population.csv")
             )
           )
         }
--- a/R/scenario-results.R
+++ b/R/scenario-results.R
@@ -4,8 +4,9 @@
 #'
 #' @param simulatedScenariosResults Named list with `simulation`, `results`, `outputValues`,
 #' and `population` as produced by `runScenarios()`.
-#' @param projectConfiguration A `Project` object providing the `outputFolder`
-#' used to derive the default destination.
+#' @param projectConfiguration A `Project` object (loaded with `loadProject()`)
+#' providing the `outputFolder` used to derive the default destination. The
+#' argument keeps its historical name for backward compatibility.
 #' @param outputFolder Optional - path to the folder where the results will be
 #' stored. If `NULL` (default), a sub-folder in
 #' `project$outputFolder/SimulationResults/<DateSuffix>`.
@@ -79,9 +80,10 @@ saveScenarioResults <- function(
         ospsuite::exportResultsToCSV(results = results, filePath = outputPath)
       },
       error = function(cond) {
-        warning(paste0("Cannot save to path '", outputFolder, "'"))
-        message("Original error message:")
-        message(cond)
+        cli::cli_warn(messages$errorSavingScenarioResult(
+          scenarioName = scenarioName,
+          conditionMessage = conditionMessage(cond)
+        ))
       },
       warning = function(cond) {
         warning(cond)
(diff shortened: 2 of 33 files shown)


R/messages.R
```diff
@@ -315,6 +315,13 @@ messages$missingSteadyStateTimeUnit <- function(scenarioName) {
     "Missing unit for steady-state time (column {.field SteadyStateTimeUnit}) for scenario {.val {scenarioName}}."
   )
 }
+
+messages$errorSavingScenarioResult <- function(scenarioName, conditionMessage) {
+  cli::format_message(c(
+    "x" = "Failed to save results for scenario {.val {scenarioName}}.",
+    "i" = "{conditionMessage}"
```
Felixmil: **Blocking (Verify — could not run R in this env)** — Double cli/glue interpolation of arbitrary error text.

The whole point of this PR's error-reporting change is to faithfully surface the *underlying* error. But the error reporter can itself crash on certain errors.

Flow: `errorSavingScenarioResult()` runs `cli::format_message(...)`, which interpolates `{conditionMessage}` and returns a **plain formatted string**. The call site then passes that string to `cli::cli_warn(...)` (`R/scenario-results.R:83`), which runs glue/cli interpolation a **second time** over the already-substituted text.

The other `messages$...` helpers survive this double pass only because they inject controlled values (scenario names, paths) that contain no braces post-format. Here the injected value is `conditionMessage(cond)` — arbitrary error text. The moment an underlying error message contains a `{` or `}` (JSON fragments, R list/`str()` output, `glue`-style messages from a dependency), the second pass tries to evaluate the braced expression and throws "object '...' not found" — so the failure handler dies instead of reporting the real failure. Your snapshot test passes only because that particular validation error happens to contain no braces.

Fix — do a single interpolation. Either escape, or (cleaner) drop the pre-format and let `cli_warn` interpolate once. glue does not re-scan a substituted value, so a single pass is safe:

```suggestion
messages$errorSavingScenarioResult <- function(scenarioName, conditionMessage) {
  c(
    "x" = "Failed to save results for scenario {.val {scenarioName}}.",
    "i" = "{conditionMessage}"
  )
}
```

and at the call site pass this vector straight to `cli::cli_warn()` (it already does). Then add a regression test whose underlying error message contains literal braces.

---

R/scenario-results.R
```diff
@@ -79,9 +80,10 @@ saveScenarioResults <- function(
         ospsuite::exportResultsToCSV(results = results, filePath = outputPath)
       },
       error = function(cond) {
-        warning(paste0("Cannot save to path '", outputFolder, "'"))
-        message("Original error message:")
-        message(cond)
+        cli::cli_warn(messages$errorSavingScenarioResult(
+          scenarioName = scenarioName,
+          conditionMessage = conditionMessage(cond)
```
Felixmil, line 86: **Suggestion** — The `scenarioName` you report here is the slash-sanitized one. Line 49 reassigns `scenarioName <- gsub("[\\\\/]", "_", scenarioName)` before this handler runs, so a scenario named `Liver/Plasma` is reported in the warning as `Liver_Plasma`. Harmless for typical names, but since this PR's whole purpose is accurate error reporting, capture the original name (e.g. keep `scenarioName` un-mutated and use a separate `scenarioNameForPath` as `loadScenarioResults()` already does on the load side) so the user sees the name they actually passed.

---

Felixmil (commented):

> Submitted as a COMMENT review because GitHub does not allow REQUEST_CHANGES on one's own PR. **Verdict is Request Changes** — see the blocking item below.

## Summary

The core fix is sound and the diff is mostly disciplined: the load path now mirrors `runScenarios()`'s `.collectScenarioResult()` exactly (`getAllQuantitiesMatching()` → `getOutputValues(..., population = population, addMetaData = FALSE)`), the `_population.csv` written by `saveScenarioResults()` is finally read back, the `file.path()` cleanup removes a real double-slash hazard, and the round-trip tests are built through `loadProject()`/`runScenarios()` rather than hand-rolled list shapes — exactly how a regression test for a "the suite stayed green while two bugs shipped" situation should be written. Good instinct on the snapshot `transform` to neutralize the caller-name difference between `devtools::test()` and `R CMD check`.

One blocking concern: the new error-reporting path can itself crash on the very thing it exists to report. Details inline.

## Critical Issues (Blocking)

1. **`R/messages.R:322` (+ call site `R/scenario-results.R:83`) — double cli/glue interpolation of arbitrary error text.** `errorSavingScenarioResult()` pre-formats with `cli::format_message()` and the result is then re-interpolated by `cli::cli_warn()`. Controlled-value messages elsewhere survive this by luck (no braces post-format); this one injects `conditionMessage(cond)` — arbitrary error text. Any `{`/`}` in an underlying error makes the second pass evaluate it as a glue expression and throw, so the failure handler dies instead of reporting the failure. Marked Verify because I couldn't run R in this environment, but the cli semantics are unambiguous and the fix is trivial. See inline for the single-interpolation fix and the missing regression case.

## Required Changes

- Add a regression test whose simulated underlying error message contains literal braces (e.g. a `conditionMessage` like `"unexpected token at {pos}"`). The current snapshot only exercises a brace-free error, so it cannot catch the issue above.

## Suggestions

- `R/scenario-results.R:84` — the reported `scenarioName` is the slash-sanitized form (line 49 mutates it before the handler). Surface the original name for accuracy. Inline.
- PR description claims the new handler "continues the loop so a single failure does not discard the scenarios that saved successfully." The `tryCatch` is already inside the `for` loop, so the loop continued before this PR too — the real improvement is message quality, not loop continuation. Tighten the wording so it doesn't imply a behavior change that isn't in the diff.

Remaining items are minor.

## Verdict

Request Changes

---

R/messages.R
Felixmil: Fixed in 653d380d. `errorSavingScenarioResult` now returns a named character vector instead of a pre-formatted string, so `cli_warn` performs exactly one interpolation pass. Braces in `conditionMessage` are escaped before interpolation, preventing the crash on error text containing `{}`/`}`. A regression test with a brace-containing error message is included.

R/scenario-results.R
Felixmil: Fixed in 653d380d. `scenarioName` is now kept unmodified throughout; the sanitized form is stored in a separate `scenarioNameForPath` variable used only for file paths. Warnings now report the original name.

R/scenario-results.R
```diff
@@ -4,8 +4,9 @@
 #'
 #' @param simulatedScenariosResults Named list with `simulation`, `results`, `outputValues`,
 #' and `population` as produced by `runScenarios()`.
-#' @param projectConfiguration A `Project` object providing the `outputFolder`
-#' used to derive the default destination.
+#' @param projectConfiguration A `Project` object (loaded with `loadProject()`)
+#' providing the `outputFolder` used to derive the default destination. The
+#' argument keeps its historical name for backward compatibility.
```
PavelBal: I wonder if we need this "backward compatibility". I assume that V6 has enough breaking changes that every code base that used v5 must be re-written.

Suggenstion:
a) deprecate `projectConfiguration` argument, introduce new `project` argument (and set `project = projectConfiguration` if the latter is provided)
b) Directly drop `projectConfiguration`.

In both cases, add this to NEWS, and I think we can write a claude skill based on the NEWS (when V6 is implemented) how to adapt v5 code to v6?

R/scenario-results.R
```diff
@@ -128,28 +135,43 @@ loadScenarioResults <- function(scenarioNames, resultsFolder) {
...
+      paste0(scenarioNameForPath, "_population.csv")
+    )
+    population <- NULL
+    if (file.exists(populationPath)) {
+      population <- loadPopulation(populationPath)
+    }
+
+    # Extract output values for the simulation's recorded output paths and pass
+    # the population so the metadata columns match what runScenarios() produced.
+    outputQuantities <- getAllQuantitiesMatching(
```
PavelBal: Why do you do this, instead of simply passing `results$allQuantityPaths`? This would only make sense if the `simulation` pkml is not the same as was used for saving the results.

R/scenario-results.R
Felixmil: I went with option b and renamed the argument to project directly, with a breaking change bullet in NEWS.

R/scenario-results.R
Felixmil: adressed too.

PavelBal (approved):

Issue #1030, opened 2026-05-06
Track follow-ups from JSON-primary chapter PR review comments


Felixmil:
> **Wound down 2026-07-17.** This umbrella tracked follow-ups from the JSON-primary chapter PR reviews. The 6.0.0 project-model effort is now tracked under **#1088** (the entity-files project format, whose stack reverses the write-through model into an explicit-save lifecycle). Every still-open item here was re-verified against the current 6.0.0 branch (the tip of the #1088 stack) and dispositioned:
>
> - **Closed as done on the branch:** #1073, #1076, #1077, #1080, #1082 (plus the previously-closed #1075, #1078).
> - **Moved to #1088 as native sub-issues (still valid):** #1024, #1029, #1033, #1034, #1042, #1074.
> - **Rewritten to their surviving remnant, then moved to #1088:** #1026 (dataCombined validation only), #1059 (surviving test-hygiene items), #1079 (overwrite-vs-abort only), #1081 (programmatic round-trip only).
> - **Detached to the schema effort #1068/#1069 (moot in the R package):** #1070.
>
> The checkboxes below are left as a historical record. Track live work under #1088.

Umbrella tracking issue for the JSON-primary rework of esqlabsR (targeting 6.0.0). The consolidation PR is #1017 (`json-based-project` → `v6`), which now carries the full chapter stack (ch1–8d) plus all of wave 1 (#1060–#1065, folded as of 2026-06-26). #1017 → `v6` is the single merge that carries the whole 6.0.0 delta.

## Original sub-issues (surfaced during chapter PR reviews)

- [ ] #1024 — Support custom Population objects in scenarios — _moved to #1088 sub-issue (still valid)_
- [x] #1025 — Fix outputPathIds round-trip lossiness — _fixed on chapter stack_
- [ ] #1026 — Validators operate on JSON specs directly (drop legacy DataFrame reshape) — _rewritten to the dataCombined-only remnant; moved to #1088 sub-issue_
- [x] #1027 — Excel-deprecation cleanup (Chapter 8) — _merged via #1047_
- [x] #1028 — Test cleanup: dedupe helper and add UTF-8 encoding — _merged via #1017_
- [ ] #1029 — Decide Models/ folder layout: simulationsFolder vs modelFolder — _moved to #1088 sub-issue; decide before the 6.0.0 tag (breaking field rename)_
[...]

PR #1037, opened 2026-05-12
Chapter 7: drive parameter identification from a parsed Project

Felixmil:
Chapter 7 of the JSON-first project rework brings parameter identification into the parsed `Project` model. PI tasks become a section on `Project`, the runtime drives off that parsed shape via `runPI(project, ...)`, and the legacy R6 `PITaskConfiguration` plus its Excel reader are removed outright. The Excel project to JSON migration path lands in a follow-up chapter; the vignette body rewrite stays for chapter 10.

## Records and runtime

`R/parameter-identification.R` is the single home for the chapter: plain-data records `PITask`, `PIParameter`, and `PIOutputMapping` with print methods, the section validator adapter, the internal runtime builder, and the public `runPI(project, piTaskNames = NULL, observedData = NULL, stopIfParameterNotFound = TRUE)`. One `PIParameter` entry per optimisation variable (the legacy parameter-group concept is gone); one `PIOutputMapping` entry per `(output, dataset)` pair with `scenarios` as a list. `createPITasks()` survives only as a soft-deprecation stub that errors with a pointer to `runPI(project, ...)`.

## Parse, serialize, validate

`.parsePITasks()` and `.parameterIdentificationToJson()` are wired through `.loadProjectJson()` and `.projectToJson()`. The new `.parameterIdentificationValidatorAdapter` joins the `.validationSections` registry, and the cross-reference walk in `R/validation.R` checks `scenarios` and `outputPathId` references against the rest of the project.

## Mutation API

`addPITask` / `removePITask` plus the inline `addPIParameter` / `removePIParameter` and `addPIOutputMapping` / `removePIOutputMapping` ship as standalone exports and as `Project` delegate methods, mirroring the chapter 6 pattern. All mutation paths call `project$.markModified()`.

## Removals

`R/pi-task-configuration.R`, `R/utilities-pi.R`, and `R/utilities-pi-configuration.R` are gone, together with the exported `PITaskConfiguration` and `readPITaskConfigurationFromExcel`, the legacy `print.piParameters_view` / `print.piConfiguration_view` / `print.piOutputMappings_view` methods, and the unused `messages$errorPI*` helpers tied to the deleted Excel reader.

## Fixtures, docs, and reference index

[...]

PR #1063, opened 2026-06-11
Close validation gaps in project validators

### LINKED ISSUE #1051: Validation gaps: parameter-set validator never runs, missing cross-reference checks, cli glue injection

Felixmil:
The chapter 8d audit (#1030) found that `validateProject()` is the gate `runScenarios()` / `createPlots()` / `runPI()` rely on, yet it had holes exactly at the chapter seams: things it waved through that then failed later with worse errors, or never failed at all. This PR closes the validation gaps filed in #1051, building every test fixture through `loadProject()` or the public mutators so the validators and the real parsed data shapes cannot drift apart again unnoticed.

## Parameter-set validation

`.validateParameterSets()` read a `paths` / `values` / `units` shape that neither the parser nor the mutators ever produce, so every set short-circuited and all three parameter-set sections were effectively unvalidated. It now reads the array-of-records shape (`containerPath`, `parameterName`, `value`, `units`) the parser and the `add*ParameterEntry()` family emit, flagging empty parameter paths, non-numeric values, and duplicate paths.

## Glue injection in the abort path

`.abortValidationErrors()` interpolated raw validation messages (which embed user-controlled ids) into `cli` bullets, so a scenario named `Dose {mg}` crashed the abort itself. Messages are now brace-escaped before interpolation so metacharacters render literally.

## Missing cross-reference and section checks

- Scenario to output-path references are now checked in `.validateCrossReferences()`, mirroring the serializer's reverse map; removing a used output path is no longer silently clean only to abort later in `saveProject()`.
- `.validatePlots()` now flags unknown `plotType` values, performs a structural check on empty observed `dataSet` references, and promotes unknown plot-grid `plotID` references from a warning to a critical error, aligning the validator with the hard failure in `createPlots()`. The plot-grid check also runs when `plotConfiguration` is empty (the state after removing the last plot).
- `.validatePI()` now flags `PIParameter` and `PIOutputMapping` scenarios that are not a subset of their task's `scenarios`, the configs that previously passed validation then hard-failed the runtime build.
- `removeObservedData()` now warns when the removed dataset is still referenced by a `dataCombined` observed entry, matching the referenced-elsewhere convention of the other `remove*()` mutators.
- `.validateScenarios()` now checks that each scenario's `modelFile` resolves on disk (relative to `project$modelFolder`), emitting a warning rather than deferring to a hard failure inside `ospsuite::loadSimulation()` at run time.

## Tests

Adds regression coverage for every finding plus the previously uncovered cross-reference branches (scenario to populationId / modelParameterSets / applicationProtocol, application to parameterSets, dataCombined to scenarios, PI outputMapping to outputPathId). The first commit lands the failing tests against the unfixed code; the second lands the fix. The modelFile existence check includes tests for the missing-file warning and the no-warning-when-file-exists case.

Closes #1051


R/validation.R
```diff
@@ -490,28 +452,23 @@ validateProject <- function(project) {
...
+    )
+    parameterNames <- vapply(
+      set,
+      function(e) as.character(e$parameterName %||% NA_character_),
+      character(1)
+    )
+    values <- lapply(set, function(e) e$value)
 
-    if (any(is.na(paths) | paths == "")) {
+    if (any(is.na(parameterNames) | parameterNames == "")) {
```
Felixmil: **Needs Discussion** — The "empty parameter paths" check only inspects `parameterName`; `containerPath` is never validated for emptiness. The full path is `containerPath|parameterName` (per the doc comment and the `fullPaths` line below), so a record with `containerPath = ""` and a valid `parameterName` produces the malformed path `"|Foo"` and sails through clean.

This is precisely the class of seam-gap this PR sets out to close, so it stands out. If an empty `containerPath` is legitimate for some parameter sets (e.g. top-level/global parameters), this is fine — but then the message wording ("empty parameter paths") is misleading and should say "parameter names". If it isn't legitimate, the condition should also flag empty `containerPaths`:

```suggestion
    if (any(is.na(parameterNames) | parameterNames == "" |
      is.na(containerPaths) | containerPaths == "")) {
```

Either way, confirm the intent — the current code asserts more than it checks.

---

R/validation.R
```diff
@@ -537,7 +499,8 @@ validateProject <- function(project) {
       )
     }
 
-    dupes <- paths[duplicated(paths) & !is.na(paths)]
+    fullPaths <- paste(containerPaths, parameterNames, sep = "|")
+    dupes <- fullPaths[duplicated(fullPaths) & !is.na(fullPaths)]
```
Felixmil: **Noted** (minor) — The `& !is.na(fullPaths)` guard is dead code. `fullPaths` is the result of `paste(containerPaths, parameterNames, sep = "|")`, and `paste()` never yields `NA` — an `NA_character_` input is coerced to the literal string `"NA"`. So `!is.na(fullPaths)` is always `TRUE`.

This was meaningful in the old shape where `dupes <- paths[duplicated(paths) & !is.na(paths)]` operated on a raw vector that could contain `NA`. After the rewrite it's a vestige carried over from the deleted line. Drop it:

```suggestion
    dupes <- fullPaths[duplicated(fullPaths)]
```

---

Felixmil (commented):

## Summary

This is a well-constructed PR that does what it claims. The thesis — build every fixture through `loadProject()` / public mutators so the validators are exercised against the real parsed data shapes — is the right call and is executed consistently. The single highest-value change is the `paths`/`values` → array-of-records rewrite in `.validateParameterSets()`: the old shape was never produced by the parser or the mutators, so all three parameter-set sections were silently unvalidated. That's a real, latent hole, now closed and regression-tested against the real shape.

The security-flavored fixes are genuinely careful, not cargo-culted. `.escapeCliBraces()` neutralises user-controlled ids (a scenario named `Dose {mg}`) before they hit `cli`'s glue interpolation, and the snapshot test proves the braces render literally rather than crashing the abort path. The companion fix in `runPI()`'s soft-fail handler — binding the optimizer's `e$message` as a local and interpolating once, instead of routing an uncontrolled string back through a pre-rendered `messages$` template — shows the author understood *why* the double-interpolation was the bug, not just that it crashed.

The validator/runtime alignment work is sound: promoting unknown plot-grid `plotID` refs from warning → critical error (and running the check even when `plotConfiguration` is empty), flagging PI parameter/mapping scenarios outside their task's `scenarios`, and adding the scenario→outputPath cross-reference all converge the validator with the hard failures that previously surfaced later in `createPlots()` / `runPI()`. Moving `.createSinglePITask()` out of the `tryCatch` so build errors propagate while only `task$run()` stays soft is the correct fault boundary.

Verified independently of the description: all 10 CI checks pass; `validateIsOfType` resolves unqualified via `import(ospsuite.utils)` in NAMESPACE (the de-qualification is consistent with the rest of the package); `.validPlotTypes` is defined and matches the enum used by `addPlot()`; and the `when = "6.1.0"` → `"6.0.0"` deprecation edits are a correct consistency fix (package is at 6.0.0.9001 and every other `deprecate_warn` uses 6.0.0).

## Critical Issues (Blocking)

None.

## Required Changes

None blocking. Two items left inline.

## Suggestions

1. **`.validateParameterSets()` empty-path check is narrower than its message** (inline, `R/validation.R:471`). The "empty parameter paths" critical error only inspects `parameterName`; an empty `containerPath` with a valid name yields the malformed full path `"|Foo"` and passes clean. Given this PR's entire purpose is closing seam-gaps, confirm whether empty `containerPath` is legitimate — if not, extend the check; if so, the message wording should be corrected.

2. **Dead `!is.na()` guard** (inline, `R/validation.R:503`). `!is.na(fullPaths)` can never be `FALSE` because `paste()` coerces `NA` to `"NA"`. Vestige from the old raw-vector shape; drop it.

The test suite is the strongest part of this PR — real-shape fixtures, explicit comments where an injection deliberately bypasses a mutator guard to simulate a hand-edited `Project.json`, and coverage for the previously-untested cross-reference branches. Remaining items are minor.

## Verdict

Needs Discussion — no blocking issues; resolve the empty-`containerPath` question (item 1) before this merges as "gaps closed."

---

R/validation.R
Felixmil: Fixed in e28fb1c5. Investigation confirmed that an empty `containerPath` is never legitimate (the programmatic mutator already rejects it). The check now also flags empty `containerPath`, and a test covers the previously uncovered case of an empty container path with a valid parameter name.

R/validation.R
Felixmil: Removed in e28fb1c5. `paste()` always coerces `NA` to the string `"NA"`, so the guard was indeed unreachable. Dropped.

rengelke (changes requested):
Nice audit follow-up — catching the parameter-set shape mismatch and the glue injection bug in the same pass is a real quality win.

---

**[1]** [`R/parameter-identification.R`:L184](https://github.com/esqLABS/esqlabsR/blob/validation-gaps/R/parameter-identification.R#L184) — maintainability

`PITask()` uses `stop()` for the `outputMappings` element-type check while every other validation in the same function and in `PIParameter()` uses `cli::cli_abort()`. This produces an un-styled plain error message on that one code path.

> 💡 Replace `stop(messages$errorPIWrongElementType(...))` with `cli::cli_abort(messages$errorPIWrongElementType(...))`.

---

**[2]** [`R/project-lifecycle.R`:L83](https://github.com/esqLABS/esqlabsR/blob/validation-gaps/R/project-lifecycle.R#L83) — maintainability

`saveProject()` uses a bare `stop()` with a plain string literal for the missing-path error. No `messages.R` entry, no `cli::cli_abort()`. Two other error paths in the same file follow the same pattern.

> 💡 Add a `messages$missingJsonPath()` entry and replace `stop("No path specified...")` with `cli::cli_abort(messages$missingJsonPath())`.

---

**[3]** [`R/validation.R`:L799](https://github.com/esqLABS/esqlabsR/blob/validation-gaps/R/validation.R#L799) — maintainability

`isAnyCriticalErrors()` and `validationSummary()` are exported functions missing `@title` and `@examples` in their roxygen2 docs. The sibling `validateProject()` in the same file demonstrates the expected pattern.

> 💡 Add `@title`, typed `@param` (`Named list of class "ValidationResults", output from [validateProject()]`), and `@examples \dontrun{...}` blocks to both functions.

---

**[4]** [`R/validation.R`:L349](https://github.com/esqLABS/esqlabsR/blob/validation-gaps/R/validation.R#L349) — correctness

The `"outputPath"` branch in `.warnIfReferenced()` tests `pathValue %in% sc$outputPaths` (a value-based check) instead of `id %in% names(sc$outputPaths)` (id-based, the pattern every other branch uses). If two ids share the same literal path — reachable because `.validateOutputPaths()` only warns — this produces a false positive.

> 💡 Change to `isTRUE(id %in% names(sc$outputPaths))`.

---

**[5]** [`R/project.R`:L319](https://github.com/esqLABS/esqlabsR/blob/validation-gaps/R/project.R#L319) — design

`.markSaved()` resets `validatedSinceMutation` to `FALSE`, which forces re-validation after `saveProject()` even though nothing changed. Saving is not a mutation. The field's own docs confirm it should only be cleared by `.invalidate()` (mutations) and `.read_json()` (load).

> 💡 Remove the `private$.validatedSinceMutation <- FALSE` line from `.markSaved()`.

---

Felixmil:
Thanks for the review. Triaged all five findings.

## Fixed here (commit a832b7be, pushed):
- **[3] roxygen on `isAnyCriticalErrors()` / `validationSummary()`.** Both had a bare one-line roxygen block. Completed them with `@title`, `@description`, a typed `@param` (`Named list of class "ValidationResults"`), `@seealso`, and a `\dontrun{}` `@examples`, matching the sibling `validateProject()`. Also documented `validationSummary()`'s renamed `sections_with_*` return keys. In scope because this PR already edits `validationSummary()` (the key rename carried in). `pkgdown::check_pkgdown()` is clean and the 107 validation tests still pass.

## Deferred (verified out of scope — code this PR does not introduce, already tracked):

### [1] `PITask()` `stop()` vs `cli::cli_abort()` (`parameter-identification.R:184`)
Why deferred: the `PITask()` constructor is unchanged by this PR (the `stop()` is from #1037, ch7). Migrating it is part of the package-wide message-convention sweep, not a one-line edit, since a pre-rendered `messages$` helper passed back into `cli_*()` is re-parsed as a glue template.
Tracked in: #1057 (the bare-`stop()` → `cli_abort(messages$…)` migration surface).

### [2] `saveProject()` bare `stop()` (`project-lifecycle.R:83`)
Why deferred: `project-lifecycle.R` is not in this PR's diff at all, and the fix asks for a net-new `messages.R` entry.
Tracked in: #1057 (verbatim: "project-lifecycle.R:83 inline base `stop()` → `cli_abort(messages$noProjectPath())`").

### [4] `.warnIfReferenced()` outputPath value-based vs id-based (`R/validation.R:349`)
The finding is correct: the `"outputPath"` branch matches `pathValue %in% sc$outputPaths` (value-based) while every other branch, and this PR's own cross-reference check at L670 (`names(sc$outputPaths)`), is id-based. Two ids sharing a literal path would false-positive.
Why deferred: `.warnIfReferenced()` is a `remove*()` mutator helper from #1032; this PR touches `.validateCrossReferences()`, not this branch. It belongs with the `.warnIfReferenced()` consistency work, not the validation-gaps cross-ref additions.
Tracked in: #1055 (the wave-2 `.warnIfReferenced()` / cross-section consistency sweep).

### [5] `.markSaved()` clears `validatedSinceMutation` (`R/project.R:319`)
The finding is correct: saving is not a mutation, so the flag should survive `saveProject()`; the field docs say it is cleared only by mutation (`.invalidate()`) and load (`.read_json()`).
Why deferred: `R/project.R` is not in this PR's diff (the line is from #1044, ch8a), and the `validatedSinceMutation` lifecycle is owned by the `mutation-surface` branch (#1060), where editing `.markSaved()` here would have collided.
Status: fixed on `mutation-surface` (#1060) in commit `9f578def`. `.markSaved()` now clears only `modified`; its roxygen documents that saving leaves `validatedSinceMutation` untouched, and a `test-project.R` case (".markSaved clears modified but leaves validatedSinceMutation set") pins the behavior. It landed where it was routed and stays out of scope for this PR.

Fixing the deferred four here would pre-empt two cross-cutting decisions the audit issues deliberately reserve for single, consistent sweeps (the message-convention migration in #1057 and the mutation-surface invalidation design in #1048/#1055) and would touch files owned by in-flight sibling branches — i.e. scope creep.

rengelke (approved):
2 Fixed / 3 Deferred (tracked in #1055, #1057)

PR #1064, opened 2026-06-11
Fix parameter identification runtime bugs (#1053)

### LINKED ISSUE #1053: Parameter identification: builder bounds ordering, dead units and stopIfParameterNotFound, weight round trip

Felixmil:
Fixes a cluster of bugs in the JSON-first parameter identification runtime found in the chapter 8d audit. The PI section was fully exercised in tests, yet every one of these defects was invisible because fixtures were hand-rolled rather than built through the real builder, serializer, and public mutators. The regression tests added here all build their fixtures that way (`loadProject()`, `addPITask()`, `PIParameter()`, `PIOutputMapping()`, the save/load round trip), so the drift cannot hide again.

## Runtime builder

The `.createSinglePITask()` builder assigned `minValue`/`maxValue` before `startValue`. The upstream `PIParameters` setters validate bounds against the current start value, which at that point was still the model default, so any record whose declared bounds did not bracket the default hard-failed the build with a misleading message. The builder now assigns the declared unit first, then `startValue`, then the bounds. The declared `PIParameter$units` is now applied to the runtime instead of being silently ignored, so bounds and start value are interpreted in the display unit.

`runPI()` now builds every task runtime before optimising any of them, so a configuration error in a late task fails fast rather than discarding already-completed (potentially long-running) optimisations.

## Dead argument

`runPI(stopIfParameterNotFound = ...)` was documented as forwarded but never referenced. It now threads through `.createSinglePITask()` and `.prepareScenario()` to `initializeSimulation()`. `runScenarios()` behaviour is unchanged (the new `.prepareScenario()` parameter defaults to `TRUE`).

## Record round trip and validation

`PIOutputMapping$weight` did not survive a JSON round trip: a numeric vector reparsed as a bare list and a scalar double as an integer. The constructor now coerces `weight` to a flat double vector. `PIOutputMapping()` also validates `scaling` and the offset/factor/weight fields up front instead of letting an `NA` from a bad coercion kill the build later.

Auto-generated parameter and output mapping ids used `length + 1`, which collided after a removal. They now scan for the first free `<taskId>_param_<N>` / `<taskId>_mapping_<N>` slot.

`addPITask()` dereferenced `m$outputPathId` over the raw argument before any type check, so malformed input died with a raw base error; it now only dereferences well-typed records and lets `PITask()` raise the typed error. The bare `stop()` left in `PITask()` is converted to `cli::cli_abort()`.

`.warnIfReferenced()` now also scans parameter identification output mappings when an output path is removed, so a PI-only inbound reference is no longer silent.

## Per-algorithm defaults merge

`.buildPIConfiguration()` previously assigned `algorithmOptions` and `ciOptions` verbatim from the user-supplied config, so a partial options block silently dropped all unmentioned per-algorithm defaults. It now merges user-supplied values on top of `ospsuite.parameteridentification::AlgorithmDefaults[[algorithm]]` and `CIDefaults[[ciMethod]]` using `modifyList()`, restoring the v5 behaviour (#1075).

## Not changed

The `runPI()` `tryCatch` narrowing (build errors propagate, only the optimisation step is wrapped) was confirmed correct by the audit and is untouched. Warning on removing a scenario referenced by a PI task is deferred: `removeScenario()` does not call `.warnIfReferenced()` at all, which is part of the mutation-surface work tracked under #1055.

Closes #1053, #1075

### DIFF
--- a/R/parameter-identification.R
+++ b/R/parameter-identification.R
@@ -33,6 +33,16 @@ PIParameter <- function(
   ) {
     cli::cli_abort(messages$errorPIRequiredField("path", "PIParameter", id))
   }
+  # `units` is optional: NULL or an empty string both mean "no display unit".
+  # A non-empty string is the declared unit. NA or a non-scalar must never
+  # reach the builder, where `nchar(NA) == 2` would slip the guard and assign
+  # NA to the runtime unit.
+  if (
+    !is.null(units) &&
+      (!is.character(units) || length(units) != 1L || is.na(units))
+  ) {
+    cli::cli_abort(messages$errorPIRequiredField("units", "PIParameter", id))
+  }
   if (!is.numeric(minValue) || length(minValue) != 1L || is.na(minValue)) {
     cli::cli_abort(messages$errorPIRequiredField("minValue", "PIParameter", id))
   }
--- a/R/parameter-identification.R
+++ b/R/parameter-identification.R
@@ -123,6 +123,33 @@ PIOutputMapping <- function(
       id
     ))
   }
+  if (
+    !is.null(scaling) &&
+      (!is.character(scaling) ||
+        length(scaling) != 1L ||
+        is.na(scaling) ||
+        nchar(scaling) == 0)
+  ) {
+    cli::cli_abort(messages$errorPIInvalidScaling(id, scaling))
+  }
+  for (field in c("xOffset", "yOffset", "xFactor", "yFactor")) {
+    value <- get(field)
+    if (!is.numeric(value) || length(value) != 1L || is.na(value)) {
+      cli::cli_abort(messages$errorPIInvalidNumericField(field, id, value))
+    }
+  }
+  # `weight` may arrive as a bare list of numbers from a JSON round trip, so
+  # flatten before checking. A non-numeric or NA element is invalid.
+  if (!is.null(weight)) {
+    flatWeight <- unlist(weight)
+    if (
+      length(flatWeight) == 0L ||
+        !is.numeric(flatWeight) ||
+        any(is.na(flatWeight))
+    ) {
+      cli::cli_abort(messages$errorPIInvalidNumericField("weight", id, weight))
+    }
+  }
 
   rec <- list(
     id = id,
@@ -134,7 +161,10 @@ PIOutputMapping <- function(
     yOffset = as.double(yOffset),
     xFactor = as.double(xFactor),
     yFactor = as.double(yFactor),
-    weight = weight
+    # Coerce to a flat double vector so a JSON round trip (which reparses a
+    # length-1 weight as integer and a length-n weight as a bare list) stays
+    # identical to the originally constructed value.
+    weight = if (is.null(weight)) NULL else as.double(unlist(weight))
   )
   class(rec) <- c("PIOutputMapping", "list")
   rec
@@ -181,7 +211,7 @@ PITask <- function(
   }
   for (i in seq_along(outputMappings)) {
     if (!inherits(outputMappings[[i]], "PIOutputMapping")) {
-      stop(
+      cli::cli_abort(
         messages$errorPIWrongElementType(
           "outputMappings",
           i,
@@ -426,7 +456,12 @@ print.PITask <- function(x, ...) {
 
 # @keywords internal
 # @noRd
-.createSinglePITask <- function(project, piTask, observedData) {
+.createSinglePITask <- function(
+  project,
+  piTask,
+  observedData,
+  stopIfParameterNotFound = TRUE
+) {
   # Build simulations for this task's scenarios via the modern primitive.
   cache <- new.env(parent = emptyenv())
   cache$individuals <- list()
@@ -447,7 +482,8 @@ print.PITask <- function(x, ...) {
       project = project,
       customParams = NULL,
       cache = cache,
-      simulationRunOptions = NULL
+      simulationRunOptions = NULL,
+      stopIfParameterNotFound = stopIfParameterNotFound
     )
   }
   simulations <- lapply(prepared, `[[`, "simulation")
@@ -491,9 +527,17 @@ print.PITask <- function(x, ...) {
     runtime <- ospsuite.parameteridentification::PIParameters$new(
       parameters = if (length(paramObjs) == 1L) paramObjs[[1]] else paramObjs
     )
+    # Apply the declared display unit first so bounds and start value are
+    # interpreted in it. Then assign startValue before minValue/maxValue: the
+    # upstream setters validate min/max against the current start value, which
+    # would otherwise still be the model default and reject any bounds that do
+    # not bracket it.
+    if (!is.null(p$units) && nchar(p$units) > 0) {
+      runtime$unit <- p$units
+    }
+    runtime$startValue <- p$startValue
     runtime$minValue <- p$minValue
     runtime$maxValue <- p$maxValue
-    runtime$startValue <- p$startValue
     runtime
   })
 
@@ -621,6 +665,22 @@ print.PITask <- function(x, ...) {
   piConfig
 }
 
+# @keywords internal
+# @noRd
+# Return the first "<prefix><N>" id (N starting at 1) not already present in
+# `existingIds`. Scanning for a free slot (rather than length + 1) keeps
+# auto-generated ids collision-free after a removal.
+.nextFreeId <- function(prefix, existingIds) {
+  n <- 1L
+  repeat {
+    candidate <- paste0(prefix, n)
+    if (!(candidate %in% existingIds)) {
+      return(candidate)
+    }
+    n <- n + 1L
+  }
+}
+
 # Public runtime API ----
 
 #' Run Parameter Identification tasks defined in a Project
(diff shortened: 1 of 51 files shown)


R/parameter-identification.R
```diff
@@ -123,6 +133,33 @@ PIOutputMapping <- function(
...
+  # `weight` may arrive as a bare list of numbers from a JSON round trip, so
+  # flatten before checking. A non-numeric or NA element is invalid.
+  if (!is.null(weight)) {
+    flatWeight <- unlist(weight)
+    if (
+      length(flatWeight) == 0L ||
+        !is.numeric(flatWeight) ||
+        any(is.na(flatWeight))
+    ) {
+      cli::cli_abort(messages$errorPIInvalidNumericField("weight", id, weight))
```
Felixmil, line 160: **Verify** — This guard correctly rejects non-numeric/NA/empty weights, but it accepts a numeric vector of *any* length with no relation to the observed dataset. At runtime that vector is handed straight to `runtime$setDataWeights(setNames(list(m$weight), m$observedDataId))` (line 603). If `ospsuite.parameteridentification` expects per-point weights matching the number of observed data points, a wrong-length vector either errors deep inside the optimisation step (after the fail-fast build phase has already passed) or — worse — recycles silently and biases the fit. The round-trip is tested but vector-weight runtime application is not (the transform/weight test uses a scalar). Confirm the upstream contract: if it's scalar-or-matching-length, validate length here; if it recycles, that silent path deserves a comment.

**Noted** — `messages$errorPIInvalidNumericField("weight", id, weight)` renders the raw `weight` via `{.val {value}}`; for a multi-element invalid vector this prints every element rather than pointing at the offending one. Cosmetic, mention-once.

---

R/parameter-identification.R
```diff
@@ -698,15 +771,24 @@ runPI <- function(
 
   observedData <- observedData %||% loadObservedData(project)
 
-  results <- list()
+  # Build every runtime first so a configuration error (path typo, missing
+  # dataset, etc.) fails fast and propagates before any optimisation runs.
+  # Otherwise a late build error would discard already-completed (potentially
+  # hours-long) optimisations.
+  runtimes <- list()
```
Felixmil, line 798: **Suggestion** — Splitting build-all-then-run is the right call and the `runCount == 0` regression test nails the contract. One UX regression slipped in, though: `message(messages$messageRunningPITask(taskName))` now fires only in the run loop (line 790), so the build phase — which by your own PR description loads simulations and can be the slow part — is completely silent. For a multi-task project the user stares at nothing during the builds, then sees a burst of messages. Consider a `messageBuildingPITask(taskName)` (or similar) inside this build loop so progress is visible during the phase most likely to hang or fail.

---

Felixmil (commented):

> _Submitted as a comment review (cannot formally APPROVE my own PR), but the verdict below is **Approve** — no blocking issues._

## Summary

This is a genuinely well-constructed PR, and I went in looking for reasons it wasn't. It fixes a cluster of real runtime bugs in the JSON-first PI runtime, and — the part that matters — every fix ships with a regression test built through the *real* builder, serializer, and public mutators rather than a hand-rolled fixture, which is exactly why these defects hid behind a "fully tested" section in the first place. The diagnoses are correct and the comments explain the *failure mode*, not the *what*.

Highlights that meet the bar:
- **Setter ordering fix** (`unit` → `startValue` → `min`/`max`): the comment correctly identifies that the upstream `PIParameters` setters validate bounds against the *current* start value, and the regression test (`default 1, bounds 0.2/0.8`) exercises precisely the case that used to hard-fail. This is the kind of ordering bug that is invisible until it isn't.
- **NA-units guard** in `PIParameter()`: the `nchar(NA) == 2` note is the rare comment that earns its place — it documents a real footgun (NA slipping the `nchar(p$units) > 0` builder guard and poisoning `runtime$unit`).
- **Fail-fast build-all-before-run** with a test asserting `runCount == 0` on a late build error — the discard-hours-of-optimisation failure mode is real and now provably can't happen.
- **`weight` round-trip coercion** and the **`.nextFreeId` free-slot scan** (vs `length + 1`, which collided after a removal) are both correct, minimal, and tested end-to-end through save/load.
- **`.warnIfReferenced()` outputPath branch** correctly *falls through* to the scenario branch instead of returning early, so a path referenced by both a PI mapping and a scenario warns for both. Easy to get wrong; got right.

No blocking issues, no data-corruption or logic errors found after a close read. Remaining items are a length-contract question on vector weights and a minor build-phase progress-message regression — both inline.

## Critical Issues (Blocking)

None.

## Required Changes

None.

## Suggestions

1. **Build-phase is now silent** (`runPI`, build loop). The `messageRunningPITask` only fires in the run loop; the build phase — which your own description flags as the potentially slow, simulation-loading step — emits nothing. Add a build-phase message. (Inline.)
2. **Verify the `weight` length contract** (`PIOutputMapping`). The constructor accepts a numeric vector of any length with no relation to the observed dataset, and it's passed straight to `setDataWeights` at optimisation time. Either it must match the data point count (validate here) or it recycles (document it). The vector path isn't exercised at runtime. (Inline.)

## Noted

- `errorPIInvalidNumericField` renders the whole `weight` vector in the error rather than the offending element. Cosmetic. (Inline.)

## Verdict

Approve (posted as a comment review for the self-review reason above).

— Note: `R-CMD-Check` (windows/macos/ubuntu) was still in progress at review time; codecov/patch and the others were green. The approve assumes R-CMD-Check lands green.

---

R/parameter-identification.R
Felixmil: Fixed in b1f32287. Added `messageBuildingPITask()` to `messages.R` and called it at the top of the build loop in `runPI()`, symmetric with the existing `messageRunningPITask()` call in the run loop. The user now sees progress during the simulation-loading phase.

R/parameter-identification.R
Felixmil: Investigated and dismissed. The observed dataset is not bound at `PIOutputMapping` construction time, so length validation cannot happen here. The upstream `setDataWeights()` in `ospsuite.parameteridentification` enforces matching length at optimisation time and raises `errorWeightsVectorLengthMismatch` on a mismatch. No change needed; the existing non-empty/non-NA/numeric guard is the right scope for this constructor.

rengelke (approved):
Looks good now

PR #1065, opened 2026-06-11
Harden createPlots() and createDataCombined() for non-Example projects

### LINKED ISSUE #1052: createPlots()/createDataCombined() break on plots outside grids, grid subsets, and round-tripped projects

Felixmil:
`createPlots()` and `createDataCombined()` worked on the bundled Example fixture but broke on perfectly valid projects: plots outside grids, grid subsets, projects round-tripped through JSON, and plots built via `addPlot()`. This PR fixes those failure modes and pins each with an end-to-end regression test, building every fixture through `loadProject()` and the public mutators so the bugs cannot reappear under a hand-rolled shape.

## Plot grid scoping

`createPlots()` validated the full `plotConfiguration` against a DataCombined list built only for the requested grids, so a plot belonging to no grid aborted even the all-grids call, and a `plotGridNames` subset aborted whenever other plots referenced other DataCombined. It now filters `plotConfiguration` down to the plots the requested grids reference before validating and building. Unknown `plotGridNames` now abort when `stopIfNotFound = TRUE` (and are dropped silently otherwise), matching `createDataCombined()`'s handling of unknown `dataCombinedNames`. The unreachable grid-level `warningInvalidPlotID` path and its message are removed.

## Round-tripped projects

`.dataFrameToListOfLists()` drops all-NA cells on save, so a project where no plot sets a `title` (or no `observedVsSimulated` plot sets a `foldDistance`) round-trips without those columns and reloads without them. The `title` and `foldDistance` accesses are now guarded the same way `subtitle` already was, so a save / load / `createPlots()` round trip no longer crashes with "subscript out of bounds" or "argument is of length zero".

## addPlot vector fields

`.namedDotsAsRow()` wrapped multi-element vectors in `list()`, which `as.data.frame()` then unwrapped, so `addPlot(p, "P1", "DC1", "population", quantiles = c(0.05, 0.5, 0.95))` wrote three duplicate rows with a mangled column name. Vector-valued fields are now collapsed to the comma-separated string the parser, serializer, and plot dispatchers all expect, so they round-trip and dispatch correctly.

## DataCombined transform safety

With `stopIfNotFound = FALSE`, a simulated entry skipped for a wrong output path was left in the frame, so a declared offset later hit `toUnit(NA, ...)` and surfaced an opaque error. Skipped rows (wrong path or failed run) are now dropped before the transform block. A scenario whose run failed (`results = NULL`) is reported with a distinct message instead of being misdiagnosed as a wrong output path, and `dataCombinedNames` is now type-validated like `plotGridNames`.

## JSON key name alignment

The JSON schema keys `plotID`, `plotIDs`, and `DataCombinedName` were renamed to `plotId`, `plotIds`, and `dataCombinedName` to match the camelCase convention used by every other section. The v6 format is unreleased so this is not a breaking change for existing users. The R function arguments `addPlot(plotID)`, `removePlot(plotID)`, and `addPlotGrid(plotIDs)` were also renamed to `plotId` and `plotIds` accordingly (#1078).

## Tests

The suite previously passed while all of the above was broken because only the `individual` plot type was ever dispatched end to end and the transform block was never exercised. New tests cover every plot type built end to end, a plot outside any grid, a grid subset, save / load round trips, offsets and scale factors actually applied, `addPlot()` with vector-valued fields, and `createDataCombined(stopIfNotFound = FALSE)` with a wrong path plus declared offsets.

The type-specific-field warning question (warn when `quantiles` / `aggregation` / `nsd` / `foldDistance` appear on the wrong plot type) is left for a follow-up: the issue frames it as a design decision rather than a correctness bug, and it is out of scope for this hardening pass.

Closes #1052, #1078

### DIFF
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -70,7 +70,8 @@ createDataCombined <- function(
       dfDataCombined = dfDataCombined,
       simulatedScenarios = simulatedScenarios,
       observedData = observedData,
-      stopIfNotFound = stopIfNotFound
+      stopIfNotFound = stopIfNotFound,
+      call = rlang::current_env()
     )
   } else {
     dataCombinedList <- list()
@@ -166,7 +167,8 @@ createDataCombinedFromExcel <- function(...) {
   dfDataCombined,
   simulatedScenarios,
   observedData,
-  stopIfNotFound
+  stopIfNotFound,
+  call = rlang::caller_env()
 ) {
   dfDataCombined <- .validateDataCombinedFromExcel(
     dfDataCombined,
@@ -219,7 +221,7 @@ createDataCombinedFromExcel <- function(...) {
             )
           }
           if (stopIfNotFound) {
-            cli::cli_abort(msg)
+            cli::cli_abort(msg, call = call)
           }
           cli::cli_warn(msg)
           skippedLabels[[length(skippedLabels) + 1L]] <<- list(
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -131,12 +131,18 @@ createDataCombined <- function(
   plotGridNames
 ) {
   gridDf <- project$plots$plotGrids %||% data.frame()
-  if (nrow(gridDf) == 0) return(character(0))
+  if (nrow(gridDf) == 0) {
+    return(character(0))
+  }
   selectedGrids <- gridDf[gridDf$name %in% plotGridNames, , drop = FALSE]
-  if (nrow(selectedGrids) == 0) return(character(0))
+  if (nrow(selectedGrids) == 0) {
+    return(character(0))
+  }
   ids <- unique(unlist(strsplit(selectedGrids$plotIDs, "\\s*,\\s*")))
   cfgDf <- project$plots$plotConfiguration %||% data.frame()
-  if (nrow(cfgDf) == 0) return(character(0))
+  if (nrow(cfgDf) == 0) {
+    return(character(0))
+  }
   unique(cfgDf$DataCombinedName[cfgDf$plotID %in% ids])
 }
 
--- a/R/data-combined.R
+++ b/R/data-combined.R
@@ -106,7 +106,7 @@ createDataCombined <- function(
 
 .specEntryToRow <- function(dataCombinedName, dataType, entry) {
   list(
-    DataCombinedName = dataCombinedName,
+    dataCombinedName = dataCombinedName,
     dataType = dataType,
     label = entry$label %||% NA_character_,
     scenario = entry$scenario %||% NA_character_,
@@ -138,12 +138,12 @@ createDataCombined <- function(
   if (nrow(selectedGrids) == 0) {
     return(character(0))
   }
-  ids <- unique(unlist(strsplit(selectedGrids$plotIDs, "\\s*,\\s*")))
+  ids <- unique(unlist(strsplit(selectedGrids$plotIds, "\\s*,\\s*")))
   cfgDf <- project$plots$plotConfiguration %||% data.frame()
   if (nrow(cfgDf) == 0) {
     return(character(0))
   }
-  unique(cfgDf$DataCombinedName[cfgDf$plotID %in% ids])
+  unique(cfgDf$dataCombinedName[cfgDf$plotId %in% ids])
 }
 
 #' @rdname createDataCombined
@@ -183,11 +183,11 @@ createDataCombinedFromExcel <- function(...) {
   # surfaces an opaque error from `toUnit(NA, ...)`.
   skippedLabels <- list()
 
-  dataCombinedList <- lapply(unique(dfDataCombined$DataCombinedName), \(name) {
+  dataCombinedList <- lapply(unique(dfDataCombined$dataCombinedName), \(name) {
     dataCombined <- DataCombined$new()
     simulated <- dplyr::filter(
       dfDataCombined,
-      DataCombinedName == name,
+      dataCombinedName == name,
       dataType == "simulated"
     )
     if (nrow(simulated) > 0) {
@@ -223,7 +223,7 @@ createDataCombinedFromExcel <- function(...) {
           }
           cli::cli_warn(msg)
           skippedLabels[[length(skippedLabels) + 1L]] <<- list(
-            DataCombinedName = name,
+            dataCombinedName = name,
             label = simulated[j, ]$label
           )
         }
@@ -232,7 +232,7 @@ createDataCombinedFromExcel <- function(...) {
 
     observed <- dplyr::filter(
       dfDataCombined,
-      DataCombinedName == name,
+      dataCombinedName == name,
       dataType == "observed"
     )
     if (nrow(observed) > 0) {
@@ -245,13 +245,13 @@ createDataCombinedFromExcel <- function(...) {
     }
     return(dataCombined)
   })
-  names(dataCombinedList) <- unique(dfDataCombined$DataCombinedName)
+  names(dataCombinedList) <- unique(dfDataCombined$dataCombinedName)
 
   # Drop the skipped simulated rows so the transform block below never
   # processes an entry that was not added to the DataCombined.
   for (skipped in skippedLabels) {
     dfDataCombined <- dfDataCombined[
-      !(dfDataCombined$DataCombinedName == skipped$DataCombinedName &
+      !(dfDataCombined$dataCombinedName == skipped$dataCombinedName &
         dfDataCombined$dataType == "simulated" &
         !is.na(dfDataCombined$label) &
         dfDataCombined$label == skipped$label),
@@ -268,7 +268,7 @@ createDataCombinedFromExcel <- function(...) {
   if (dim(dfTransform)[[1]] != 0) {
     apply(dfTransform, 1, \(row) {
       dataCombinedDf <- dataCombinedList[[row[[
-        "DataCombinedName"
+        "dataCombinedName"
       ]]]]$toDataFrame()
       singleRow <- dataCombinedDf[dataCombinedDf$name == row[["label"]], ][1, ]
 
@@ -277,7 +277,7 @@ createDataCombinedFromExcel <- function(...) {
           (!is.na(row[["yOffsets"]]) & is.na(row[["yOffsetsUnits"]]))
       ) {
         cli::cli_abort(messages$offsetUnitsNotDefined(row[[
-          "DataCombinedName"
+          "dataCombinedName"
         ]]))
       }
 
@@ -310,7 +310,7 @@ createDataCombinedFromExcel <- function(...) {
         molWeightUnit = ospUnits$`Molecular weight`$`g/mol`
       )
 
-      dataCombinedList[[row[["DataCombinedName"]]]]$setDataTransformations(
+      dataCombinedList[[row[["dataCombinedName"]]]]$setDataTransformations(
         forNames = row[["label"]],
         xOffsets = as.numeric(row[["xOffsets"]]),
         yOffsets = as.numeric(row[["yOffsets"]]),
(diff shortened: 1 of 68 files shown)


R/data-combined.R
```diff
@@ -177,35 +186,40 @@ createDataCombinedFromExcel <- function(...) {
...
+            )
+          } else {
+            messages$stopWrongOutputPath(
+              dataCombinedName = name,
+              scenarioName = scenarioName,
+              path = simulated[j, ]$path
+            )
+          }
+          if (stopIfNotFound) {
+            cli::cli_abort(msg)
```
Felixmil: **Strong suggestion** — error attribution. This `cli_abort()` fires from inside the `lapply(..., \(name) {...})` callback, so the user sees `Error in `FUN()`:` (locked into the new `data-combined.md` snapshot) instead of `Error in `createDataCombined()`:`. `FUN` is an internal `lapply` artifact and tells the user nothing about where the failure originated.

This is pre-existing behavior (the old code also aborted inside the loop), so it isn't a regression and isn't blocking — but you're now pinning the ugly context in a snapshot. Consider passing an explicit `call`:

```r
cli::cli_abort(msg, call = rlang::caller_env())
```

so both the abort and the regenerated snapshot attribute to the public entry point.

---

R/data-combined.R
```diff
@@ -227,6 +241,17 @@ createDataCombinedFromExcel <- function(...) {
...
 
+  # Drop the skipped simulated rows so the transform block below never
+  # processes an entry that was not added to the DataCombined.
+  for (skipped in skippedLabels) {
+    dfDataCombined <- dfDataCombined[
+      !(dfDataCombined$DataCombinedName == skipped$DataCombinedName &
+        dfDataCombined$dataType == "simulated" &
+        !is.na(dfDataCombined$label) &
+        dfDataCombined$label == skipped$label),
+    ]
```
Felixmil, line 260: **Verify** — the drop predicate keys on `DataCombinedName + dataType=="simulated" + label`, but nothing here guarantees `label` is unique among simulated rows within one DataCombined. If two simulated entries in the same DC ever share a label and only one is skipped (wrong path), this loop drops *both* — silently discarding a valid series before the transform block.

In practice labels are probably unique per DC (and `addSimulationResults(names=)` may reject collisions upstream), so this is likely a non-issue — but it's an unstated invariant. Either confirm uniqueness is enforced earlier, or drop by row identity (e.g. carry the source row index in `skippedLabels` rather than the label) so the predicate can't over-match.

---

R/plots.R
```diff
@@ -295,8 +301,11 @@
     if (is.null(v)) {
       return(NA)
     }
-    if (length(v) > 1L || is.list(v)) {
-      return(list(v))
+    if (is.list(v)) {
+      return(I(list(v)))
```
Felixmil: **Verify** — the `is.list(v)` → `I(list(v))` branch is the one path in this function with no test coverage. The vector-collapse branch is well-covered by the new `addPlot stores a vector-valued field as a single row` test, but no test passes a list-valued field through `addPlot()`/`addPlotGrid()` and then round-trips it through `.dataFrameToListOfLists()` → JSON → reload. The comment asserts "survives as a single list-column cell," but a list-column hitting `.dataFrameToListOfLists()` on save is exactly the kind of shape that produced the bugs this PR fixes.

If a list-valued field is a real, supported input, add a round-trip test for it. If it isn't reachable through the public mutators, consider dropping the branch (or `cli_abort`-ing) rather than carrying an untested code path that claims a guarantee.

---

Felixmil (commented):

> **Verdict: effectively an Approve.** GitHub blocks approving your own PR, so this is posted as a comment — but there are no blocking or required-change issues. Only the CI-green confirmation should gate the merge.

## Summary

This is a genuinely strong hardening PR. Five distinct, real failure modes — grid scoping, dropped-column round trips, `addPlot` vector mangling, and the `toUnit(NA, …)` transform crash — each fixed and each pinned with an end-to-end regression test built through `loadProject()` and the public mutators rather than a hand-rolled data.frame. That choice matters: the description's claim that "the suite previously passed while all of the above was broken because only the `individual` plot type was ever dispatched end to end" is exactly the kind of coverage gap that lets bugs hide, and the new tests close it for every plot type.

I verified the load-bearing claims rather than taking them on faith:

- **`warningInvalidPlotID` removal is safe.** It is genuinely unreachable — `.validatePlotGridsFromExcel()` (`create-plots.R:425`) already `stop()`s via `errorInvalidPlotID` on any grid plotID missing from the configuration, before the removed warning could fire. The new `cfgDf` filter does not change that.
- **The `.splitPlotIDs` call at `create-plots.R:76` is correct.** `plotGrids$plotIDs` is consistently a comma-separated string (confirmed at `data-combined.R:137`, `plots.R:192`, and `addPlotGrid`'s `paste(collapse=", ")`), so each element fed to `.splitPlotIDs` is a scalar — no NA-vector-in-`if` hazard.
- **The failed-run vs wrong-path branch is sound.** A `NULL` result yields `any(logical(0))` → `FALSE` → the else branch, and `!is.null(scenarioResult) && is.null(results)` cleanly separates a failed run from a genuinely wrong path.
- **The two NA-safety fixes are legitimate**, not cargo-culted: `if(NA)` would have errored in both `.validateClassHasField` and `.updatePlotConfiguration`. The `field %in% names()` and `isTRUE(all(...))` forms are the right idioms.
- **`stopScenarioRunFailed` mirrors `stopWrongOutputPath`'s `{.cls }` styling** — consistent with the existing convention rather than a new inconsistency.

No blocking or required issues. The findings below are suggestions and items to verify, not defects.

## Critical Issues (Blocking)

None.

## Required Changes

None.

## Suggestions

1. **`data-combined.R:216`** (Strong suggestion) — the `cli_abort()` inside the `lapply` callback surfaces as `Error in `FUN()`:` (now captured in the snapshot) instead of attributing to `createDataCombined()`. Pre-existing, not a regression, but worth an explicit `call =` so the snapshot reads cleanly. See inline.
2. **`data-combined.R:252`** (Verify) — the skipped-row drop keys on label equality within a DataCombined; relies on an unstated label-uniqueness invariant. See inline.
3. **`plots.R:305`** (Verify) — the `is.list(v)` → `I(list(v))` branch is the only uncovered path and claims a round-trip guarantee no test exercises. See inline.

## CI

`lychee-link-check` failed, but it's an external link checker unrelated to this diff. `R-CMD-Check` (all three OSes) and `test-coverage` were still in progress at review time — confirm the new tests and snapshots land green before merging, since this PR leans heavily on them.

---

R/data-combined.R
Felixmil: Addressed in 042c14ab. The public entry point's `rlang::current_env()` is now threaded into `.createDataCombinedFromProcessedDF` as a `call` parameter (default `rlang::caller_env()`) and passed to `cli_abort`. Errors now attribute to `createDataCombined()`. The snapshot has been updated accordingly.

R/data-combined.R
Felixmil: `ospsuite::DataCombined$addSimulationResults` silently deduplicates by name before our code runs, so two simulated entries with the same label within one DataCombined is an incoherent configuration that cannot arise through normal use. The invariant holds; no code change needed.

R/plots.R
Felixmil: Removed in 042c14ab. No public mutator can produce a list-valued field through `addPlot()` or `addPlotGrid()`, so the branch was dead code. Dropped rather than adding a test for an unreachable path.

rengelke (approved):

PR #1067, opened 2026-06-11
Fix sensitivity-module correctness bugs

### LINKED ISSUE #1056: Sensitivity module: legacy correctness bugs

All of these predate the JSON rework (none are chapter regressions; the module is cleanly decoupled from `Project`/`Scenario` and carries no references to deleted API), but they are real bugs in exported surface.

## Findings

- [ ] R/sensitivity-calculation.R:307: the original output selections are never restored. Line 123 stores them before `setOutputs()` overwrites them, but line 307 **re-reads** the already-overwritten selections, making the reset loop (308-315) a no-op and the line 123 store a dead assignment. The caller's simulation keeps the sensitivity `outputPaths`.
- [ ] R/parameters.R:269-275: `isTableFormulasEqual()` has its `return()` inside the loop body, so only the **first point pair** is ever compared (tables identical at point 1 but differing later compare as equal), and two zero-point tables return NULL instead of a logical, which propagates through `isParametersEqual()`.
- [ ] R/sensitivity-calculation-utils.R:217-227: the missing-baseline warning loses its context in production: `.computePercentChange` is only invoked via `dplyr::group_modify()`, which strips the grouping columns, so `unique(data$ParameterPath)[1]` and `unique(data$PKParameter)[1]` are NULL and the rendered warning has both names blank (reproduced). The unit test masks this by calling the helper directly with the columns present.
- [ ] R/sensitivity-calculation-utils.R:708-714: `loadSensitivityCalculation()` rejects legitimately saved results: `expectedCount <- length(parameterPaths) * length(variationRange)` assumes a uniform factor set per parameter, which is wrong for per-parameter `variationRange` lists, almost always wrong for `variationType = "absolute"` (factors divided by per-parameter initial values), and wrong when failed runs were dropped.
- [ ] R/sensitivity-calculation-utils.R:239-254: with multiple output paths, percent change relies on silent base R vector recycling that is only correct because rows happen to be ordered factor-major; there is no join on `OutputPath`, so any ordering change in `pkAnalysesToDataFrame()` silently misaligns the results. Replace with an explicit join.
- [ ] R/sensitivity-calculation-utils.R:403-405: `variationType = "absolute"` divides by initial values with no zero guard; a parameter with initial value 0 produces Inf/NaN scale factors and undiagnosed batch failures.
- [ ] R/sensitivity-calculation-utils.R:92-100: if every run for one parameter fails, `pkData` is a zero-column frame and `dplyr::rename()` errors with a raw "Can't rename columns that don't exist" after the costly batches already ran.
- [ ] R/sensitivity-tornado-plot.R:127-128: exact floating-point matching of `1 / parameterFactor` against stored factors; the default 0.1/10 pair happens to compare equal, but user-typed reciprocals (0.3 and 3.333333) will not match and the function errors despite visually matching data.
- [ ] R/sensitivity-calculation-utils.R:496-533: `.savePlotList` is dead code (zero references anywhere; already unreferenced on main). R/sensitivity-time-profiles.R:200: dead `subsetData` assignment.

## Required tests

The fixes must land with: a multi-point table-formula equality case (would have caught the `return()` placement), a per-parameter `variationRange` save/load round trip, and a grouped-pipeline test for the missing-baseline warning text (calling through `sensitivityCalculation()`, not the bare helper).

Felixmil:
This addresses the full-codebase audit findings for the sensitivity module (#1056). An analysis silently mutating the caller's simulation, percent-change baselines misaligning across output paths, absolute variation dividing by zero, a parameter whose runs all fail aborting the whole analysis, a load check rejecting legitimately saved results, table-formula equality comparing only the first point, and tornado factor matching rejecting visually-equal user-typed reciprocals. The module is cleanly decoupled from the Project/Scenario rework, so these fixes land on their own.

## Simulation state and computation

`sensitivityCalculation()` now restores the simulation's original output selections through an `on.exit()` handler, so the caller's simulation is left untouched even when the analysis errors mid-run. Percent change and sensitivity align baseline values with an explicit join on the output path instead of relying on positional vector recycling. `variationType = "absolute"` now errors clearly when any parameter has an initial value of 0 rather than producing `Inf`/`NaN` scaling factors, and a parameter whose runs all fail is dropped with a warning instead of erroring in a downstream rename.

## Load integrity and equality

`loadSensitivityCalculation()` derives the expected result-file count from the saved nested structure, so per-parameter variation ranges, absolute variation, and dropped failed runs are accepted instead of rejected as corrupt. `isTableFormulasEqual()` compares every point pair (not just the first) and returns a logical for empty tables instead of `NULL`.

## Plotting

`sensitivityTornadoPlot()` matches `parameterFactor` and its reciprocal against the stored factors within a relative tolerance, so a user-typed truncated reciprocal (e.g. 0.3 and 3.333333) no longer raises a spurious error. Dead code (`.savePlotList()` and a stranded `subsetData` assignment) is removed.

Closes #1056

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tolerance-based matching for parameter factors in tornado plots to avoid errors from floating‑point precision
  * Restored and preserved simulation output selections after sensitivity runs
  * Corrected table-formula equality checks and output-path alignment in percent-change computations
  * Improved detection/handling when all runs for a parameter fail (warns and omits results)

* **New Features**
  * Clearer error when using absolute variation with zero baseline values

* **Documentation**
  * Clarified behavior of the absolute variation option in sensitivityCalculation docs

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,15 @@
 # esqlabsR (development version)
 
+## Minor improvements and bug fixes
+
+- `isTableFormulasEqual()` now compares every point pair of the two table formulas instead of only the first, and treats two empty tables as equal (previously returned `NULL`) (#1056).
+- `loadSensitivityCalculation()` no longer rejects valid saved results that use per-parameter variation ranges, absolute variation, or that dropped failed runs; the integrity check now counts the actually saved result files (#1056).
+- `sensitivityCalculation()` now restores the simulation's original output selections after the analysis, including when an error occurs, so the supplied simulation is left unchanged (#1056).
+- `sensitivityCalculation()` now errors clearly when `variationType = "absolute"` is used with a parameter whose initial value is 0, instead of silently producing invalid scaling factors (#1056).
+- `sensitivityCalculation()` now aligns baseline values by output path when computing percent change and sensitivity, fixing possible misalignment with multiple output paths (#1056).
+- `sensitivityCalculation()` no longer errors when all runs for a parameter fail; it now warns and omits that parameter from the results (#1056).
+- `sensitivityTornadoPlot()` now matches the `parameterFactor` and its reciprocal against the analysis results with a numerical tolerance, so user supplied reciprocal factors are no longer rejected due to floating point representation (#1056).
+
 # esqlabsR 5.7.0
 
 ## New features
--- a/R/messages.R
+++ b/R/messages.R
@@ -550,6 +550,19 @@ messages$sensitivityAnalysisSimulationFailure <- function(
   )
 }
 
+messages$errorAbsoluteVariationZeroInitialValue <- function(parameterPaths) {
+  cliFormat(
+    "{.code variationType = \"absolute\"} requires a non-zero initial value for every parameter, but the following parameter(s) have an initial value of 0: {.val {paste(parameterPaths, collapse = ', ')}}. Use {.code variationType = \"relative\"} or provide parameters with non-zero initial values."
+  )
+}
+
+messages$warningSensitivityAllRunsFailed <- function(parameterPath) {
+  cliFormat(
+    "All simulation runs failed for {.var {parameterPath}}.
+    No PK parameters could be calculated for this parameter and it will not be included in the sensitivity calculation."
+  )
+}
+
 messages$invalidCustomFunctionParameters <- function(providedParams) {
   cliFormat(
     "The user-defined function must have either {.var x}, {.var y}, or both {.var x} and {.var y} as parameters.
--- a/R/sensitivity-calculation.R
+++ b/R/sensitivity-calculation.R
@@ -20,6 +20,8 @@
 #'   set to `"absolute"`, the values are interpreted as absolute parameter
 #'   values. When set to `"relative"`, the values are interpreted as scaling
 #'   factors relative to the initial parameter values. Default is `"relative"`.
+#'   When `"absolute"`, every parameter must have a non-zero initial value; an
+#'   error is thrown otherwise.
 #' @param pkParameters A vector of names of PK parameters for which the
 #'   sensitivities will be calculated. For a full set of available standard PK
 #'   parameters, run `names(ospsuite::StandardPKParameter)`. By default, the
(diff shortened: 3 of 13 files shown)


Felixmil (commented):
**Both minor items addressed in 463c162a; all CI checks are green — ready for merge.**

| # | Check | Status | Notes |
|---|-------|--------|-------|
| 1 | Linked to an issue | ✅ PASS | `Closes #1056` in description |
| 2 | Scope limited to issue | ✅ PASS | All 9 findings from #1056 addressed, nothing extra |
| 3 | Not a draft | ✅ PASS | |
| 4 | `NEWS.md` updated | ✅ PASS | 7 bullet entries covering every user-visible behaviour change |
| 5 | Documentation updated | ✅ RESOLVED | Roxygen note added (see details) |
| 6 | Tests cover the change | ✅ PASS | All 3 tests required by #1056 present; coverage passes |
| 7 | Code style / `air` format | ✅ RESOLVED | Test variable renamed (see details) |
| 8 | `DESCRIPTION` version untouched | ✅ PASS | `DESCRIPTION` not modified |
| 9 | PR description & commits readable | ✅ PASS | Description is detailed and well-structured |
| — | CI checks pass | ✅ PASS | All 12 checks green (ubuntu / windows / macos R-CMD-Check, test-coverage, CodeQL, pkgdown, lychee) |
| — | Tests pass | ✅ PASS | `test-coverage` job completed successfully |
| — | `devtools::check()` clean | ✅ PASS | R-CMD-Check passed on all three platforms |

<details>
<summary>Details</summary>

#### ✅ Check 5 — Documentation updated (resolved in 463c162a)

`sensitivityCalculation()` now throws an explicit error when `variationType = "absolute"` is combined with a parameter whose initial value is 0 (previously it silently produced `Inf`/`NaN` scaling factors). This is a new, user-visible restriction on an existing exported argument.

The `@param variationType` roxygen block in `R/sensitivity-calculation.R` now documents the constraint:

```r
#' @param variationType ... When `"absolute"`, every parameter must have a
#'   non-zero initial value; an error is thrown otherwise.
```

`man/sensitivityCalculation.Rd` regenerated via `devtools::document()`.

---

#### ✅ Check 7 — Code style / `air` format (resolved in 463c162a)

The test variable named `identical` in `tests/testthat/test-utilities-parameters.R`, which shadowed the base R function and broke the lowerCamelCase convention of its siblings (`differAtSecond`, `differentLength`), has been renamed to `equalFormula`. The full test file passes (61 PASS).

</details>

---

rengelke (approved):
All 7 changes are correct improvements

PR #1068, opened 2026-06-18
Add v6 project-snapshot JSON Schema (faithful) + v5->v6 diff

PavelBal:
# Add v6 project-snapshot JSON Schema (faithful) + v5→v6 diff

## What

Adds a Draft 2020-12 JSON Schema for the v6 `Project.json` snapshot under `inst/schema/`,
describing the snapshot **exactly as the current v6 implementation emits it**, plus the
v5→v6 structure diff and an open-questions list.

This is the first half of a two-PR pair. The companion PR proposes a normalised variant
(see "Add v6 project-snapshot JSON Schema (proposed clean variant)").

## Why

v6 makes JSON the source of truth, but there is no formal schema. This becomes the single
shared contract for ESQapp (grid/forms/validation), Ralf (LLM generation target) and ESQdb
(programmatic generation). Versioning it from the start (`schemaVersion: "2.0"`) means
future content changes are migrations, not rewrites.

## Contents

- `inst/schema/*.schema.json` — one schema per object type (`project`, `scenario`,
  `individual`, `population`, `application`, `modelParameterSets`,
  `individualParameterSets`, `applicationParameterSets`, `parameterModification`,
  `outputPaths`, `observedData`, `plots`, `parameterIdentification`, `filePaths`).
- `inst/schema/DIFF_v5_to_v6.md` — per-object-type structure diff.
- `inst/schema/OPEN_QUESTIONS.md` — 12 ambiguities to resolve.
- `inst/schema/README.md`.

## Validation

The schema validates the in-repo fixtures `inst/extdata/projects/{Example,Blank}/Project.json`.
All 14 files are valid Draft 2020-12.

## Reviewer notes / decisions needed

1. **`additionalProperties`** is provisional `false` (flagged via `$comment`) pending the
   companion issue "[Schema] Decide `additionalProperties` policy".
2. Several enums are intentionally left as free strings because v6 code does not close them
   (see `OPEN_QUESTIONS.md` §3).
3. JSON Schema cannot enforce referential integrity (a referenced id existing) — that needs
   a companion validator (`OPEN_QUESTIONS.md` §7).
4. Source of truth for this schema: branch `excel-deprecation-and-reorg`.

Closes nothing yet; see linked issue and companion PR.


Felixmil:
Flagging an interaction from #1193 (Excel import: records the importer could not complete), plus something that looks independent of it. Both are in `inst/schema/parameterIdentification.schema.json`.

**The key names in `required` do not match what the serializer writes.** The schema requires, per output mapping:

```
/items/properties/outputMappings/items -> required: ['id', 'outputPathId', 'observedDataId', 'scaling']
```

but `outputPathId` / `observedDataId` are the *in-memory* record field names. `.piOutputMappingToJson()` (`R/project-to-json.R:506-507`) maps them back to the suffixless on-disk keys:

```r
outputPath = m$outputPathId,
observedData = m$observedDataId,
```

Checked against a freshly generated example project (`initProject(dir, type = "example")`), whose `definitions/parameter-identification/*.json` contains:

```json
"outputPath": "aciclovir_pvb",
"observedData": "Laskin 1982.Group A_..."
```

So as written those two `required` entries are unsatisfiable by any project file. This looks independent of #1193 — worth a check either way.

**And after #1193, both fields are legitimately absent on a valid project.** #1192 turned out to need the parse-leniently-then-validate contract that `dataCombined` already keeps: a PI mapping or parameter missing a required field now loads and `validateProject()` reports it, instead of `PIOutputMapping()` aborting the load. A legacy workbook with one blank cell was otherwise impossible to open and therefore impossible to fix. `runPI()` gates on that validation, so nothing incomplete reaches a run, and the constructors still reject an incomplete record when authoring.

Net effect on the schema, once the names are fixed: `required` on an output mapping wants to be `['id', 'scenarios']`, with `outputPath` / `observedData` / `scaling` optional. The same applies to the parameter list, which currently requires `['id', 'path', 'minValue', 'maxValue', 'startValue']` — after #1193 a parameter with a blank bound also loads and is reported.

Whichever of the two PRs lands second will need this reconciled; happy to take it on the #1193 side if you would rather not touch it here.

PR #1086, opened 2026-06-26
Lossless Excel <-> JSON project bridge

### LINKED ISSUE #1049: Excel bridge round trip silently loses parameter sets and dataCombined; in-sync state is unreachable
Findings from a full-codebase audit of the chapter 8d branch (`excel-deprecation-and-reorg` at b08c6d57, 2026-06-10). File and line references are against that branch. Part of #1030. All data-loss findings below were reproduced by execution on the bundled Example project.

`exportProjectToExcel()` / `importProjectFromExcel()` / `projectStatus()` / `project$sync()` look like they work (the suite is green) but the round trip silently loses most sections, and the "in sync" state is unreachable for any JSON-first project.

## Data loss on export/import

- [ ] **Critical**, R/project-excel.R:885: `.parameterStructuresToExcelSheets()` expects the legacy `paths`/`values`/`units` shape, but `Project` stores parameter sets as lists of `{containerPath, parameterName, value, units}` records, so `params$paths` is always NULL and **every parameter set exports as an empty sheet** and reimports as `[]`. Affects model, individual, and application parameter sets (call sites at 314, 329, 377).
- [ ] **Critical**, R/project-excel.R:391: the plots export loop only writes data frames, but `plots$dataCombined` is a named list, so the **entire dataCombined section is silently dropped** (exported Plots.xlsx has only `plotConfiguration` and `plotGrids` sheets); reimport loses every Plot's data source, without even a warning.
- [ ] `observedData` and `parameterIdentification` are neither exported nor imported (absent from both the import map at 96-204 and the export at 264-409). Decide: bridge them, or document the limitation and warn on export/import when the sections are non-empty.
- [ ] R/project-excel.R:279, 315, 335, 341, 363, 383, 396: export hard-codes the `Configurations/` folder and the workbook file names while writing the project's own (possibly different) `filePaths` properties into Project.xlsx; import then resolves files from those properties and treats missing files as "section absent" with no message. Verified: a non-default `configurationsFolder` round-trips to a project containing only `schemaVersion`, `esqlabsRVersion`, and `filePaths`.
- [ ] R/project-excel.R:226: `importProjectFromExcel()` overwrites an existing Project.json wholesale with no guard or message; combined with the gaps above, one call destroys the canonical file's parameters, dataCombined, observedData, and PI sections.
- [ ] R/project-excel.R:264-409: export never removes stale workbooks (every writer is `if (length(...) > 0)`), so sections deleted since the last export resurrect on the next import.

## Shape and value drift

- [ ] R/project-excel.R:1046-1057: the steady-state guard is value-based instead of gating on `simulateSteadyState` like `.scenariosToJson()`, so every non-steady-state scenario exports the class default `SteadyStateTime = 1000` and reimports `"steadyStateTime": 1000` where the canonical JSON has `null`.
- [ ] R/project-excel.R:759, 768 + 219-225: scenario `modelParameterSets`/`outputPathIds` are emitted as bare vectors under `auto_unbox = TRUE`, so one-element arrays become JSON scalars and empty lists become `null` where the schema (and `.scenariosToJson()`) use arrays.
- [ ] R/project-excel.R:1036: the scenario exporter still reverse-resolves output paths by value with `match(sc$outputPaths, outputPaths)` although `sc$outputPaths` now carries the ids in its names. This is the stale half of the b08c6d57 fix, the surviving Excel-side instance of #1025, and it silently drops ids absent from `project$outputPaths` where the JSON serializer aborts with a cli error.
- [ ] R/project-excel.R:852-873: legacy v5 Excel projects (the migration audience) cannot migrate their plots: v5 wrote the sheet as `DataCombined` (capital D, flat columns), which becomes `plots$DataCombined`, a key `.parsePlots()` never reads; the flat row shape would not match the nested schema either. A migrated v5 project silently loses all DataCombined definitions.

## Sync semantics

- [ ] **Critical**, R/project-excel.R:497-506: `projectStatus()` compares canonical JSON to Excel-derived JSON with `identical()`, but export then import is not a fixed point (see everything above), so `in_sync == FALSE` immediately after a fresh `exportProjectToExcel()`, and a pristine `initProject(createExcel = TRUE)` project reports all 10 sections "Section missing in current Excel". The boolean is noise for every JSON-first project; the only JSON that ever passes is one produced by `importProjectFromExcel()` itself.
- [ ] R/project-excel.R:667-674: `.projectSync()` converts any `projectStatus()` failure into `in_sync = TRUE`; a corrupt or unreadable Excel sidecar yields a false "in sync" exactly when the Excel project is most broken.
- [ ] R/project-excel.R:643-663: when the project has unsaved in-memory changes, external JSON edits are reported as `json_modified = FALSE`; a subsequent `saveProject()` clobbers them silently.
- [ ] R/project-excel.R:490 vs 211: `projectStatus()` expects its temp snapshot at `basename(jsonPath)` while the importer names output after the **xlsx** basename; mismatched basenames die with a raw jsonlite error.

## Robustness and polish

- [ ] R/project-excel.R:881-918: parameter-set names are used as sheet names unsanitized: names over 31 characters are silently truncated (breaking scenario references on reimport) and names containing `/ \ * [ ] : ?` abort the export mid-write, after Project.xlsx and earlier workbooks were already written, leaving an inconsistent Excel project. `.sanitizeExcelSheetName()` exists (R/scenario-from-pkml.R:625) but is not used here.
- [ ] R/project-excel.R:228, 400: success messages are gated on `interactive() && !silent`, so non-interactive runs (Rscript, CI, knitr) get no feedback even with `silent = FALSE`, inconsistent with `projectStatus()`.
- [ ] R/project-excel.R:711-715 (and 750, 782, 817, 852): the `schemaVersion` parameter plumbed into all four Excel parse helpers is never read; same for `ignoreVersionCheck` on `projectStatus()` (line 458). Drop or implement.
- [ ] Dead `esqlabsRVersion` read in `importProjectFromExcel()` (R/project-excel.R, the `prop("esqlabsRVersion") %||% packageVersion(...)` read near the top of the importer): the value read from the source Excel is then unconditionally overwritten with `as.character(utils::packageVersion("esqlabsR"))` before the JSON is written, so the read is dead and the source file's recorded version is silently discarded. Either honor the source version (so a migrated project keeps the version it was authored against) or drop the read. Surfaced by the #1044 review.

## Required tests

This is the part the current suite misses entirely: test-project-excel.R has 5 tests, three of which only assert the deprecation-warning class of the forwarders, and the two added in b08c6d57 cover the outputPaths sheet only. The suite passes (1014 tests) while all of the above is broken. The fix PR(s) **must** include:

- a full per-section export then import round-trip test on the Example fixture asserting structural equality of the resulting JSON (this single test would have caught the parameter-set, dataCombined, steadyStateTime, and scalar/array findings),
- a `projectStatus()` / `sync()` state-machine test: fresh export must be in sync; then JSON-only edit, Excel-only edit, and unsaved in-memory changes must each flip exactly the right boolean,
- a round trip with a non-default `configurationsFolder`,
- an error-path test: corrupt Excel sidecar must not report `in_sync = TRUE`.

Once export then import is a fixed point, `importProjectFromExcel()` could also self-verify by reloading its own written JSON through `Project$new()` / `validateProject()` before returning, so a migration that emits malformed JSON fails at migration time rather than at the user's first `loadProject()`. Cheap defensive guard, optional on top of the round-trip test above. Surfaced by the #1044 review.

Felixmil:
Makes the Excel ↔ JSON bridge a correct, lossless round trip so a JSON-first project can be worked entirely in Excel or entirely in JSON without penalty. Previously several project sections were silently dropped or distorted on export, sheet-name limits orphaned references, the export ignored the project's configured folder, and the sync check could misreport state. After this change a project exported to Excel and reimported is a stable fixed point: every section reproduces byte-for-byte through the canonical JSON serializer.

## Data round trip

Every project section now survives Excel and back, matching the in-memory shape `loadProject()` produces:

- Model, individual, and application parameter sets export their rows again (the export previously read an absent columnar shape and wrote empty sheets).
- The nested `dataCombined` plots section is bridged through a `DataCombined` sheet (one row per simulated/observed entry, tagged with a `dataType` discriminator); it was previously dropped entirely because the export only wrote data-frame members.
- `observedData` and `parameterIdentification` are fully bridged (they had no Excel handling at all). PI uses three `taskId`-joined sheets (`PITasks`, `PIParameters`, `PIOutputMappings`) with the task `configuration` flattened to `config.*` columns.
- Scenario export is aligned with the canonical serializer: the steady-state time/unit is emitted only when a unit is declared (no more resurrecting a stale time with a fabricated `min`), `outputPathIds` are read from the named output-path vector, and `populationId` is emitted verbatim.

## Sheet-name sanitization

Parameter-set names that exceed Excel's 31-character sheet-name limit or contain forbidden characters (`/ \ * [ ] : ?`) are carried in a `Set Name` cell rather than relying on the sheet name. The sheet name is sanitized and de-duplicated for Excel, and the importer keys each set by the canonical name, so a long or awkward name no longer crashes the writer or orphans the scenario references that point at it.

## Folder, stale sheets, and overwrite guard

`exportProjectToExcel()` now writes into the project's configured `configurationsFolder` instead of a hardcoded `Configurations/`, and clears the section workbooks it manages before rewriting, so a section removed from the project does not survive as a stale file and reappear on the next import. `importProjectFromExcel()` gains an `overwrite` argument (default `FALSE`) and refuses to clobber an existing `Project.json` unless asked.

## Sync state machine

`projectStatus()` / `Project$sync()` now report the right status for each edit type: a freshly exported project reads as in sync; unsaved in-memory edits, an external JSON edit, and a differing Excel side-car each flip exactly their own flag, including the case where an in-memory edit and an external JSON edit happen together. A corrupt or unreadable Excel source is no longer misreported as in sync, and the temp-snapshot path is read from where the importer actually writes it.

## Tests and docs

The bridge test file asserts a per-section and whole-project round-trip fixed point, the sync state machine across all edit types, the non-default `configurationsFolder` round trip, the corrupt-Excel case, and the sheet-name-sanitization reference round trip; all fixtures are built through `loadProject()`. NEWS and the public roxygen are updated.

Closes #1049


Felixmil:
closed in favor of new pr stack starting at #1089

Issue #1088, opened 2026-07-02
Definitions-tree project format for 6.0.0 (design & decision record)


Felixmil:
> **Status (2026-07-17).** This issue is the **design and decision record** for the 6.0.0 project-model rework. It is not the live PR tracker: for the current state of the branches and stacked PRs, see the consolidation PR **#1017**.
>
> **Two things below have been superseded by later decisions and are kept as the rationale trail, not as a description of the shipped code:**
>
> 1. **Write-through authoring was reversed to an explicit-save lifecycle (see #1117).** The walkthrough below pitches "no save step" as the point: `add*` writing to disk immediately. The shipped model is the opposite. Memory is the source of truth for a loaded project; authoring edits stay in memory until `saveProject(project)` reconciles the on-disk tree. This buys a real "edit, then save" mental model, an undo (`reloadProject()`), a dirty-state concept, and in-place rollback. So wherever the text below says "no `saveProject()`", "write-through", or "the unsaved-changes concept is gone", read the inverse: there is an explicit save, and an unsaved-changes state is a first-class concept.
> 2. **The "entity-files" / "entity" vocabulary was renamed to "definitions" (see #1104, #1114).** On-disk the tree is `definitions/<kind>/*.json` (that part was always true); the code-level term "entity" (files, ids, lists) is now "definition" throughout. Read "entity-files" below as "the definitions tree".
>
> Everything else below (the definitions tree as one source of truth, one file per definition, Excel demoted to import/export, validation at authoring time) still holds. The go/no-go "ask" at the end was answered yes; the work is built and consolidating on #1017.

---

This proposes the project format we would ship in 6.0.0. It builds directly on the JSON-primary model we already approved (the branch that makes a project's JSON the single stored form and loads it into an in-memory `Project`), so this is not a fresh design; it is one increment on top of that approved baseline. The increment is small in concept and large in payoff: instead of one monolithic `Project.json`, a project is a tree of one small JSON file per definition, and authoring is write-through, so every edit is validated and persisted the moment you make it. There is no separate save step, and no way for the in-memory project, the files on disk, and any Excel export to silently disagree.

## What changes from the approved JSON-primary model

The audience for this issue already said yes to JSON-primary, so the important thing is what moves from there, not from the 5.x Excel-first workflow. Three concrete deltas:

| | JSON-primary (approved) | entity-files (proposed) |
|---|---|---|
[...]

PR #1090, opened 2026-07-02
Project class core: write-through bindings, syncStatus, snapshot verbs, container/metadata reshape

Felixmil:
Full context: https://github.com/esqLABS/esqlabsR/issues/1088

This PR reshapes the in-memory `Project` R6 class onto PR1's entity-file persistence engine. Every section accessor becomes a write-through active binding (reading returns the section; assigning validates structurally and persists to the tree, while a write on a clone stays in memory), the old `saveProject`/`modified`/`sync` surface is replaced by `syncStatus` plus the class-side `saveSnapshot`/`loadSnapshot` verbs, and the `Project.json` container is split into live working folders and an optional Excel-bridge block. It also adds project metadata (`name`, `description`), a configurable `definitionsFolder`, and `defaultSimulationRunOptions`, and carries the JSON (de)serializer that produces the inlined snapshot. It sits directly on PR1 because the class binds to PR1's tree engine.

## Project class

`R/project.R` gains per-section private stores backed by write-through active bindings for all nine sections, with the public accessors locked read-only so mutation routes through the authoring functions / `.setSection` rather than direct binding writes. The container path fields are split into live working folders (`modelFolder`, `dataFolder`, `outputFolder`, `populationsFolder`) versus the optional Excel-bridge block. `syncStatus` replaces the removed `modified`/`sync()`/`unsaved_changes` surface (an Excel-vs-entity-files comparison), and `name`, `description`, `definitionsFolder` (default `definitions/`), and `defaultSimulationRunOptions` are surfaced in `print(project)`. Because the whole file is owned here, its net delta also carries the parameter-set-unify binding change, the plots-trio to three-top-level-section binding split, the initial-conditions fields, and the `DefinitionList` wrap-on-read getter as riders.

## JSON snapshot serializer

`R/project-to-json.R` carries the snapshot / inlined-container (de)serializer. It emits the split container, the unified `parameterSets`, the three top-level plots sections, and the initial-conditions data.

## Tests and snapshots

`tests/testthat/test-project.R` and `tests/testthat/test-project-to-json.R` cover the reshaped class and serializer; `tests/testthat/_snaps/project.md` captures the print output. Note that some of the print snapshot reflects the `DefinitionList` and per-definition print methods authored in a later stacked PR (this PR only carries the wrap-on-read getter that calls them), so the captured output is the final rendering rather than code present on this branch.

## Stacking notes

Stacked on `pr01-persistence-engine`; this PR is intentionally not CI-green in isolation, since the section authoring bodies and the print method definitions land in later stacked PRs. `man/` and `NAMESPACE` regenerate in PR13 because this PR changes roxygen source.

## Summary by CodeRabbit

* **New Features**
  * Project export now supports a cleaner v2.0 JSON structure, including separate container metadata and optional Excel-related settings.
  * Section views are now read-only and display clearer counts and names when printed.

* **Bug Fixes**
  * Improved scenario and output-path serialization for more consistent saved projects.
  * Fixed handling of empty or missing sections so projects reload more reliably.
  * Updated clone and save behavior to avoid unintended writes between project copies.

### DIFF
--- a/R/project-to-json.R
+++ b/R/project-to-json.R
@@ -3,7 +3,7 @@
 # Inverse of `Project$.read_json()` (called by `Project$new()`). Walks a
 # `Project` and emits a v2.0 `Project.json` file. The contract is:
 #
-#   loadProject(path) |> saveProject(out)  produces a JSON file that, when
+#   loadProject(path) |> saveSnapshot(out)  produces a JSON file that, when
 #   re-loaded, yields a `Project` structurally identical to the first one.
 #
 # Layered to mirror the end-state shape: a top-level `.projectToJson()`
@@ -11,10 +11,10 @@
 # etc.). Today every per-section helper is essentially a passthrough — the
 # parser stores each section JSON-faithfully, so there is no transformation
 # to perform on the way out. The seams exist so future migrations
-# (relative-path resolution in `.filePathsToJson`, `outputPaths` →
-# `outputPathIds` rewriting in `.scenariosToJson`, unit conversions, plot
-# nesting, ...) can land in one section at a time without rearranging the
-# top-level call shape.
+# (relative-path resolution in `.filePathsToJson`, the named-vector
+# `outputPaths` record field rewritten to the `outputPaths` id array in
+# `.scenariosToJson`, unit conversions, plot nesting, ...) can land in one
+# section at a time without rearranging the top-level call shape.
 
 #' Internal: render a `Project` to a JSON-shaped R list in the v2.0 schema.
 #'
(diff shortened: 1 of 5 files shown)


Felixmil:
@coderabbitai review, take into account the stacked PR structure (some things may already be fixed in downstream PRs).

Felixmil:
@coderabbitai Confirmed and expected under the stacked-PR / tree-identity contract: `.projectSyncStatus()` is provided downstream by #1097 (pr09-excel-bridge) at `R/project-excel.R:777`, not on this branch. It carries through to the stack tip unchanged, so `syncStatus()` resolves once #1097 has landed. These PRs are not required to be CI-green individually; correctness is asserted on the sum at the stack tip. Leaving this here for confirmation at the tip. The two doc/comment nits you also raised (the stale `plotConfiguration` reference in the `.persistKindChanges` keyed fast-path comment, and the `@field excel` roxygen omitting `initialConditionsFile` while `.excelFilePathFields` lists it) are tracked and will be fixed in the review round 1 layer (#1104).

nautilus69 (approved):
according to what happen in #1104 Approved.

PR #1093, opened 2026-07-02
Entity authoring reshape: individuals/populations/applications/output-paths/observed-data + set*/vectorize

Felixmil:
Full context: #1088 

This PR brings the remaining entity types (individuals, populations, applications, output paths, and observed-data sources) onto the same homogeneous, id-first, write-through, vectorized authoring API that scenarios and parameter sets already use, and introduces the shared vectorization engine (`R/authoring-vectorize.R`) that every authoring surface relies on. It sits on `pr04-parameter-sets` in the stack, referencing the unified `parameterSets` section as a foreign-key target when authoring individuals.

## Homogeneous entity authoring

Each of the five entity types gets id-first `add*`/`set*`/`remove*` bodies with consistent semantics: the leading argument is the id (or a vector of ids), writes go through the tree, and structural validation fails fast. `setIndividual`/`setPopulation`/`setOutputPath` support partial updates with `.unset` semantics (a `NULL` clears an optional field). Individuals carry a `parameterSets` foreign-key check against the unified `parameterSets` section introduced upstream. Entities are classed at build time so downstream layers can dispatch on them.

## Shared vectorization engine

`R/authoring-vectorize.R` is a single internal engine (`@keywords internal`, not exported) used by all entity types plus the earlier sections. A vector of ids sets N entities in one write-through: scalar fields must be length 1 or N and recycle/align to N, whole-vector fields (such as an individual's `parameterSets`) apply to all N, and the operation is all-or-nothing (every entity is validated before any write). Abort messages name the offending argument.

## Observed data

Observed-data source CRUD (`addObservedData`/`removeObservedData`) authors and persists the four source types through the same write-through path; the on-the-fly source resolution logic is existing and reused.

## Tests and snapshots

New and expanded tests cover the reshaped authoring for each type plus the vectorization engine (`test-authoring-vectorize.R`). Snapshots reflect final content, so they already include per-definition print output (the print methods themselves land downstream) and the unified `parameterSets` foreign-key references from the predecessor PR.

## Stacking notes

Stacked on `pr04-parameter-sets`; this PR is intentionally not CI-green in isolation, since per-definition print methods (downstream, PR8) and validation consumers are not present on this branch. `man/` and `NAMESPACE` regenerate in the docs-regeneration PR (PR13), so no generated docs are touched here.

## Summary by CodeRabbit

* **New Features**
  * Added printable summaries for applications, individuals, populations, and observed data sources.
  * Enabled batch editing for applications, individuals, populations, observed data, and output paths.
  * Added a new way to update existing output paths.

* **Bug Fixes**
  * Improved validation and error messages for invalid, missing, or duplicate IDs.
  * Better handling of referenced items during removal, with warnings for potential dangling references.
  * More consistent saving behavior after edits, including atomic updates for batch operations.

### DIFF
--- a/R/applications.R
+++ b/R/applications.R
@@ -1,8 +1,198 @@
 # Applications: parse + mutation API for the `applications` and
-# `applicationParameterSets` sections.
+# `parameterSets` references on applications.
 
-# Parse ----
+# Print ----
+
+#' @exportS3Method
+#' @noRd
+print.Application <- function(x, ...) {
+  ospsuite.utils::ospPrintClass(x)
+  ospsuite.utils::ospPrintItems(
+    list(
+      "Parameter Sets" = paste(x$parameterSets, collapse = ", ")
+    ),
+    print_empty = TRUE
+  )
+  invisible(x)
+}
+
+# Public CRUD: applications ----
+
+#' Add one or more application protocols to a Project
+#'
+#' Add protocols to `project$applications`, vectorizing over a vector of ids
+#' (see the recycling rule under Details). `parameterSets` is
+#' vector-valued-per-entity: it is applied whole to every protocol; to give a
+#' different set per protocol, pass a list of the same length as `id` (one
+#' character vector per protocol).
+#'
+#' @inherit vectorizedAuthoring details
+#'
+#' @param project A `Project` object.
+#' @param id Character vector of unique protocol ids (the number of protocols
+#'   to add). Each is canonicalized to a safe, lowercase id (a warning names
+#'   the result if it changed).
+#' @param parameterSets Optional character vector of set ids referencing
+#'   `project$parameterSets`, applied whole to every protocol. Defaults to
+#'   `NULL`. Use a list of the same length as `id` for a per-protocol set.
+#' @returns The `project` object, invisibly.
+#' @export
+#' @family application
+addApplication <- function(project, id, parameterSets = NULL) {
+  validateIsOfType(project, "Project")
+  .assertIdVector(id)
+  id <- .canonicalizeId(id)
+  n <- length(id)
+  perId <- .wholeField(parameterSets, n)
+
+  clash <- intersect(id, names(project$applications))
+  if (length(clash) > 0L) {
+    cli::cli_abort("application {.val {clash}} already exists")
+  }
+  call <- rlang::current_env()
+  apps <- .collectCanonicalizedRefs(lapply(seq_len(n), function(i) {
+    .buildApplicationEntry(project, perId[[i]], call = call)
+  }))
+
+  applications <- project$.getSection("applications") %||% list()
+  for (i in seq_len(n)) {
+    applications[[id[[i]]]] <- apps[[i]]
+  }
+  project$.setSection("applications", applications)
+  invisible(project)
+}
+
+#' Remove one or more application protocols from a Project
+#'
+#' Drop the protocols with matching ids in one write-through. Warns (and
+#' skips) any id not present, and warns when a removed protocol is still
+#' referenced.
+#'
+#' @param project A `Project` object.
+#' @param id Character vector of application ids to remove. Each is
+#'   canonicalized the same way [addApplication()] canonicalizes it.
+#' @returns The `project` object, invisibly.
+#' @export
+#' @family application
+removeApplication <- function(project, id) {
+  validateIsOfType(project, "Project")
+  .assertIdVector(id)
+  id <- .canonicalizeId(id)
+
+  missingIds <- setdiff(id, names(project$applications))
+  if (length(missingIds) > 0L) {
+    cli::cli_warn("application {.val {missingIds}} not found; no-op.")
+  }
+  toRemove <- intersect(id, names(project$applications))
+  if (length(toRemove) == 0L) {
+    return(invisible(project))
+  }
+  for (one in toRemove) {
+    .warnIfReferenced(project, "application", one)
+  }
+  applications <- project$.getSection("applications")
+  applications[toRemove] <- NULL
+  project$.setSection("applications", applications)
+  invisible(project)
+}
+
+#' Replace the parameter-set references on one or more applications
+#'
+#' @inherit vectorizedAuthoring details
+#'
+#' @param project A `Project` object.
+#' @param id Character vector of application ids. Each is canonicalized the
+#'   same way [addApplication()] canonicalizes it.
+#' @param parameterSets Character vector of set ids (from
+#'   `project$parameterSets`), applied whole to every application; use
+#'   `character(0)` to clear. To set a different list per application, pass a
+#'   list of the same length as `id` (one character vector per application).
+#' @returns The `project` object, invisibly.
+#' @export
+#' @family application
+setApplicationParameterSets <- function(
+  project,
+  id,
+  parameterSets
+) {
+  validateIsOfType(project, "Project")
+  .assertIdVector(id)
+  id <- .canonicalizeId(id)
+  n <- length(id)
+  missingIds <- setdiff(id, names(project$applications))
+  if (length(missingIds) > 0L) {
+    cli::cli_abort("application {.val {missingIds}} not found")
+  }
+  perId <- .wholeField(parameterSets, n)
+
+  call <- rlang::current_env()
+  resolved <- .collectCanonicalizedRefs(lapply(seq_len(n), function(i) {
+    sets <- perId[[i]]
+    if (!is.character(sets)) {
+      cli::cli_abort(
+        "{.arg parameterSets} must be a character vector",
+        call = call
+      )
+    }
+    sets <- .canonicalizeIdRef(sets)
+    bad <- setdiff(sets, names(project$parameterSets %||% list()))
+    if (length(bad) > 0L) {
+      cli::cli_abort(
+        c(
+          "{.arg parameterSets} references undefined parameter sets:",
+          "x" = "{.val {bad}}"
+        ),
+        call = call
+      )
+    }
+    sets
+  }))
+
+  applications <- project$.getSection("applications")
+  for (i in seq_len(n)) {
+    applications[[id[[i]]]]$parameterSets <- resolved[[i]]
+  }
+  project$.setSection("applications", applications)
+  invisible(project)
+}
+
+# Internal helpers ----
+
+# Build one classed `Application` entry from its (optional) parameterSets
+# references, validating the references resolve. Aborts on a problem.
 #
+# @keywords internal
+# @noRd
+.buildApplicationEntry <- function(
+  project,
+  parameterSets,
+  call = rlang::caller_env()
+) {
+  app <- list()
+  if (!is.null(parameterSets)) {
+    if (!is.character(parameterSets)) {
+      cli::cli_abort(
+        "{.arg parameterSets} must be a character vector of set ids",
+        call = call
+      )
+    }
+    parameterSets <- .canonicalizeIdRef(parameterSets)
+    bad <- setdiff(parameterSets, names(project$parameterSets %||% list()))
+    if (length(bad) > 0L) {
+      cli::cli_abort(
+        c(
+          "{.arg parameterSets} references undefined parameter sets:",
+          "x" = "{.val {bad}}"
+        ),
+        call = call
+      )
+    }
+    app$parameterSets <- parameterSets
+  }
+  class(app) <- c("Application", "list")
+  app
+}
+
 # Parse the `applications` JSON object. Each entry is stamped with
 # `class = c("Application", "list")`. The current schema stores
 # applications as a map of name -> object containing only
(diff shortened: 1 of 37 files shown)


Felixmil:
@coderabbitai review - Keep in mind that this PR is part of several stacked PR: Some of your findings may already be fixed in downstream PRs.

R/individuals.R
```diff
@@ -190,221 +250,261 @@ addIndividual <- function(project, individualId, species, ...) {
...
+        ),
+        call = call
+      )
+    }
+  }
+
+  entry <- project$individuals[[id]]
+  for (field in names(fields)) {
+    if (field %in% c("weight", "height", "age")) {
+      entry[[field]] <- as.double(fields[[field]])
```
Felixmil, line 479: **Blocking** — `setIndividual(project, id, weight = NULL)` does not clear `weight`; it sets it to `numeric(0)`.

```r
entry <- project$individuals[[id]]
for (field in names(fields)) {
  if (field %in% c("weight", "height", "age")) {
    entry[[field]] <- as.double(fields[[field]])
```

`as.double(NULL)` returns `numeric(0)`, a non-`NULL` zero-length vector, so `entry[["weight"]] <- numeric(0)` **keeps the key** instead of deleting it (contrast the `else` branch two lines down, where `entry[[field]] <- fields[[field]]` correctly deletes the key when the value is `NULL`). The add-path guard three lines above this function's sibling, `.buildIndividualEntry` (R/individuals.R:266-270), explicitly does `if (!is.null(fields[[field]])) entry[[field]] <- as.double(...)` — that guard was dropped here.

Consequences: `.individualsToJson()` serializes this as `"weight": []` on disk (not absent, not `null`), and a later `validateProject()` call runs `is.na(weight)`/comparisons on that zero-length vector inside `if()`/`&&` — which errors outright under R ≥ 4.3 ("argument is of length zero"). This directly contradicts the PR description's own stated contract: *"`setIndividual`/`setPopulation`/`setOutputPath` support partial updates with `.unset` semantics (a `NULL` clears an optional field)"*.

No test in `tests/testthat/test-individuals.R` exercises `setIndividual(..., weight = NULL)` or the height/age equivalents — this would ship silently.

Fix: mirror the add-path guard, e.g. `if (!is.null(fields[[field]])) entry[[field]] <- as.double(fields[[field]]) else entry[[field]] <- NULL`.

---

R/populations.R
```diff
@@ -361,87 +449,207 @@ addPopulation <- function(
...
+        "{.arg numberOfIndividuals} must be a positive number",
+        call = call
+      )
+    }
+  }
+
+  entry <- project$populations[[id]]
+  for (field in names(fields)) {
+    if (field %in% numericFields) {
+      entry[[field]] <- as.double(fields[[field]])
```
Felixmil, line 648: **Blocking** — identical bug to `R/individuals.R:479`: `setPopulation(project, id, ageMin = NULL)` (or `weightMin`, `weightMax`, `proportionOfFemales`, any `numericFields` entry) sets the field to `numeric(0)` instead of clearing it.

```r
entry <- project$populations[[id]]
for (field in names(fields)) {
  if (field %in% numericFields) {
    entry[[field]] <- as.double(fields[[field]])
```

Same root cause as individuals.R: `as.double(NULL)` is `numeric(0)`, not `NULL`, so the key survives with an empty value instead of being deleted — while string fields (`gender`, `weightUnit`, `population`, …) go through the `else` branch and clear correctly. `.buildPopulationEntry` has the same missing-guard-on-set-vs-add asymmetry. No test in `test-populations.R` passes `NULL` for a numeric field to `setPopulation()`.

Given this is the exact same defect shape in two sibling files, worth fixing once in a shared helper (e.g. in `R/authoring-vectorize.R`) rather than patching both call sites independently — the duplication is what let it regress in the first place.

---

R/observed-data.R
```diff
@@ -247,10 +266,13 @@ addObservedData <- function(project, entry) {
       )
     }
     state$.programmaticDataSets[[name]] <- entry
```
Felixmil, line 268: **Blocking** — runtime state is mutated before the fallible persist call, so a write failure leaves an orphaned in-memory `DataSet` with no corresponding disk entry.

```r
state$.programmaticDataSets[[name]] <- entry
sentinel <- .asObservedDataSource(list(type = "programmatic", name = name))
project$.setSection(
  "observedData",
  c(project$.getSection("observedData"), list(sentinel))
)
```

`state$.programmaticDataSets[[name]] <- entry` (line 268) commits to the runtime store *before* `project$.setSection(...)` (272-275) is even attempted. `.setSection` can genuinely throw here — `.serializeObservedDataSet` aborts on an id collision, and the new test at `test-observed-data.R:274-294` ("declarations sharing a basename fail the write-through") proves it. The pre-check at line 263 (`name %in% existingNames`) only catches a collision against already-loaded *names*, not a basename collision against an existing file-based entry's on-disk id, so the throw path is reachable in normal use, not just in theory.

If `.setSection` throws here, `state$.programmaticDataSets` keeps the entry while `observedData` on disk has no sentinel — the exact mirror image of the "sentinel without dataset" case this same PR deliberately tests at `test-observed-data.R:71-91`. This breaks the core invariant this whole PR is built on (per #1088: *"the in-memory object and the files cannot disagree; there is no unsaved state to lose or to reconcile"*).

Fix: build the sentinel and call `.setSection(...)` first; only assign into `state$.programmaticDataSets[[name]]` after the write succeeds.

---

R/observed-data.R
```diff
@@ -302,75 +330,98 @@ addObservedData <- function(project, entry) {
...
+  }
+  if (length(dropIdx) == 0L && length(programmaticNames) == 0L) {
     return(invisible(project))
   }
 
-  .warnIfObservedDataReferenced(project, name)
-  project$observedData <- project$observedData[-matchIdx[[1]]]
+  for (name in programmaticNames) {
+    state$.programmaticDataSets[[name]] <- NULL
+  }
```
Felixmil, line 406: **Blocking** — same mutate-before-persist ordering bug as the `addObservedData` DataSet branch, on the removal path:

```r
for (name in programmaticNames) {
  state$.programmaticDataSets[[name]] <- NULL
}
if (length(dropIdx) > 0L) {
  observedData <- observedData[-unique(dropIdx)]
}
project$.setSection("observedData", observedData)
```

The runtime store is cleared (404-406) before `.setSection` (410) is attempted. If the write throws, disk still has the sentinel but the in-process `DataSet` is already gone — memory and disk disagree in the opposite direction from the add-path bug, but it's the same defect: mutate the fallible-persist-guarded state outside of the guard.

Fix: call `project$.setSection("observedData", observedData)` first, then clear `state$.programmaticDataSets` for `programmaticNames` only after the write succeeds.

---

R/output-paths.R
```diff
@@ -5,6 +5,151 @@
...
+  validateIsOfType(project, "Project")
+  # Route the id-vector check through the shared helper every sibling add* uses,
+  # then canonicalize (which aborts on an in-batch collision, so no separate
+  # duplicate guard is needed).
+  .assertIdVector(id)
+  id <- .canonicalizeId(id)
+
+  if (
+    !is.character(path) ||
+      !(length(path) == 1L || length(path) == length(id))
```
Felixmil, line 37: **Required Changes** — `addOutputPath` never rejects an empty or `NA` path, unlike its sibling `setOutputPath` (R/output-paths.R:133-140, which explicitly aborts on `is.na(one) || nchar(one) == 0`).

```r
if (
  !is.character(path) ||
    !(length(path) == 1L || length(path) == length(id))
) {
  cli::cli_abort(...)
}
```

`addOutputPath(project, "X", "")` or `addOutputPath(project, "X", NA_character_)` passes this check and writes straight through via `.setSection` at line 57. The invalid value is only caught later, by `.validateOutputPaths()` on the next `validateProject()` call — a real window where a structurally-invalid path persists to disk from `add*`. This is exactly the "addX errors on bad input" contract the PR is meant to uphold everywhere else (per #1088: *"Validation runs at authoring time, so a bad reference … is caught as you add it, not at run time"*), and it's a known, previously-flagged gap for this exact function (issue #1055: *"`path` values escape the NA/empty check applied to ids"*) that this rewrite had the opportunity to close but didn't.

`tests/testthat/test-output-paths.R:89-98` has "setOutputPath rejects an empty path" but no equivalent test for `addOutputPath`.

Fix: apply the same `is.na`/`nchar == 0` check used in `setOutputPath` to every element of `path` before persisting.

---

R/applications.R
```diff
@@ -1,262 +1,219 @@
...
+  }
+  perId <- .wholeField(parameterSets, n)
 
-# Public CRUD: applicationParameterSets ----
+  call <- rlang::current_env()
+  resolved <- .collectCanonicalizedRefs(lapply(seq_len(n), function(i) {
+    sets <- perId[[i]]
+    if (!is.character(sets)) {
+      cli::cli_abort(
+        "{.arg parameterSets} must be a character vector",
```
Felixmil, line 133: **Strong Suggestion** — Duplicated Code: `setApplicationParameterSets` re-implements the same "must be a character vector" + "undefined parameter sets" checks as `.buildApplicationEntry` (lines 172-189), and the wording has already drifted apart:

- here: `"{.arg parameterSets} must be a character vector"` (line 133)
- `.buildApplicationEntry`, line 175: `"{.arg parameterSets} must be a character vector of set ids"`

The "references undefined parameter sets" block is byte-identical in both, so this isn't a case where the two checks are meant to differ — it's the same validation copy-pasted and half-updated. The same add/set pair in `R/individuals.R` (lines 274 and 456) keeps this wording in sync today, which shows the convention is normally hand-maintained and just slipped here. Given this pattern repeats per entity type with no shared helper, it's a real, escalating maintenance hazard rather than a one-off nitpick.

Suggested fix: extract a shared `.resolveParameterSetRefs(project, sets, call)` and call it from both `.buildApplicationEntry` and `setApplicationParameterSets`.

---

R/authoring-vectorize.R
```diff
@@ -0,0 +1,189 @@
...
+
+# Resolve a whole-vector-per-entity field to N per-entity values. A length-N
+# list aligns by position (one element per entity); any other value (a scalar,
+# an atomic vector, or `NULL`) is applied verbatim to every entity. This is the
+# "applied whole, never split positionally" rule for vector-valued fields.
+#
+# @keywords internal
+# @noRd
+.wholeField <- function(value, n) {
+  if (is.list(value) && length(value) == n) {
```
Felixmil, line 134: **Strong Suggestion** — `.wholeField`'s "is it per-entity or applied-to-all" decision hinges entirely on a length coincidence:

```r
.wholeField <- function(value, n) {
  if (is.list(value) && length(value) == n) {
    return(value)
  }
  rep(list(value), n)
}
```

If a caller means "apply this same set of two parameter-set ids to every entity" and expresses it as `parameterSets = list("global", "aciclovir")` against `id = c("p1", "p2")` (a natural mistake — using `list()` to bundle two strings instead of `c()`), and `n` happens to equal `length(list)`, this silently reinterprets it as *per-entity* assignment: `p1` gets only `"global"`, `p2` gets only `"aciclovir"`, with no error, since each element is independently a valid length-1 character vector and passes `.buildApplicationEntry`'s `is.character` check. This is the documented behavior (`R/applications.R:107-109`: "pass a list of the same length as `id`"), so it's arguably by design — but nothing in the API distinguishes an intentional per-entity list from this accidental one, and `test-applications.R` has no test that exercises the ambiguity. Worth at least a doc callout or an explicit sentinel (e.g. requiring `I(list(...))` or similar) for "apply this list to all," since "same length as `id`" and "coincidentally the same length as `id`" are indistinguishable today.

---

Felixmil (commented):

## Summary

This is a large, well-organized reshape (5 entity types + a new shared vectorization engine), and the "all-or-nothing" vectorized-write guarantee is genuinely well implemented everywhere I traced it: every `add*`/`set*`/`remove*` path builds/validates the full per-entity batch before a single `.setSection()` write, and `.setSection()` itself persists to disk before mutating in-memory state, so a serialization failure can't desync `project$<section>` from disk. `.recycleField`'s length handling is also strict — no silent partial recycling.

Two real defects survived that discipline, though, both directly touching the two guarantees this PR (and the parent RFC, #1088) is explicitly built around:

- **`.unset` semantics are broken for every numeric field on `setIndividual`/`setPopulation`.** `NULL` is supposed to clear an optional field (this is the PR description's own stated contract) but `as.double(NULL)` → `numeric(0)` keeps the key instead of deleting it, writing `[]` to disk and likely crashing a later `validateProject()` call. Untested.
- **The `observedData` "programmatic DataSet" side-channel isn't covered by the write-through guarantee.** Both `addObservedData`'s DataSet branch and `removeObservedData` mutate `state$.programmaticDataSets` outside of, and on the wrong side of, the fallible `.setSection()` call — the exact "orphan sentinel" desync this PR's own tests prove is otherwise guarded against.

A third, lower-severity gap: `addOutputPath` is missing the empty/NA path validation its sibling `setOutputPath` has, letting a structurally-invalid path persist until the next full `validateProject()` — a known, previously-flagged gap (#1055) this rewrite had the chance to close.

No `CLAUDE.md` or other documented coding standard exists in this repo, so the Standards axis here is judged against general R/code-smell heuristics rather than a house style guide.

(Note: GitHub does not allow requesting changes on one's own pull request, so this review is submitted as a comment; treat the verdict below as "Request Changes" in substance.)

## Critical Issues (Blocking)

1. `R/individuals.R:479` — `setIndividual(..., weight = NULL)` sets `weight` to `numeric(0)` instead of clearing it (same bug for `height`, `age`).
2. `R/populations.R:648` — identical bug for every `numericFields` entry (`ageMin`, `weightMin`, `proportionOfFemales`, …) via `setPopulation()`.
3. `R/observed-data.R:268` — `addObservedData()`'s DataSet branch mutates `state$.programmaticDataSets` before the fallible `.setSection()` call; a write failure (proven reachable by this PR's own collision test) orphans the runtime entry.
4. `R/observed-data.R:404-406` — `removeObservedData()` clears `state$.programmaticDataSets` before `.setSection()`; the same desync risk in the removal direction.

See inline comments for code, evidence, and fixes.

## Required Changes

5. `R/output-paths.R:35-44` — `addOutputPath` doesn't reject empty-string/`NA` paths the way `setOutputPath` does, letting invalid state persist until the next `validateProject()`.

## Spec Compliance

Primary spec source: issue #1088 ("Entity-files project format for 6.0.0"), explicitly referenced in the PR body ("Full context: #1088"), plus the PR body's own stated contract for this diff.

- **Implemented but wrong**: PR body states *"`setIndividual`/`setPopulation`/`setOutputPath` support partial updates with `.unset` semantics (a `NULL` clears an optional field)"* — true for string fields, false for every numeric field on individuals/populations (Blocking #1, #2 above).
- **Implemented but wrong**: #1088 states *"Because authoring is write-through, the in-memory object and the files cannot disagree; there is no unsaved state to lose or to reconcile."* — true everywhere except the `observedData` programmatic-DataSet side-channel (Blocking #3, #4).
- **Implemented but wrong**: #1088 states *"Validation runs at authoring time, so a bad reference … is caught as you add it, not at run time."* — `addOutputPath` violates this for empty/NA paths (Required Change #5).
- No scope creep found; the diff stays within the stated reshape (individuals/populations/applications/output-paths/observed-data + the shared vectorization engine).

## Suggestions

6. `R/applications.R:113-194` — duplicated, already-drifted `parameterSets` validation between `.buildApplicationEntry` and `setApplicationParameterSets`; extract a shared helper (this pattern repeats across entity types with no shared validator).
7. `R/authoring-vectorize.R:133-138` — `.wholeField`'s list-vs-per-entity classification hinges on a length coincidence; a `list()` of N strings meant to apply to N entities as a whole is silently reinterpreted as one string per entity. Likely "as designed" per the docstring, but untested and easy to trigger by accident.
8. `R/observed-data.R:365-377` — the removal-resolution loop queues a name into `programmaticNames` without confirming a matching sentinel exists in `observedData`; low-severity on its own (only surfaces once already desynced by #3), but worth tightening while fixing #3/#4.

Remaining items are minor. The vectorization engine itself, the all-or-nothing guarantee, and the write-then-commit ordering in `Project$.setSection()` are all solid — the defects above are narrow and mechanical to fix, not structural rework.

## Verdict
Request Changes (submitted as Comment — see note above)

---

R/individuals.R
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104) rather than here, to keep this slice byte-identical to `entity-files-scenarios`. The set path now routes numeric fields through a shared `NULL`-aware coercion helper, so `setIndividual(..., weight = NULL)` deletes the key instead of storing `numeric(0)`. Commit `ce95ace1`, with a test asserting key absence.

R/populations.R
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104). The same shared `NULL`-aware coercion helper is applied to `.setOnePopulation()`, so a `NULL` numeric field (e.g. `ageMin`) clears rather than becoming `numeric(0)`. Commit `ce95ace1`, with a `setPopulation(..., ageMin = NULL)` test.

R/observed-data.R
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104). `addObservedData()` now writes the sentinel through `project$.setSection("observedData", ...)` first and assigns into `state$.programmaticDataSets` only after the write succeeds, so a failed write no longer leaves an orphaned in-memory `DataSet`. Commit `0dc80b8a`, with a test that the runtime store is untouched when the write aborts.

R/observed-data.R
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104). `removeObservedData()` now calls `.setSection()` first and clears `state$.programmaticDataSets` only on success, so a failed write no longer drops the in-memory `DataSet` while the disk sentinel remains. Commit `0dc80b8a`, with a mirror test.

R/output-paths.R
Felixmil: Confirmed valid on this slice. It is already resolved on `pr15-review-round-1` / `entity-files-scenarios`: `addOutputPath()` there rejects empty and `NA` paths (`anyNA(path) || any(nchar(path) == 0)`), matching `setOutputPath()`, and a dedicated `addOutputPath rejects an empty or NA path` test is present. No separate commit was needed; this slice inherits the fix when the stack lands.

R/applications.R
Felixmil: Confirmed and addressed on `pr15-review-round-1` (PR #1104). The duplicated "must be a character vector" + undefined-set checks in `setApplicationParameterSets()` and `.buildApplicationEntry()` are consolidated behind one behavior-preserving helper (`.resolveParameterSetRefs()`), so both paths now share a single canonical message. Commit `5e955f24`.

R/authoring-vectorize.R
Felixmil: Acknowledged. This one is deferred pending an API-stance decision: the choice between a documentation-only callout and an explicit opt-in marker (which would change the locked recycling semantics other authoring functions depend on) is a design call, so no code change was made yet. Tracked for a follow-up decision on `pr15-review-round-1`.

R/individuals.R
```diff
@@ -190,221 +250,261 @@ addIndividual <- function(project, individualId, species, ...) {
...
+    } else {
+      entry[[field]] <- fields[[field]]
+    }
+  }
+  class(entry) <- c("Individual", "list")
+  entry
 }
```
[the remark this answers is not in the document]
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104): `setIndividual(..., weight = NULL)` now clears the field via a shared `NULL`-aware numeric coercion helper instead of storing `numeric(0)`. Commit `ce95ace1`, with a key-absence test. Landed there rather than on this slice to preserve the tree-identity contract.

R/populations.R
```diff
@@ -361,87 +449,207 @@ addPopulation <- function(
...
+    } else {
+      entry[[field]] <- fields[[field]]
+    }
+  }
+  class(entry) <- c("Population", "list")
+  entry
+}
```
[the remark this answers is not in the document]
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104): the same shared helper clears `NULL` numeric fields in `.setOnePopulation()`. Commit `ce95ace1`, with a `setPopulation(..., ageMin = NULL)` test.

R/observed-data.R
```diff
@@ -302,75 +330,98 @@ addObservedData <- function(project, entry) {
...
+  if (length(dropIdx) > 0L) {
+    observedData <- observedData[-unique(dropIdx)]
+  }
+  project$.setSection("observedData", observedData)
   state$.observedDataNamesCache <- NULL
   invisible(project)
 }
```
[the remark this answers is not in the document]
Felixmil: Confirmed and fixed on `pr15-review-round-1` (PR #1104): `removeObservedData()` now performs the `.setSection()` write-through before clearing `state$.programmaticDataSets`. Commit `0dc80b8a`. On the secondary note (a programmatic name queued for removal without a matching sentinel): with the add path also reordered so memory and disk stay in sync, that state can no longer arise, so the reordering resolves it without a separate guard.

R/output-paths.R
```diff
@@ -5,6 +5,151 @@
...
+  # Recycle a single path to every id (the scalar-per-entity rule).
+  if (length(path) == 1L) {
+    path <- rep(path, length(id))
+  }
+  newPaths <- as.list(path)
+  names(newPaths) <- id
+  outputPaths <- c(project$.getSection("outputPaths"), newPaths)
+  project$.setSection("outputPaths", outputPaths)
+  invisible(project)
+}
```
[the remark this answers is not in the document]
Felixmil: Confirmed valid; already resolved on `pr15-review-round-1` / `entity-files-scenarios`, where `addOutputPath()` rejects empty/`NA` paths like `setOutputPath()` and has a dedicated test. No separate commit needed; this slice inherits it.

tests/testthat/test-populations.R
```diff
@@ -226,3 +226,222 @@ test_that("extendPopulationFromXLS throws an error if specified sheet is empty o
...
+    expect_equal(after[[f]], before[[f]])
+  }
+})
+
+test_that("setPopulation clears validatedSinceMutation", {
+  project <- testProject()
+  project$.markValidated()
+  setPopulation(project, "testpopulation", numberOfIndividuals = 10)
+  expect_false(project$validatedSinceMutation)
+})
```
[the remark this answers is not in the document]
Felixmil: Done on `pr15-review-round-1` (PR #1104): a `setPopulation(..., ageMin = NULL)` regression test now asserts the field is cleared (key absent) while the other population fields are unchanged, locking in the `NULL`-clears behavior. Commit `ce95ace1`.

nautilus69 (changes requested):
Some of these are marked resolved but I couldnt find the fix in the upstream, the order has no meaning. 


[1] setIndividual persists numeric(0) instead of clearing a field — R/individuals.R

  .setOneIndividual() (L408) coerces numeric fields unconditionally in the SET apply loop (L476–483):
  ## R/individuals.R:479
  entry[[field]] <- as.double(fields[[field]])   # as.double(NULL) -> numeric(0)
  setIndividual(project, id, weight = NULL) should clear the field but stores a zero-length numeric. 

[4] setPopulation persists numeric(0) instead of clearing a field — R/populations.R

  Same bug in .setOnePopulation() (L604), SET apply loop (L645–653):
  ## R/populations.R:648
  entry[[field]] <- as.double(fields[[field]])   # numeric(0) on NULL

[2] observedData write-through is non-atomic → orphan sentinel/dataset — R/observed-data.R

  The runtime store is mutated before the fail-able .setSection, so a serialize/write failure leaves the JSON section and the .programmaticDataSets store inconsistent.

Felixmil:
Thanks for flagging these. All three are already fixed, but on the **review round 1** layer at the top of the stack (#1104, `pr15-review-round-1`), not in this PR. That is deliberate: per the migration tree-identity contract, PRs 1-13 must sum to the same tree, so a fix cannot land inside its originating PR (it would break that contract) and instead lands as a post-migration layer, exactly as the polish layer does. So on this branch you correctly see the unfixed code; the resolution lives downstream.

Where each is fixed on `pr15-review-round-1`:

**[1] `setIndividual(weight = NULL)` storing `numeric(0)`** and **[4] the same on `setPopulation()`** — fixed in commit ce95ace110bf9c8a2bd25e1fb02a4c3582701a2b. The set-path apply loops now go through a shared `.coerceNumericField()` helper that returns `NULL` for a `NULL` input (clearing the field) instead of `as.double(NULL)` → `numeric(0)`:
- helper: https://github.com/esqLABS/esqlabsR/blob/6cfad2457270e3821dd77b2794c0aef0afe160d6/R/authoring-vectorize.R#L222-L227
- individuals set apply: https://github.com/esqLABS/esqlabsR/blob/6cfad2457270e3821dd77b2794c0aef0afe160d6/R/individuals.R#L525-L531
- populations set apply: https://github.com/esqLABS/esqlabsR/blob/6cfad2457270e3821dd77b2794c0aef0afe160d6/R/populations.R#L682-L688

**[2] non-atomic `observedData` write-through** — fixed in commit 0dc80b8a0bf4ecd93173233b9c8c7a6dae3cee13. `addObservedData()` now calls the fail-able `.setSection()` **first** and mutates the `.programmaticDataSets` runtime store only after the write succeeds, so an aborted write no longer leaves the store and the JSON sentinel disagreeing:
- https://github.com/esqLABS/esqlabsR/blob/6cfad2457270e3821dd77b2794c0aef0afe160d6/R/observed-data.R#L300-L311

You were right that they weren't reachable from the upstream you were looking at: the canonical branch had drifted and hadn't been advanced to carry the round-1 fixes. We're collapsing that so the stack tip (#1104) is the single source of truth, which removes the drift.

nautilus69 (approved):

PR #1094, opened 2026-07-02
Plots restructure: three per-entity kinds + keyed lists + create-plots

Felixmil:
BEWARE: This PR is part of a stacked PR merge train. It was split by concept and was not meant to pass CI. Also, when reviewing, make sure to double check findings against the downstream PRs. More at #1088 

This PR reshapes the plots subsystem for the entity-files project model. It splits the old composite plots kind into three per-entity tree kinds (`data-combined`, `plots`, `plot-grids`), stores plot and grid entries as keyed lists rather than NA-padded data frames, rewrites `createPlots` to iterate those keyed lists directly, rationalizes the id fields (`plotId`, `plotGridId`, `dataCombinedId`), vectorizes the plot mutators, and adds `createPlots(plotIds=)` for rendering standalone plots. It is the plots *logic*; the three top-level section bindings that back it (`project$dataCombined`, `project$plots`, `project$plotGrids`) are defined in `project.R` and land in an earlier PR of the stack.

## Three per-entity plots kinds

The single composite plots kind is split into three per-entity tree kinds, each keyed by its own id: `data-combined/` keyed by `dataCombinedId`, `plots/` keyed by `plotId`, and `plot-grids/` keyed by `plotGridId`. In memory each is a keyed list of entries rather than a data frame, so there is no NA padding anywhere in the plots path and each entry stands alone.

Id fields are rationalized alongside the split: `plotId` is unchanged, `plotGridId` replaces the former `plotGrids$name`, `dataCombinedId` replaces the former `dataCombined` `name`, and the cross-reference `plotConfiguration$dataCombinedName` becomes `dataCombinedId`. Plot reference args and keys drop their old suffixes.

## createPlots runtime

`createPlots` is rewritten to iterate the keyed lists directly (`.createConfigurationFromEntry`), removing the row-wise `apply(df, 1, ...)` walk over a data frame. It gains a `plotIds` argument that renders standalone individual plots and unions them with the grid results; `stopIfNotFound` now also covers unknown `plotIds`. This file also carries two riders kept here rather than split out: the `createPlots` / `createDataCombined` argument rename to `scenarioResults`, and the cli-assertion change.

## Data-combined

`R/data-combined.R` moves data-combined parse and serialize onto the keyed kind, and carries the `scenarioResults` rename and the suffixless rename of reference keys.

## Plot mutators

The add and remove plot mutators are vectorized: they operate over a vector of ids in a single write-through instead of one call per id.

## Tests and snapshots

`test-plots.R`, `test-create-plots.R`, `test-data-combined.R` and their `_snaps/*.md` are updated to the new keyed-list shape, the rationalized ids, and the vectorized mutators. The snapshots reflect final content, which includes per-definition print output introduced in a later PR of the stack and the suffixless-rename keys.

## Stacking notes

Stacked on `pr05-entity-authoring`. The three top-level section bindings (`project$dataCombined`, `project$plots`, `project$plotGrids`) and their private stores live in `project.R` and land in an earlier PR of the stack; this PR is the logic that reads them. This PR is intentionally not CI-green in isolation. `man/` and `NAMESPACE` are regenerated in the docs-regeneration PR at the tail of the stack.

## Summary by CodeRabbit

* **New Features**
  * Plot creation now supports selecting both plot grids and standalone plots directly, with clearer handling of requested outputs.
  * Plot and data-combination setup now works with more flexible, named configuration entries.

* **Bug Fixes**
  * Improved validation for missing or unknown plots, grids, and data combinations.
  * Fixed handling of plot IDs containing commas and other configuration edge cases.

* **Refactor**
  * Updated plot-related storage and processing to be more consistent and reliable across create, remove, and print actions.

### DIFF
--- a/R/create-plots.R
+++ b/R/create-plots.R
@@ -2,42 +2,58 @@
 #
 # Public createPlots / createPlotsFromExcel plus the parsing, validation,
 # and configuration helpers they use. Stateless: pulls a Project's
-# dataCombined / plotConfiguration / plotGrids and returns a named list of
-# plot-grid objects.
+# dataCombined / plots / plotGrids sections and returns a named list of
+# rendered plot-grid (and optionally standalone plot) objects keyed by id.
 
 #' Generate plots from a Project
 #'
 #' @description
-#' Reads `project$plots$plotConfiguration` and `project$plots$plotGrids`
-#' (both data.frames) to build the requested plot grids. DataCombined
-#' objects are resolved via [createDataCombined()] internally unless
-#' supplied via `dataCombinedList`.
+#' Reads `project$plots` and `project$plotGrids` (both keyed lists, one entry
+#' per plot / grid) to build the requested plot grids and, optionally,
+#' standalone single plots. DataCombined objects are resolved via
+#' [createDataCombined()] internally unless supplied via `dataCombinedList`.
+#'
+#' With neither `plotGrids` nor `plots`, all plot grids declared in
+#' `project$plotGrids` are built (the default). The two arguments are
+#' independent selectors: `plotGrids` selects plot grids (keyed by
+#' `plotGridId` in the result), `plots` selects standalone single plots
+#' (keyed by `plotId`). A `plotId` that is also part of a requested grid still
+#' gets its own standalone entry.
 #'
 #' @param project A `Project` (see [loadProject()]).
-#' @param plotGridNames Names of plot grids to build. If `NULL` (default),
-#'   all grids declared in `project$plots$plotGrids` are built.
-#' @param simulatedScenarios Named list of simulated scenarios from
-#'   [runScenarios()].
+#' @param plotGrids Names of plot grids to build. If `NULL` (default) and
+#'   `plots` is also `NULL`, all grids declared in `project$plotGrids` are
+#'   built.
+#' @param plots Ids of standalone single plots to render directly (not laid
+#'   out in a grid), each resolved against `project$plots`. `NULL` (default)
+#'   renders no standalone plots; standalone plots are opt-in.
+#' @param scenarioResults Named list of Scenario Results from
+#'   [runScenarios()] (each entry has `simulation`, `results`,
+#'   `outputValues`, `population`). Not the OSPS `SimulationResults`.
 #' @param dataCombinedList Optional pre-built named list of `DataCombined`
 #'   objects. If `NULL`, the function builds them via [createDataCombined()].
 #' @param stopIfNotFound If `TRUE`, errors when a referenced DataCombined or
-#'   simulated/observed entry cannot be resolved.
+#'   simulated/observed entry cannot be resolved, or when a requested
+#'   `plotGrids` / `plots` id is not defined in the project.
 #' @param validate Logical. If `TRUE` (default), runs the relevant
 #'   section validators via [validateProject()] before building the
 #'   plots and aborts with a formatted summary on critical errors. Set
 #'   to `FALSE` to skip the pre-flight check (e.g. when the caller has
 #'   already validated the project).
 #'
-#' @returns A named list of plot-grid objects (one per `plotGridName`), or
-#'   an empty list when the project has no plots section.
+#' @returns A named list keyed by id: one entry per requested plot grid
+#'   (keyed by `plotGridId`) unioned with one entry per requested standalone
+#'   plot (keyed by `plotId`). An empty list when the project has no plots to
+#'   build.
 #'
 #' @import tidyr
 #'
 #' @export
 createPlots <- function(
   project,
-  plotGridNames = NULL,
-  simulatedScenarios = NULL,
+  plotGrids = NULL,
+  plots = NULL,
+  scenarioResults = NULL,
   dataCombinedList = NULL,
   stopIfNotFound = TRUE,
   validate = TRUE
@@ -50,51 +66,67 @@ createPlots <- function(
       opName = "createPlots"
     )
   }
-  if (is.null(project$plots)) {
-    return(list())
+  allPlotConfig <- .unwrapDefinitionList(project$plots) %||% list()
+  allPlotGrids <- .unwrapDefinitionList(project$plotGrids) %||% list()
+
+  # Only default to "all grids" when neither selector is given. A caller that
+  # asks only for standalone `plots` should not also get every grid.
+  requestSpecified <- !is.null(plotGrids) || !is.null(plots)
+  if (is.null(plotGrids) && !requestSpecified) {
+    plotGrids <- names(allPlotGrids)
   }
-  cfgDf <- project$plots$plotConfiguration %||% data.frame()
-  gridDf <- project$plots$plotGrids %||% data.frame()
-  if (nrow(gridDf) == 0) {
-    return(list())
-  }
-  if (is.null(plotGridNames)) {
-    plotGridNames <- gridDf$name
-  }
-  # Unknown grid names are silently dropped below; surface them when the
-  # caller asked us to stop on unresolved references.
+
+  # Surface unknown requested ids when the caller asked us to stop on
+  # unresolved references (mirrors the per-kind lookups below).
   if (isTRUE(stopIfNotFound)) {
-    unknownGrids <- setdiff(plotGridNames, gridDf$name)
+    unknownGrids <- setdiff(plotGrids, names(allPlotGrids))
     if (length(unknownGrids) > 0) {
       cli::cli_abort(messages$stopPlotGridNamesNotFound(unknownGrids))
     }
-  }
-  # Filter to only the requested grids and the plot configs they reference,
-  # so validation and DataCombined building stay scoped to this build. An
-  # unfiltered cfgDf would validate plots whose DataCombined is not built
-  # (e.g. plots in other grids, or plots in no grid at all) and abort.
-  gridDf <- gridDf[gridDf$name %in% plotGridNames, , drop = FALSE]
-  if (nrow(gridDf) == 0) {
-    return(list())
-  }
-  referencedPlotIDs <- unique(unlist(lapply(gridDf$plotIds, .splitPlotIDs)))
-  if (nrow(cfgDf) > 0) {
-    cfgDf <- cfgDf[cfgDf$plotId %in% referencedPlotIDs, , drop = FALSE]
+    unknownPlots <- setdiff(plots, names(allPlotConfig))
+    if (length(unknownPlots) > 0) {
+      cli::cli_abort(messages$stopPlotIdsNotFound(unknownPlots))
+    }
   }
 
-  # Build DataCombined for the configs referenced by the requested grids.
+  # Filter to only the requested grids and the plot configs they reference,
+  # plus the explicitly requested standalone plots. Scoping the build to what
+  # is asked for keeps validation and DataCombined building from touching plots
+  # whose DataCombined is not built (which would abort).
+  selectedGrids <- allPlotGrids[intersect(names(allPlotGrids), plotGrids)]
+  selectedPlotIds <- intersect(plots, names(allPlotConfig))
+
+  gridReferencedPlotIds <- unique(unlist(lapply(
+    selectedGrids,
+    function(g) .splitPlotIDs(g$plotIds)
+  )))
+  neededPlotIds <- union(gridReferencedPlotIds, selectedPlotIds)
+  plotConfig <- allPlotConfig[intersect(names(allPlotConfig), neededPlotIds)]
+
+  if (length(selectedGrids) == 0 && length(selectedPlotIds) == 0) {
+    return(list())
+  }
+
+  # Build DataCombined for the configs referenced by both selectors: the
+  # requested grids and the explicitly requested standalone plots.
   if (is.null(dataCombinedList)) {
+    standaloneDataCombinedNames <- unique(unlist(lapply(
+      plotConfig[selectedPlotIds],
+      function(p) p$dataCombinedId
+    )))
     dataCombinedList <- createDataCombined(
       project,
-      plotGridNames = plotGridNames,
-      simulatedScenarios = simulatedScenarios,
+      dataCombined = standaloneDataCombinedNames,
+      plotGrids = plotGrids,
+      scenarioResults = scenarioResults,
       stopIfNotFound = stopIfNotFound
     )
   }
 
-  .createPlotGridsFromDataFrames(
-    dfPlotConfigurations = cfgDf,
-    dfPlotGrids = gridDf,
+  .createPlotGridsFromEntries(
+    plotConfigurations = plotConfig,
+    plotGrids = selectedGrids,
+    standalonePlotIds = selectedPlotIds,
     dataCombinedList = dataCombinedList
   )
 }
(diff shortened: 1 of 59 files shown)


Felixmil:
@coderabbitai review - Keep in mind that this PR is part of several stacked PR: Some of your findings may already be fixed in downstream PRs.

R/create-plots.R
```diff
@@ -50,51 +66,67 @@ createPlots <- function(
...
+  # unresolved references (mirrors the per-kind lookups below).
   if (isTRUE(stopIfNotFound)) {
-    unknownGrids <- setdiff(plotGridNames, gridDf$name)
+    unknownGrids <- setdiff(plotGrids, names(allPlotGrids))
     if (length(unknownGrids) > 0) {
       cli::cli_abort(messages$stopPlotGridNamesNotFound(unknownGrids))
     }
+    unknownPlots <- setdiff(plots, names(allPlotConfig))
+    if (length(unknownPlots) > 0) {
+      cli::cli_abort(messages$stopPlotIdsNotFound(unknownPlots))
```
Felixmil, line 88: **Blocking** — `messages$stopPlotIdsNotFound` is never defined anywhere in the package. `R/messages.R` is untouched by this PR and only defines the plot-grid analogue, `stopPlotGridNamesNotFound` (messages.R:521). Since `messages` is a plain list, `messages$stopPlotIdsNotFound` evaluates to `NULL`, so this line becomes `NULL(unknownPlots)`, which throws `"attempt to apply non-function"` — not the intended abort, and not the message your own new snapshot expects:

```
# tests/testthat/_snaps/create-plots.md
Error in `createPlots()`:
! The following plots are not defined in the project: "ghost_plot"
```

The PR description explicitly promises this: "`stopIfNotFound` now also covers unknown `plotIds`." As shipped it doesn't — every caller passing an unknown standalone plot id with `stopIfNotFound = TRUE` (the default) gets a cryptic base-R error instead, and the test added in this same commit will fail against its own snapshot.

Fix: add the missing helper to `R/messages.R`, mirroring the existing pattern:
```r
messages$stopPlotIdsNotFound <- function(plotIds) {
  cliFormat(
    "The following plots are not defined in the project:
    {.val {paste(plotIds, collapse = ', ')}}"
  )
}
```

---

Felixmil (commented):

_(Submitted as a comment review — GitHub does not allow this account to request changes on its own pull request. Verdict below is Request Changes.)_

## Summary
The restructuring itself is solid: the three per-entity keyed-list kinds (`data-combined`, `plots`, `plot-grids`) are cleanly split with no NA padding, `createPlots` genuinely iterates keyed lists instead of `apply(df, 1, ...)`, the `scenarioResults` rename and id-field rationalization (`plotId`/`plotGridId`/`dataCombinedId`) are applied consistently across parse/serialize/validate/mutate, and the vectorized `addPlot`/`removePlot` mutators check out against their tests.

One real bug survives independent verification on both the Standards and Spec axes (see inline comment): the new `stopIfNotFound` path for unknown `plots` ids calls a message function that was never defined, so it crashes instead of aborting cleanly — and the snapshot test this PR itself adds asserts an error message that can never actually be produced. There's also a small amount of scope creep beyond the PR's stated description.

## Critical Issues (Blocking)
1. **`R/create-plots.R:88`** — calls the undefined `messages$stopPlotIdsNotFound`. See inline comment for detail and fix. This breaks the exact behavior the PR description promises ("`stopIfNotFound` now also covers unknown `plotIds`") and makes the new snapshot test in this same commit unreachable as written.

## Required Changes
None beyond the fix above. Several lower-confidence nitpicks were considered and left to the author's judgement rather than blocked on: a redundant boolean condition in the `requestSpecified` check at `R/create-plots.R:74-76` (`!requestSpecified` already implies `is.null(plotGrids)`), an asymmetric defensiveness gap between `.parsePlotEntries` and `.parseNestedDataCombined` (the latter guards malformed/missing ids with a friendly abort, the former doesn't — currently unreachable via the normal load path, but worth double-checking), and a silent-flatten fallback in `addPlotGrid`'s per-grid `plots` recycling when list length doesn't match grid count.

## Spec Compliance
- **Missing/broken**: "`stopIfNotFound` now also covers unknown `plotIds`" — implemented but broken, see Blocking #1.
- **Scope creep**: the PR description scopes itself to splitting the plots kinds, rewriting `createPlots`, id rationalization, and vectorizing mutators. Two additions aren't mentioned in that scope:
  - New critical-error validation for a missing `label` field on both simulated and observed `dataCombined` entries in `.validatePlots()` (`R/plots.R`, e.g. `"Simulated entry in dataCombined '", dcName, "' is missing 'label'"`). Looks like a legitimate correctness improvement, but it's a new validation rule, not a restructuring of an existing one — worth calling out explicitly in the PR description.
  - Comma/backslash-escaping added to `.joinPlotIDs()`/`.splitPlotIDs()` so a plot id containing a literal comma survives the grid's delimited `plotIds` round-trip. Reasonable, but not mentioned in the "vectorizes the plot mutators" framing.
- **Minor wording mismatch**: the description says the new argument is "`createPlots(plotIds=)`", but the shipped parameter is named `plots` (`R/create-plots.R:55`). Not a functional issue — `plots` is arguably more consistent with the existing `plotGrids` argument — just flagging the description/implementation naming mismatch.
- Everything else — the three-kind split, the keyed-list rewrite of `createPlots`, the `scenarioResults` rename on both `createPlots` and `createDataCombined`, and the id-field rationalization — is faithful to spec.

## Suggestions
- `R/create-plots.R:74-76`: `!requestSpecified` already implies `is.null(plotGrids)`, so `if (is.null(plotGrids) && !requestSpecified)` can simplify to `if (!requestSpecified)`.

## Verdict
Request Changes

---

R/create-plots.R
```diff
@@ -340,200 +370,185 @@ createPlotsFromExcel <- function(...) {
...
+  missingPlots <- setdiff(referencedPlotIds, plotIDs)
   if (length(missingPlots) != 0) {
-    stop(messages$errorInvalidPlotID(missingPlots))
+    msg <- messages$errorInvalidPlotID(missingPlots)
+    cli::cli_abort("{msg}")
   }
-
-  return(dfPlotGrids)
+  invisible(plotGrids)
 }
```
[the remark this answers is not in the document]
Felixmil: Addressed in `6cfad245` (now on the top-of-stack branch, #1104). `.assertPlotGridsBuildable()` now aborts on a grid missing its `plotGridId`, before the duplicate-id check, via a new `messages$missingPlotGridId()` helper ("Every plot grid must declare a `plotGridId`."). This also stops two id-less grids from being misreported as a duplicate-id violation. A regression test covers it (`expect_snapshot(error = TRUE)` in `test-create-plots.R`).

This slice (#1094) is not modified: per the stacked entity-files train's tree-identity contract, review fixes land on the top of the stack (#1106, since merged into #1104), not on the individual slice.

R/data-combined.R
```diff
@@ -25,38 +25,38 @@
...
-    dataCombinedNames[!is.na(dataCombinedNames)],
+    dataCombined[!is.na(dataCombined)],
     names(allSpecs)
   )
   if (length(missingNames) > 0) {
     cli::cli_abort(messages$stopDataCombinedNamesNotFound(missingNames))
   }
 
-  selectedSpecs <- allSpecs[intersect(names(allSpecs), dataCombinedNames)]
+  selectedSpecs <- allSpecs[intersect(names(allSpecs), dataCombined)]
```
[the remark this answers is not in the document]
Felixmil: Addressed in `6cfad245` (now on the top-of-stack branch, #1104). `createDataCombined()` now validates its `plotGrids` argument against the project's actual grid ids and aborts on any unknown name (reusing the existing `stopPlotGridNamesNotFound()`), before `.extractDataCombinedNamesForPlotsFromProject()` is called, so a typo no longer yields an incomplete or empty result. A regression test covers it (`expect_snapshot(error = TRUE)` in `test-data-combined.R`).

This slice (#1094) is not modified: per the stacked train's tree-identity contract, review fixes land on the top of the stack (#1106, since merged into #1104), not on the slice.

R/create-plots.R
```diff
@@ -340,200 +370,185 @@ createPlotsFromExcel <- function(...) {
...
+    if (!is.null(entry$subtitle)) {
+      plotConfiguration$subtitle <- entry$subtitle
     }
-    .validateLogScaleAxisLimits(plotConfiguration, row[["plotId"]])
-    return(plotConfiguration)
+    .validateLogScaleAxisLimits(plotConfiguration, entry$plotId)
+    plotConfiguration
   })
-  names(plotConfigurationList) <- dfPlotConfigurations$plotId
+  names(plotConfigurationList) <- configPlotIds
```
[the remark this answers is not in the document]
Felixmil: Addressed earlier in the stack, in `e0c28803` (reachable from #1104). `xLabel` and `yLabel` are now re-applied verbatim onto the built `plotConfiguration` after the style-field pass (the same treatment `title`/`subtitle` get), so a user's axis labels reach the rendered plot instead of being silently dropped. A test asserts the labels survive onto the built configuration (`test-create-plots.R`, "createPlots carries xLabel/yLabel onto the built plot").

R/create-plots.R
Felixmil: Addressed earlier in the stack, in `e3808c54` (reachable from #1104): `messages$stopPlotIdsNotFound()` is now defined in `R/messages.R`, so the `createPlots()` unknown-`plots` abort no longer references an undefined helper. The `R/messages.R` changes ride with the validation/prints slice rather than this plots slice, which is why they were absent here at the time of review.

R/plots.R
```diff
@@ -373,311 +617,468 @@
...
-    cli::cli_abort("plot {.val {plotId}} already exists")
-  }
+  .requireNonEmptyStringVector(id, "id")
+  id <- .canonicalizeId(id)
+  n <- length(id)
+  dataCombined <- .canonicalizeIdRef(
+    .recycleScalarArg(dataCombined, n, "dataCombined")
+  )
+  plotType <- .recycleScalarArg(plotType, n, "plotType")
+  perEntityFields <- .dotsToPerEntityFields(list(...), n)
```
Felixmil, line 707: **Usability trap in the `...` recycling (medium).** The named scalar args (`dataCombined`, `plotType`) align a length-N vector *by position* via `.recycleScalarArg()`, but a `...` field that is an atomic vector of length N is treated as *whole-per-entity* by `.dotsToPerEntityFields()` and collapsed to a CSV string applied to every entity. So `addPlot(id = c("p1", "p2"), dataCombined = "dc", plotType = "individual", title = c("Plot 1", "Plot 2"))` gives *both* plots the title `"Plot 1, Plot 2"` rather than one title each.

It is documented ("pass a list to vary per plot"), and the whole-application rule is legitimate for genuinely multi-value fields like `quantiles = c(0.05, 0.5, 0.95)`. But the asymmetry with the positional scalar args, plus the *silent* CSV-collapse corruption (no error, and not even an obvious repeat), makes this easy to hit by accident. Worth either erroring when an atomic `...` field's length equals N for a known single-value field, or at minimum calling out the asymmetry more loudly in the docs. Same applies to `addPlotGrid()`'s `...`.

R/create-plots.R
```diff
@@ -340,200 +370,185 @@ createPlotsFromExcel <- function(...) {
...
+.assertPlotConfigurationsBuildable <- function(
+  plotConfigurations,
   dataCombinedNames
 ) {
-  # mandatory column dataCombinedName is empty - throw error
-  missingLabel <- sum(is.na(dfPlotConfigurations$dataCombinedName))
-  if (missingLabel > 0) {
-    stop(messages$missingDataCombinedName())
-  }
+  ids <- vapply(
```
Felixmil, line 389: **Build guard misses a single plot with no `plotId` (low).** `.assertPlotConfigurationsBuildable()` is documented as guarding the build path itself, but `ids` is derived here via `p$plotId %||% NA_character_`, and only the *duplicate* check downstream would flag a missing id (and only when two or more are missing, since `duplicated(c(NA, NA))` is `c(FALSE, TRUE)`). A single entry lacking `plotId` slips through this guard and then fails in `.createPlotGridsFromEntries()` at `configPlotIds <- vapply(..., function(p) p$plotId, character(1))` with the opaque `values must be length 1, but FUN(X[[1]]) result is length 0`.

Only reachable via `createPlots(validate = FALSE)` with a hand-built entry (the default `validate = TRUE` catches it in `.validatePlots()`), so low severity, but an explicit `plotId`-presence assertion here would honor the guard's stated contract and give a named, actionable error.

Felixmil (commented):
Review of the plots subsystem this PR introduces (three keyed kinds `dataCombined`/`plots`/`plotGrids`, the keyed-list `createPlots` runtime, the vectorized `addPlot`/`addPlotGrid`/`addDataCombined` mutators).

Overall the restructure is well-executed: dropping the NA-padded data frames for per-entity keyed lists is a clear improvement, the parse/serialize round-trip is near-identity, and the escape-aware `plotIds` join/split correctly handles the fact that a canonical id may legally contain a comma. I confirmed the `.joinPlotIDs()`/`.splitPlotIDs()` pair round-trips for every id reachable through the public API (canonicalization strips leading/trailing spaces, so the only mismatch I could construct is unreachable), every referenced `messages$` helper exists, and the subsystem tests pass under `NOT_CRAN=true`.

Two issues inline below, both low-to-medium severity, neither blocking.

Note on scope: this review was originally posted against the stack tip and has been moved here, since both findings are about code this PR introduces. A third, separate finding (a backwards code comment) concerns code added later in the review-round-1 layer and stays on that PR.

R/create-plots.R
Felixmil: Addressed in `50c4dbc2` (on the top-of-stack branch, #1104). `.assertPlotConfigurationsBuildable()` now rejects a plot missing its `plotId` up front, before the duplicate check, via a new `messages$missingPlotId()` helper ("Every plot must declare a `plotId`."). This closes the gap you named: a single id-less entry (which `duplicated()` cannot catch) no longer slips through to fail opaquely at `configPlotIds <- vapply(..., function(p) p$plotId, character(1))` in `.createPlotGridsFromEntries()`, and the guard now honours its stated build-path contract. It mirrors the sibling `plotGridId` guard already in `.assertPlotGridsBuildable()`. A regression test covers it (`expect_snapshot(error = TRUE)` in `test-create-plots.R`).

This slice (#1094) is not modified: per the stacked entity-files train's tree-identity contract, review fixes land on the top of the stack (#1104), not on the individual slice.

R/plots.R
Felixmil: Addressed in `50c4dbc2` (on the top-of-stack branch, #1104), via the docs route.

The `@param ...` roxygen on both `addPlot()` and `addPlotGrid()` now calls the asymmetry out loudly: a length-N atomic `...` field is treated as one multi-value field and applied whole to every entity (collapsed to a comma-separated string), unlike a length-N vector on the positional scalar args (`dataCombined` / `plotType`), which aligns by position. It states the concrete trap (`title = c("A", "B")` gives every plot `"A, B"`) and the fix (pass a length-N list to vary per entity).

I did not take the "error when an atomic `...` field's length equals N" route: telling a single-value field apart from a genuine multi-value one (`quantiles`, `foldDistance`) would require `.dotsToPerEntityFields()` to carry a per-field single-vs-multi registry, coupling it to the field set it deliberately stays agnostic of, and it would turn a documented, currently-legal call into an abort. The whole-application rule is intentional and load-bearing for the multi-value fields, so the asymmetry stays; the docs now make it hard to hit by accident.

This slice (#1094) is not modified: per the stacked train's tree-identity contract, review fixes land on the top of the stack (#1104), not on the slice.

R/create-plots.R
```diff
@@ -556,20 +573,21 @@ createPlotsFromExcel <- function(...) {
       )
     )
   })
-  names(plotList) <- dfPlotConfigurations$plotId
+  names(plotList) <- configPlotIds
 
   defaultPlotGridConfig <- createEsqlabsPlotGridConfiguration()
-  plotGrids <- apply(dfPlotGrids, 1, \(row) {
-    plotGridConfiguration <- .createConfigurationFromRow(
+  gridStyleFields <- c("plotGridId", "plotIds", "title")
```
nautilus69, line 579: `subtitle` is not in `gridStyleFields`, so a grid `subtitle` is routed through `.createConfigurationFromEntry()`'s
  comma-splitting `scan()` and split into a vector: `"Model A, Model B"` → `c("Model A", "Model B")`.
  `tlf::PlotGridConfiguration$subtitle` accepts a length-2 value without error, so the corruption is **silent**.
  - This is the grid-side twin of the `xLabel`/`yLabel` drop that #1104 fixed on the plot side (`e0c28803`); the grid side was
  missed. Only `title` is re-applied (`R/create-plots.R:586`).

nautilus69 (approved):
Other findings are fixed in different places.
Approved With minor changes: 
the other  one in create-plots.R and 
one is here: 
 (D) `.validatePlots()` does not require simulated `path`, but the write-gate does 
  - `R/plots.R:279-300` — the simulated-entry validation loop checks only `label` and `scenario`.
  - `R/plots.R:606` — `.checkDataCombinedEntry()` requires `c("label", "scenario", "path")` for simulated entries.
  - A DataCombined loaded from JSON (bypassing `addDataCombined()`) with a simulated entry missing `path` passes `validateProject()`,
  yet would have been rejected at add-time and fails later in the build. Validator is laxer than the writer.

R/create-plots.R
Felixmil: Addressed in `7c5a94bc` (on the top-of-stack branch, #1104). This is the grid-side twin of the plot-side `xLabel`/`yLabel` fix: `subtitle` is now in `gridStyleFields`, so it is excluded from `.createConfigurationFromEntry()`'s comma-splitting `scan()` and re-applied verbatim onto the built `PlotGridConfiguration` (the same treatment `title` already had). A grid `subtitle` like `"Model A, Model B"` now reaches the grid as one string instead of being silently shredded into `c("Model A", "Model B")`. A regression test in `test-create-plots.R` asserts a comma-bearing subtitle survives onto the built grid.

This slice (#1094) is not modified: per the stacked entity-files train's tree-identity contract, review fixes land on the top of the stack (#1104), not on the individual slice.

Felixmil:
Addressed in `7c5a94bc` (on the top-of-stack branch, #1104).

On the "validator laxer than the writer" finding: `.validatePlots()` now flags a simulated `dataCombined` entry that is missing its `path`, matching the write-gate `.checkDataCombinedEntry()`, which already requires `c("label", "scenario", "path")` for a simulated entry. Both gates route through the shared `.isMissingField()` helper, whose contract is that they treat a required field's absence identically, so the validator is no longer laxer than the writer: a `dataCombined` loaded from JSON (bypassing `addDataCombined()`) with a simulated entry missing `path` is now caught by `validateProject()` instead of passing validation and failing later at build. A regression test in `test-validation.R` covers the new critical error.

This slice (#1094) is not modified: per the stacked entity-files train's tree-identity contract, review fixes land on the top of the stack (#1104), not on the individual slice.

PR #1095, opened 2026-07-02
Parameter identification reshape

Felixmil:
BEWARE: This PR is part of a stacked PR merge train. It was split by concept and was not meant to pass CI. Also, when reviewing, make sure to double check findings against latest pr in the chain: currently #1104. Details in #1088 

Reshapes parameter identification (PI) onto the entity-file / JSON-first model. It exports the PI record constructors (`PITask`, `PIParameter`, `PIOutputMapping`) so tasks can be built programmatically, canonicalizes the task ids that enter `runPI` through the public API the same way every other id kind is canonicalized, warns when a fit reports convergence but the uncertainty (SD/CV/CI) cannot be quantified, applies the suffixless rename to PI reference arguments and JSON keys, and aligns PI validation with the shared validation helpers. It sits on PR6 in the stack.

## PI record constructors

Exports the PI record constructors `PITask`, `PIParameter`, and `PIOutputMapping`, letting parameter-identification tasks and their parameters and output mappings be assembled programmatically rather than only decoded from a project. Sub-entity authoring is vectorized and routed through the shared section-setting path.

## runPI id canonicalization

Ids supplied to `runPI` via the API are now canonicalized on the way in, so they match the canonical form used everywhere else in the model instead of being taken verbatim.

## Uncertainty warning

When a fit reports convergence as `TRUE` but the standard deviation, coefficient of variation, or confidence interval come back `NA`, the run now emits a warning naming the affected parameter and the likely cause, rather than silently returning unquantifiable uncertainty.

## Suffixless rename and validation alignment

Applies the cross-cutting suffixless rename to the PI reference arguments and JSON keys, aligns PI validation with the shared validation helpers, and corrects the related documentation.

## Tests and snapshots

Updates `test-parameter-identification.R` and the `_snaps/parameter-identification.md` snapshot to the reshaped behaviour (constructor exports, canonicalized ids, the uncertainty warning, and the renamed keys/args).

## Stacking notes

Stacked on `pr06-plots-restructure`. The Excel PI round-trip bridge lives in a later PR (PR9), not here. This PR is intentionally not CI-green in isolation: PI reads the parameter-sets and scenario model and the write-through engine introduced in earlier PRs in the stack, and the Excel PI bridge lands in PR9. The newly exported constructors add `@export`, but their generated `man/*.Rd` and `NAMESPACE` entries regenerate in PR13.

### DIFF
--- a/R/parameter-identification.R
+++ b/R/parameter-identification.R
@@ -2,8 +2,45 @@
 
 # Plain-data record constructors.
 
-#' @keywords internal
-#' @noRd
+#' Create a Parameter Identification parameter
+#'
+#' @description Builds a plain-data `PIParameter` record describing one
+#'   optimisation variable of a Parameter Identification task: the model
+#'   parameter to estimate, the scenarios it is fitted across, and its
+#'   search bounds and start value.
+#'
+#'   A `PIParameter` is a building block for a [PITask()]; pass a list of
+#'   them as the `parameters` argument of [PITask()] or [addPITask()], or
+#'   add one to an existing task with [addPIParameter()].
+#'
+#' @param id Character scalar. Identifier for this parameter, unique within
+#'   its task. Used as a free label by the PI run, not as an entity-file id.
+#' @param scenarios Character vector of scenario ids the parameter is
+#'   estimated across. Listing several scenarios fits one shared value
+#'   across all of them. The constructor does not check these against the
+#'   task's own `scenarios`; any that are not a subset of the task's
+#'   scenarios are reported later, by [validateProject()] and at run time.
+#' @param path Character scalar. Full simulation path of the model
+#'   parameter to estimate (OSPS notation).
+#' @param units Optional character scalar. Display unit the bounds and
+#'   start value are expressed in. `NULL` or `""` means the model default
+#'   unit.
+#' @param minValue,maxValue,startValue Numeric scalars. Search bounds and
+#'   the starting value; `minValue <= startValue <= maxValue` must hold.
+#'
+#' @returns A `PIParameter` object: a named list with copy semantics.
+#' @seealso [PITask()], [PIOutputMapping()], [addPIParameter()].
+#' @export
+#' @family parameterIdentification
+#' @examples
+#' PIParameter(
+#'   id = "lipophilicity",
+#'   scenarios = "aciclovir_iv",
+#'   path = "Aciclovir|Lipophilicity",
+#'   minValue = -2,
+#'   maxValue = 2,
+#'   startValue = 0
+#' )
 PIParameter <- function(
   id,
   scenarios,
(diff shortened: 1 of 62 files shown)


R/parameter-identification.R
```diff
@@ -867,6 +1000,48 @@ runPI <- function(
...
+# A `result` without a `toList()` method (e.g. a test stub) is a silent no-op.
+#
+# @keywords internal
+# @noRd
+.warnUnquantifiedUncertainty <- function(taskName, result) {
+  if (is.null(result) || !is.function(result$toList)) {
+    return(invisible(NULL))
+  }
+  info <- result$toList()
+  paramNames <- info$paramNames
```
Felixmil, line 1019: **Blocking** — `.warnUnquantifiedUncertainty()` never reads `info$convergence`. The spec (and this function's own header comment two lines up: *"`runtime$run()` returns `convergence = TRUE` even when the CI step ... produces nothing usable, so a green 'converged' status can hide a parameter with no real uncertainty"*) is explicit that this warning exists to catch the specific case of **convergence = TRUE with unquantifiable uncertainty**. As written, it fires for *any* run with all-NA SD/CV/CI, converged or not.

That means a non-converged fit (the ordinary, unremarkable case where uncertainty is naturally unavailable) now also emits: *"uncertainty could not be quantified ... even though convergence is reported"* — which is simply false when convergence was `FALSE`. This actively misinforms a caller inspecting fit quality.

Confirmed against the two new tests at `test-parameter-identification.R:1198` and `:1219` — both fixtures set `convergence = TRUE` and neither exercises `convergence = FALSE`, so the gap isn't caught.

```suggestion
  paramNames <- info$paramNames
  if (!isTRUE(info$convergence)) {
    return(invisible(NULL))
  }
```

---

R/parameter-identification.R
```diff
@@ -867,6 +1000,48 @@ runPI <- function(
...
+  if (is.null(paramNames) || length(paramNames) == 0L) {
+    return(invisible(NULL))
+  }
+  # A parameter has no usable uncertainty when SD, CV, and both CI bounds are
+  # all NA. These vectors are parallel to `paramNames` (one entry per
+  # parameter), each defaulting to NA when the CI step yielded nothing.
+  unquantified <- is.na(info$sd) &
+    is.na(info$cv) &
+    is.na(info$lowerCI) &
+    is.na(info$upperCI)
```
Felixmil, line 1029: **Verify** — the PR description says the warning fires when SD, CV, "**or**" CI come back NA, but this requires all four (`sd`, `cv`, `lowerCI`, `upperCI`) to be NA simultaneously (`&`-chained). If these are always derived together from one Hessian-based computation (so a partial-NA case can't occur in practice), this is fine and just loosely worded in the description — but if a partial-NA case is possible (e.g. SD computed, CI bounds not), a parameter with real but incomplete uncertainty would silently skip the warning. Worth a one-line confirmation from whoever wrote the upstream `toList()` result.

---

Felixmil (commented):

(Note: submitted as a comment review, not "Request changes" — GitHub blocks requesting changes on one's own PR at the API level. Treat the verdict below as Request Changes in substance.)

## Summary
Reviewed the diff on both axes (documented standards vs. the PR's own description as spec — no repo `CLAUDE.md`/`CONTRIBUTING.md` exists, so the standards axis ran against this environment's global R-package conventions). Standards axis: clean — the strongest finding (triplicated canonicalization warnings from `addPITask()`'s three separate `.canonicalizeIdRef()` calls not being batched through `.collectCanonicalizedRefs()`, confirmed by the updated snapshot) only reached 65/100 confidence and is dropped per the review's confidence bar. Spec axis: one confirmed logic bug — the new `.warnUnquantifiedUncertainty()` doesn't gate on `info$convergence`, so it fires on non-converged runs too, contradicting both the PR description and the function's own header comment about what it's meant to catch.

## Critical Issues (Blocking)
1. `R/parameter-identification.R:1019` — the uncertainty warning ignores `info$convergence`; see inline comment for the fix and the confirming test gap.

## Required Changes
None beyond the blocking item above.

## Spec Compliance
- **Uncertainty warning — implemented but wrong.** Spec: *"warns when a fit reports convergence but the uncertainty (SD/CV/CI) cannot be quantified"*. Implementation warns unconditionally on all-NA uncertainty, converged or not (see inline Blocking comment).
- **AND vs. OR wording — Verify, not confirmed wrong.** Spec: *"the standard deviation, coefficient of variation, or confidence interval come back `NA`"*. Implementation requires all four fields NA simultaneously. Flagged inline as a question, not a defect — plausible these values are always NA together in practice.
- **Canonicalization scope wider than described — not a defect.** Spec ("runPI id canonicalization") names only `runPI`, but the diff also canonicalizes ids/refs in `addPITask`/`removePITask`/`addPIParameter`/`removePIParameter`/`addPIOutputMapping`/`removePIOutputMapping`. Cross-checked against sibling entity files (`R/applications.R`, `R/individuals.R`, `R/scenarios.R`): every other `add*`/`remove*` in the codebase canonicalizes refs the same way, so this is consistent, necessary behavior — the PR description just undersells it. Worth a one-line addition to the description, not a code change.
- PI record constructors, the suffixless rename, validation alignment, and the test/snapshot updates otherwise match the description.

## Suggestions
None beyond the Verify item above — the rename and export work is otherwise clean.

## Verdict
Request Changes

---

PR #1102, opened 2026-07-02
Polish: lower-tier AUDIT + issue-ledger remediation

Felixmil:
This is a post-migration polish layer stacked on top of the entity-files migration stack (base `pr13-regenerate-docs`). Unlike PRs 1 through 13, it does not carry a slice of the migration delta; it lands the remaining lower-tier fixes from the codebase quality review plus the still-open rows of the issue ledger, i.e. the concrete defects that survived the migration. Because it fixes defects rather than reproducing the migration tree, it deliberately extends the stack beyond the migration boundary (the tree-identity contract is satisfied at PR13). The stack tip `pr15-review-round-1` (#1104) is the canonical reference for this effort.

## Authoring input validation

`addOutputPath()` now rejects an empty or `NA` path, matching `setOutputPath()`. `addIndividual()` rejects a non-finite `weight` / `height` / `age` instead of silently coercing it to `NA` (e.g. `weight = "80kg"`). `addPopulation()` rejects a non-integer `numberOfIndividuals`.

## Observed data

`loadObservedData()` now aborts when two observed-data sources resolve to the same data-set name (previously the later source silently shadowed the earlier), and a `script` source that returns a list of data sets is keyed by each data set's own name. The `loadObservedData()` roxygen is corrected: the `programmatic` type is a session-resolved sentinel, not a type that "errors fast".

## Scenario results and runs

`saveScenarioResults()` aborts up front when two scenario names collapse to the same file-safe name (e.g. `"A/B"` and `"A_B"`), which previously overwrote each other silently. `runScenarios()` gains a `stopIfParameterNotFound` argument (default `TRUE`) that is threaded through to per-scenario preparation.

## Plots

The plot-configuration override now uses an NA-safe comparison, so a legitimate override is no longer silently skipped when the default value carries an `NA`.

## API surface and cleanups

The redundant `setIndividualParameterSets()` is removed (use `setIndividual(id, parameterSets = )`); `setApplicationParameterSets()` is kept, as applications have no `setApplication`. Dead code is pruned (three unused message builders, a stale `globalVariables` entry). Roughly fifteen base `stop()` / `warning()` / `message()` sites are routed through the `cli` wrappers over the message catalog (three legitimate condition re-raises are kept). `projectStatus()`'s dual-mode first argument is documented, and a test-setup option is scoped with `withr` rather than a bare `options()`.

## Stacking notes

Stacked on `pr13-regenerate-docs`. Unlike PRs 1 through 13, this PR is internally consistent (source, generated `man/`, and tests are all present and green together), so it is CI-green on its own. Its file-set intentionally overlaps earlier PRs' file-sets, because it fixes code those PRs introduced.


Felixmil:
Superseded by #1104. Every commit on `pr14-polish` is already contained in `pr15-review-round-1` (pr15 = pr14 + review fixes + the entity->definition rename), so there is nothing in this PR that #1104 does not already carry. #1104 has been retargeted to base directly onto `json-based-project`, and `pr14-polish` is being removed. Closing this to avoid a redundant merge.

PR #1103, opened 2026-07-06
Review round 1: cross-stack review fixes

Felixmil:
This is the **review round 1** layer: it collects the fixes required by the first round of reviews across the stacked PRs, landing as a post-migration layer stacked on `pr14-polish`. Like `pr14-polish`, it extends the stack beyond `entity-files-scenarios`, so it is outside the migration tree-identity contract (PRs 1-13 must sum to `entity-files-scenarios`); fixing these inside their originating PRs would break that contract, so they land here instead. `entity-files-scenarios` is advanced to carry the same fixes so it stays canonical.

The initial content addresses the three "must fix" follow-ups raised on the persistence engine (#1089). Round-1 fixes from later PRs' reviews are added here as they are triaged (for example the `R/project.R` documentation nits raised on #1090).

## Observed-data id escape guard

The observed-data serialize path derived its on-disk id (`.observedDataEntryId()`, a programmatic `name` or a file basename) and used it as the map key without the safe-key guard every other keyed kind applies, so a programmatic `name` containing a path separator or `..` could escape `definitions/observed-data/`. A new `.validateObservedDataId()` rejects an id that is not a single safe filename segment before it becomes a filename. It rejects rather than rewrites, because the id doubles as the match key for `removeObservedData()`, so canonicalizing it would desync the on-disk filename from that match key.

## Plots / PI serialize key alignment

`.serializePlotEntrySet()` and `.serializePITaskSet()` keyed each per-file record by the in-memory map key but stored the record's own id field, which could diverge. On the next `loadProject()` the load side compares the stored id to the filename and aborts, so a divergence produced an unreadable file. Both serializers now abort up front when the stored id disagrees with the map key, so the on-disk filename stays the authoritative key.

## initProject() definitions/ scaffold

`initProject()`'s documentation said the scaffold includes a `definitions/` tree, but the scaffold only created `Models/`, `Data/`, `Populations/`, and `Results/`, so a minimal project had no `definitions/` at all. `initProject()` now creates a `definitions/` directory, matching the documented contract.


Felixmil:
Superseded by #1104. This branch was renamed `pr15-persistence-review-fixes` -> `pr15-review-round-1` (now the "review round 1" layer collecting all first-round review fixes across the stack), and the rename closed this PR; #1104 carries the same head at the same commit onto `pr14-polish`.

PR #1104, opened 2026-07-06
Review round 1: cross-stack review fixes

Felixmil:
This is the **review round 1** layer: it collects the fixes required by the first round of reviews across the stacked PRs, landing as a post-migration layer stacked on `pr14-polish`. Like `pr14-polish`, it extends the stack beyond the migration tree, so it is outside the migration tree-identity contract (PRs 1-13 reproduce the migration tree); fixing these inside their originating PRs would break that contract, so they land here instead. This branch is the stack tip and the canonical reference for the effort.

The initial content addresses the three "must fix" follow-ups raised on the persistence engine (#1089). Round-1 fixes from later PRs' reviews are added here as they are triaged.

## Observed-data id escape guard

The observed-data serialize path derived its on-disk id (`.observedDataEntryId()`, a programmatic `name` or a file basename) and used it as the map key without the safe-key guard every other keyed kind applies, so a programmatic `name` containing a path separator or `..` could escape `definitions/observed-data/`. A new `.validateObservedDataId()` rejects an id that is not a single safe filename segment before it becomes a filename. It rejects rather than rewrites, because the id doubles as the match key for `removeObservedData()`, so canonicalizing it would desync the on-disk filename from that match key.

## Plots / PI serialize key alignment

`.serializePlotEntrySet()` and `.serializePITaskSet()` keyed each per-file record by the in-memory map key but stored the record's own id field, which could diverge. On the next `loadProject()` the load side compares the stored id to the filename and aborts, so a divergence produced an unreadable file. Both serializers now abort up front when the stored id disagrees with the map key, so the on-disk filename stays the authoritative key.

## initProject() definitions/ scaffold

`initProject()`'s documentation said the scaffold includes a `definitions/` tree, but the scaffold only created `Models/`, `Data/`, `Populations/`, and `Results/`, so a minimal project had no `definitions/` at all. `initProject()` now creates a `definitions/` directory, matching the documented contract.

## Project doc fixes (#1090 review)

Two documentation nits raised on the project-core PR are corrected in `R/project.R`. The `.persistKindChanges()` keyed fast-path comment still referenced `plotConfiguration`, which was removed when the plots trio split into `dataCombined` / `plots` / `plotGrids`; it now names the current fields. The `@field excel` roxygen listed eight sheet-name fields but omitted `initialConditionsFile` (the `.excelFilePathFields` const has nine), and that const's header comment said "seven" while listing nine; the roxygen now lists `initialConditionsFile` and the header comment reads "nine".

## Scenario authoring fixes (#1091 review)

Three defects raised on the scenarios PR are corrected. `setScenario()` (via the internal `.setOneScenario()`) treated a unit-only `steadyStateTimeUnit` change as a conversion, reading the stored base value under the old unit and reinterpreting it under the new one, which silently rescaled the physical duration (a "10 h" steady state relabeled to "min" became 10 minutes). A unit-only change is now a pure relabel that preserves the stored base value; supplying `steadyStateTime` still converts under the effective unit as before. `renameScenario()` now warns about dangling references on the old id before it changes the key, matching `removeScenario()`, so a `dataCombined` still naming the scenario by its old id is no longer silently orphaned. The `initializeSimulation()` `additionalInitialConditions` test now pins `units = "µmol"` on its read-back so it verifies the applied value under the unit it was set in rather than relying on the fixture's base unit.

## Entity-id canonicalizer fixes (#1089 review)

Three defects in the entity-id canonicalizer (`R/entity-id.R`), raised on the persistence-engine review, are corrected. The Windows reserved-basename guard in `.canonicalizeOneId()` tested the whole trimmed string against the device-name list, so a reserved name carrying an extension (`con.txt`, `com1.log`) slipped through unsuffixed even though Windows reserves a device name regardless of any extension; the guard now tests the segment before the first dot, so `con.txt` is disarmed to `con.txt_` while a non-reserved dotted name such as `data.txt` is left untouched. The collision guard in `.canonicalizeId()` fired on any duplicated canonical id, so `.canonicalizeId(c("Foo", "Foo"))` aborted claiming two distinct ids collide when in fact one id was supplied twice; it now aborts only when a canonical id is reached from more than one distinct pre-image, so an identical repeat passes while a genuine collision from distinct inputs (`c("Foo", "foo")`, `c("a/b", "a_b")`) still aborts. Finally, `.canonicalizeIdRef()` dropped the empty string from canonicalization, so a reference of `""` stayed `""` while the definition side canonicalized `""` to `"_"`, breaking the invariant that the same transform runs identically on both sides; the reference side now canonicalizes `""` the same way (only `NA` still passes through untouched), so a definition and a reference made from `""` resolve to each other again.

## Entity-id and entity-files fixes (#1089 review, round 1 continued)

Five further defects raised on the persistence-engine review are corrected.

`.canonicalizeId()`'s and `.warnCanonicalizedRefs()`'s changed-id warnings pasted raw user text straight into a cli glue template, so an id or reference containing `{`/`}` (e.g. `Conc{Organ}`) crashed with a glue evaluation error instead of warning, and an id containing a valid-looking `{expr}` could execute that expression as R code. Both warnings now build their bullets through a shared helper that binds each pair's input and canonical form under bullet-indexed variables in one environment, and glue-parses them in a single outer `cli_warn()` call, so the quoted text is never re-evaluated by a second glue pass.

`.writeEntityJson()` and `.saveProjectJson()` omitted `digits = NA` from their `jsonlite::write_json()` calls, so every numeric value written to the authoritative per-entity files and the JSON container/snapshot was silently rounded to jsonlite's default of 4 decimal places (`1.234567890123` persisted as `1.2346`). Both now pass `digits = NA`, matching `loadSnapshot()` and the Excel import path, which already did.

`.writeProjectTree()` wrote its container with `includeScenarios = FALSE`, which empties only the `scenarios` array; every other tree-owned section (parameter sets, individuals, populations, applications, observed data, plots, and more) stayed fully inlined, bloating every snapshot-materialized or Excel-imported container with a redundant copy of the whole definitions tree. It now writes with `containerOnly = TRUE`, matching `.persistContainer()`.

`.normalizeSnapshotPath()` compared the `.json` extension case-insensitively but the canonical `.esqlabsR` extension case-sensitively, so `study.ESQLABSR` or `study.esqlabsr` fell into the "foreign extension" branch and triggered a mis-warning even though it is the canonical extension in a different case. The comparison is now case-insensitive for `.esqlabsR` too.

## Authoring and observed-data fixes (#1093 review)

Four correctness defects and one maintainability item raised on the entity-authoring PR are corrected here (the review round on the `pr05-entity-authoring` slice, tracked locally as review round 1 for that slice). Each defect exists identically throughout the migration stack and cannot be fixed inside `pr05-entity-authoring` without breaking the tree-identity contract, so it lands on this layer.

`setIndividual()` and `setPopulation()` stored a supplied `NULL` numeric field as `numeric(0)` instead of clearing it. The set-path loops in `.setOneIndividual()` (`R/individuals.R`) and `.setOnePopulation()` (`R/populations.R`) coerced every numeric field with an unguarded `as.double()`, and `as.double(NULL)` returns `numeric(0)`, a non-`NULL` zero-length vector, so `entry[[field]] <- numeric(0)` kept the key with an empty value rather than deleting it. This contradicted the documented partial-update `.unset` contract (a `NULL` clears an optional field, exactly as the guarded add-path builders already did) and let the field serialize as `[]`, which a later `is.na()`-style check would trip over. A shared `NULL`-aware coercion helper in `R/authoring-vectorize.R` now returns `NULL` for a `NULL` input and `as.double()` otherwise, and both set paths route through it, so a `NULL` clears the key while a value coerces as before.

`addObservedData()` and `removeObservedData()` (`R/observed-data.R`) mutated the runtime `.programmaticDataSets` store before the fallible write-through to disk (`project$.setSection("observedData", ...)`, which aborts on an id/basename collision). If the write threw, the in-memory store and the on-disk sentinels disagreed, breaking the entity-files invariant that memory and disk stay in sync. Both paths now perform the write-through first and mutate the runtime store only after it succeeds, so a failed write leaves the runtime store untouched.

The `parameterSets` validation duplicated across `setApplicationParameterSets()` and `.buildApplicationEntry()` (`R/applications.R`), whose "must be a character vector" messages had already drifted apart, is consolidated behind one behavior-preserving `.resolveParameterSetRefs()` helper, so both application paths now emit a single canonical message and share the undefined-set check.

Two further findings need no code here: `addOutputPath()` already rejects an empty or `NA` path on this branch (guard and test present), and the `.wholeField()` per-entity-versus-applied-to-all ambiguity is deferred pending an API-stance decision.

## Example

Impact not shown as a reprex: the id-canonicalization crash and the numeric-precision loss both require building a `Project` and writing it through entity files, which needs a live project fixture rather than a self-contained snippet.


R/create-plots.R
```diff
@@ -506,12 +530,22 @@ createPlotsFromExcel <- function(...) {
       defaultConfiguration = defaultPlotConfiguration,
       fields = entry[!(names(entry) %in% styleFields)]
     )
+    # Free-text scalar fields are excluded from `styleFields` and re-applied
```
Felixmil, line 543: **Comment is backwards (doc nit).** This says the free-text fields "are excluded from `styleFields`", but `title`/`subtitle`/`xLabel`/`yLabel` are in fact *in* `styleFields` (lines 517-520) — which is precisely why they are excluded from the `.createConfigurationFromEntry()` pass and re-applied verbatim right here. The wording is inverted: a maintainer reading it could conclude the fields are *not* in `styleFields` and "tidy up" by removing them, which would send them back through the comma-splitting scan and re-introduce the label-shredding bug this block exists to prevent. The code is correct; only the comment needs fixing (e.g. "These fields are listed in `styleFields` so the pass above skips them; they are re-applied verbatim here …").

Felixmil (commented):
Review focused on the plots subsystem introduced in #1094 (three keyed kinds `dataCombined`/`plots`/`plotGrids`, the keyed-list `createPlots` runtime, the vectorized mutators), as it stands at this stack tip with the round-1 fixes layered on.

Overall the restructure is well-executed and the round-1 fixes are sound: the NA-safe config override via `isTRUE(all.equal())`, the by-column reconstruction of empty DataCombined rows, the comma-safe `xLabel`/`yLabel` re-application, the escape-aware `plotIds` join/split, and the unknown-`plotType` guard are all good calls. I confirmed the `plotIds` escape machinery round-trips correctly for every id reachable through the public API (canonicalization strips leading/trailing spaces, so the only round-trip mismatch I could construct is unreachable), every referenced `messages$` helper exists, and the subsystem tests pass under `NOT_CRAN=true`.

Three issues, all low-to-medium severity, inline below. None is blocking.

Verified non-issues, for the record: the `.calculateLimits()` empty-positive-log `Inf`/`-Inf` degeneracy is pre-existing (unchanged here, and used only by the sensitivity plots), so out of scope; and `aggregation`/`quantiles`/`nsd`/`foldDistance` are correctly kept in `styleFields` since they are plot-build args, not `DefaultPlotConfiguration` bindings.

Issue #1118, opened 2026-07-17
Excel bridge: correctness & safety


Felixmil:
Umbrella for the JSON ↔ Excel bridge in the v6 project model. Part of #1088; the concrete problems below were surfaced during the #1110 / PR #1117 review. The goal is to fix them together in one PR so the import, export, and sync-status surfaces end up mutually consistent, rather than patching one axis at a time.

## Scope

In v6 a project is JSON-first, and Excel is a secondary I/O bridge: `importProjectFromExcel()` migrates an Excel project to a JSON project, `exportProjectToExcel()` writes the reverse, and `projectStatus(project)` reports whether a configured `Project.xlsx` side-car is still a faithful export of the current project. This umbrella covers the correctness and data-safety of that bridge:

- the sync-status Excel comparison telling the truth about drift;
- `saveProject()` writing to the right container so a status check is not fooled;
- `importProjectFromExcel()` not silently destroying JSON-side work on re-import.

They share a surface (the `.compareJsonToExcel()` engine, the container-naming logic, and the import/export writers) and are cheapest to fix and test as one change.

## Notes

- After PR #1117 the status surface is a single `projectStatus(project)` (the former `syncStatus()`) taking a loaded `Project`; the earlier path-based disk-to-disk comparator was removed. Sub-issue reproductions written against the old path-based call should be read against `projectStatus(project)`.
- Reworking the Excel axis to diff a temp workbook exported from the in-memory project removes both the on-disk re-import step (fixing #1125) and the persisted-container comparison (fixing #1123), so the sync-axis rework and those two are naturally one change.

## References

- Parent umbrella: #1088
- Surfaced in: #1110 / PR #1117 review

Felixmil:
Reopening: migration testing of the 6.0.0 JSON model surfaced new findings on this umbrella's surface, attached as sub-issues below (tested at commit 677f9c8e).

Issue #1110, opened 2026-07-15
Rework project load / snapshot / restore semantics (and reconsider auto-save)


PavelBal:
## Summary

The v6 JSON-first project rework (unreleased) introduces a project **load / snapshot /
restore** model built on **auto-save** (every `add*`/`set*`/`remove*` edit is written to
disk immediately). While reviewing the vignettes and roxygen I found the model confusing
in several ways — both in documentation and in design. This issue collects those notes so
we can decide the intended behavior before release.

The **central question**: should an esqlabsR project behave like a **document you
explicitly save** (explicit `save()`), or like a **live workspace that is always
persisted** (auto-save)? That answer drives most of the points below.

### 1. Auto-save vs. explicit save (core design question)
Every authoring edit persists to disk instantly. Upsides: disk always matches memory, no
lost work. Downsides: **no unsaved-edits concept, no undo, no save-points**, and no
rollback (see #4). I am not convinced auto-save is the right default; at minimum it should
only ship alongside a robust, convenient save/restore chain. I would consider an explicit
`Project$save()` instead. (Note: the v6 rework already removed `project$modified` /
`project$sync()`.)

### 2. Clone semantics are ambiguous
```r
p <- loadProject(".../Project.json")   # edits auto-save to disk (p OWNS the tree)
s <- p$clone()                          # edits are in-memory only (no-op writes)
```
Two objects of the same class behave oppositely, and the clone **retains the source's
`jsonPath`** and prints like a bound project, so it looks saveable but silently discards
edits. `saveSnapshot(s)` then fails with a message ("...already saved, so there is nothing
to save in place") that is **actively wrong for a clone** — the clone has unsaved in-memory
state. Also, `saveSnapshot()` is a one-shot export and does **not** rebind the object, so
after snapshotting a clone, further edits still go nowhere.

### 3. `saveSnapshot()` ergonomics
- `saveSnapshot(project)` with no `path` **always** errors (bound project → "own
  container"; in-memory → "no path"), so `path = NULL` is a trap default.
- The restriction is narrower than it reads: a snapshot **can** live in the project's own
  folder; only overwriting `Project.json` itself is refused. The docs implied otherwise.

### 4. No in-place restore (the save/restore chain is incomplete)
`loadSnapshot(file, dir)` aborts when `dir` already contains a project, so it can only
unpack into a new/empty directory — giving a **separate copy, not a rollback**. Combined
[...]

Issue #1059, opened 2026-06-10
Test suite hygiene: state leaks, package-dir writes, dead fixtures, masked assertions


Felixmil:
Part of the JSON-primary rework (#1088). The original list came from an audit of an earlier branch; most of those items were fixed during the rework. This body has been rewritten to cover only the test-hygiene issues that are still present on the current `Project` model.

## What is already fixed

The file-top `options()` display state in the sensitivity tests, the accidental global-environment leak (`obsData <<-`), the `sourceAll()` test writing into the global environment, the unscoped `setup.R` options, the shared top-level setup in `test-data-combined.R`, and the `skip_if_not(file.exists(...))` guard on a tracked fixture in `test-scenario-from-pkml.R` have all been addressed. The `Rplots.pdf` deposit is gone.

### Writes into the package test directory

`test-sensitivity-calculation.R` still writes `mydata.xlsx` into `tests/testthat/` (a working-directory-relative path) and relies on a trailing `on.exit(unlink(...))` for cleanup. Use `withr::local_tempfile()` so an early failure cannot orphan the file and nothing is ever written into the package directory.

### Bare `set.seed()`

There are around 66 bare `set.seed()` calls across the test suite (concentrated in the sensitivity and plot test files) against only a handful of `withr::local_seed()` usages. Bare `set.seed()` mutates the session RNG state beyond the test. Convert them to `withr::local_seed()`.

### Dead helpers

`testProjectExcelConfigurationsPath()`, `executeWithTestFile()`, and `createValidPISheets()` in `tests/testthat/helpers.R` have no callers anywhere in the package. Remove them.

### Orphan fixtures

`tests/testthat/data/ObsDataAciclovir_1.pkml`, `ObsDataAciclovir_2.pkml`, `ObsDataAciclovir_3.pkml`, and `ProjectConfiguration-V5.xlsx` are referenced by nothing in the suite. Remove them.

### macOS skip with no justification

`test-parallel.R` still carries a `skip_on_os("mac")` with no explaining comment, so `executeInParallel()` has no coverage on macOS. Either remove the skip or document why it is there.

### Duplicated settings tests

`test-esqlabsr-settings.R` and `test-esqlabs-env.R` overlap (both assert `esqlabsRSettingNames` and the package-name/version settings). The functions under test live in `R/esqlabs-env.R`. Merge the two into `test-esqlabs-env.R`.

Felixmil:
[...]

Issue #1034, opened 2026-05-07
Enforce project-folder containment when resolving project-controlled file paths


Felixmil:
## Context

Sub-issue of #1132 (project folder & file layout, under the 6.0.0 umbrella #1088). This is a cross-cutting path-safety concern: it is not specific to any one section. Several project-controlled path fields, spanning observed data, scenario model files, population CSVs, and config-export writes, are joined with a project root and passed to a file loader/writer without verifying that the resolved path stays under the root.

Originally surfaced by @sorinvoicu during review of #1022 (`R/observed-data.R:85`): _"check paths against outside of project references"_, but the same gap exists at every site in the table below.

Sibling issue: #1033 covers the documentation-side mitigation for the observed-data `script` source type (arbitrary code execution by design). This issue covers the defensive path-containment check that complements it for every non-`script` project-controlled path.

## Threat model

A third party publishes a malicious project (GitHub, shared drive, "example project"). The user clones it and runs `loadProject()` followed by any project-driven function. Project field values are author-controlled JSON strings; several of them are joined with a project root and passed to a file loader without verifying that the resolved path stays under the root. Concrete attacker JSON:

```json
{ "type": "pkml", "file": "../../../../etc/passwd" }
```

`file.path("Project/Data", "../../../../etc/passwd")` resolves to `/etc/passwd`, which the loader will then attempt to read. Information disclosure depends on the loader actually parsing the file (e.g. `loadDataSetFromPKML` will fail on non-PKML content), but the read attempt is enough to be a concern, and write paths (config-export) are worse, because the attacker can clobber files outside `configDir`.

Path containment does not stop the `script` source type (which is arbitrary code exec by design, covered in #1033). It does meaningfully reduce the surface for everything else.

## Affected sites

Each of these joins a project-root with a project-author-controlled string and passes the result to a file loader/writer. None of them currently enforces containment. The set spans several sections, which is why this belongs to the folder/layout umbrella rather than any single-section issue.

| File | Line | Root | Attacker-controlled field |
|---|---|---|---|
| `R/observed-data.R` | 89 | `dataFolder` | `entry$file` (and `entry$importerConfiguration` via the same helper) |
| `R/scenario-execution.R` | 188 | `project$modelFolder` | `scenario$modelFile` |
| `R/scenario-execution.R` | 288 | `project$populationsFolder` | `scenario$populationId` (suffixed with `.csv`) |
[...]

PR #1121, opened 2026-07-17
Tidy the Project internal API surface: method-first, private readers, grouped accessors

### LINKED ISSUE #1112: Internal `.`-prefixed methods on the `Project` class are exposed in the public API
## Problem

The `Project` R6 class (`R/project.R`) declares seven methods inside its `public = list(...)`
block that are named with a **leading dot** and documented as internal
(`@keywords internal`, "Not intended for end-user use"):

- `.rebindPath`
- `.markValidated`
- `.markModified`
- `.getSection`
- `.setSection`
- `.getFilePathsData`
- `.getExcelData`

Because they live in the public list, they are part of the object's public API: a user can
call `project$.setSection(...)` directly and they appear in autocomplete on the object. The
"internal" contract is therefore **convention-only** (the leading dot plus the roxygen note),
not enforced. A modeler can reach these and corrupt project state by bypassing the sanctioned
authoring functions (`addScenario()`, `setParameter()`, etc.).

## Why they are public today

These are the package-internal read/write entry points used by the **free authoring
functions** — `add*` / `set*` / `remove*` in `scenarios.R`, `parameters.R`, `plots.R`,
`individuals.R`, `populations.R`, `applications.R`, `output-paths.R`,
`observed-data.R`, `parameter-identification.R`, etc. — plus the Excel bridge and the JSON
writer. That is roughly **93 call sites across 14 files**.

R6 has **no package-private scope**: a `private$` method can only be called from another
method of the same instance, not from a free function elsewhere in the package. So the
"dot-prefixed method in the public list" pattern is the standard R6 workaround for
"package-internal, not user-facing". The finding is real, but the fix is not a trivial
move to `private`.

## Options to weigh

1. **Keep the convention, document it once (cheapest).** Add a short, single statement that
   dot-prefixed members are package-internal and unsupported, and make sure they are excluded
   from the pkgdown reference index. The contract stays unenforced but becomes explicit.
2. **Make the authoring functions methods of `Project`.** The real logic could then live in
   `private$`, and the `.`-prefixed public shims disappear. Largest change; rethinks the
   free-function authoring API.
3. **A "friend accessor" indirection** (e.g. an internal helper reaching a hidden
   environment). Encapsulates, but adds machinery — probably over-engineered for the benefit.

## Scope

Currently **`Project` only**. The package's other R6 class, `ValidationResult`, has no
dot-prefixed public methods. Whatever is decided is worth adopting as a stated convention so
future R6 classes stay consistent.

## Note

Part of the v6 JSON-first project rework (unreleased), so there is freedom to change the
public surface now without a deprecation cycle. This ties into the broader authoring-API /
auto-save design discussion in #1110.

### LINKED ISSUE #1127: Make Project$validatedSinceMutation internal — it is a validation-cache flag, not user API
Part of the v6 JSON-project rework (#1110 line). Related to #1112 (internal members on the public `Project` surface).

## Problem

`Project$validatedSinceMutation` is a public, documented active binding ([project.R:71-81](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/project.R#L71-L81)), but it is an internal optimization detail: a cache flag meaning "a full `validateProject()` has passed since the project was last changed", used only to let automatic validation skip a redundant re-run.

Its only consumers are package internals:

- `.ensureValid()` — the skip-revalidation short-circuit ([validation.R:423](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/validation.R#L423))
- `validateProject()` — sets it via `.markValidated()` ([validation.R:250](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/validation.R#L250))
- the error-rollback in `createScenarioConfigurationsFromPKML()` ([scenario-from-pkml.R:535](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/scenario-from-pkml.R#L535), [570](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/scenario-from-pkml.R#L570))

No vignette or user workflow reads it. Exposing it invites modelers to consult a cache flag as if it were project state, next to the actual user-facing surface: `validateProject()` answers "is my project valid?", and `syncStatus()` / `project$status` answer "what is unsaved or out of sync?". The name also leaks internal vocabulary ("mutation") into the documented API.

## Suggestion

Remove the public active binding and expose the flag internally only — e.g. an internal dot-method `.isValidated()` alongside the existing `.markValidated()` / `.isModified()` family (a dot-prefixed *binding* is not an option: the private field is already named `.validatedSinceMutation`). Update the three internal consumers and the tests that read `project$validatedSinceMutation`.

Since the whole v6 rework is unreleased, the binding can be removed outright without a deprecation cycle (same reasoning as the `ValidationResult` rename).

Whatever convention #1112 settles on for package-internal members of `Project` should cover this field too.

Felixmil:
> **Rebased onto `json-based-project` (post-#1117 merge).** The explicit-save lifecycle work (PR #1117) is now merged into `json-based-project`, and this PR is rebased on top of it, so the diff shown here is only this PR's own delta.

Tidies the `Project` R6 class's internal API surface (umbrella #1129), in three layered steps. First, it makes the class method-first: `Project` gains real public authoring methods and the exported free functions become thin forwarders, removing the `.`-prefixed public methods and the `.__enclos_env__$private` back-door (#1112). Second, it makes the last three package-internal readers genuinely internal (#1127-style): `rawFilePaths`, `rawExcel`, and `isModified` stop being public active bindings and become `@keywords internal` methods that read `private$` directly. Third, it gathers the class's ~30 flat accessors into four named groups (`project$info`, `project$paths`, `project$definitions`, `project$excel`), so the object surface reads as a small set of grouped fields instead of a flat wall of members. This is unreleased v6 work, so the old members are removed outright with no deprecation cycle; every call site in the package, tests, vignettes, and README is migrated.

## Project class: method-first

`Project` exposes a public method for each authoring operation (`project$addScenario(...)`, the whole `add*` / `set*` / `remove*` family, `project$loadObservedData()`, and the lifecycle methods `project$save()` / `project$reload()` / `project$validate()`). Each method body is a one-line forwarder to a `.<name>_impl(self, private, ...)` function in the relevant domain file, so `project.R` stays a thin class and the logic stays where it already was. The read/write seam is a set of genuinely-private `private$` methods reachable only from within a method of the same instance. There are no `.`-prefixed methods in the public list, and there is no `.__enclos_env__` reach anywhere in `R/`.

## Free functions

Every exported authoring function (`addScenario(project, ...)`, `saveProject(project)`, ...) keeps its signature and becomes a thin forwarder: it validates that `project` is a `Project` and calls the matching method. Error messages still name the public function the user called, so an abort reads `Error in addScenario()`, not the internal `_impl`.

## Internal readers made private

Three read-only active bindings that only package-internal code consumed are removed from the object surface and reimplemented as `@keywords internal` methods that read `private$` directly: `rawFilePaths()` (the raw `{value, description}` working-folder records, read by the JSON writer and the Excel exporter), `rawExcel()` (the raw Excel-bridge sheet-name records), and `isModified()` (the in-memory dirty bit, read by the sync reporter). The JSON writer, Excel exporter, and `projectStatus()` now call these methods. Users read the sync state through `project$status` or `projectStatus()`; the dirty bit no longer appears on the object or in autocomplete.

## Grouped accessors

The class's flat accessors are gathered into four groups:

- `project$info` — identity and metadata: `name` and `description` (writable), and read-only `schemaVersion`, `esqlabsRVersion`, `projectFilePath`, `projectDirPath`. The `jsonPath` alias is dropped in favour of the single `projectFilePath`, and the two version fields become read-only (they are set by the load/save machinery, not by users).
- `project$paths` — the working folders `modelFolder`, `dataFolder`, `outputFolder`, `populationsFolder`, plus `definitionsFolder` (all writable). Each folder is resolved against the project directory on read and stored verbatim on write, exactly as the former flat getters were.
- `project$definitions` — the twelve definition sections (`scenarios`, `outputPaths`, `individuals`, ...), read-only from the handle. Assigning any form aborts with a message pointing at the authoring functions.
- `project$excel` — the nine Excel-bridge sheet-name fields (writable).

The writable groups are backed by a single small proxy (`ProjectFieldGroup`): a classed environment whose fields are active bindings built from a per-group spec of get/set closures. The proxy is constructed inside `Project`, and the closures capture `private`, so external code still cannot reach a project's private state, but the proxy can read and write it. Writing a field runs the same invalidation the flat setters did (sets the dirty bit and clears the validation cache). A field group cannot be replaced wholesale (`project$info <- ...` aborts); a read-only field aborts on assignment. Each group prints its own section through the same block renderer `print(project)` uses, so the whole-project summary and a group's own print never drift.

## Documentation and call sites

Every reference to a flat accessor is migrated across `R/`, the test suite, the vignettes, and the README (`project$scenarios` -> `project$definitions$scenarios`, `project$name` -> `project$info$name`, `project$modelFolder` -> `project$paths$modelFolder`, and so on). `NEWS.md` records the grouped surface as part of the v6 breaking-changes section, and `devtools::document()` is clean.

## Example

``` r
pkgload::load_all(quiet = TRUE)

# Grouped reads
project$info$name
#> [1] "Example"
project$definitions$scenarios
#> <DefinitionList>
#> scenarios (3 definitions):
#>   • aciclovir_iv
#>   • aciclovir_iv_population
#>   • aciclovir_iv_steadystate

# Writable fields write back and mark the project as having unsaved changes
project$info$name <- "Study A"
project$paths$modelFolder <- "Models"
project$isModified()
#> [1] TRUE

# A read-only field and a whole-group replacement both abort
try(project$info$schemaVersion <- "9")
#> Error : info$schemaVersion is read-only and cannot be assigned into.
try(project$definitions$scenarios <- list())
#> Error : scenarios is read-only and cannot be assigned into.
#> ℹ To change a definition, edit its '.json' file or use an authoring function
#>   (e.g. `addScenario()` / `setScenario()` / `removeScenario()` and their
#>   per-section siblings).
#> ℹ To edit one record, read it, change the copy, then re-submit it with an
#>   authoring function: `sc <- project$definitions$scenarios[["id"]]; sc$field <-
#>   value; setScenario(project, "id", ...)`.
```

Closes #1112
Closes #1127


R/project-sections.R
```diff
@@ -0,0 +1,143 @@
...
+# as its first argument and reaches its private environment through the single
+# narrow accessor `.projectInternals()`. Because these are plain package
+# functions (not attached to the object), a user cannot call them and cannot
+# reach `private` through them, so no internal capability leaks onto the
+# object's public API.
+
+# The single narrow accessor onto a `Project`'s private environment. Returns the
+# live `private` environment (not a copy), so field mutations by the seam
+# functions below target the object's real state. This is the only place in the
+# package that reaches into an R6 object's `.__enclos_env__`.
```
Felixmil: **Required Changes.** This claims `.projectInternals()` is "the only place in the package that reaches into an R6 object's `.__enclos_env__`" — that's not accurate. Four test files still reach `.__enclos_env__` directly, bypassing the seam:
- `tests/testthat/test-project.R:137,158`
- `tests/testthat/test-validation.R:1173`
- `tests/testthat/test-definition-files.R:310`

Leaving those alone (they target unrelated private fields like `.clean_path`/`.scenarios`) is a reasonable call, and the PR description says as much — but that nuance isn't here. A reader who trusts this comment on its own will conclude the invariant is enforced package-wide when it isn't. Since the entire point of this refactor is "exactly one sanctioned mechanism instead of two," suggest scoping the claim (e.g. "the only place in package `R/` code...") or naming the test-only exceptions.

---

Felixmil (commented):

*(Submitted as a comment review, not "Request changes" — GitHub doesn't allow requesting changes on one's own pull request. Verdict below stands as written.)*

## Summary

This reviews commit `f4da8bc` ("Encapsulate Project internal seam behind free functions"), the change that closes #1112. (The rest of the PR's diff against `pr15-review-round-1` is carried-forward, already-reviewed work from the explicit-save lifecycle rework and is out of scope here.)

Mechanically the refactor is sound: the ~90 call sites rerouting `project$.X(a)` → `.X(project, a)` is complete (verified by grep — zero remaining `project$.` call sites), each free function reproduces its removed method's body without silent behavior change, and nothing new is exported. Standards axis is clean — nothing survived at high confidence. Spec axis has one real defect worth fixing before merge: the seam's central architectural comment makes a claim that is demonstrably false against the same codebase, which undercuts the exact guarantee this PR is supposed to establish.

## Critical Issues (Blocking)

None.

## Required Changes

1. **`R/project-sections.R:22-23`** — see inline comment. The claim that `.projectInternals()` is "the only place in the package that reaches into an R6 object's `.__enclos_env__`" is false: four test files still reach `.__enclos_env__` directly. Reasonable to leave those, but the comment should say so instead of overclaiming.

## Spec Compliance

Faithful to spec on the core ask: all dot-methods #1112 named that still exist at this commit are removed from `Project$public`, and `project$.setSection(...)` etc. are genuinely unreachable (zero call sites remain, no new exports). `.rebindPath`, the one method #1112 named that isn't touched here, was already deleted in an earlier commit on this branch (`1a8b8e2`), so its absence isn't a gap in this diff.

Two things worth flagging that the issue and PR description don't account for:

- Issue #1112's "Options to weigh" calls option 3 (friend-accessor indirection — what's implemented here) *"probably over-engineered for the benefit."* The PR picks it with no note addressing why it beats option 1 (document the convention) or option 2 (methods on `Project`). The issue's "Scope" section also asks that *"whatever is decided is worth adopting as a stated convention so future R6 classes stay consistent"* — no such convention was written down anywhere.
- Regenerating `man/Project.Rd` drops the class's `\keyword{internal}` tag entirely (verified against `daab597:man/Project.Rd`). It was never on the class's own roxygen block — it was inferred purely from the now-deleted methods' `@keywords internal` tags. Net effect: `Project` may now surface in the pkgdown reference index / non-internal help search, a documentation-visibility change the PR description doesn't mention (it only claims "NAMESPACE is unchanged," which is true but doesn't cover the Rd keyword). Worth a maintainer's explicit call on whether that's intended.

## Suggestions

- No regression test asserts the fix's own central claim — that `project$.setSection(...)` (or any of the nine) errors/is no longer reachable on a `Project` instance.

## Verdict
Request Changes (submitted as Comment — see note above)

---

man/syncStatus.Rd
```diff
@@ -0,0 +1,58 @@
...
+is dirty). Reported as \code{NA} for an unbound in-memory project.
+\item memory vs. Excel: when a \code{Project.xlsx} side-car is configured, whether
+it is a stale export of the current project (one-way: would re-exporting
+change it). Reported as \code{NA} when no side-car is configured or it cannot
+be read.
+}
+
+\code{syncStatus()} never reconciles either axis. To sync the tree, call
+\code{\link[=saveProject]{saveProject()}}; to sync Excel, call \code{\link[=exportProjectToExcel]{exportProjectToExcel()}} or
+\code{\link[=importProjectFromExcel]{importProjectFromExcel()}}.
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. This PR (#1121) closes #1112/#1127 (Project internal API surface); the load/snapshot/restore semantics and the "reload after external tree changes" documentation are carried-forward, already-reviewed lifecycle work owned by #1110. Tracking the doc clarification there.

R/project-excel.R
```diff
@@ -777,26 +788,89 @@ projectConfigurationStatus <- function(...) {
...
+}
+
+#' Two-axis sync-status engine behind `syncStatus()` and `project$status`
+#'
+#' Reports both sync axes of an explicit-save project:
+#'   - memory vs. tree: whether there are unsaved in-memory edits, driven by
+#'     the project's internal dirty bit (`NA` for an unbound in-memory project);
+#'   - memory vs. Excel: whether a configured `Project.xlsx` side-car is a stale
+#'     export of the current project, reusing the `.compareJsonToExcel()`
+#'     comparison (`NA` when there is no side-car or it cannot be read).
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. The two-axis `syncStatus()`/Excel-comparison engine is carried-forward Excel-bridge work, owned by #1118 (Excel bridge: correctness & safety). The memory-vs-Excel comparison fix belongs there, not in the #1112/#1127 encapsulation change.

R/project-lifecycle.R
```diff
@@ -17,38 +17,154 @@
...
+  validateIsOfType(project, "Project")
+
+  if (is.null(project$projectFilePath)) {
+    cli::cli_abort(messages$saveProjectNoTree())
+  }
+
+  # The dirty bit is the memory-vs-tree divergence signal. A clean save is a
+  # reassuring, idempotent no-op, never an error.
+  if (!.isModified(project)) {
+    cli::cli_inform(messages$projectAlreadyUpToDate())
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. The dirty-bit-vs-disk messaging wording is carried-forward explicit-save lifecycle work, owned by #1110. Tracking the wording fix there.

R/project-snapshot.R
```diff
@@ -0,0 +1,233 @@
...
+    digits = NA
+  )
+
+  # Read the canonicalized snapshot into an in-memory `Project`, then explode it
+  # into the `definitions/<kind>/` tree at `dir`. Loading the materialized
+  # container back returns a fresh tree-backed `Project` bound to `dir`. The
+  # local is named `inMemory` (not `snapshotProject`) so it does not shadow the
+  # exported `snapshotProject()` function.
+  inMemory <- loadProject(canonFile)
+  containerPath <- .writeProjectTree(inMemory, dir)
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. `project-snapshot.R` path handling is carried-forward snapshot/restore work, owned by #1110 (load/snapshot/restore) and #1132 (project folder & file layout). The traversal guard belongs there, not in the #1112/#1127 encapsulation change.

tests/testthat/helpers.R
```diff
@@ -117,6 +116,47 @@ executeWithTestFile <- function(actionWithFile) {
...
+  # `/tmp/...`. Either way, an unrelated absolute path is left intact.
+  prefix <- .escapeRegex(tmp)
+  # A quoted temp path ending in the snapshot extension keeps that suffix; any
+  # other quoted temp path collapses to a bare placeholder.
+  lines <- gsub(
+    paste0("'", prefix, "[^']*\\.esqlabsR'"),
+    "'<tmp-path>.esqlabsR'",
+    lines
+  )
+  gsub(paste0("'", prefix, "[^']*'"), "'<tmp-path>'", lines)
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. The tmpdir snapshot-redaction helper is test-hygiene, owned by #1059 (test suite hygiene). Tracking there.

tests/testthat/test-definition-files-sections.R
```diff
@@ -79,45 +81,51 @@ test_that("addPopulation / addApplication / addOutputPath write definition files
...
+  # Mutate only setb, then save; only setb.json's mtime should advance.
   addParameterEntry(project, "setb", "Organism", "Param", 1.5, "mg")
-  expect_identical(readLines(siblingFile), sentinel)
-  expect_true(file.exists(file.path(dir, "setb.json")))
+  saveProject(project)
+
+  after <- file.info(list.files(dir, full.names = TRUE))$mtime
+  names(after) <- list.files(dir)
+  changed <- names(after)[after > before[names(after)]]
+  expect_identical(changed, "setb.json")
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. The mtime-resolution flakiness is test-hygiene, owned by #1059. Tracking there.

tests/testthat/test-parameter-identification.R
```diff
@@ -1784,6 +1785,7 @@ test_that("removing the last PI parameter and mapping deletes the task file", {
   )
 
   expect_false("aciclovirsimple" %in% names(project$parameterIdentification))
+  saveProject(project)
   expect_false(file.exists(file.path(dir, "aciclovirsimple.json")))
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. Asserting deferred deletion before save is test-hygiene on carried-forward explicit-save tests, owned by #1059 (with the lifecycle behavior itself under #1110). Tracking there.

tests/testthat/test-project.R
```diff
@@ -906,71 +896,76 @@ test_that("jsonPath is read-only and aliases projectFilePath", {
...
-  expect_identical(clone$individuals[["indiv1"]]$weight, 99)
-  expect_false(identical(project$individuals[["indiv1"]]$weight, 99))
-})
+# The print output carries per-run temp-dir prefixes (the JSON path and the
+# resolved working-folder paths); redact everything up to and including the
+# throwaway `TestProject_<hash>` directory so the class line's marker and the
+# section structure stay reviewable and portable.
+.redactJsonPathLine <- function(lines) {
+  gsub(".*(TestProject_[^/]+)", "<tmp>", lines)
+}
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. The Windows path-separator redaction is test-hygiene, owned by #1059. Tracking there.

tests/testthat/test-scenario-from-pkml.R
```diff
@@ -554,7 +554,7 @@ test_that("end-to-end: seed from PKML, snapshot, loadProject round-trips", {
     scenarios = "seeded"
   ))
 
-  path <- saveSnapshot(project, local_projectPath())
+  path <- snapshotProject(project, dir = withr::local_tempdir())
   reloaded <- loadProject(path)
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. Exercising `restoreProject()` in these round trips is carried-forward snapshot/restore test coverage, owned by #1110 (with test hygiene under #1059). Tracking there.

tests/testthat/test-snapshot.R
```diff
@@ -199,24 +295,75 @@ test_that("loadSnapshot creates dir when absent", {
...
+  # handles on the overwrite action.
+  expect_snapshot(
+    rolledBack <- restoreProject(out, dir, overwrite = TRUE),
+    transform = .redactTmpDir
+  )
+  expect_named(
+    rolledBack$scenarios,
+    names(source$scenarios),
+    ignore.order = TRUE
+  )
```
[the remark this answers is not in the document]
Felixmil: Out of scope for this PR. Exercising an actual rollback before the overwrite assertion is carried-forward snapshot/restore test coverage, owned by #1110 (test hygiene under #1059). Tracking there.

tests/testthat/test-definition-files.R
```diff
@@ -325,61 +295,62 @@ test_that("structural validation rejects a serializer-hostile scenario", {
...
-  # new scenario. The materialize must abort before writing any file, so the
-  # reload still yields the original full set (no partial tree).
+  # Make one loaded scenario serializer-hostile in the backing store (bypassing
+  # the authoring API), then save. The save must abort before writing any file.
   poke <- project$.__enclos_env__$private
   hostile <- poke$.scenarios[[before[[1]]]]
   hostile$simulateSteadyState <- TRUE
   hostile$steadyStateTimeUnit <- NULL
   poke$.scenarios[[before[[1]]]] <- hostile
+  poke$.modified <- TRUE
```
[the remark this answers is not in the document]
Felixmil: Fixed. This test now reaches the section state through the `.getSection()` / `.setSection()` seam helpers instead of `project$.__enclos_env__$private`, so it no longer preserves the back-door this PR removes (`.setSection()` also marks the project modified, replacing the manual `.modified <- TRUE`). Also addressed the related outside-diff rollback finding in `R/scenario-from-pkml.R`: the Phase B rollback now restores the pre-call dirty bit (via `.clearModified()` when the project was clean), so a failed `createScenariosFromPKML()` no longer leaves an initially-clean project reporting unsaved changes.

R/project-sections.R
Felixmil: Moot: `R/project-sections.R` was removed by this PR (the seam is now private `Project` methods), so this comment no longer applies to any file in the diff. Resolving.

R/project-snapshot.R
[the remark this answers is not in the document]
Felixmil: Follow-up: the `definitionsFolder` path-containment concern is now tracked as **#1034** ("Enforce project-folder containment when resolving project-controlled file paths"), a sub-issue of #1132. #1034 covers the general project-path traversal surface (definitions folder, model files, population CSVs, config-export writes), so this finding lands there.

R/project.R
```diff
@@ -898,6 +749,430 @@ Project <- R6::R6Class(
...
+    # closures, and the closures re-read the live backing fields every call).
+
+    # Handle an assignment into a group active binding (`project$info <- ...`).
+    # A nested field write (`project$info$name <- "x"`) mutates the proxy
+    # environment in place through the field's active binding, then R re-assigns
+    # the (same) proxy back into the group binding; that write-back must be
+    # accepted as a no-op. Any other value is a genuine attempt to replace the
+    # whole group, which is not allowed: the group is a live view, not a slot.
+    .acceptGroupWriteback = function(value, group) {
+      isWriteback <- inherits(value, "ProjectFieldGroup") &&
```
Felixmil, line 774: **[correctness, CONFIRMED]** `.acceptGroupWriteback()` accepts any `ProjectFieldGroup` proxy tagged with the matching `group` attribute as a legitimate internal write-back — regardless of which `Project` instance it actually came from — and silently no-ops instead of aborting.

```r
isWriteback <- inherits(value, "ProjectFieldGroup") &&
  identical(attr(value, "group"), group)
```

**Repro:** `p1 <- loadProject(...); p2 <- loadProject(...); p2$paths <- p1$paths`. `isWriteback` is `TRUE` (class and `group` tag both match), so the assignment returns `invisible(value)` with no error and **no change to `p2`** — contradicting the function's own documented contract that a genuine replacement attempt should abort with "field group cannot be replaced wholesale." The guard should verify the proxy's origin (e.g. identity of the captured `private` env), not just its class/tag.

---

R/scenario-from-pkml.R
```diff
@@ -530,22 +578,23 @@ createScenariosFromPKML <- function(
...
       if (length(pending) > 0) {
         addOutputPath(
-          project,
+          self,
           id = names(pending),
           path = unname(pending)
         )
       }
       for (spec in specs) {
         addScenario(
```
Felixmil: **[correctness, CONFIRMED]** `.createScenariosFromPKML_impl()` calls the exported free functions `addOutputPath()`/`addScenario()` reentrantly here (and at line 589) instead of `self$addScenario()`/`.addScenario_impl()` directly. This adds extra call-stack frames that break the `rlang::caller_env(2)` error-attribution convention every `_impl` function relies on.

**Repro:** `createScenariosFromPKML(project, pkmlFiles)` where two PKML-derived scenarios share the same `scenarioName`. The call chain is `createScenariosFromPKML()` → `.createScenariosFromPKML_impl()` → `addScenario()` (free fn, here) → `Project$addScenario()` (R6 method) → `.addScenario_impl()`, whose `rlang::local_error_call(rlang::caller_env(2))` resolves to *this* `addScenario()` call, not the user's `createScenariosFromPKML()` call. The duplicate-id abort in `R/scenarios.R` then prints `"Error in \`addScenario()\`:"` instead of `"Error in \`createScenariosFromPKML()\`:"` — misattributing the error to a function the caller's script never invoked. Consider calling `.addOutputPath_impl()`/`.addScenario_impl()` directly here.

---

R/project.R
```diff
@@ -898,6 +749,430 @@ Project <- R6::R6Class(
...
+    # (resolved) `configurationsFolder`, as the former flat getters did. All are
+    # writable (raw on write, resolved on read); every write sets the dirty bit.
+    .excelGroup = function() {
+      configResolved <- function() {
+        private$.clean_path(
+          private$.excelData$configurationsFolder$value,
+          private$.projectDirPath
+        )
+      }
+      fileField <- function(name) {
```
Felixmil, line 901: **[correctness, PLAUSIBLE]** `project$excel$<field>` (e.g. `modelParamsFile`) used to return the verbatim configured string; the identical expression now returns a fully resolved absolute path via `private$.clean_path(private$.excelData[[name]]$value, configResolved())`.

Confirmed by the old vs. new test assertions: pre-PR `expect_identical(project$excel$modelParamsFile, "ModelParameters.xlsx")`, post-PR the equivalent test expects `fs::path_abs(file.path(projectDirPath, "ModelParameters.xlsx"))`. No caller inside this repo relies on the old raw form (`R/project-excel.R` uses the separate `rawExcel()` accessor, unaffected), but this is a real external-behavior change for the public `project$excel$*` surface — any downstream/external caller reading it for a literal filename (logging, comparison, further path-joining) would silently break or double-resolve. Worth a NEWS.md mention if not already covered.

---

R/project.R
```diff
@@ -33,87 +33,121 @@
...
-      if (!missing(value)) {
-        cli::cli_abort("{.field validatedSinceMutation} is readonly")
-      }
-      private$.validatedSinceMutation
-    },
-
-    #' @field name Human-readable project name (the `name` JSON field). May be
-    #'   `NULL` when the project has no name.
-    name = function(value) {
+    #' @field info Project identity and metadata, as a writable field group.
```
Felixmil, line 98: **[correctness, CONFIRMED]** The public, documented read-only field `project$jsonPath` (an alias of `projectFilePath`) is dropped here with no deprecation shim. This is intentional per the PR description ("the `jsonPath` alias is dropped in favour of the single `projectFilePath`"), and the old test was renamed rather than simply deleted, so the removal itself is clean and deliberate.

The gap: NEWS.md's *existing* line naming `project$jsonPath` specifically was deleted rather than converted into a "dropped" note, and the adjacent new NEWS entry enumerating the `info` group's surviving fields never flags this one as removed. A caller who reads only NEWS.md (not the PR description, which disappears after merge) would not learn `project$jsonPath` now silently returns `NULL` instead of erroring or redirecting. Small ask: add one clause to the NEWS entry.

---

R/scenario-results.R
```diff
@@ -144,7 +144,7 @@ saveScenarioResults <- function(
 #'
 #' # Now load the results, restricting to each scenario's declared output paths
 #' simulatedScenariosResults <- loadScenarioResults(
-#'   scenarios = names(project$scenarios),
+#'   scenarios = names(the scenarios definitions),
```
Felixmil: **[correctness, CONFIRMED, low severity]** A mechanical `project$scenarios` → `project$definitions$scenarios` rename regex also matched plain-English roxygen prose here, producing nonsense text like `names(the scenarios definitions)`.

This one sits inside a `\dontrun{}` `@examples` block, so it's never executed by `R CMD check`/`devtools::document()` — purely a documentation-quality regression, not a build breaker. The same artifact also corrupted `@param`/`@description` prose at `R/parameter-identification.R:1093,1096,1427` and `R/scenarios.R:994`. Worth a quick pass to fix the wording in all four spots.

---

R/scenarios.R
```diff
@@ -728,6 +728,53 @@ addScenario <- function(
...
+  simulationTimeUnit = "h",
+  steadyState = FALSE,
+  steadyStateTime = 1000,
+  steadyStateTimeUnit = "min",
+  overwriteFormulasInSS = FALSE,
+  readPopulationFromCSV = FALSE
+) {
+  # Attribute any abort to the public authoring function the user called
+  # (the free-function forwarder), not this internal `_impl`.
+  rlang::local_error_call(rlang::caller_env(2))
```
Felixmil, line 777: **[reuse, CONFIRMED]** This two-line boilerplate — `rlang::local_error_call(rlang::caller_env(2))` plus its explanatory comment — is copy-pasted byte-identically as the first statement of ~45 `_impl` functions across 12 files.

Every `_impl` is invoked from exactly one place: its corresponding `project.R` method, a bare one-line forwarder `.xxx_impl(self, private, ...)`. That means the call-depth relationship is architecturally constant across all 45 sites (including the reentrant PKML path flagged separately, which routes through the same free-function/method shape). This could be collapsed into one shared private dispatch helper in `R/project.R` (e.g. wrapping the `_impl` call itself), removing 45 duplicated copies that any future `add*`/`set*`/`remove*` method must otherwise remember to reproduce exactly, depth argument included.

---

R/project.R
```diff
@@ -33,87 +33,121 @@
...
+    #'   project name), `description` (free-text description). Read-only fields:
+    #'   `schemaVersion` (schema version declared in the JSON, always `"2.0"`
+    #'   for projects this package loads), `esqlabsRVersion` (informational
+    #'   version string from the JSON), `projectFilePath` (absolute path to the
+    #'   JSON file the project was loaded from, `NULL` for an in-memory
+    #'   project), and `projectDirPath` (the directory containing that file, the
+    #'   base for resolving relative paths). Assigning a writable field sets the
+    #'   dirty bit; assigning a read-only field aborts. The two version fields
+    #'   are managed by the load/save machinery, not by users.
+    info = function(value) {
```
Felixmil, line 110: **[efficiency, CONFIRMED]** `project$info`/`$paths`/`$excel`/`$definitions` rebuild a brand-new `ProjectFieldGroup` proxy (fresh environment + one `makeActiveBinding()` per field) on every single read, with no caching.

Hit repeatedly in hot paths: `R/project-to-json.R` reads `project$info`/`project$definitions` ~10+ times per `saveProject()`; `R/validation.R`'s cross-reference validator and `.warnIfReferenced()` rebuild the `definitions` proxy up to 7 times per validation pass / per single `removeX()` call; `R/definition-files.R`'s `.sectionForKind()` rebuilds it once per definition kind inside a 12-iteration loop. Building each group's proxy once per `Project` instance (the closures already capture `private` by reference and stay live) would give identical semantics without the repeated allocation. Likely microsecond-scale overhead per call — probably not user-perceptible at typical project scale, but architecturally wasteful and easy to fix (e.g. memoize into a private field on first access).

---

R/project.R
```diff
@@ -898,6 +749,430 @@ Project <- R6::R6Class(
...
+      )
+    },
+
+    # `project$paths`: the working-folder paths. Each folder is writable;
+    # reading resolves the stored value against `projectDirPath` (raw on write,
+    # resolved on read), exactly as the former flat getters did. Every write
+    # sets the dirty bit. `definitionsFolder` is stored separately (it is not a
+    # `filePaths` record) and defaults to `"definitions"`.
+    .pathsGroup = function() {
+      folderField <- function(name) {
```
Felixmil, line 863: **[reuse, CONFIRMED]** `folderField()` here (in `.pathsGroup()`) and `fileField()` (in `.excelGroup()`, line 881) are two independently-written, structurally identical get/set/invalidate field-factory closures — differing only in backing store and parent-resolution expression. `.infoGroup()`'s `name`/`description` fields repeat the same skeleton a third time by hand (getter reads a private field, setter writes it then calls `private$.invalidateContainer()`).

No shared per-field factory exists today to unify these — would need a small new helper, not adopting an unused existing one — but a future fourth group's field would likely hand-copy this shape a fourth time otherwise.

---

R/plots.R
```diff
@@ -819,9 +830,9 @@ addPlot <- function(project, id, dataCombined, plotType, ...) {
       perDefinitionFields[[i]]
     )
   }
-  project$.setSection("plots", plotConfig)
-  project$.markModified()
-  invisible(project)
+  private$.setSection("plots", plotConfig)
+  private$.markModified()
```
Felixmil, line 834: **[simplification, CONFIRMED, harmless]** This `_impl` (and five siblings at lines 911, 1014, 1058, 1161, 1239) calls `private$.markModified()` immediately after `private$.setSection()`, which already performs the identical work internally (`private$.modified <- TRUE; private$.invalidate()` — `.markModified()`'s body is textually the same). `R/plots.R` is the sole outlier among ~8 other domain files that also call `.setSection()` and don't add this extra call.

It's a no-op re-assignment of already-true state, not a correctness bug — but the inconsistency could mislead a future maintainer into thinking `.setSection()` alone is insufficient, or into copying the redundant call into a new mutator. Worth deleting the six redundant calls for consistency.

---

R/project.R
```diff
@@ -898,6 +749,430 @@ Project <- R6::R6Class(
...
+      private$.validatedSinceMutation <- TRUE
+      invisible(self)
+    },
+
+    # Read the validation-cache flag: `TRUE` when a full validation has
+    # succeeded since the last mutation or load, so the pre-op validation gate
+    # (`.ensureValid()`) can skip a redundant re-run. The mirror of
+    # `.isModified()` for the validation axis; a genuinely-private method with
+    # no public binding, so the flag never appears on the object surface.
+    .isValidated = function() {
```
Felixmil, line 1129: **[correctness, CONFIRMED]** The previously public, documented field `project$validatedSinceMutation` (a real `@field` in the old roxygen, not `@keywords internal`) is removed entirely here with no public replacement — unlike its sibling `isModified`, which was deliberately kept public via `project$isModified()`.

Old code calling `project$validatedSinceMutation` now silently gets `NULL` back (R6/environment `$` on an undefined name returns `NULL`, not an error) instead of `TRUE`/`FALSE` — any `if (project$validatedSinceMutation)` check misbehaves or throws "argument is of length zero." This `.isValidated()` method is deliberately private with no public binding, so the flag never appears on the object surface anymore. The old read-only test for this field was deleted outright rather than migrated, and NEWS.md documents other public-surface changes from this PR but never mentions this removal. Worth at least a NEWS.md line calling out the breaking removal.

---

Felixmil (commented):
Automated xhigh-effort code review of this PR's delta against `json-based-project` (10 finder angles → verification → gap sweep). 10 findings posted inline: 5 correctness issues (most notably a silent no-op in `.acceptGroupWriteback()` for cross-project group assignment, and an error-misattribution bug in the PKML scenario-import path), 2 undocumented breaking removals (`validatedSinceMutation`, `jsonPath`), 1 cosmetic doc-rename artifact, and 3 cleanup items (duplicated boilerplate, proxy rebuild overhead, one harmless redundant call). No CLAUDE.md conventions apply to this repo. Full detail in each inline comment.

---

R/project.R
Felixmil: Fixed in 558ebd7a. `.acceptGroupWriteback()` now also compares the proxy's `owner` attribute (the owning instance's `private` env, by identity) to `private`, so a proxy from a different `Project` is rejected. Verified: `p2$paths <- p1$paths` now aborts, while a same-instance nested/captured-handle write-back still no-ops correctly.

R/scenario-from-pkml.R
Felixmil: Fixed in 558ebd7a. `.createScenariosFromPKML_impl()` now calls `.addOutputPath_impl()` / `.addScenario_impl()` directly instead of the exported free functions, so a failure is attributed to the operation the user invoked. The snapshot now reads `Error in \`project$createScenariosFromPKML()\`:` instead of `Error in \`addScenario()\`:`.

R/project.R
Felixmil: The resolved-on-read behavior is the intended design for the grouped surface (read resolves against the project dir, write stores verbatim; `rawExcel()` / `rawFilePaths()` expose the raw strings). Not adding a NEWS line: the whole v6 `Project` surface is unreleased, so this is intra-cycle churn, not a released-behavior delta.

R/scenario-results.R
Felixmil: Fixed in 558ebd7a. Corrected the mangled prose in all the flagged spots: `R/scenario-results.R:147`, `R/parameter-identification.R:1093/1096/1427`, and `R/scenarios.R:992/994` now read `names(project$definitions$scenarios)` / `project$definitions$outputPaths` / `sc <- project$definitions$scenarios[[name]]`. Swept the rest of R/ for the same artifact; none remain.

R/scenarios.R
Felixmil: Valid but out of scope here. The `local_error_call(caller_env(2))` boilerplate is from the method-first work (#1112, commit d9605511), not the accessor grouping, and collapsing it into a shared dispatch helper touches all ~45 `_impl` sites across 12 files and shifts error-attribution frame depths API-wide, too broad and risky to fold into this change. Better as its own follow-up; happy to file one.

R/project.R
Felixmil: Deferring by design. Building a fresh proxy per read was a deliberate choice (avoids the identity/staleness questions a cached proxy raises) and, as noted, the cost is microsecond-scale and not user-perceptible. The actual hot paths are already addressed: I hoisted the definitions section out of the per-iteration loops in `R/validation.R` and `R/scenario-execution.R` (commit 13ec7734). Not memoizing the proxy itself, to keep the fresh-each-access semantics.

R/project.R
Felixmil: Fixed in 558ebd7a. Extracted `.recordFieldSpec(store, name, parentFn)` and use it from both `.pathsGroup()` and `.excelGroup()` in place of the duplicated `folderField` / `fileField` factories. Left `info`'s `name`/`description` as-is: they are a plain-scalar shape (no `{value, description}` record, no path resolution), so folding them into the record factory would over-generalize it.

R/plots.R
Felixmil: Valid but out of scope / pre-existing. The redundant `.markModified()` after `.setSection()` in `R/plots.R` predates this PR (present on the base at 743ee3a9, six sites), and it is unrelated to the internal-API-surface work here. Leaving it for a separate cleanup to keep this PR's diff scoped; happy to file one.

R/project.R
Felixmil: The removal is intentional (`validatedSinceMutation` was the validation-cache flag; #1127 makes it internal-only, read via the private `.isValidated()`). Not adding a NEWS line: the v6 `Project` surface is unreleased, so this is intra-cycle churn rather than a released-API removal (same reasoning applied to the other unreleased-surface changes in this PR).

R/scenarios.R
Felixmil: Tracked as #1136 (sub-issue of the #1129 umbrella).

R/project.R
Felixmil: Tracked as #1137 (sub-issue of the #1129 umbrella).

R/plots.R
Felixmil: Tracked as #1138 (sub-issue of the #1129 umbrella).

R/project.R
Felixmil: Right that `project$jsonPath` is dropped in favour of `project$info$projectFilePath`, and the old test was renamed rather than deleted. On the NEWS ask: the existing line that named `project$jsonPath` has been removed (it was the `\`project$jsonPath\` is also read-only` clause in the read-only-accessors bullet, reconciled away), so there is no dangling reference. Not adding a "dropped" note, though: the whole v6 `Project` surface is unreleased, so `jsonPath` never shipped and its removal is intra-cycle churn, not a released-API change (same convention applied to the other unreleased-surface changes in this PR). Nothing in NEWS should reference it either way.

Issue #1033, opened 2026-05-07
Warn users that `type = "script"` observed-data sources execute arbitrary R code


Felixmil:
## Context

Sub-issue of #1030 (umbrella for follow-ups from JSON-primary chapter PRs). Surfaced by @sorinvoicu during review of #1022 (R/observed-data.R:121 and :85).

## Current state on `plots-from-project` (PR #1022)

`loadObservedData()` resolves `type = "script"` observed-data sources by calling `source(filePath, local = TRUE)` with no warning to the user, no opt-in flag, and no documentation pointing out the trust implication:

```r
result <- source(filePath, local = TRUE)$value
```

The threat model that motivates this is: a third party publishes a malicious project (GitHub, shared drive, "example project"); a user clones it and runs `loadProject()` + `loadObservedData()` (or any downstream call that resolves observed data, e.g. `createDataCombined()` / `createPlots()`); attacker R code executes on the user's machine.

This is consistent with the package's existing trust model — `sourceAll()` (R/utilities-file.R) and the historical PI-script workflow already source user-supplied R from a project — so the chapter PR is not introducing a new attack surface. But none of those entry points carries a documented warning either.

The path-resolution helper `.resolveDataPath()` (R/observed-data.R:85) does plain `file.path(dataFolder, file)` with no containment check. A malicious JSON could declare `entry$file = "../../somewhere"`. In practice, the `script` source already provides arbitrary code exec, so containment offers nothing extra against an attacker who controls the JSON. It is mentioned here so this issue resolves both review comments.

## End-state (`json-as-primary-input-v2`)

End-state keeps the same `source(filePath, local = TRUE)$value` call (`R/observed-data.R:248`) and the same plain `file.path()` resolution (lines 220, 236, 241). No security warning prose in the function docs or vignettes. The design decision to support the `script` source type is locked in.

## Resolution

Add documentation-only mitigations in a follow-up chapter:

1. `?loadObservedData`: a short security note (e.g. a `@section Security:` block) flagging that `type = "script"` sources execute arbitrary R code, and that users should only run projects from sources they trust.
2. The Project schema docs / vignette: same warning on the `script` source-type description.
3. Consider a one-line `cli::cli_warn()` the first time a project sources a `script`-type observed-data entry — gated so it is not noisy for the legitimate case (e.g. fires once per session, or only when `interactive()`). Open question whether this is worth the noise; decide during implementation.

[...]

PR #1134, opened 2026-07-20
Settle v6 project folder & file layout (rename modelFolder, README scaffold, path containment)

### LINKED ISSUE #1029: Decide Models/ folder layout: simulationsFolder vs modelFolder
Open design discussion from #1017, never resolved. End-state currently keeps `modelFolder = "Models/Simulations/"` but neither side of the discussion has been picked.

Surfaced in:
- https://github.com/esqLABS/esqlabsR/pull/1017#discussion_r3182734421 — @PavelBal: rename to `simulationsFolder`?
- https://github.com/esqLABS/esqlabsR/pull/1017#discussion_r3182887288 — @PavelBal: idea is to store all model types (PK-Sim/MoBi projects, future RxODE) under `Models/`, with pkml simulations in a `Simulations/` subfolder
- https://github.com/esqLABS/esqlabsR/pull/1017#discussion_r3182935069 — @Felixmil: `Models/PKML/` + `pkmlFolder`?

## Decision needed

Pick one before the next breaking schema change, since this affects `Project.json` field naming:

1. **Status quo** — keep `modelFolder` pointing at `Models/Simulations/`. Simple, but the field name doesn't reflect that it holds *simulations* specifically rather than "models" in general.
2. **Rename to `simulationsFolder`** — describes the contents accurately. Frees `Models/` as a future container for other model types (RxODE, etc.) without overloading `modelFolder`.
3. **Subtype subfolders** — `Models/PKML/` + `pkmlFolder`, `Models/RxODE/` + `rxodeFolder` (or similar). Most flexible, most ceremony.

## Scope

Schema decision. Once decided, fixture (`inst/extdata/projects/Example/Project.json`) and any field references update accordingly.

### LINKED ISSUE #1034: Enforce project-folder containment when resolving project-controlled file paths

### LINKED ISSUE #1108: V6 user review - folder structure
1. When a new project is created, it contains several empty folders (e.g., Results, Results/Figures). These folders would not be tracked on GitHub if they are empty. In v5 projects, we supplied README.md files to keep these folders under version control. I think this should be done for v6, too.
2. Why was the "Populations" folder moved outside of "Configurations"? I am not requesting a change; I just want to understand the rationale.
3. The ProjectConfiguration has a field "modelFolder" that points to "Models/Simulations", so it is already an inconsistency.
  a) Either rename to "simulationsFolder", or point to "Models" folder itself
  b) We will introduce the option to load not only from simulation pkml exports, but also from
     - PK-Sim (and later MoBi) snapshots (json), most likely located in "Models/Snapshots" folder
     - Snapshots from github
     - Modules (as pkml exports) from the "Modules" folder
     - MoBi projects (mbp3, most likely living in "Modules" folder)
  So we could store only the "Models" folder in the configuration and then select the sub-folder based on the type we are loading, or specify them all explicitly in the configuration. Or other suggestions?
3. The "Models" folder should have a "Snapshots" subfolder (even if emtpy for now)
4. Does the project configuration still support variables for folder definitions? Will it still be possible to move the "Data" folder to SharePoint and use the `PROJECT_DATA_FOLDER` variable?

### LINKED ISSUE #1132: Project folder & file layout
Umbrella for the on-disk project folder and file layout in the v6 project model. Part of #1088.

The v6 project directory raises several layout questions that should be settled together before the next breaking schema change: keeping empty scaffold folders under version control, the placement of the `Populations` folder, the `Models/` layout and its `Project.json` field naming, and room for future model sources (PK-Sim / MoBi snapshots, modules). These are one coherent layout decision plus the fixture and field updates that follow from it.

Felixmil:
Settles the on-disk project folder and file layout for the v6 project model, resolving the three sub-issues of the layout umbrella together so the fixture and field changes land as one coherent decision: the `Models/` folder is reorganized and its `Project.json` field renamed, empty scaffold folders are kept under version control, and every project-controlled file path is now confined to the project folder. Targets `json-based-project` (rebased on top of it after the method-first work #1121 merged).

## Models/ folder layout and field rename (#1029)

The working-folder field `modelFolder` is renamed to `simulationsFolder`, still resolving to `Models/Simulations` and read through the grouped `project$paths$simulationsFolder` accessor. The old name described the field as holding "models" when it holds simulations specifically; the new name frees `Models/` to hold other model sources (snapshots, modules) under sibling fields added when that work lands, rather than overloading one field. Because v6 is unreleased there is no released name to migrate from, but a legacy `Project.json` (or an Excel-imported project) whose `filePaths` still carries `modelFolder` is accepted on load and stored under `simulationsFolder`, so existing on-disk projects keep loading with no manual edit. A file that carries both keys warns and keeps `simulationsFolder` rather than letting iteration order decide.

## Folder scaffold, README placeholders, and Models/Snapshots (#1108)

`initProject()` now writes a short `README.md` into each working folder it scaffolds, so an otherwise-empty folder (`Results/Figures`, `Results/SimulationResults`, `Populations`, ...) stays tracked under version control instead of disappearing from git, matching the v5 behavior. An existing `README.md` a user has edited is left untouched on a rerun (including `overwrite = TRUE`), so the placeholder never clobbers a user's own notes. The same placeholders ship in the bundled Example project. A `Models/Snapshots` folder is added to both the scaffold and the Example fixture; it is for PK-Sim and MoBi snapshots and is reserved for a future release (the package does not load from it yet), but it is part of every project's structure.

The environment-variable form for working-folder paths (`${PROJECT_DATA_FOLDER}` and friends) is confirmed working and now documented: a variable may sit anywhere in a `filePaths` path and an unset variable is left in place, so a folder such as `Data/` can point at a shared drive per machine. The `Populations/` folder sits at the top level (alongside `Data/` and `Models/`) rather than under the optional `Configurations/` folder, because population CSVs need a home that is always present even in a JSON-only project that has no `Configurations/`.

## Project-folder path containment (#1034)

A project loaded from an untrusted source (a shared drive, a downloaded example) carries author-controlled path strings that were joined with a project folder and handed to a file loader without checking that the result stays inside the project. A value such as `"../../../../etc/passwd"` resolved outside the project and the loader would attempt to read it. Two internal helpers in `R/file-utils.R` close this: `.pathEscapesRoot()` is a purely-lexical containment predicate, and `.resolveProjectPath()` resolves a path against its root and aborts (naming the offending field) unless the result stays under the root. The check runs before the existence check, so a traversal is rejected with a clear message rather than a misleading "file not found".

Containment is enforced at two levels:

- **Leaf paths** joined onto a working folder: a scenario's `modelFile`, a CSV population id, and an observed-data source's `file` / `importerConfiguration`.
- **The working folders themselves** (`simulationsFolder`, `dataFolder`, `populationsFolder`, `outputFolder`) read from `Project.json` must resolve under the project directory, so an author cannot point a folder at an arbitrary location (`"dataFolder": "/etc"`) and then reference a plainly-"contained" file inside that attacker-chosen root. The `${VAR}` form remains the sanctioned way to place a folder outside the tree and is exempt.

`validateProject()` reports the same escape as a critical error, so it agrees with load/run behavior instead of giving false assurance, and `createScenariosFromPKML()` warns when a PKML kept outside `simulationsFolder` would be stored as an escaping relative `modelFile`. Containment is a lexical check on `..` and absolute paths; a symlink inside the project tree is resolved by the file loader as before.

Two write sites named in the issue (`R/utilities-config-json.R`) no longer exist on this branch (that file was removed and its logic folded into the Excel bridge, which joins only literal filenames), so there is nothing author-controlled to confine there. The `importProjectFromExcel()` entry point resolves paths from the same author-controlled Excel `Property` column and is not migrated here to keep this PR scoped; it is tracked as #1139 under the Excel-bridge umbrella #1118. The observed-data `script` source type remains arbitrary code execution by design and is out of scope here (its documentation-side mitigation is the sibling issue #1033); an absolute `modelFile` produced by `createScenariosFromPKML()` keeps its existing verbatim-use branch.

## Example

``` r
pkgload::load_all(quiet = TRUE)

# initProject() scaffolds a project whose empty folders stay tracked: each gets
# a README.md, and a Models/Snapshots folder ships for PK-Sim / MoBi snapshots.
dir <- withr::local_tempdir()
initProject(destination = dir, type = "minimal", createExcel = FALSE)
fs::dir_tree(dir)
#> /tmp/RtmpXXXX/projectXXXX
#> ├── Data
#> │   └── README.md
#> ├── Models
#> │   ├── Simulations
#> │   │   └── README.md
#> │   └── Snapshots
#> │       └── README.md
#> ├── Populations
#> │   └── README.md
#> ├── Project.json
#> ├── Results
#> │   ├── Figures
#> │   │   └── README.md
#> │   └── SimulationResults
#> │       └── README.md
#> └── definitions

# The simulations folder is now `simulationsFolder` (was `modelFolder`),
# under the grouped `paths` accessor.
names(project$paths)
#> [1] "dataFolder"        "definitionsFolder" "populationsFolder"
#> [4] "outputFolder"      "simulationsFolder"

# A project-controlled path that escapes the project folder is rejected
# (validateProject() flags it before the run, matching load/run behavior).
sc_project <- loadProject(exampleProjectPath())
setScenario(sc_project, "aciclovir_iv", modelFile = "../../../../etc/passwd")
try(runScenarios(sc_project, scenarios = "aciclovir_iv"))
#> Error : Cannot runScenarios: project has 1 critical validation error.
#> ✖ [scenarios] Scenario 'aciclovir_iv' references a modelFile outside the
#>   project folder: ../../../../etc/passwd
#> ℹ Run `validateProject(project)` for a full report.
```

</details>

Closes #1029
Closes #1108
Closes #1034
Closes #1132


R/project-lifecycle.R
```diff
@@ -366,6 +357,43 @@ initProject <- function(
...
+.scaffoldProjectFolders <- function(destination) {
+  for (folder in names(.projectScaffoldFolders)) {
+    dir.create(
+      file.path(destination, folder),
+      recursive = TRUE,
+      showWarnings = FALSE
+    )
+    readmeText <- .projectScaffoldFolders[[folder]]
+    if (!is.na(readmeText)) {
+      writeLines(readmeText, file.path(destination, folder, "README.md"))
```
Felixmil: **Silent data loss: this unconditionally overwrites `README.md` in every working folder, on every `initProject()` call.**

`.scaffoldProjectFolders()` is called unconditionally from `initProject()` (including `overwrite = TRUE` reruns on an *existing* project directory), and this `writeLines()` has no existence check. `.clearProjectArtifacts()`'s own doc comment states the invariant this is supposed to honor: "Everything else in `destination` (working folders, and any unrelated user file) is left untouched" — but a user who naturally edits `Data/README.md` (or any of the other 5 scaffold READMEs) to document their own project loses that content the moment `initProject(overwrite = TRUE)` is rerun, e.g. to pick up a newer template. It even happens on a *first* `initProject()` call into a directory that already happens to contain a file named `README.md` in one of these folders, since `isProjectInitialized()` doesn't check working folders at all.

The new regression test (`test-project-lifecycle.R`, "keeping unrelated user files") doesn't catch this — its `userFile` is a top-level `my_notes.txt`, never inside a working folder and never named `README.md`.

Suggest: only write the placeholder if the file doesn't already exist (`if (!file.exists(...))`).

---

R/file-utils.R
```diff
@@ -58,6 +58,49 @@ readExcel <- function(path, sheet = NULL, ...) {
...
+#' @param path Author-controlled path, relative to `root`.
+#' @param root Project folder the path must stay under. Need not exist yet;
+#'   containment is checked lexically so a declared-but-not-created folder is
+#'   fine.
+#' @param fieldName Name of the project field `path` came from, used in the
+#'   error message.
+#' @returns The resolved absolute path (a string), on success.
+#' @keywords internal
+#' @noRd
+.resolveProjectPath <- function(path, root, fieldName = "path") {
```
Felixmil: Two things worth calling out about this new containment guard:

1. **The root itself is never checked against `projectDirPath`.** This function (correctly) checks that `path` stays under `root` — but `root` here is `project$dataFolder` / `project$simulationsFolder` / `project$populationsFolder`, each resolved by `Project$private$.clean_path()` (R/project.R ~1377-1401), which happily accepts an absolute path or a `../`-escaping relative path from the same untrusted `Project.json` with **no containment check against the project directory**. So a crafted `Project.json` doesn't even need a `../` in `modelFile`/`file` — it can just set `"dataFolder": "/etc"` directly, and `"file": "passwd"` then "resolves inside" that (attacker-controlled) root, defeating the fix. Given the whole point of this PR is closing this exact class of issue, I think the root fields need the same containment check (against `projectDirPath`) that this function gives to the leaf paths.

2. **Symlinks bypass this by design** (documented in the roxygen block above — lexical-only, no symlink resolution). That's a reasonable tradeoff to make explicitly, but it does mean NEWS.md's "can no longer read a file elsewhere on the machine via a `../` path" slightly overstates what's closed: a project tree containing a symlink (e.g. cloned via git, which preserves symlinks) still lets an author-controlled path escape at read time.

Line 98 specifically: `paste0(absRoot, sep)` will double-slash if `absRoot` is a filesystem root (`"/"`, or `"D:/"` on Windows), which would make `startsWith()` never match and wrongly reject every legitimate path under such a root. Low real-world odds a project root is literally `/`, but worth a one-line guard (`absRoot == "/"` case) for correctness.

---

R/scenario-execution.R
```diff
@@ -223,12 +223,18 @@
...
+  # is resolved against the project's simulations folder, which must exist for
+  # the join to be meaningful (`file.path(NULL, x)` yields `character(0)`). A
+  # relative value must stay under the simulations folder, so a shared project
+  # cannot point `modelFile` at a file outside itself via `../`.
   if (fs::is_absolute_path(scenario$modelFile)) {
     modelFilePath <- scenario$modelFile
-  } else if (!is.null(project$modelFolder)) {
-    modelFilePath <- file.path(project$modelFolder, scenario$modelFile)
+  } else if (!is.null(project$simulationsFolder)) {
+    modelFilePath <- .resolveProjectPath(
```
Felixmil, line 236: **Possible regression for a previously-supported layout.** If a scenario's PKML lives outside `simulationsFolder` (e.g. models kept on a shared drive alongside `Models/Simulations`), `createScenariosFromPKML()` (R/scenario-from-pkml.R, `fs::path_rel(pkmlPath, start = self$simulationsFolder)`) will happily compute and store an escaping relative `modelFile` like `"../../SharedModels/Custom.pkml"` with no warning — unlike the `NULL`-`simulationsFolder` branch, which already falls back to an absolute path with a `cli_warn`. `validateProject()` won't catch it either (see the `R/scenarios.R:358` comment). The scenario is added successfully and only aborts here, at `runScenarios()` time, with "modelFile ... resolves outside the project folder" — a confusing failure for something that used to just work.

Worth deciding explicitly: should `createScenariosFromPKML()` warn (like the absolute-path branch does) or refuse to store an out-of-root relative path, rather than let it silently round-trip until the next run fails?

---

R/scenarios.R
```diff
@@ -353,8 +354,8 @@ print.Scenario <- function(x, ...) {
         "Missing Fields",
         paste0("Scenario '", name, "' has no modelFile")
       )
-    } else if (!is.null(modelFolder)) {
-      modelFilePath <- file.path(modelFolder, sc$modelFile)
+    } else if (!is.null(simulationsFolder)) {
+      modelFilePath <- file.path(simulationsFolder, sc$modelFile)
```
Felixmil: `.validateScenarios()` (the `validateProject()` adapter) still joins `modelFile` with plain `file.path()`, not `.resolveProjectPath()` — same gap in `.validateObservedData()` (R/observed-data.R:150,161) for `file`/`importerConfiguration`. The actual security backstop is sound (`.prepareScenario()` / `loadObservedData()` never route through `validateProject()`, so a traversal is always rejected before any file is touched), but this means `validateProject()` gives **false assurance**: if an escaping `modelFile`/`file` happens to exist on disk, this existence-only check reports zero issues, while the identical field aborts at load/run time. Given this PR's entire purpose is closing this class of gap, might be worth also flagging it here (or explicitly documenting that `validateProject()` intentionally doesn't check containment).

---

R/project.R
```diff
@@ -1438,10 +1441,18 @@ Project <- R6::R6Class(
       private$.filePathsData <- list()
       private$.excelData <- list()
       for (n in names(fp)) {
-        if (n %in% .excelFilePathFields) {
-          private$.excelData[[n]] <- list(value = fp[[n]], description = "")
+        # Accept the pre-6.0.0 key `modelFolder` and store it under the current
+        # name `simulationsFolder`, so a legacy `Project.json` (or an
+        # Excel-imported project whose `Property` column still says
+        # `modelFolder`) resolves without a manual edit.
+        key <- if (identical(n, "modelFolder")) "simulationsFolder" else n
```
Felixmil, line 1352: Minor edge case: if a `Project.json` somehow carries both the legacy `modelFolder` key and the current `simulationsFolder` key in the same `filePaths` block, whichever one `names(fp)` iterates last silently wins with no conflict warning. Unlikely in practice (nothing writes both), but since this loop is explicitly handling untrusted/hand-edited input, a `cli_warn` on the collision might be cheap insurance.

---

R/project-lifecycle.R
```diff
@@ -366,6 +357,43 @@ initProject <- function(
...
 }
 
+# The working folders a project ships with, each mapped to the one-line
+# `README.md` written into it. The README keeps the otherwise-empty folder
+# tracked under version control (git does not track empty folders) and tells
+# the reader what belongs there. `definitions/` is created without a
+# placeholder because it holds the authored project content, never empty in a
+# real project. Order is deepest-first only for readability; `dir.create()`
+# with `recursive = TRUE` creates parents regardless.
+.projectScaffoldFolders <- c(
```
Felixmil: Nit: this named character vector conflates "folders to create" with "folders needing a README" via an `NA_character_` sentinel for `definitions`. A plain character vector of folder names plus a separate readme-text lookup (or just special-casing `definitions` in the loop) would avoid the sentinel and read a bit more directly — minor, not blocking.

---

R/messages.R
```diff
@@ -242,10 +242,10 @@ messages$scenariosAddedToProject <- function(scenarioNames) {
 
 messages$noModelFolderUsingAbsolutePath <- function(pkmlPath) {
```
Felixmil: Cosmetic: `messages$noModelFolderUsingAbsolutePath()` and `messages$noModelFolderForRelativeModelFile()` (below) keep their old "ModelFolder" identifiers even though their message text was updated to say `simulationsFolder`. Purely internal/unexported, so no functional impact — just flags for a future `grep modelFolder` rename-completeness pass.

---

Felixmil (commented):
Reviewed at xhigh effort (10 finder angles + verification). Solid PR overall — the containment fix, rename, and README scaffold are each well-tested for the paths they cover. Inline comments cover 7 issues anchored to this diff's own lines (most notably a silent README.md clobber in `.scaffoldProjectFolders()`, and a `validateProject()`/execution-time inconsistency around the new containment check). Two more issues don't anchor to lines *this* diff touches, but matter for the PR's stated goal of closing path-traversal, so flagging here instead:

**1. The working-folder roots themselves aren't containment-checked (R/project.R, `Project$private$.clean_path()`, ~line 1377-1401).** `.resolveProjectPath()` correctly checks that `modelFile`/`populationId`/observed-data `file` stay under their declared folder — but the folder itself (`dataFolder`, `simulationsFolder`, `populationsFolder`) comes from the same untrusted `Project.json` via `.clean_path()`, which accepts an absolute path or a `../`-escaping relative path with no check against `projectDirPath`. A crafted `Project.json` can set `"dataFolder": "/etc"` directly and reference a plain `"file": "passwd"` — trivially "contained" by the new check, but the containment boundary itself was attacker-chosen. Given the PR's whole purpose is closing this class of issue, the root fields probably need the same treatment as the leaf paths.

**2. `importProjectFromExcel()`'s path resolution (R/project-excel.R, `configsFolder`/`resolveConfigFile()`, ~lines 87-104) wasn't migrated.** It still uses plain `file.path()` + `normalizePath(mustWork = FALSE)` for `configurationsFolder` and the per-section Excel file properties, read from the same kind of author-controlled `Property` column this PR closes off elsewhere. Same traversal shape, different entry point.

**3. Heads-up, not a bug in this diff: an imminent merge hazard with sibling PR #1121.** This PR is based on `1112-project-private-methods` at `743ee3a` — a point in #1121's branch that predates its later "grouped accessors" commits. Those commits hardcode `project$paths$modelFolder` (`.pathsGroup()`, `.prepareScenario()`, `createScenariosFromPKML()`, `.scenariosValidatorAdapter()`, the print block) against the storage key this PR renames to `simulationsFolder`. Whichever of #1121/#1134 gets rebased onto the other, every one of those becomes a silent NULL-read (not an error) — breaking relative-`modelFile` resolution project-wide. Worth flagging in #1121 now so it's fixed at reconciliation time rather than discovered after merge.

---

R/project-lifecycle.R
Felixmil: Fixed. `.scaffoldProjectFolders()` now writes a folder's `README.md` only when one does not already exist (`if (!file.exists(...))`), so a user-edited placeholder survives an `initProject(overwrite = TRUE)` rerun and the "working folders are left untouched" invariant holds. Added a regression test that edits `Data/README.md`, reruns `initProject(overwrite = TRUE)`, and asserts the edited content is preserved.

R/file-utils.R
Felixmil: Addressed all three points.

1. **Root folders now contained.** Working-folder values (`simulationsFolder` / `dataFolder` / `populationsFolder` / `outputFolder`) read from `Project.json` are now checked against the project directory: a bare absolute (`"dataFolder": "/etc"`) or `../`-escaping folder value aborts on read (new `Project$private$.resolveWorkingFolder()`). The `${VAR}` env-var form remains the sanctioned way to point a folder outside the tree, so it is exempt. Tests cover both the rejection and the `${VAR}` allowance. The `importProjectFromExcel()` entry point you flagged in PR-level point 2 is the one path I did not migrate here (its `configurationsFolder` / workbook-property resolution) — tracked as #1139 under the Excel-bridge umbrella #1118 to keep this PR scoped.

3. **Double-slash guard added** on line 98: the contained-child prefix now special-cases a root that already ends in the separator (`/`, `D:/`), so a filesystem-root project directory no longer wrongly rejects every child.

2. **Symlink caveat** is a deliberate lexical-only tradeoff; you are right that NEWS overstated it. Reworded the NEWS bullet to say the check confines `../`-style traversal and absolute paths, and to note symlinks inside the tree are resolved by the loader (not by the containment check).

R/scenario-execution.R
Felixmil: Fixed by warning at creation time. When `createScenariosFromPKML()` computes an escaping `../` relative `modelFile` (a PKML kept outside `simulationsFolder`), it now emits a `cli_warn` naming the PKML and the stored path and explaining it will fail at run time — mirroring the existing absolute-path fallback warning in the `NULL`-`simulationsFolder` branch. The scenario is still stored (behavior unchanged), so a previously-working shared-drive layout keeps working but is no longer silent. Added a test asserting the warning and the stored `../`-prefixed `modelFile`.

R/scenarios.R
Felixmil: Fixed. `.validateScenarios()` and `.validateObservedData()` now run the same containment check the runtime uses (via a shared, non-aborting `.pathEscapesRoot()` predicate) and record a **critical error** ("Path Containment") when a `modelFile` / observed-data `file` / `importerConfiguration` escapes its folder, instead of only checking existence. So `validateProject()` now agrees with load/run behavior rather than reporting clean on a path the loader would reject. Added a `.validateScenarios` containment test; the observed-data run-time rejection is already covered by the `loadObservedData` traversal test.

R/project.R
Felixmil: Fixed. If a `filePaths` block carries both the legacy `modelFolder` and the current `simulationsFolder`, the loader now emits a `cli_warn` and deterministically keeps `simulationsFolder` (dropping the legacy key) rather than letting iteration order decide silently. Added a test asserting the warning and that the current key wins.

R/project-lifecycle.R
Felixmil: Done. Split the single `NA_character_`-sentinel vector into two: `.projectReadmeFolders` (a plain name -> README-text map) and `.projectPlainFolders` (`definitions`, created without a placeholder). `.scaffoldProjectFolders()` iterates each directly, no sentinel.

R/messages.R
Felixmil: Done. Renamed `messages$noModelFolderUsingAbsolutePath()` -> `noSimulationsFolderUsingAbsolutePath()` and `messages$noModelFolderForRelativeModelFile()` -> `noSimulationsFolderForRelativeModelFile()` at all four sites (both definitions and both call sites), so a `grep modelFolder` no longer turns them up. Message text was already correct.

PR #1141, opened 2026-07-21
Collapse _impl error-attribution boilerplate; drop redundant .markModified() in plots

### LINKED ISSUE #1136: Collapse the duplicated `local_error_call()` boilerplate in the ~45 authoring `_impl` functions
Every exported authoring function (`addScenario()`, `setScenario()`, `removeScenario()`, and the whole `add*` / `set*` / `remove*` family) forwards to a matching `Project` method, which in turn forwards to a `.<name>_impl(self, private, ...)` free function that holds the logic. Each `_impl` begins with the same two lines:

```r
# Attribute any abort to the public authoring function the user called
# (the free-function forwarder), not this internal `_impl`.
rlang::local_error_call(rlang::caller_env(2))
```

This boilerplate is copy-pasted byte-identically as the first statement of roughly 45 `_impl` functions across about 12 files. Because every `_impl` is invoked from exactly one place (its one-line `project.R` method forwarder), the call-depth relationship (`caller_env(2)`) is architecturally constant across all sites, so the duplication is mechanical rather than meaningful.

The risk is maintenance drift: any new `add*` / `set*` / `remove*` method must remember to reproduce these two lines exactly, including the depth argument, or its errors will be misattributed. A single shared private dispatch helper on the `Project` class (wrapping the `_impl` call and setting the error call once) would remove the 45 duplicated copies and make the convention impossible to get wrong.

This surfaced during the review of the accessor-grouping work but predates it: the pattern comes from the method-first change (#1112). It was left out of that PR to keep its diff scoped, since collapsing it touches every `_impl` site and shifts error-attribution frames across the whole authoring API, so it deserves its own focused change with its own test pass.

Scope:

- Introduce one shared private helper that performs the `_impl` dispatch and sets `rlang::local_error_call()` once at the correct depth.
- Route the `add*` / `set*` / `remove*` methods (and the reentrant PKML path) through it, removing the per-`_impl` boilerplate.
- Confirm error attribution is unchanged: an abort from any authoring function still reads `Error in <that function>()`.

### LINKED ISSUE #1138: Drop the redundant `.markModified()` calls after `.setSection()` in R/plots.R
Six mutator `_impl` functions in `R/plots.R` call `private$.markModified()` immediately after `private$.setSection()`. This is redundant: `.setSection()` already sets the in-memory dirty bit and clears the validation-cache flag internally (its body does the same work `.markModified()` does), so the extra call is a no-op re-assignment of already-true state.

`R/plots.R` is the sole outlier: the roughly eight other domain files that also mutate through `.setSection()` do not add this extra call. It is not a correctness bug, but the inconsistency is a maintenance hazard, it could mislead a future maintainer into thinking `.setSection()` alone is insufficient, or into copying the redundant call into a new mutator.

This predates the internal-API-surface work (it is present on the pre-existing base) and is unrelated to the accessor grouping, so it was left out of that PR to keep the diff scoped.

Scope:

- Delete the six redundant `private$.markModified()` calls in `R/plots.R` that immediately follow a `private$.setSection()` call.
- No behavior change; the existing plots tests should stay green.

Felixmil:
Two scoped cleanups on the v6 `Project` internal API surface (sub-issues of the internal-API-surface umbrella), each left out of earlier PRs to keep those diffs focused. Both are pure internal refactors with no user-facing behaviour change: the full test suite stays green and every error-attribution snapshot is byte-identical.

## Shared `_impl` dispatch helper

Every exported authoring function forwards to a `Project` method, which forwarded to a `.<name>_impl(self, private, ...)` free function that opened with the same two-line comment plus `rlang::local_error_call(rlang::caller_env(2))`, so an abort would name the public function the user called rather than the internal `_impl`. That boilerplate was copy-pasted, byte-identical, into 45 `_impl` functions across 12 files, with the call-depth `2` hard-coded at each site.

This introduces one private dispatch method, `private$.impl(fn, ...)`, that resolves the attribution call once and hands it to the `_impl` as a `.call` argument. Each `_impl` now simply installs it with `rlang::local_error_call(.call)`, and the depth arithmetic lives in exactly one place. The eight sites that also threaded the call into a `.build*Entry()` helper reuse `.call` instead of recomputing it, and the few reentrant paths where one `_impl` calls another directly (the `createScenariosFromPKML` and observed-data chains) thread `.call` explicitly so their attribution is unchanged. `.call` is a required argument, so a forgotten thread fails loudly at the call site instead of silently misattributing an error.

## Redundant `.markModified()` calls in `R/plots.R`

Six mutator `_impl` functions in `R/plots.R` called `private$.markModified()` immediately after `private$.setSection()`. That is a no-op: `.setSection()` already sets the in-memory dirty bit and invalidates the validation cache, doing the same work `.markModified()` does. `R/plots.R` was the last file still carrying this; the other domain files omit it. The six calls are removed; `.markModified()` itself stays, as it remains a valid standalone method.

Closes #1136
Closes #1138

### DIFF
--- a/R/project.R
+++ b/R/project.R
@@ -267,9 +267,8 @@ Project <- R6::R6Class(
       overwriteFormulasInSS = FALSE,
       readPopulationFromCSV = FALSE
     ) {
-      .addScenario_impl(
-        self,
-        private,
+      private$.impl(
+        .addScenario_impl,
         id,
         modelFile,
         individual,
@@ -290,101 +289,101 @@ Project <- R6::R6Class(
 
     #' @description Remove scenarios. See [removeScenario()].
     removeScenario = function(id) {
-      .removeScenario_impl(self, private, id)
+      private$.impl(.removeScenario_impl, id)
     },
 
     #' @description Modify fields of existing scenarios. See [setScenario()].
     #'   The `...` carries only the fields to change (partial update); a field
     #'   passed as `NULL` is cleared, an omitted field is left untouched.
     setScenario = function(id, ...) {
-      .setScenario_impl(self, private, id, ...)
+      private$.impl(.setScenario_impl, id, ...)
     },
 
     #' @description Rename a scenario. See [renameScenario()].
     renameScenario = function(id, newId) {
-      .renameScenario_impl(self, private, id, newId)
+      private$.impl(.renameScenario_impl, id, newId)
     },
 
     #' @description Duplicate a scenario. See [duplicateScenario()].
     duplicateScenario = function(id, newId) {
-      .duplicateScenario_impl(self, private, id, newId)
+      private$.impl(.duplicateScenario_impl, id, newId)
     },
 
     #' @description Create scenarios from PKML model files. See
     #'   [createScenariosFromPKML()].
     createScenariosFromPKML = function(pkmlFilePaths, ...) {
-      .createScenariosFromPKML_impl(self, private, pkmlFilePaths, ...)
+      private$.impl(.createScenariosFromPKML_impl, pkmlFilePaths, ...)
     },
 
     #' @description Add an individual. See [addIndividual()].
     addIndividual = function(...) {
-      .addIndividual_impl(self, private, ...)
+      private$.impl(.addIndividual_impl, ...)
     },
 
     #' @description Remove individuals. See [removeIndividual()].
     removeIndividual = function(id) {
-      .removeIndividual_impl(self, private, id)
+      private$.impl(.removeIndividual_impl, id)
     },
 
     #' @description Modify an existing individual. See [setIndividual()].
     setIndividual = function(...) {
-      .setIndividual_impl(self, private, ...)
+      private$.impl(.setIndividual_impl, ...)
     },
 
     #' @description Add a population. See [addPopulation()].
     addPopulation = function(...) {
-      .addPopulation_impl(self, private, ...)
+      private$.impl(.addPopulation_impl, ...)
     },
 
     #' @description Remove populations. See [removePopulation()].
     removePopulation = function(id) {
-      .removePopulation_impl(self, private, id)
+      private$.impl(.removePopulation_impl, id)
     },
 
     #' @description Modify an existing population. See [setPopulation()].
     setPopulation = function(...) {
-      .setPopulation_impl(self, private, ...)
+      private$.impl(.setPopulation_impl, ...)
     },
 
     #' @description Add an application. See [addApplication()].
     addApplication = function(id, parameterSets = NULL) {
-      .addApplication_impl(self, private, id, parameterSets)
+      private$.impl(.addApplication_impl, id, parameterSets)
     },
 
     #' @description Remove applications. See [removeApplication()].
     removeApplication = function(id) {
-      .removeApplication_impl(self, private, id)
+      private$.impl(.removeApplication_impl, id)
     },
 
     #' @description Set an application's parameter sets. See
     #'   [setApplicationParameterSets()].
     setApplicationParameterSets = function(id, parameterSets) {
-      .setApplicationParameterSets_impl(self, private, id, parameterSets)
+      private$.impl(.setApplicationParameterSets_impl, id, parameterSets)
     },
 
     #' @description Add an output path. See [addOutputPath()].
     addOutputPath = function(id, path) {
-      .addOutputPath_impl(self, private, id, path)
+      private$.impl(.addOutputPath_impl, id, path)
     },
 
     #' @description Remove output paths. See [removeOutputPath()].
     removeOutputPath = function(id) {
-      .removeOutputPath_impl(self, private, id)
+      private$.impl(.removeOutputPath_impl, id)
     },
 
     #' @description Modify an existing output path. See [setOutputPath()].
     setOutputPath = function(id, path) {
-      .setOutputPath_impl(self, private, id, path)
+      private$.impl(.setOutputPath_impl, id, path)
     },
 
     #' @description Add a parameter set. See [addParameterSet()].
     addParameterSet = function(id) {
-      .addParameterSet_impl(self, private, id)
+      private$.impl(.addParameterSet_impl, id)
     },
 
     #' @description Remove parameter sets. See [removeParameterSet()].
     removeParameterSet = function(id) {
-      .removeParameterSet_impl(self, private, id)
+      private$.impl(.removeParameterSet_impl, id)
     },
 
     #' @description Add an entry to a parameter set. See [addParameterEntry()].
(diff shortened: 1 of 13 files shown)


R/project.R
```diff
@@ -1097,6 +1091,24 @@ Project <- R6::R6Class(
...
+    # called `.impl()`, two frames up is the exported free function the user
+    # actually called (`addScenario()` etc.), so an abort reads
+    # `Error in \`addScenario()\`:`, never the internal `_impl`. The impl only
+    # installs the value it is handed (`rlang::local_error_call(.call)`); the
+    # depth arithmetic lives here, in one place, instead of being copied into
+    # every `_impl`. An impl that calls a sibling impl directly (the reentrant
+    # PKML and observed-data paths) does not route through `.impl()`; it threads
+    # its own `.call` on, so its attribution is unchanged.
+    .impl = function(fn, ...) {
+      fn(self, private, ..., .call = rlang::caller_env(2))
```
Felixmil, line 1115: **Possible new failure mode: `.call` name collision with forwarded `...`**

`fn(self, private, ..., .call = rlang::caller_env(2))` supplies `.call` by name. Every R6 method that forwards a user-facing `...` through here (`addIndividual`, `setIndividual`, `addPopulation`, `setPopulation`, `addPlot`, `addPlotGrid`, `setScenario`, `createScenariosFromPKML`, …) is now vulnerable: if that `...` ever happens to carry an element literally named `.call` (e.g. `do.call(addIndividual, args)` where `args` carries a leftover `.call` key from some upstream list, or just a typo'd user argument), this call becomes `fn(self, private, ..., .call = <whatever>, .call = rlang::caller_env(2))` — two actual arguments matching the same formal — which R raises as `formal argument ".call" matched by multiple actual arguments`. Before this diff `.call` wasn't a formal name on any `_impl`, so a stray `.call` in `...` was harmless.

Narrow edge case, but new, and there's no guard against it in the `...`-forwarding chain. Worth at least a comment noting the constraint, or reserving a collision-proof internal name.

---

R/scenario-from-pkml.R
```diff
@@ -596,11 +595,19 @@ createScenariosFromPKML <- function(
...
+        # caller (`private$.impl`), two up is the `createScenariosFromPKML`
+        # method, so an abort still reads `Error in
+        # \`project$createScenariosFromPKML()\`:`, as before the shared helper.
         .addOutputPath_impl(
           self,
           private,
           id = names(pending),
-          path = unname(pending)
+          path = unname(pending),
+          .call = rlang::caller_env(2)
```
Felixmil, line 610: **Fragile recompute instead of reusing the already-resolved `.call`**

This site (and the matching one at line 631) recomputes `.call = rlang::caller_env(2)` at the call site, even though `.createScenariosFromPKML_impl` already has a correctly-resolved `.call` parameter in scope (it's this function's own `.call`, pointing at the free function `createScenariosFromPKML()`). The recompute lands one frame shallower — at `project$createScenariosFromPKML()` — purely because that happens to cancel out the one extra frame `private$.impl()` now inserts between the R6 method and this `_impl`. That's a coincidental match, not a robust invariant: it differs from `.call` itself, is inconsistent with the equivalent reentrant sites in `R/observed-data.R` (`.loadObservedData_impl(..., .call = .call)`, `.getObservedDataNames_impl(..., .call = .call)`), which just forward the resolved value, and will silently drift out of sync if the frame depth between the public method and this `_impl` ever changes again (e.g. another dispatch layer added later) — nothing would catch a misattribution regression like that.

Not a regression from this PR (old code had the identical gap), but since this diff already threads `.call` as a real parameter here, swapping both sites to `.call = .call` would remove the fragility for free and match the pattern this same PR uses in `observed-data.R`.

---

Felixmil (commented):
Automated review (high effort, recall-biased) of the `_impl` dispatch collapse + `.markModified()` cleanup. The core mechanism checks out: I traced the frame-depth arithmetic for `private$.impl()`'s `rlang::caller_env(2)` by hand against R's promise-environment semantics, and it resolves to the same target the old per-`_impl` `caller_env(2)` did in every direct call path. The `.markModified()` removals in `R/plots.R` are also safe — `.setSection()` already sets `.modified` and invalidates validation on every path that used to call it.

Two inline comments below on things worth a look. Plus, for the record, a couple of items that don't map to a single diff line:

- **`R/scenarios.R` (`do.call(project$setScenario, c(list(id), supplied))`, ~line 1110, pre-existing, not introduced by this PR):** `do.call()` inserts its own stack frame, so `.impl()`'s `caller_env(2)` resolves one frame short for this call path — an abort from `setScenario()` will attribute to the `do.call(...)` line rather than `setScenario()`. This predates this PR (same gap existed before, just via a differently-shaped frame count), but since this PR centralizes exactly this bookkeeping and states the goal is consistent attribution, it seemed worth flagging as a known gap in the mechanism this PR builds.
- **Minor cleanup, non-blocking:** `.markModified()` now has zero call sites left in production `R/` code after the six `plots.R` removals — it survives only via a test helper poking it directly. Might be worth a one-line note that it's now effectively test-fixture-only, so a future reader doesn't assume it's still on a live authoring path. Also, the `.call = rlang::caller_env(2)` recompute in the `scenario-from-pkml.R` loop (line 631) re-derives an identical value on every `spec` iteration — harmless but wasted work.

None of these block the refactor — the two flagged items are pre-existing latent behavior, not regressions, and the rest are polish. Nice cleanup collapsing 45 copies of that boilerplate down to one dispatcher.

---

R/project.R
Felixmil: Valid, and confirmed: a stray `.call` in a forwarded `...` produces `formal argument ".call" matched by multiple actual arguments`. It only happens on misuse (no public authoring function has a `.call` argument), and it fails loudly rather than silently misattributing, so I've documented `.call` as a reserved formal on `private$.impl()` (commit `9c6be1fb`) rather than adding a runtime guard for a pathological, self-announcing case.

R/scenario-from-pkml.R
Felixmil: Declining this one: `.call = .call` and `.call = rlang::caller_env(2)` are not equivalent here, and swapping to `.call = .call` would change the attribution and break the pinned snapshot.

`.createScenariosFromPKML_impl`'s own `.call` points at the exported free function `createScenariosFromPKML()`. But `tests/testthat/_snaps/scenario-from-pkml.md` pins the reentrant abort as `Error in \`project$createScenariosFromPKML()\`:` (the R6 method form), which is the pre-existing behavior this PR must preserve byte-for-byte. Verified against R's frame semantics:

- `.call = rlang::caller_env(2)` (current) -> `project$createScenariosFromPKML()` ✓ matches the snapshot
- `.call = .call` (suggested) -> `createScenariosFromPKML()` ✗ changes the snapshot

So the recompute is deliberate, not coincidental: these reentrant sites must attribute to the R6 method, whereas `observed-data.R`'s `.call = .call` forwarding is correct there because those inner `_impl`s raise no attributed abort of their own, so their `.call` value is not observable. The two reentrant chains genuinely need different threading, and the comment at the call site now says so. The full suite (including this snapshot) stays green.

PR #1142, opened 2026-07-21
Excel bridge: correctness & safety (#1118)

### LINKED ISSUE #1123: projectStatus() flags every section as out-of-sync after the first dirty saveProject()
Part of the v6 JSON-project rework (#1110 line; reproduced on branch `1117-modeler-friendly-docs`).

## Problem

`importProjectFromExcel()` deliberately writes the `Project.json` container **inlined** (all sections carried in the file), because the Excel sync check diffs the raw container against a fresh temp re-import of the Excel files — `R/project-excel.R` documents this next to the write: *"emptying the inline sections would blind that comparison"* ([project-excel.R:298-302](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/project-excel.R#L298-L302), comparison in `.compareJsonToExcel()`, [project-excel.R:660-694](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/project-excel.R#L660-L694)).

But `saveProject()` drives `.writeProjectTree()`, which rewrites the container with **every tree-owned section emptied** (`containerOnly = TRUE`, [definition-files.R:1238-1241](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/definition-files.R#L1238-L1241)). The two halves of the design contradict each other: after the first *dirty* save (a clean save no-ops via the dirty bit), the comparison is exactly the blinded one the import code warns about.

## Reproduction (confirmed)

Using the `TestProjectExcel` fixture with the Excel file renamed to `Project.xlsx`:

1. `importProjectFromExcel("Project.xlsx")` → `projectStatus()` reports **in sync** (control ✅)
2. `project <- loadProject("Project.json")`
3. `project$description <- "edited"` — the only change
4. `saveProject(project)` — container is rewritten with all sections emptied
5. `projectStatus("Project.xlsx")` → **out-of-sync for every section**: scenarios, parameterSets, individuals, populations, outputPaths, dataCombined, plots, plotGrids, …

Only the description changed; Excel and all definitions are untouched. Expected: out-of-sync verdict overall (the description did change), but per-section `file_status` staying `in-sync` for the untouched definition sections.

## Fix directions to weigh

- Compare against what `loadProject()` actually reads (the `definitions/` tree), not the raw container; or
- serialize the loaded project to the inlined snapshot shape for the comparison; or
- have `saveProject()` keep the inline sections when an Excel side-car exists (perpetuates the dual container shape — least attractive).

A failing testthat test pinning the correct behavior is ready in `tests/testthat/_problems/test-projectstatus-after-save-ISSUE-11-12.R` (local review branch).

Related: the legacy-container-name variant of a post-save status check fails in the opposite direction (false *in-sync*) — filed separately.

### LINKED ISSUE #1124: saveProject() writes a hardcoded Project.json, orphaning a legacy-named container (silent data loss + false in-sync)
Part of the v6 JSON-project rework (#1110 line; reproduced on branch `1117-modeler-friendly-docs`).

## Problem

`.writeProjectTree()` hardcodes the container filename: `containerPath <- file.path(dir, "Project.json")` ([definition-files.R:1240](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/definition-files.R#L1240)), ignoring `project$projectFilePath`.

`importProjectFromExcel()` derives the JSON container name from the Excel filename ([project-excel.R:281](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/project-excel.R#L281)), so importing a legacy-named `ProjectConfiguration.xlsx` produces `ProjectConfiguration.json` — a container `saveProject()` then never updates.

## Reproduction (confirmed)

Using the `TestProjectExcel` fixture as-is (legacy `ProjectConfiguration.xlsx`):

1. `importProjectFromExcel("ProjectConfiguration.xlsx")` → `ProjectConfiguration.json`
2. `project <- loadProject("ProjectConfiguration.json")`
3. `project$description <- "edited"`
4. `saveProject(project)`

Observed:

- `ProjectConfiguration.json` is **byte-identical** to before the save — the edit never lands in the loaded container;
- a **second container** `Project.json` appears next to it, carrying the edit;
- `projectStatus("ProjectConfiguration.xlsx")` reports **in sync** although the saved project differs from Excel (false negative);
- a later `loadProject("ProjectConfiguration.json")` silently loses the edit.

## Suggested fix

`.writeProjectTree()` should take the container path (or derive it from `project$projectFilePath`) instead of hardcoding `Project.json`. Both `saveProject()` and `restoreProject()` drive it; `restoreProject()` legitimately wants a fresh `Project.json`, so passing the path in keeps both callers correct.

A failing testthat test pinning the correct behavior (loaded container updated on save, no stray `Project.json`, honest out-of-sync verdict) is ready in `tests/testthat/_problems/test-projectstatus-after-save-ISSUE-11-12.R` (local review branch).

Related: with the canonical container name the same save blinds `projectStatus()` in the opposite direction (false out-of-sync on every section) — filed separately.

### LINKED ISSUE #1125: exportProjectToExcel() + projectStatus() aborts on id canonicalization collision when the project has dangling protocol refs
Part of the v6 JSON-project rework (#1110 line; observed on branch `1117-modeler-friendly-docs`).

## Problem

After exporting an imported legacy project back to Excel, checking sync **aborts with an error** instead of returning a status report.

## Reproduction (observed)

Using the `TestProjectExcel` fixture (which carries dangling `applicationProtocol` references — the import only warns about them):

1. `jsonPath <- importProjectFromExcel("ProjectConfiguration.xlsx")`
2. `project <- loadProject(jsonPath)`
3. `exportProjectToExcel(project, outputDir = ...)` → writes `Project.xlsx` + workbooks
4. `projectStatus("Project.xlsx", jsonPath = jsonPath)` →

```
Error in `.canonicalizeId()`:
! Ids collide after canonicalization:
✖ "aciclovir_iv_250mg" and "Aciclovir_iv_250mg" -> "aciclovir_iv_250mg"
✖ "protocol_250mg" and "Protocol_250mg" -> "protocol_250mg"
✖ "protocol_500mg" and "Protocol_500mg" -> "protocol_500mg"
```

The abort happens inside the temp re-import that `.compareJsonToExcel()` runs, while canonicalizing `parameterSets` ([project-excel.R:986-996](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/project-excel.R#L986-L996)).

## Suspected mechanism (not yet fully traced)

Id canonicalization on import rewrites definitions and *resolvable* references, but the fixture's **dangling** protocol references appear to survive in their original spelling. The export writes them back to Excel verbatim, and the re-import then materializes them alongside the canonical ids → collision. Needs a trace of where the original-case ids enter the exported workbooks.

## User impact

A modeler who imports a legacy Excel project with dangling protocol references (import succeeds with a warning), exports it back to Excel, and checks sync gets a hard error rather than a status report — and the error message talks about id collisions, not about the actual root cause (dangling references).

### LINKED ISSUE #1126: importProjectFromExcel(): add an overwrite = FALSE argument — re-import silently replaces an existing JSON project
Part of the v6 JSON-project rework (#1110 line; behavior verified on branch `1117-modeler-friendly-docs`).

## Current behavior

`importProjectFromExcel()` writes into the output directory with no existence check, prompt, or backup:

- The container is written unconditionally (`writeLines(jsonText, outputPath)`, [project-excel.R:310](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/project-excel.R#L310)) — an existing `Project.json` is silently replaced.
- The `definitions/` tree is then **fully reconciled** to the Excel content: `.writeDefinitionTree()` owns each `definitions/<kind>/` directory and deletes every `.json` not in the freshly written keep-set ([definition-files.R:435-445](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/definition-files.R#L435-L445), deletion at [definition-files.R:479-487](https://github.com/esqLABS/esqlabsR/blob/1117-modeler-friendly-docs/R/definition-files.R#L479-L487)).

So re-running the import over an existing JSON project doesn't just overwrite matching files — it **deletes definitions that exist only on the JSON side**, and empties every `definitions/<kind>/` directory whose kind has no workbook in the Excel project.

## Why this bites

A modeler migrates from Excel, authors new scenarios in the JSON project for a while, and later re-runs the import — plausibly following `projectStatus()`'s own out-of-sync suggestion (*"`importProjectFromExcel()` — Update JSON from Excel files"*). The JSON-only work is silently erased. Neither the roxygen docs nor `vignette("migrate-from-excel")` currently mention the overwrite/delete semantics.

## Suggestion

Add an `overwrite = FALSE` argument:

- `overwrite = FALSE` (default): abort with a clear message when the target container file (or a non-empty `definitions/` tree) already exists in the output directory, telling the user to pass `overwrite = TRUE` to replace the existing JSON project with the Excel state.
- `overwrite = TRUE`: current behavior (full replace + tree reconcile).

Follow-ups to align if adopted:

- `projectStatus()`'s suggested-actions text should show `importProjectFromExcel(overwrite = TRUE)` for the update-from-Excel path.
- Document the replace semantics in the roxygen block and in `vignette("migrate-from-excel")`.
- `.compareJsonToExcel()` imports into a fresh temp dir, so the internal re-import is unaffected.

### LINKED ISSUE #1128: Compute the sync-status Excel axis from the in-memory project
## Context

Part of the Excel bridge umbrella (#1118). Follow-up to the explicit-save lifecycle rework (#1110, PR #1117), which deliberately left the sync-status Excel comparison unchanged.

`projectStatus(project)` (and `project$status$excel_in_sync`) reports a two-axis status: memory ↔ tree (unsaved edits) and memory ↔ Excel (is the workbook a stale export of the current project?). The design doc and the #1110 spec document the Excel axis as a **one-way "would re-exporting from the current in-memory project change the workbook?"** comparison.

The current implementation does not do that. It reuses the `.compareJsonToExcel()` engine, which **re-imports the on-disk Excel into a temporary JSON and diffs it against the persisted `Project.json` container**. Because it compares the persisted JSON rather than the in-memory project, the axis disagrees with its own contract whenever the project has unsaved in-memory edits, and the re-import step and the emptied-container-on-disk shape are what make the other bugs in this umbrella (#1123, #1125) possible.

## Goal

Compute the Excel axis of `projectStatus(project)` from the current **in-memory** project, honouring the documented "would re-exporting change it" contract, instead of re-importing the on-disk Excel and comparing to the persisted JSON.

## Sketch

- Export a temporary Excel workbook from the current in-memory `project` and diff it against the configured `Project.xlsx` side-car (one-way: does re-export change it?), instead of re-importing the on-disk Excel and comparing to the persisted `Project.json`.
- This is an internal change to how the Excel axis is computed; `projectStatus(project)`'s signature and the `list(tree_in_sync, excel_in_sync, details)` contract are unchanged.

## Regression coverage

- A clean project whose workbook matches memory reports the Excel axis in-sync.
- A project with unsaved in-memory edits reports the Excel axis as out-of-sync (drift).
- A normal tree project with empty container sections does not report false drift after a dirty `saveProject()`.

## Notes

- After PR #1117 there is a single status surface, `projectStatus(project)` (the former `syncStatus()`), taking a loaded `Project`; the path-based, disk-to-disk comparator was removed.
- Removing the on-disk Excel re-import step is also what eliminates the #1125 canonicalization abort; not comparing against the emptied persisted container is what eliminates the #1123 false drift. This issue is the shared root fix those two build on.

### LINKED ISSUE #1139: importProjectFromExcel(): enforce project-folder containment on Excel-side paths
Part of the Excel-bridge umbrella #1118 (JSON ↔ Excel bridge correctness & safety), sibling to the JSON-side path-containment work in #1034.

## Context

#1034 added project-folder path containment for the JSON-driven read paths (a scenario's `modelFile`, a CSV `populationId`, an observed-data source's `file` / `importerConfiguration`), so an untrusted `Project.json` can no longer point a project-controlled path outside the project folder. The Excel import entry point was left out of that PR to keep it scoped, and is tracked here.

`importProjectFromExcel()` resolves paths read from the Excel side (the `configurationsFolder` and the per-section workbook file properties, from the author-controlled `Property` column) with plain `file.path()` + `normalizePath(mustWork = FALSE)` and no containment check against the project directory. This is the same traversal shape #1034 closed elsewhere, at a different entry point: a crafted `ProjectConfiguration.json` / `Project.xlsx` can name a `configurationsFolder` or a workbook path that escapes the project root.

## Affected surface

- `R/project-excel.R`, around the `configsFolder` / `resolveConfigFile()` path resolution (~lines 87-104 at time of writing; re-locate against the current branch).
- Any other Excel-import site that joins a project root with an author-controlled `Property`/filename and hands it to a reader.

## Suggested fix

Route the Excel-import path resolution through the same containment primitives #1034 introduced in `R/file-utils.R`:

- `.resolveProjectPath(path, root, fieldName)` — aborts when a resolved path escapes its root (for the file/workbook properties resolved against `configurationsFolder`).
- the root `configurationsFolder` itself should be contained under the project directory the same way the working folders now are (the `.resolveWorkingFolder()` treatment from #1034), with the `${VAR}` env-var form remaining the sanctioned escape hatch.

## Notes

- Surfaced in the #1134 (sub-issue #1034) review, xhigh-effort pass, PR-level point 2.
- Cheapest to land alongside the other #1118 import/export changes, since it touches the same import path-resolution code.

Felixmil:
Fixes the correctness and data-safety problems of the v6 Excel bridge as one change, so the import, export, and sync-status surfaces end up mutually consistent (umbrella #1118). Each sub-issue is one commit on this branch; the two Excel-axis regressions that the sync-status rework fixes by construction (#1123, #1125) land as their own test-only commits. Targets the `json-based-project` consolidation branch.

## saveProject() container name (#1124)

`.writeProjectTree()` no longer hardcodes `Project.json`; it takes the container path. `saveProject()` passes the path the project was loaded from, so a project opened from a legacy-named container (`ProjectConfiguration.json`, or the `<xlsx-stem>.json` an Excel import produced) is saved in place instead of forking a stray `Project.json` that left the loaded container stale. `restoreProject()` keeps writing a fresh `Project.json` (its own destination), and the `migrate-from-excel` vignette's Step 2 now says the container keeps its imported name.

## Excel sync axis from the in-memory project (#1128, #1123, #1125)

The Excel axis of `projectStatus()` is now computed from the current in-memory project, honouring its documented "would re-exporting change the workbook?" contract, instead of re-reading the on-disk `Project.json` container. Both sides of the comparison are serialized through `.projectToJson()` (the memory side directly, the Excel side by re-importing the side-car and loading it back), so the comparison is independent of how either container is stored on disk. As part of this, `importProjectFromExcel()` now writes the same canonical slim container + `definitions/` tree that `saveProject()` and `initProject()` produce, dropping the second fully-inlined container variant that existed only to feed the old raw-container diff (per the review design position on #1118).

Reading the persisted container raw was the shared root cause of two follow-on bugs, both fixed by construction and pinned with regression tests:

- **#1123**: a dirty `saveProject()` on a tree project no longer reports false drift on every section (the container the save left section-emptied no longer blinds the comparison); only the field that actually changed is flagged.
- **#1125**: a project with dangling references, exported back to Excel, no longer hard-aborts the status check on an id-canonicalization collision; it is reported as the `NA` "cannot compare" state with a warning.

## Overwrite guards on the Excel bridge (#1126)

`importProjectFromExcel()` and `exportProjectToExcel()` both gain `overwrite = FALSE`. Import aborts when a JSON project already exists in `outputDir` (a re-import reconciles the whole `definitions/` tree and deletes any definition authored only on the JSON side); export aborts when `Project.xlsx` or any `Configurations/` workbook already exists (an export replaces each workbook wholesale, and its default `outputDir` is the project's own directory). `overwrite = TRUE` restores the previous behavior on each. The `projectStatus()` suggested-action text now shows the `overwrite = TRUE` form, `initProject()`'s internal export passes `overwrite = TRUE` (it owns its scaffold), and the vignette documents the replace semantics.

## Excel-import path containment (#1139)

`importProjectFromExcel()` now contains the `configurationsFolder` and every per-section workbook filename read from the author-controlled Property column under the project folder, aborting on a value that escapes it (a `../` climb or an absolute path). This extends the JSON-side path containment from #1034 to the Excel-import entry point it left out, reusing the same `.resolveProjectPath()` primitive and keeping the `${VAR}` environment-variable escape hatch.

Closes #1123
Closes #1124
Closes #1125
Closes #1126
Closes #1128
Closes #1139


R/project-excel.R
```diff
@@ -83,23 +97,43 @@ importProjectFromExcel <- function(
...
+    if (!declaresEnvVar) {
+      configsFolder <- .resolveProjectPath(
+        configsFolder,
+        pcDir,
+        "configurationsFolder"
+      )
+    } else if (!fs::is_absolute_path(configsFolder)) {
       configsFolder <- file.path(pcDir, configsFolder)
     }
     configsFolder <- normalizePath(configsFolder, mustWork = FALSE)
```
Felixmil, line 129: The `${VAR}` escape hatch never actually expands the variable. `declaresEnvVar` (line 112) only skips the new `.resolveProjectPath()` containment check — nothing here calls `.replace_env_var()` / `Sys.getenv()` on `configsFolder` the way `.resolveWorkingFolder()` does in `R/project.R` for the analogous `filePaths` case. So when `declaresEnvVar` is `TRUE` and the value isn't literally absolute (it never is — it starts with `$`), line 120 joins the **literal, unexpanded** string onto `pcDir`:

```r
configsFolder <- file.path(pcDir, configsFolder)  # e.g. ".../${MY_CONFIGS}/Configurations"
```

That path doesn't exist on disk regardless of whether `MY_CONFIGS` is actually set in the environment, so every `resolveConfigFile()` lookup under it silently fails its `file.exists()` check and returns `NULL` — the import proceeds (per the new roxygen doc's "still allowed" claim) but silently drops every section that lives under the env-var folder, with no error and no warning under `silent = TRUE`.

The existing test (`importProjectFromExcel allows a ${VAR} configurationsFolder (escape hatch)`) only asserts `expect_no_error`, so it doesn't catch this — its own comment even says "resolves against the Excel directory when the variable is unset," which is true regardless of whether the variable is set, confirming expansion never happens.

Worth either implementing expansion here (mirroring `.clean_path()`/`.replace_env_var()`) or softening the doc claim if the env-var form isn't actually meant to work yet.

---

R/project-excel.R
```diff
@@ -83,23 +97,43 @@ importProjectFromExcel <- function(
...
       return(NULL)
     }
     if (is.null(configsFolder)) {
       return(NULL)
     }
+    # Abort if the author-controlled filename escapes the configurations folder
+    # (a `../`-climbing or absolute value); a legitimate missing file is left to
+    # the caller's own existence check, so containment must run before that.
+    .resolveProjectPath(fileName, configsFolder, fieldName)
     normalizePath(file.path(configsFolder, fileName), mustWork = FALSE)
```
Felixmil: Minor: `.resolveProjectPath()`'s return value is discarded here (line 136) and the path is then recomputed a second, different way via `normalizePath(file.path(...))` (line 137). Since `.resolveProjectPath()` already returns the resolved absolute path (`fs::path_abs`-based), it'd be simpler and less fragile to use its return value directly instead of validating with one function and building the actual path with another — the two could diverge on separator/`.`-segment edge cases if either implementation changes later.

---

R/definition-files.R
```diff
@@ -1217,18 +1217,24 @@
...
+# owns its `<kind>/` directory (it deletes any `.json` not in the
+# freshly-written keep-set), so a re-run reconciles the tree to memory exactly,
+# leaving no stale entries. Shared by `saveProject()`, `restoreProject()`, and
+# the Excel import.
+#
+# `containerPath` names the container file to write. `saveProject()` passes the
+# path the project was loaded from (`project$info$projectFilePath`), so a save
+# updates that file in place rather than writing a stray `Project.json` next to
+# a legacy-named container. `restoreProject()` and the Excel import pass a fresh
+# `file.path(dir, "Project.json")`, the canonical name for a new tree project.
```
Felixmil: Nit: this says "the Excel import" passes a fresh `file.path(dir, "Project.json")`, but it actually passes `outputPath` (`R/project-excel.R` — named after the source `.xlsx` file's stem, e.g. `MyProject.json` for `MyProject.xlsx`), not the canonical `Project.json`. Only `restoreProject()` passes the literal `Project.json`. Worth correcting so the doc doesn't mislead about the Excel-import container name.

---

Felixmil (commented):
Automated code review (high effort). One real bug found: the new `${VAR}` escape hatch for `configurationsFolder` never expands the environment variable, so it silently resolves to a nonexistent literal path and drops sections instead of reading them from the intended out-of-project location — see inline comment on `R/project-excel.R`. Two minor cleanup/doc nits also flagged inline. Everything else in this diff (overwrite guards, path containment, `.compareJsonToExcel()` rework, `containerPath` threading) checked out against callers and tests.

---

R/project-excel.R
Felixmil: Fixed in c58cbaf1. Real bug, thanks. The `${VAR}` detection was copied from `.resolveWorkingFolder()` but not the expansion, so the literal joined onto `pcDir` and every section under the env-var folder silently dropped.

Extracted the package's expansion (`${VAR}`/`$VAR`, `$PATH` exempt, unset left in place) into one shared internal `.replaceEnvVarPath()` in `R/file-utils.R`; `Project$.replace_env_var()` now forwards to it so the two resolvers can't drift, and `importProjectFromExcel()` expands before resolving. The escape-hatch test now sets the variable and asserts the sections import, so a non-expanding literal fails it.

R/project-excel.R
Felixmil: Already addressed in f890bc0a: `resolveConfigFile()` now uses the path `.resolveProjectPath()` returns instead of recomputing the join.

R/definition-files.R
Felixmil: Fixed in c58cbaf1. The comment now says the Excel import passes the `<xlsx-stem>.json` container it wrote, not a fresh `Project.json` (only `restoreProject()` passes the literal `Project.json`).

Issue #1131, opened 2026-07-19
Observed data


Felixmil:
Umbrella for observed-data handling in the v6 project model. Part of #1088.

Observed-data sources in a v6 project span a documentation/security gap (a `type = "script"` source runs arbitrary R code on load), a path-safety gap (author-controlled path fields can escape the project root), and an unsettled round-trip contract (a programmatically added `DataSet` is not persisted across save/reload). All live on the same `R/observed-data.R` surface and its path-resolution helpers, so they are best addressed as one change with a coherent contract.

Felixmil:
Reopening: migration testing of the 6.0.0 JSON model surfaced new findings on this umbrella's surface, attached as sub-issues below (tested at commit 677f9c8e).

PR #1145, opened 2026-07-22
Observed data: persist programmatic DataSets, warn on script sources (#1081, #1033)

### LINKED ISSUE #1033: Warn users that `type = "script"` observed-data sources execute arbitrary R code

### LINKED ISSUE #1081: v6: observed data — fix "errors fast" roxygen and define programmatic round-trip
Part of the JSON-primary rework (#1088). The documentation defect this issue originally raised is fixed; what remains is a design decision about round-tripping.

## What is already done

The `loadObservedData()` roxygen no longer claims the `programmatic` observed-data type "errors fast" or is "reserved for a later milestone". It now describes the actual behaviour: a `programmatic` declaration is a sentinel for a `DataSet` added at runtime with `addObservedData()`, its data lives in the session rather than on disk, and it is resolved from the in-memory store on load.

## What remains

A programmatically added `DataSet` does not survive a save and reload. `addObservedData(project, dataSet)` keeps the actual `DataSet` only in the private runtime store (`state$.programmaticDataSets`) and writes just a `{type: "programmatic", name: ...}` sentinel to the JSON. After a reload the sentinel is present but the data is gone unless the same session re-adds it.

The open question is whether this is the intended contract or a gap to close:

1. Accept it as by-design and steer users to a `script` source when they need reproducibility across reloads (the roxygen already hints at this). This is likely a "won't fix, documented" outcome.
2. Define a real round-trip, for example by persisting the `DataSet` to a data file next to the project on save and rehydrating it on load.

## Resolution

Pick one of the two, document the chosen contract explicitly on `addObservedData()` and `loadObservedData()`, and add a test that asserts the round-trip behaviour that was decided.

Felixmil:
Umbrella #1131 covers observed-data handling in the v6 project model. This addresses the two open sub-issues on the same `R/observed-data.R` surface: a programmatically added `DataSet` is now persisted so it survives a save and reload (#1081), and a `script` observed-data source now carries a security note and a runtime warning because it executes arbitrary R on load (#1033). Along the way it closes a related gap the umbrella implies: a `programmatic` sentinel that resolves to no data no longer vanishes silently. The path-containment concern the umbrella also names is already handled on this branch (a source path that escapes the project folder is rejected at load and flagged by `validateProject()`), so no work was needed there.

## Programmatic round-trip (#1081)

A `DataSet` passed to `addObservedData()` previously lived only in the session that added it: the on-disk entry was a `{type: "programmatic", name}` sentinel, and the `DataSet` behind it was lost on reload. It now round-trips. On `saveProject()` each session-added `DataSet` is written to a PKML file named `<DataSet name>.pkml` under the project's data folder, and its entry is rewritten to an ordinary `pkml` source. From then on it is file-backed: it survives a reload and its remove id becomes the PKML file basename.

- Persisting turns a `programmatic` source into a `pkml` file, so the data folder must be declared before saving; a project with no data folder aborts the save with a clear message rather than silently dropping the data.
- The persist step rewrites and validates the whole `observedData` section in memory before writing any PKML, so a save that aborts (an unsafe name, or a name whose PKML file would collide with an existing source) never overwrites an existing file. This keeps the save all-or-nothing for the data files, not just the definition tree.

## Script-source security (#1033)

A `script` source runs the R file it names, with `source()`, whenever the data is resolved, so any R code in that file executes on the user's machine. This is unchanged behaviour, now documented and surfaced:

- `?loadObservedData` gains a `@section Security` note, and the observed-data vignette gains a matching callout, both stating that a `script` source runs arbitrary R and that observed data should only be resolved from a trusted project.
- The first time a session sources a `script` source, the package warns (once per session) that arbitrary R is being executed.
- The vignette now frames the choice honestly: since both `programmatic`/`pkml` and `script` survive a reload, `programmatic`/`pkml` is the safe default for fixed measured data (nothing runs on load), and `script` is for data that must be recomputed from code (the reason to accept the trust implication).

## Unresolved programmatic sentinel

A `programmatic` sentinel read from disk with no matching in-session `DataSet` (a hand-authored declaration, or a project opened before it was ever saved) used to disappear from `loadObservedData()`'s result with no message, surfacing later as an obscure dangling DataCombined reference. `loadObservedData()` now warns by name when this happens.

## Dependency

`rlang` moves from `Suggests` to `Imports`. It is already used unconditionally across the package (`rlang::local_error_call()` in essentially every `_impl`), and the once-per-session warning uses `rlang`'s frequency gating, so it can no longer be a soft dependency.

## Example

<details><summary>Reprex: a programmatic DataSet now survives a save and reload</summary>

``` r
devtools::load_all(".", quiet = TRUE)

# A fresh project that declares a data folder.
dir <- withr::local_tempdir()
jsonlite::write_json(
  list(schemaVersion = "2.0", filePaths = list(dataFolder = "Data"),
    outputPaths = structure(list(), names = character(0)),
    scenarios = list(), observedData = list()),
  file.path(dir, "Project.json"), auto_unbox = TRUE, null = "null")

# Build a DataSet in the session and add it.
ds <- ospsuite::DataSet$new(name = "MyMeasurement")
ds$setValues(xValues = c(0, 1, 2), yValues = c(5, 6, 7))
addObservedData(project, ds)
#> ℹ Added programmatic observed-data source "MyMeasurement". It lives in this
#> session until you save. ℹ On `saveProject()` it is written to
#> 'MyMeasurement.pkml' under the data folder, so it survives a reload.

# Until you save it is a programmatic source.
project$definitions$observedData[[1]]$type
#> [1] "programmatic"

# saveProject() writes the DataSet to a PKML file and rewrites the entry.
saveProject(project)
project$definitions$observedData[[1]]
#> <ObservedDataSource>
#>   • Type: pkml
#>   • File: MyMeasurement.pkml
#>   • Name: <empty string>
#>   • Importer Configuration: <empty string>
#>   • Sheets: <empty string>

# The data now survives a reload (fresh session, empty runtime store).
reloadProject(project)
loadObservedData(project)[["MyMeasurement"]]$yValues
#> [1] 5 6 7
```

</details>

Closes #1033
Closes #1081


R/observed-data.R
```diff
@@ -634,3 +681,85 @@ removeObservedData <- function(project, id) {
...
+  if (!dir.exists(dataFolder)) {
+    dir.create(dataFolder, recursive = TRUE, showWarnings = FALSE)
+  }
+  for (i in seq_along(toPersist)) {
+    filePath <- .resolveProjectPath(fileNames[[i]], dataFolder, "file")
+    ospsuite::saveDataSetToPKML(
+      private$.programmaticDataSets[[persistedNames[[i]]]],
+      filePath
+    )
+  }
```
Felixmil, line 805: **Correctness (high): the PKML persist step breaks `saveProject()`'s all-or-nothing guarantee, with a real data-loss path.**

`.persistProgrammaticObservedData()` is called from `.saveProject_impl()` (project-lifecycle.R:124) *before* `.writeProjectTree()` (project-lifecycle.R:133). It writes real PKML files to disk (this loop) and commits `private$.setSection("observedData", ...)` / clears `private$.programmaticDataSets[[name]]` (lines 757-760) — all before the other 11 definition kinds (scenarios, individuals, populations, ...) have been written by `.writeProjectTree()`.

`.writeProjectTree()` can still abort partway through for an unrelated reason (e.g. a corrupt scenario — this exact failure mode is already tested in `tests/testthat/test-definition-files.R:304`, "saveProject() aborting on one scenario leaves the tree intact"). If that happens *after* this function has already run:
- The PKML file is already on disk.
- The original in-session `DataSet` object is already gone from `private$.programmaticDataSets`.
- `saveProject()` reports failure (never reaches `.clearModified()`), so the user reasonably retries or calls `reloadProject()`.
- `reloadProject()` re-reads the tree from disk, which never got the observedData entry written (that kind's own write never ran) — so the entry disappears entirely, and the once-in-session `DataSet` is unrecoverable. The PKML file is left as an orphan, referencing nothing.

This is a genuine data-loss scenario despite the save having been reported as failed. Consider running this persist step only after `.writeProjectTree()` has succeeded (or make it fully rollback-able), so a failed save never has partial, un-referenced side effects on disk.

No existing test exercises "programmatic persist succeeds, a later kind's write aborts" — worth a regression test either way.

---

R/observed-data.R
```diff
@@ -634,3 +681,85 @@ removeObservedData <- function(project, id) {
...
+#
+# @keywords internal
+# @noRd
+.persistProgrammaticObservedData <- function(self, private) {
+  observedData <- private$.getSection("observedData")
+  toPersist <- which(vapply(
+    observedData,
+    function(e) {
+      identical(e$type, "programmatic") &&
+        !is.null(private$.programmaticDataSets[[e$name]])
```
Felixmil, line 752: **Correctness (high): crashes `saveProject()` for a `programmatic` entry with no `name`.**

`{"type": "programmatic"}` with no `name` is schema-legal — `.observedDataRequiredFields("programmatic")` returns `character(0)`, so neither `validateProject()` nor the add/load guard rejects it, and this is exactly the "hand-authored declaration" case this PR's own docs anticipate as an orphan sentinel.

`private$.programmaticDataSets[[e$name]]` here indexes a list with `[[NULL]]` when `e$name` is `NULL`, which throws `Error: subscript out of bounds` in base R (the `&&` on the previous line only short-circuits on `type`, not on `name`). This contradicts the function's own documented contract ("a programmatic entry with no DataSet in the store ... is left untouched") — it should be, but instead the save crashes.

The sibling helper `.unresolvedProgrammaticNames()` a few lines below (line ~303) already guards this exact case with `e$name %||% NA_character_` — the same guard is needed here, e.g. `!is.null(e$name) && !is.null(private$.programmaticDataSets[[e$name]])`.

Not covered by any existing test.

---

R/observed-data.R
```diff
@@ -634,3 +681,85 @@ removeObservedData <- function(project, id) {
...
+  # with another source's on-disk id) never overwrites an existing file: every
+  # write below is known to land on a fresh, collision-free path. This upholds
+  # the all-or-nothing save contract for the on-disk data files, not just the
+  # definition tree.
+  persistedNames <- vapply(
+    observedData[toPersist],
+    function(e) e$name,
+    character(1)
+  )
+  fileNames <- paste0(persistedNames, ".pkml")
```
Felixmil, line 778: **Correctness (high): a `DataSet$name` containing `/` produces a nested PKML path whose parent directory is never created, and the safety check doesn't catch it.**

`persistedNames` comes straight from `DataSet$name` with no path-segment validation. If `name = "Cohort/A"` (plausible — free-form names, e.g. containing units like "mg/kg", or names carried over from Excel imports), `fileNames[[i]] = "Cohort/A.pkml"`.

The intended safety net doesn't fire: `.observedDataEntryId()` (definition-files.R) computes the id for the *rewritten* pkml entry as `basename(entry$file)`, which strips the slash before `.validateObservedDataId()` (line 738 below) ever sees it — so the "must be a single safe path segment" check silently passes on a path that still has a subdirectory component.

Then `dir.create(dataFolder, recursive = TRUE, ...)` (line ~747) only materializes `dataFolder` itself, not `dataFolder/Cohort`. `ospsuite::saveDataSetToPKML()` does no directory creation on its own (confirmed against the OSPSuite-R source — it forwards straight to the .NET backend), so the write fails with an opaque, low-level path error instead of one of this PR's own clear diagnostics.

Consider validating `name` as a single safe filename segment at `addObservedData()` time (matching `.validateObservedDataId()`'s existing rule), rather than only checking the post-basename-stripped id here.

---

R/observed-data.R
```diff
@@ -634,3 +681,85 @@ removeObservedData <- function(project, id) {
...
+  for (i in seq_along(toPersist)) {
+    observedData[[toPersist[[i]]]] <- .asObservedDataSource(list(
+      type = "pkml",
+      file = fileNames[[i]]
+    ))
```
Felixmil, line 783: **Correctness (medium): after a save, `removeObservedData()`'s documented "remove by DataSet name" contract silently stops working for this entry.**

Once this rewrite runs, a programmatic entry named e.g. `"X"` becomes `type = "pkml", file = "X.pkml"`, and `private$.programmaticDataSets[["X"]]` is deleted (lines 758-760). `.removeObservedData_impl()` matches programmatic entries by name via `private$.programmaticDataSets` first, then falls back to matching file-based entries by `basename(file)` — so `removeObservedData(project, "X")` after a save hits neither branch and silently no-ops with just a warning ("not found; no-op"), even though `"X"` is exactly the id the function's own roxygen docs say to use for a programmatic entry.

The correct post-save call is now `removeObservedData(project, "X.pkml")`. This is mentioned once in the vignette ("its remove id becomes the PKML file basename rather than the DataSet name") but not in `removeObservedData()`'s own `@param id` docs, and no test exercises save-then-remove-by-original-name. Worth at least documenting on `removeObservedData()` itself, since that's the primary reference a caller would consult.

---

R/observed-data.R
```diff
@@ -634,3 +681,85 @@ removeObservedData <- function(project, id) {
...
+  }
+  # Same checks the serializer runs, but up front: each id a safe filename
+  # segment, and no two entries sharing an id.
+  ids <- vapply(observedData, .observedDataEntryId, character(1))
+  lapply(ids, .validateObservedDataId)
+  duplicates <- unique(ids[duplicated(ids)])
+  if (length(duplicates) > 0) {
+    cli::cli_abort(messages$observedDataPersistIdCollision(duplicates))
```
Felixmil, line 791: **Correctness (medium): a previously-enforced basename-collision case now silently succeeds instead of aborting, and isn't tested either way.**

Before this PR, a programmatic DataSet whose raw `name` exactly equalled an *existing* file-based entry's basename (e.g. `name = "Aciclovir_TimeValuesData.xlsx"` colliding with an Excel entry `file = "Aciclovir_TimeValuesData.xlsx"`) aborted at save via `.serializeObservedDataSet()`'s duplicate-id check — this was the old, now-deleted test `"addObservedData mutates memory only; a basename collision aborts saveProject()"`.

Because this function rewrites the programmatic entry to `type = "pkml", file = "<name>.pkml"` *before* the collision check runs, the id is now `"<name>.pkml"`, not the raw `name` — so that exact scenario no longer collides and silently succeeds, writing `Aciclovir_TimeValuesData.xlsx.pkml` alongside the untouched `.xlsx`. The replacement test only covers a different case (name-plus-`.pkml` colliding with an existing `.pkml`), so this specific reversal is currently untested in either direction. If this behavior change is intentional (they genuinely are different files now), it'd be worth a one-line test to lock it in; if not, worth reconsidering.

---

R/observed-data.R
```diff
@@ -612,9 +652,16 @@ removeObservedData <- function(project, id) {
...
-    "i" = "Sourcing observed-data script: {.path {filePath}}"
-  ))
+  # A script source executes arbitrary R (see the loadObservedData() Security
+  # section). Warn the user that this is happening, once per session so a
+  # project with several script sources, or a repeated resolve, stays quiet
+  # after the first.
+  cli::cli_warn(
+    messages$observedDataScriptSecurityWarn(),
+    .frequency = "once",
+    .frequency_id = "esqlabsR_observed_data_script_source"
```
Felixmil: **Correctness/security (medium): the once-per-session security warning is gated globally, not per project, so it can silently fail to warn for an untrusted project.**

`.frequency_id` here is a fixed literal string, and rlang's frequency-throttling keys off that string alone with no per-project component (confirmed against rlang's `warning_freq_env`, and corroborated by this PR's own test resetting via `rlang::reset_warning_verbosity("esqlabsR_observed_data_script_source")` — a flat global key).

So: resolve observed data from a trusted project A with a `script` source in a session (warning fires once), then later in the *same* session resolve observed data from a different, untrusted project B that also has a `script` source (e.g. a script/loop iterating over several projects) — B's arbitrary-code execution will **not** warn, because the global gate was already tripped by A. That defeats the stated purpose ("only resolve observed data from a project you trust") for exactly the realistic case of working with more than one project per session.

Consider scoping the frequency id to something project-identifying (e.g. incorporate `self$info$projectFilePath` or similar) so each project gets its own one-time warning.

---

R/observed-data.R
```diff
@@ -261,10 +272,40 @@ loadObservedData <- function(project) {
...
+  # DataSet: the data lives only in the runtime store, so a sentinel read from
+  # disk (a reload, or a project opened elsewhere) has no backing DataSet and
+  # would silently vanish from the result. Warn by name so the drop is loud
+  # instead of surfacing later as an obscure dangling DataCombined reference.
+  unresolved <- .unresolvedProgrammaticNames(
+    self$definitions$observedData,
+    names(allDataSets)
+  )
+  if (length(unresolved) > 0) {
+    cli::cli_warn(messages$observedDataProgrammaticUnresolved(unresolved))
```
Felixmil: **Correctness/UX (medium): this warning is unthrottled and reentrant, unlike its sibling warning added in the same PR — it can fire as a surprising side effect of unrelated calls.**

Unlike `.loadObservedScript()`'s security warning (line ~660, `.frequency = "once"`), this "resolved to no data" warning has no throttling at all. `.loadObservedData_impl()` is reentered from `getObservedDataNames()` (when its cache is empty) and, critically, from `addObservedData()`'s own existing-name check — and `private$.observedDataNamesCache` is reset to `NULL` on every `loadProject()`/`reloadProject()`. So the *very first* `addObservedData(project, newUnrelatedDataSet)` call in a session, on a project that happens to carry one orphaned `programmatic` sentinel, will emit this warning as a side effect of an unrelated call.

It also fires on every `loadObservedData()` / `createDataCombined()` / `createPlots()` call (all unconditional, no caching of the actual DataSets) — so a loop building several plots against a project with one orphan sentinel re-emits the same warning every iteration. Given the sibling warning in this same PR was deliberately throttled, this one might benefit from the same treatment (or at least a per-project frequency gate, same as the comment on the script-source warning above).

---

R/messages.R
```diff
@@ -989,12 +989,54 @@ messages$observedDataScriptWrongReturnType <- function(filePath, klass) {
...
     "{.field dataFolder} is not declared in {.code filePaths}; cannot resolve {.path {file}}."
   )
 }
 
+messages$observedDataProgrammaticAdded <- function(name) {
+  cli::format_message(c(
+    "i" = "Added programmatic observed-data source {.val {name}}. It lives in this session until you save.",
+    "i" = "On {.fn saveProject} it is written to {.path {paste0(name, '.pkml')}} under the data folder, so it survives a reload."
+  ))
+}
```
Felixmil, line 1028: **Minor: this add-time message unconditionally promises persistence, even when `saveProject()` will later abort.**

`.addObservedData_impl()`'s `DataSet` branch never checks `self$paths$dataFolder` before showing this message — that check only happens later, in `.persistProgrammaticObservedData()`, which aborts with `observedDataPersistNoDataFolder` if `dataFolder` is unset. A project with no `dataFolder` declared (a very reachable state — a fresh `Project$new()`, or a minimal hand-authored `Project.json`) will show this reassuring "it is written to `<name>.pkml`... so it survives a reload" message immediately, then hit a hard abort at save time with no hint at add-time that a precondition was unmet. Consider checking `dataFolder` here too (or softening the wording to note the precondition).

---

REFACTORING-LOG.md
```diff
@@ -1,5 +1,10 @@
 # Refactoring log
 
-One bullet per commit, newest at the bottom: `- YYYY-MM-DD HH:MM — plain-language one-liner`. See the "Log every commit" rule in `REFACTORING.md`.
+One bullet per commit, newest at the bottom: `- YYYY-MM-DD HH:MM: plain-language one-liner`. See the "Log every commit" rule in `REFACTORING.md`.
```
Felixmil, line 3: **Nit (convention): this diverges from `REFACTORING.md`'s mandated log format.**

`REFACTORING.md`'s "Log every commit" rule specifies the separator as an em dash: `` `- YYYY-MM-DD HH:MM — <plain-language one-liner of what changed>` `` (confirmed at the byte level — it's U+2014, not a hyphen or colon). This diff changes the log's own header to use a colon instead, and retroactively rewrites the pre-existing 2026-07-22 11:08 entry (and all 5 new entries) to match — so `REFACTORING-LOG.md` now permanently disagrees with the format `REFACTORING.md` specifies, with no note that the format itself changed. Either restore the em dash, or update the rule in `REFACTORING.md` to match.

---

Felixmil (commented):
Automated high-effort review (8 finder angles, 1-vote adversarial verification per candidate — 9 of 10 candidates confirmed, 1 refuted and dropped).

Two are severity-high and worth prioritizing:
- **Save-atomicity break**: the new PKML-persist step writes real files and drops the in-session `DataSet` reference *before* the rest of `saveProject()`'s tree write runs, so an unrelated failure later in the same save (a bad scenario, etc.) can leave an orphaned PKML file and an unrecoverable `DataSet` even though the save was reported as failed.
- **Crash on a name-less `programmatic` sentinel**: schema-legal, unguarded, and contradicts the function's own "orphan sentinels are left untouched" contract — `saveProject()` throws `subscript out of bounds` instead.

The rest are real but lower severity: a nested-path `DataSet` name breaking the PKML write, `removeObservedData()` silently no-op'ing on the original name after a save, a silently-reversed (and now untested) basename-collision case, a security warning that's throttled globally instead of per-project, an unthrottled warning that can fire as a side effect of unrelated calls, an add-time message that overpromises when no data folder is declared, and a `REFACTORING-LOG.md` format drift from the em-dash convention `REFACTORING.md` specifies.

Great test coverage on the new happy paths (round-trip, no-dataFolder abort, basename collision) — the gaps above are mostly at the edges (malformed/hand-authored entries, multi-project sessions, partial-failure interleaving) rather than the core flow.

---

R/observed-data.R
Felixmil: Fixed. The persist step now keeps the session `DataSet` in the runtime store and returns the persisted names; `.saveProject_impl()` drops them only after `.writeProjectTree()` succeeds. So a tree write that aborts (e.g. a serializer-hostile scenario) leaves the `DataSet` recoverable in the store; worst case is an orphan PKML file, never lost data. Added a regression test 'a save that aborts mid-tree-write keeps the programmatic DataSet recoverable' that forces exactly this abort and asserts the DataSet survives.

R/observed-data.R
Felixmil: Fixed. The persist predicate now guards `!is.null(e$name)` before indexing the store, so a name-less `{type: programmatic}` sentinel is left untouched instead of aborting on `store[[NULL]]`. Such an entry still can't be *saved* (it has no id to name its definition file), so the save now aborts with the serializer's clear 'no id to name its definition file' message rather than an opaque `subscript out of bounds`. Added a test asserting that clear error.

R/observed-data.R
Felixmil: Fixed at the root: `addObservedData()` now validates the `DataSet` name as a safe filename segment (`.validateObservedDataId(paste0(name, '.pkml'))`) at add time, so a name like `Cohort/A` is rejected up front with a clear message instead of failing later with a low-level path error at save. Added a test for the rejection.

R/observed-data.R
Felixmil: Documented on `removeObservedData()` itself: its `@description` and `@param id` now state that a programmatic source changes id once saved (`saveProject()` rewrites it to `type = pkml, file = <name>.pkml`), so after a save you remove it by the `<name>.pkml` basename, not the original DataSet name. Kept as a documentation fix, since the behaviour itself is intended.

R/observed-data.R
Felixmil: Intended (they genuinely are different files now: the programmatic entry persists to `<name>.pkml`, a distinct id from the raw basename). Locked it in with a test 'a programmatic name equal to an existing file basename saves cleanly' that adds a DataSet named `Aciclovir_TimeValuesData.xlsx` (the fixture's Excel basename) and asserts the save succeeds and writes `Aciclovir_TimeValuesData.xlsx.pkml` alongside the untouched `.xlsx`.

R/observed-data.R
Felixmil: Fixed. The `.frequency_id` is now per-project: `paste0('esqlabsR_observed_data_script_source_', .projectWarningKey(self))` where the key is the project file path. So resolving a second, untrusted project's script source in the same session warns even if a trusted project already tripped its own gate.

R/observed-data.R
Felixmil: Fixed. The 'resolved to no data' warning is now throttled `.frequency = 'once'` with a per-project `.frequency_id`, matching the script-source warning, so it no longer fires on every reentrant `getObservedDataNames()` / `addObservedData()` / `createDataCombined()` / `createPlots()` call.

R/messages.R
Felixmil: Fixed. `observedDataProgrammaticAdded()` now takes `hasDataFolder` and, when no `dataFolder` is declared, tells the user to declare one before saving (noting `saveProject()` aborts otherwise) instead of unconditionally promising 'it is written to <name>.pkml... so it survives a reload'. `.addObservedData_impl()` passes `!is.null(self$paths$dataFolder)`.

REFACTORING-LOG.md
Felixmil: Resolved by updating the rule to match: `REFACTORING.md`'s 'Log every commit' rule now specifies the colon separator (`- YYYY-MM-DD HH:MM: ...`), so `REFACTORING.md` and `REFACTORING-LOG.md` agree. (Chose colon over the em dash deliberately: the project's writing convention avoids em dashes as punctuation.)

PR #1152, opened 2026-07-22
Unify R6 method, helper, and message-catalog casing

### LINKED ISSUE #1144: Unify R6 method and internal-helper casing across the 6.0.0 project model
The 6.0.0 project-model rework left R6 method and internal-helper casing inconsistent across the codebase: some code is camelCase, some is snake_case, and the two styles sit side by side. This issue tracks unifying it to one convention so future work (and review) has a single rule to follow instead of matching whatever the neighbouring file happens to do.

This is deliberately deferred cleanup, not a migration blocker: the mixed casing is cosmetic and touches internal surfaces. It is filed so the decision and the sweep happen on purpose later rather than being silently locked in.

## What is inconsistent

- **R6 method casing across classes.** `Project` uses camelCase methods; the `validationResult` R6 class uses snake_case: `add_critical_error()`, `add_warning()`, `is_valid()`, `has_critical_errors()`, `get_summary()` (`R/validation.R:41,55,66,71,88`).
- **`Project` private-method casing.** Mostly dot-prefixed camelCase (`.getSection`, `.setSection`, `.markModified`), but the parsing-era ones are dot-prefixed snake_case: `.read_json()`, `.replace_env_var()`, `.clean_path()` (`R/project.R:1314,1215,1238`).
- **Validation-internal helper casing.** Shared check helpers are snake_case, matching the `validationResult` style rather than the rest of the package: `.check_no_duplicates()`, `.check_required_fields()` (`R/validation.R:619,644`).
- **`messages` catalog entry name prefixes.** Entry names have no standardized semantic prefix: a mix of `error*`, `stop*`, `warning*`, `validation*`, many with no prefix, and a few PascalCase (`PlotGridsNamesMustBeUnique`, `PlotIDsMustBeUnique`) in `R/messages.R`.
- **`@family` roxygen tag values.** Mix of singular (`individual`, `scenario`, `output path`) and plural (`parameters`, `plots`), and spaced (`project persistence`) vs camelCase (`observedData`, `dataCombined`).

## What to decide

1. One casing convention for R6 public and private methods (the package default is camelCase; `validationResult` and the validation helpers are the outliers).
2. Whether `messages` entry names get a standardized prefix scheme, or the prefix is dropped as noise.
3. A single `@family` value style (singular vs plural, spaced vs camelCase).

## Scope note

Renaming R6 methods is API-affecting only where the method is called externally; `validationResult` is internal, so its methods can be renamed freely, but every call site must move in the same change. This is a mechanical sweep once the convention is decided, best done as one PR per surface (R6 methods, `messages` names, `@family` values) so each diff stays reviewable.

Part of the 6.0.0 project-model rework tracked in #1088.

Felixmil:
Unify the mixed camelCase / snake_case naming that the 6.0.0 project-model rework left across internal surfaces, so future work follows one rule instead of matching the neighbouring file. This is a mechanical rename sweep with no change to observable behaviour; each surface is a separate commit to keep the diff reviewable.

## R6 method and helper casing

Move the snake_case outliers to the package-default camelCase:
- `validationResult` methods: `add_critical_error`/`add_warning`/`is_valid`/`has_critical_errors`/`get_formatted_messages`/`get_summary` become camelCase. The class is internal, so all call sites move in the same change. The snake_case `@field`s and the `getSummary()` return-list keys are data shape, not methods, and are left unchanged.
- `Project` parsing-era private methods `.read_json`/`.replace_env_var`/`.clean_path` become camelCase (and the `.cleanPath` argument `replace_env_vars` follows).
- Shared validation helpers `.check_no_duplicates`/`.check_required_fields` become camelCase.

## Message-catalog entry names

Drop the inconsistent severity/kind prefixes (`error`/`stop`/`warning`/`validation`/`message`) from the `messages` catalog, naming each entry by what it describes. Acronyms stay uppercase (`PIParameterNotFound`), PascalCase entries are lowercased. Where dropping the prefix would collide two distinct messages, a short context qualifier keeps them apart (`combineInvalidDataSetName`, `validatorObservedDataFileNotFound`, `validatorObservedDataMissingField`).

## @family roxygen values

Normalize the spaced and kebab-case `@family` values to camelCase (`project persistence` to `projectPersistence`, `output path` to `outputPath`, `sensitivity-calculation` to `sensitivityCalculation`, `create-plotting-configurations` to `createPlottingConfigurations`), leaving singular/plural as-is. The matching `has_concept()` entry in `_pkgdown.yml` is updated so the reference index stays in sync.

Closes #1144

### DIFF
--- a/R/create-plots.R
+++ b/R/create-plots.R
@@ -87,11 +87,11 @@ createPlots <- function(
   if (isTRUE(stopIfNotFound)) {
     unknownGrids <- setdiff(plotGrids, names(allPlotGrids))
     if (length(unknownGrids) > 0) {
-      cli::cli_abort(messages$stopPlotGridNamesNotFound(unknownGrids))
+      cli::cli_abort(messages$plotGridNamesNotFound(unknownGrids))
     }
     unknownPlots <- setdiff(plots, names(allPlotConfig))
     if (length(unknownPlots) > 0) {
-      cli::cli_abort(messages$stopPlotIdsNotFound(unknownPlots))
+      cli::cli_abort(messages$plotIdsNotFound(unknownPlots))
     }
   }
 
@@ -264,7 +264,7 @@ createPlotsFromExcel <- function(...) {
       for (limitsField in check$limits) {
         limitsValue <- plotConfiguration[[limitsField]]
         if (!is.null(limitsValue) && 0 %in% limitsValue) {
-          msg <- messages$warningLogScaleWithZeroLimit(
+          msg <- messages$logScaleWithZeroLimit(
             plotID = plotID,
             axisLimitsField = limitsField,
             axis = check$axis
@@ -414,7 +414,7 @@ createPlotsFromExcel <- function(...) {
   }
   duplicatedPlotIds <- ids[duplicated(ids)]
   if (length(duplicatedPlotIds) > 0) {
-    msg <- messages$PlotIDsMustBeUnique(duplicatedPlotIds)
+    msg <- messages$plotIDsMustBeUnique(duplicatedPlotIds)
     cli::cli_abort("{msg}")
   }
   if (any(vapply(plotTypes, is.null, logical(1)))) {
@@ -439,7 +439,7 @@ createPlotsFromExcel <- function(...) {
     dataCombinedNames
   )
   if (length(missingDataCombined) != 0) {
-    msg <- messages$stopInvalidDataCombinedName(missingDataCombined)
+    msg <- messages$invalidDataCombinedName(missingDataCombined)
     cli::cli_abort("{msg}")
   }
   invisible(plotConfigurations)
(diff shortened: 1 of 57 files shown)


Felixmil (commented):
Reviewed this rename sweep (all 3 commits) at high effort with a multi-angle pass: line-by-line correctness, removed-behavior audit, cross-file caller tracing, reuse/duplication, simplification/efficiency, altitude, and CLAUDE.md conventions.

**Result: no actionable issues.** Every renamed R6 method/helper (`add_warning`→`addWarning`, `.read_json`→`.readJson`, `.clean_path`→`.cleanPath`, `.check_no_duplicates`→`.checkNoDuplicates`, etc.) has all call sites updated consistently across `R/`, `tests/`, and `vignettes/` — no stale references, no dynamic-dispatch (`do.call`/`get`/`__enclos_env__`) hazards, no collisions.

Two casing-style questions came up during review (PI* message names keeping a leading uppercase letter, and three different qualifier styles — `validator`/`combine`/bare — used to resolve prefix-drop collisions), but the PR description already explains both deliberately ("Acronyms stay uppercase (PIParameterNotFound)"; "a short context qualifier keeps them apart (...)"), so neither needed a comment here.

The one thing worth a heads-up rather than a blocker: the `@family` commit touches `R/sensitivity-calculation*.R` and `R/sensitivity-*-plot.R`, which are under REFACTORING.md's "Do not touch" list for this migration. The edits there are roxygen-tag-only (no logic), which fits the stated exception ("if something internal it depends on changes, update the wiring so it keeps working"), so this isn't a concern — just flagging it was noticed.

---

PR #1153, opened 2026-07-22
Uniform overwrite policy for the v6 add* authoring API

### LINKED ISSUE #1079: v6 mutation API: unify duplicate-entry handling policy
Part of the JSON-primary rework (#1088). Narrowed to the one divergence that is still present on the current `Project` model.

## What is already done

Auto-cleanup of an emptied parent is now uniform. `removeParameterEntry()` removes a parameter set once its last entry is gone, and both `removePIParameter()` and `removePIOutputMapping()` now remove a parameter-identification task once it has no parameters and no output mappings left (each warns when it does so). This was previously inconsistent.

## What remains

Duplicate-key handling is still split between the two authoring families:

- `addParameterEntry()` (`R/parameters.R`) is last-write-wins: adding an entry whose `(containerPath, parameterName)` already exists overwrites the existing one silently.
- `addPITask()`, `addPIParameter()`, and `addPIOutputMapping()` (`R/parameter-identification.R`) abort when the id already exists.

So a user adding a duplicate gets a silent overwrite in one family and a hard error in the other.

## Resolution

Decide one duplicate-collision policy (error, or overwrite with a warning) and apply it to both families so the authoring API behaves predictably regardless of which section is being edited.

Felixmil:
Gives every `add*` authoring function on the v6 `Project` model one duplicate-collision policy, so the authoring surface behaves the same regardless of which section is edited. Previously the reaction to adding a definition whose id already existed was split three ways: most `add*` functions aborted with no way to overwrite, the parameter-identification family aborted, and `addParameterEntry()` / `addInitialConditionEntry()` silently overwrote. Now a duplicate is always an error unless you pass `overwrite = TRUE`, which replaces the existing definition (last-write-wins). This resolves the divergence tracked by the mutation-API umbrella and extends it to the whole authoring surface.

## Uniform overwrite policy

Every `add*` function gains an `overwrite` argument (default `FALSE`):

- Adding a definition whose id already exists (or, for a parameter / initial-condition entry, whose `(containerPath, parameterName)` / `path` already exists) aborts with a message pointing at `overwrite = TRUE`.
- `overwrite = TRUE` replaces the existing definition in place (last-write-wins), including in-batch duplicates in a vectorized call.

This flips the previous silent last-write-wins default of `addParameterEntry()` and `addInitialConditionEntry()`, and adds the opt-in to the previously abort-only `addScenario()`, `addIndividual()`, `addPopulation()`, `addApplication()`, `addOutputPath()`, `addParameterSet()`, `addInitialConditions()`, `addPlot()`, `addPlotGrid()`, `addDataCombined()`, `addObservedData()`, `addPITask()`, `addPIParameter()`, and `addPIOutputMapping()`.

The three-layer authoring chain is preserved throughout: the free function guards and forwards, the `Project` method forwards through the private seam, and the `_impl` holds the logic. Functions that vectorize over ids via `...` (individuals, populations, plots) pull `overwrite` out of `...` before aligning per-definition fields so it is never stored as a field.

## Observed data

`addObservedData()` has no id argument and keeps an unkeyed source list plus a separate runtime store of programmatic `DataSet`s, so it needs specific handling: with `overwrite = TRUE` it replaces the matching source in place (a programmatic sentinel of the same name, together with its runtime `DataSet`; or a config entry of the same file basename), and a non-overwrite add still appends. The existing save-time guarantee that two sources sharing a file basename fail fast at `saveProject()` is unchanged.

## Example

``` r
devtools::load_all(quiet = TRUE)
project <- exampleProject()

# A duplicate id now errors by default (previously addParameterEntry silently overwrote):
suppressMessages(addParameterEntry(project, "global", "Organism", "K", 1, "1/min"))
try(addParameterEntry(project, "global", "Organism", "K", 2, "1/min"))
#> Error in .addParameterEntry(parameters, containerPath[[i]], parameterName[[i]],  : 
#>   parameter "Organism|K" already exists in the set.
#> ℹ Pass `overwrite = TRUE` to replace it.

# Opt in to replace with overwrite = TRUE:
suppressMessages(addParameterEntry(project, "global", "Organism", "K", 2, "1/min", overwrite = TRUE))

# Same policy on every add*, e.g. addScenario:
try(addScenario(project, "aciclovir_iv", modelFile = "Aciclovir.pkml"))
#> Error in addScenario(project, "aciclovir_iv", modelFile = "Aciclovir.pkml") : 
#>   scenario "aciclovir_iv" already exists.
#> ℹ Pass `overwrite = TRUE` to replace it.
```

</details>

Closes #1079


R/observed-data.R
```diff
@@ -419,20 +428,41 @@ addObservedData <- function(project, entry) {
...
-      c(private$.getSection("observedData"), list(sentinel))
-    )
+    observedData <- private$.getSection("observedData")
+    # On overwrite, replace the existing programmatic sentinel that carries this
+    # name in place; otherwise append. A DataSet is keyed by its `name` (the
+    # `existingNames` space the abort above guards), so only a programmatic
+    # sentinel of the same name is replaced. A basename clash with a file-based
+    # source lives in a different id space and is surfaced by the serializer at
+    # saveProject(), not here.
+    replaceIdx <- if (overwrite) {
```
Felixmil, line 446: **Correctness (confirmed):** `overwrite = TRUE` on a `DataSet` only searches for an existing entry with `type == "programmatic"` to replace (line 449-459). It never matches a name collision against an already-saved file-based (`pkml`/`excel`/`script`) entry, even though the pre-check at line 431 (`name %in% existingNames`) tests against that same combined name space.

Concretely: if a project already has a file-based observed-data entry resolving to DataSet name `"Laskin 1982.Group A"`, then `addObservedData(project, DataSet$new(name = "Laskin 1982.Group A"), overwrite = TRUE)` succeeds silently but falls into the append branch (line 463) instead of replacing — leaving two entries that resolve to the same name. The *next* `loadObservedData()`/`getObservedDataNames()`/`createDataCombined()`/`createPlots()`/`saveProject()` call then hits `.mergeObservedDataSets()`'s duplicate-name guard and aborts with an unrelated-looking "name collision" error, with no apparent link back to the earlier `overwrite = TRUE` call that caused it.

---

R/parameters.R
```diff
@@ -1214,27 +1247,41 @@ removeParameterEntry <- function(
...
   id <- .canonicalizeId(id)
-  .assertNoDuplicateIds(id, "initial-condition set")
-  clash <- intersect(id, names(self$definitions$initialConditions))
-  if (length(clash) > 0L) {
-    cli::cli_abort("initial-condition set {.val {clash}} already exists")
+  if (!overwrite) {
+    .assertNoDuplicateIds(id, "initial-condition set")
+    clash <- intersect(id, names(self$definitions$initialConditions))
+    if (length(clash) > 0L) {
+      cli::cli_abort(c(
```
Felixmil: **Correctness (confirmed):** same gap as `.addParameterSet_impl` (see comment on line 766) — the overwrite branch discards an existing initial-condition set's entries with no `.warnIfReferenced()` call, unlike `.removeInitialConditions_impl` (~line 1328), which warns before dropping a referenced set. `addInitialConditions(project, "icset", overwrite = TRUE)` on a set still referenced by a scenario silently wipes it with no warning.

---

R/parameters.R
```diff
@@ -1101,6 +1127,13 @@ removeParameterEntry <- function(
     units = if (nchar(units) == 0L) NULL else units
   )
   if (length(existingIdx) > 0L) {
+    if (!overwrite) {
```
Felixmil, line 1138: **Correctness / convention (confirmed):** this new overwrite-duplicate abort lives in the plain helper `.addParameterEntry()`, which has no `.call` parameter and never calls `rlang::local_error_call()`. Per REFACTORING.md's `_impl` convention, aborts should read as `Error in \`addParameterEntry()\`:` (the public name); instead this reads as `Error in \`.addParameterEntry()\`:`, leaking the internal dot-prefixed helper name.

This is confirmed by the diff's own new snapshot in `tests/testthat/_snaps/parameters.md`, which asserts `Error in \`.addParameterEntry()\`:` as the expected/passing output. The identical issue exists in `.addInitialConditionEntry()` around line 1594.

---

R/individuals.R
```diff
@@ -165,6 +167,10 @@ addIndividual <- function(project, id, species, ...) {
   n <- length(id)
 
   dots <- list(...)
+  # `overwrite` arrives through `...`; pull it out before aligning so it is not
+  # mistaken for a per-definition field.
+  overwrite <- isTRUE(dots[["overwrite"]])
```
Felixmil: **Correctness (confirmed):** `overwrite <- isTRUE(dots[["overwrite"]])` silently coerces any non-bare-`TRUE` value (a string like `"TRUE"`, `1`, `NA`, a vector) to `FALSE` with no error. The same pattern recurs in `R/populations.R:399`, `R/plots.R:813` (addPlot), and `R/plots.R:994` (addPlotGrid).

This is inconsistent with every other touched function (addScenario, addApplication, addOutputPath, addParameterSet, ...), which takes `overwrite` as a direct formal argument used in `!overwrite` — that throws a hard R type error on the same malformed input rather than silently defaulting to `FALSE`. E.g. `addIndividual(project, "dup", species = "Human", overwrite = "TRUE")` silently behaves as `overwrite = FALSE` (just aborts "already exists" with no hint the argument was malformed), while `addScenario(..., overwrite = "TRUE")` errors immediately pointing at the bad argument.

---

R/parameter-identification.R
```diff
@@ -1131,8 +1142,15 @@ addPITask <- function(
     errors <- c(errors, "id must be a non-empty string")
   } else {
     id <- .canonicalizeId(id)
-    if (id %in% names(self$definitions$parameterIdentification)) {
-      errors <- c(errors, paste0("PI task '", id, "' already exists"))
+    if (!overwrite && id %in% names(self$definitions$parameterIdentification)) {
+      errors <- c(
+        errors,
+        paste0(
+          "PI task '",
```
Felixmil, line 1155: **Conventions/consistency (confirmed):** `addPITask`'s duplicate-id message is a flat `paste0("PI task '", id, "' already exists; pass overwrite = TRUE to replace it")` folded into a generic errors vector, while every other of the ~13 touched functions uses `cli::cli_abort(c("<thing> already exists.", "i" = "Pass {.code overwrite = TRUE} to replace it."))`.

Confirmed via `tests/testthat/_snaps/parameter-identification.md`: this function's snapshot reads `x PI task 'dup' already exists; pass overwrite = TRUE to replace it` vs. every sibling snapshot's two-line `! ... already exists.` / `i Pass \`overwrite = TRUE\` to replace it.` form. This divergence predates this PR (addPITask already aggregated errors this way), but the PR had a chance to reconcile the wording across all 13 sites via a shared message and didn't, so a user (or anything asserting on error text/shape) sees inconsistent formatting depending on which `add*` function raised the duplicate-id error.

---

R/scenarios.R
```diff
@@ -798,10 +805,15 @@ addScenario <- function(
     )
   )
 
-  .assertNoDuplicateIds(id, "scenario")
-  clash <- intersect(id, names(self$definitions$scenarios))
-  if (length(clash) > 0L) {
-    cli::cli_abort("scenario {.val {clash}} already exists")
+  if (!overwrite) {
```
Felixmil: **Reuse (confirmed):** this ~8-line clash-check block (`if (!overwrite) { .assertNoDuplicateIds(...); clash <- intersect(...); cli_abort(...) }`) is copy-pasted near-identically into `R/applications.R:72`, here, `R/individuals.R:188`, `R/populations.R:409`, `R/parameters.R:754` and `:1276`, and `R/plots.R:820/1000/1151` — varying only the label string and section accessor.

`R/authoring-vectorize.R` already holds the shared vectorization helpers this convention calls for (`.assertNoDuplicateIds()`, `.assertIdVector()`) but wasn't extended to absorb this new overwrite-aware clash-check. A single shared helper (e.g. `.assertNoOverwriteClash(id, existingIds, label, overwrite)`) would collapse ~10 call sites to one line each and prevent drift — the addPITask message-format inconsistency flagged elsewhere in this review is a direct symptom of there being no single shared implementation to keep in sync.

---

R/applications.R
```diff
@@ -35,29 +35,49 @@ print.Application <- function(x, ...) {
...
 
-  .assertNoDuplicateIds(id, "application")
-  clash <- intersect(id, names(self$definitions$applications))
-  if (length(clash) > 0L) {
-    cli::cli_abort("application {.val {clash}} already exists")
+  if (!overwrite) {
+    .assertNoDuplicateIds(id, "application")
+    clash <- intersect(id, names(self$definitions$applications))
+    if (length(clash) > 0L) {
+      cli::cli_abort(c(
```
Felixmil: **Conventions (confirmed):** REFACTORING.md: *"User-facing text goes through the `messages` catalog... Don't `stop()` / `warning()` on inline literals."* This PR adds ~14 new `cli::cli_abort(c("<label> already exists.", "i" = "Pass {.code overwrite = TRUE} to replace it."))` calls as inline literals across R/applications.R, R/individuals.R, R/output-paths.R, R/populations.R, R/scenarios.R, R/plots.R (x3), R/parameters.R (x4), R/observed-data.R (x2) — `R/messages.R` itself is untouched.

11 of the 14 sites extend an already-inline pre-existing pattern in this codebase, but 2 (`.addParameterEntry`/`.addInitialConditionEntry` in R/parameters.R) are genuinely new abort call sites where none existed before (previously silent overwrite). Worth adding one `messages$...()` catalog entry for this hint so future wording changes don't require editing up to 14 call sites in lockstep — the addPITask formatting drift flagged elsewhere in this review is exactly what happens when they aren't centralized.

---

R/parameters.R
```diff
@@ -732,26 +732,34 @@ print.InitialConditionSet <- function(x, ...) {
...
+    clash <- intersect(id, names(self$definitions$parameterSets))
+    if (length(clash) > 0L) {
+      cli::cli_abort(c(
+        "parameter set {.val {clash}} already exists.",
+        "i" = "Pass {.code overwrite = TRUE} to replace it."
+      ))
+    }
   }
   parameterSets <- private$.getSection("parameterSets") %||% list()
   for (one in id) {
```
Felixmil, line 761: **Correctness (confirmed):** the overwrite branch here (line 766: `parameterSets[[one]] <- .asParameterSet(list())`) replaces an existing parameter set with an empty one with no `.warnIfReferenced()` call, unlike the equivalent `.removeParameterSet_impl` (line 805), which warns before dropping a referenced set.

`addParameterSet(project, "global", overwrite = TRUE)` where a scenario/individual/application still references parameter set `"global"` silently wipes all of its entries with zero warning, whereas `removeParameterSet(project, "global")` on the same project would warn about the dangling reference. This path is functionally remove-then-recreate-empty but bypasses the invariant the remove path enforces. (Same gap exists in `.addInitialConditions_impl`, flagged separately in this review.)

---

R/observed-data.R
```diff
@@ -419,20 +428,41 @@ addObservedData <- function(project, entry) {
...
+    }
+    if (length(replaceIdx) > 0L) {
+      observedData[[replaceIdx[[1]]]] <- sentinel
+    } else {
+      observedData <- c(observedData, list(sentinel))
+    }
+    private$.setSection("observedData", observedData)
     private$.programmaticDataSets[[name]] <- entry
     # The observedData setter resets the names cache, so rebuild it after the
     # write, from the names known before it plus the newly added name.
```
Felixmil: **Correctness (confirmed):** the line right below this comment (`private$.observedDataNamesCache <- c(existingNames, name)`) runs unconditionally after an overwrite. On a genuine overwrite, `existingNames` already contains `name` (that's how the collision was detected), so this appends a duplicate into the cache instead of leaving it unchanged.

Repeated `addObservedData(project, sameNamedDataSet, overwrite = TRUE)` calls (plausible in iterative fitting/tuning workflows) grow `.observedDataNamesCache` by one duplicate per call, unbounded. `getObservedDataNames()` returns this cache verbatim, and the next `addObservedData()` call's own `existingNames` lookup is built from this already-duplicated vector, compounding further. Consider guarding with `if (!name %in% existingNames)` or `union(existingNames, name)`.

---

Felixmil (commented):
Automated high-effort code review of the uniform overwrite-policy rollout (11 commits, `origin/json-based-project...HEAD`). 8 independent finder passes + 1-vote verification per candidate; 9 of 10 candidates confirmed (1 refuted — the output-paths.R loop change turned out to be the consistent house pattern, not a regression).

**Two real data-corruption/data-loss bugs** worth prioritizing before merge:
- `addObservedData(overwrite = TRUE)` doesn't detect a name collision against a file-based (non-programmatic) entry, silently creating a duplicate that later aborts with an unrelated "name collision" error.
- `addParameterSet`/`addInitialConditions` with `overwrite = TRUE` silently discard a set's entries with no `.warnIfReferenced()` check, unlike their `remove*` siblings.

Plus a confirmed error-attribution bug (violates the `_impl`/`.call` convention, caught by the PR's own test snapshot), an inconsistent `overwrite` argument validation across sibling functions, and some real but lower-severity consistency/reuse observations (duplicated clash-check boilerplate across ~10 functions, inline literals bypassing the messages catalog, addPITask's divergent error format). Details inline.

---

R/observed-data.R
Felixmil: Fixed in 003e0fca. `addObservedData(dataSet, overwrite = TRUE)` now aborts with a clear message when the name collides with a file-based (pkml/excel/script) source instead of falling into the append branch and leaving two entries that only fail later at load/save. The message names the fix (remove the file-based source by its basename first). A programmatic-vs-programmatic name collision still replaces in place as before.

R/parameters.R
Felixmil: Fixed in 2e38c66f. `.addInitialConditions_impl` now calls `.warnIfReferenced()` before replacing a still-referenced set with an empty one on `overwrite = TRUE`, matching `removeInitialConditions()`. Same fix applied to `addParameterSet` (the sibling thread).

R/parameters.R
Felixmil: Fixed in 2e38c66f. Threaded `.call` from the `_impl` through `.addParameterEntries()`/`.addParameterEntry()` and `.addInitialConditionEntries()`/`.addInitialConditionEntry()`, so both the duplicate-entry abort and the invalid-entry abort now read `Error in \`addParameterEntry()\`:` (the public name). Snapshots in `_snaps/parameters.md` updated accordingly.

R/individuals.R
Felixmil: Fixed in 51dc5e36. Added `.validateOverwriteFlag()` in `R/authoring-vectorize.R` and routed the four `...`-based functions (addIndividual, addPopulation, addPlot, addPlotGrid) through it, so a non-logical/non-scalar `overwrite` now errors (`\`overwrite\` must be a single TRUE or FALSE.`) instead of silently coercing to FALSE, matching the direct-formal functions. Added a regression test.

R/parameter-identification.R
Felixmil: Partially reconciled in 51dc5e36. Every single-id `add*` now raises the shared `messages$definitionAlreadyExists()` two-line message via `.assertNoOverwriteClash()`. `addPITask` keeps its flat single-line entry because it aggregates all field errors (id + scenarios + output-path references) into one abort rather than raising the standalone two-line form; converting it would mean rewriting its error-aggregation. Left it as-is with a comment explaining why, and aligned the wording ("<thing> already exists" + overwrite hint).

R/scenarios.R
Felixmil: Fixed in 51dc5e36. Added `.assertNoOverwriteClash(id, existingIds, definition, overwrite)` to `R/authoring-vectorize.R` and collapsed the copy-pasted clash-check to one call in addScenario, addApplication, addIndividual, addPopulation, addOutputPath, addPlot, addPlotGrid, addDataCombined, and the two set-adds in parameters.R. The helper wraps `.assertNoDuplicateIds()` plus the existing-id abort, gated on `overwrite`.

R/applications.R
Felixmil: Fixed in 51dc5e36. Added `messages$definitionAlreadyExists(label)` to `R/messages.R` (the overwrite hint's single source of wording), raised by the shared `.assertNoOverwriteClash()` helper. The ~10 inline clash literals are gone; the two genuinely-new entry aborts in parameters.R keep their own wording (they name the (containerPath, parameterName)/path, a different subject) but share the same `overwrite = TRUE` hint text.

R/parameters.R
Felixmil: Fixed in 2e38c66f. `.addParameterSet_impl` now calls `.warnIfReferenced()` before replacing a still-referenced set with an empty one on `overwrite = TRUE`, matching `removeParameterSet()`. Added a test: overwriting the referenced "global" set warns; overwriting a fresh unreferenced set does not. Same fix on `addInitialConditions` (sibling thread).

R/observed-data.R
Felixmil: Fixed in 003e0fca. The cache rebuild now uses `union(existingNames, name)` instead of `c(existingNames, name)`, so a repeated `addObservedData(sameNamedDataSet, overwrite = TRUE)` no longer grows `.observedDataNamesCache` by a duplicate each call. Added a test that 3 repeated overwrites keep a single cached name.

PR #1154, opened 2026-07-23
Simulation build & run API: buildSimulations(), custom Population, stopIfFails

### LINKED ISSUE #1024: Support custom Population objects in scenarios (programmatic injection)
The new JSON-primary workflow loses v6's ability to build a `Population` in R, mutate it, and run a scenario against that mutated object. `project$populations` only stores demographic specs and `runScenarios()` rebuilds a fresh `Population` on every run.

Surfaced by @PavelBal in #1018:
- https://github.com/esqLABS/esqlabsR/pull/1018#issuecomment-4389511241
- https://github.com/esqLABS/esqlabsR/pull/1018#issuecomment-4390951754 (explicit ask for an issue)

## Plan

Mirror the `observedData` `programmatic` source pattern. Make `addPopulation()` polymorphic:

- accepts an `ospsuite::Population` object → stored in a private runtime store on `Project`, JSON gets a `{type: "programmatic", populationId: ...}` sentinel (does not survive save/reload)
- accepts a demographics spec (current behavior) → stored as today
- optionally accepts a `script` source pointing at an R file that returns a `Population` (reproducible alternative to `programmatic`)

`.prepareScenario()` switches on `populationEntry$type` at run time. The CSV path can be brought into the same dispatcher as a follow-up.

## Scope

Chapter 6 (project mutation API). Same pattern that already exists for `addObservedData()` in `R/observed-data.R`.

### LINKED ISSUE #1036: `runScenarios` should get an argument `stopOnFail` and throw an error when one simulation fails
Currently, if one of the scenarios fails, a warning is issued and the other scenarios are returned. The default behavior should be changed, so the function throws an error by default.

### LINKED ISSUE #1074: v6: decouple scenario simulation construction from execution (build-without-run, custom population)
From the v5→v6 API review of #1017.

**Type:** feature / design

v6 `Scenario` is a plain-data record; the initialized `ospsuite::Simulation` (+ `Population`) is built transiently inside `.runScenariosFromProject()` and only surfaced after a run. There is no public way to obtain a fully-parameterized-but-not-run simulation.

**Use cases lost vs v5 (where `Scenario$simulation` / `$population` existed pre-run):**
- Inspect the final parameterization before running.
- Save the configured simulation to PKML without running it.
- Pre-/post-process the simulation object.
- Feed the built simulation into parameter identification or sensitivity analysis.
- Inject a custom in-memory `Population` (see below).

**Custom population:** `runScenarios()` builds the population only from `project$populations[[id]]` or, when `readPopulationFromCSV = TRUE`, from `populationsFolder/<populationId>.csv` (`R/scenario-execution.R`). A programmatically-created `ospsuite::Population` cannot be passed in directly. Workaround today = write it to CSV and reference it; a custom population not expressible as a population table has no path.

**Proposed:** a public constructor, e.g. `createSimulation(project, scenarioName, ...)` / `buildScenario()`, returning the initialized `Simulation` (+ `Population`); and/or let `runScenarios()` accept pre-built simulations / a `population` override.

### LINKED ISSUE #1130: Simulation build & run API
Umbrella for the scenario build-and-run API in the v6 project model. Part of #1088.

In v6 a `Scenario` is a plain-data record and the `ospsuite::Simulation` (plus `Population`) is built transiently inside the run path, then discarded. This loses capabilities the v5 API had: injecting a programmatically built or mutated `Population`, obtaining a fully-parameterized-but-not-run simulation, and controlling how a failed simulation is surfaced. These are the same build/run code path and are cheapest to reshape together.

Felixmil:
Reshapes the v6 scenario build-and-run path to restore capabilities the v5 API had but the JSON-first model lost, and to make simulation failure fail loudly by default. These are the same build/run code path, so the umbrella (#1130) groups the three sub-issues; they are cheapest to reshape together. Excel and the JSON entrypoints both keep working unchanged.

## Build without running (#1074)

`buildSimulations(project, scenarios = NULL, ...)` returns the fully-parameterized-but-not-run `Simulation` (and, for a population scenario, the `Population`) per scenario, as `list(simulation, population)`. Use it to inspect the final parameterization, save a configured simulation to PKML, or hand it to another OSP-suite routine before simulating. Internally the build phase is extracted into a shared `.buildScenarioSimulations()` helper that both `buildSimulations()` and `runScenarios()` drive, so there is no duplicated prep logic and the run-scoped build cache (one `IndividualCharacteristics` / `Population` per id per call) is preserved.

## Custom population injection (#1024)

`addPopulation(project, id, population)` now accepts an `ospsuite::Population` object in the `species` position, injecting a programmatically built or mutated population that a scenario runs against directly. It mirrors the existing programmatic-`observedData` pattern:

- The object lives in a session-only runtime store; the JSON entry is a `{type: "programmatic"}` sentinel.
- A population entry may now carry an optional `type` (`programmatic` / `csv`); an entry with no `type` stays a demographics spec, exactly as before. The run path resolves a population by the entry's own type, so it no longer depends on the referencing scenario's `readPopulationFromCSV` flag (that flag still works for a spec entry, for back-compat).
- `saveProject()` freezes an injected population to a `<id>.csv` file under the populations folder and rewrites the entry to a `csv` source, so it survives a reload; rerun the code that built it to reproduce it in a new session. An orphan `programmatic` sentinel (read from disk with no backing object) serializes verbatim and only fails, with an actionable message, if a scenario actually tries to use it.

## Fail-fast on scenario failure (#1036)

`runScenarios()` gains a `stopIfFails` argument that defaults to `TRUE`: a scenario whose simulation produces no results now aborts the run instead of warning and returning the others. Set `stopIfFails = FALSE` for the previous warn-and-continue behaviour. This is a breaking default change, noted in `NEWS.md`.

## Example

New usage of `buildSimulations()`:

``` r

# Build a scenario's simulation without running it
built <- buildSimulations(project, scenarios = "aciclovir_iv")
sim <- built$aciclovir_iv$simulation
class(sim)
#> [1] "Simulation"    "ObjectBase"    "DotNetWrapper" "NetObject"    
#> [5] "R6"

# Inspect / edit the fully-parameterized simulation, then run it yourself
results <- ospsuite::runSimulations(sim)[[sim$id]]
class(results)
#> [1] "SimulationResults" "DotNetWrapper"     "NetObject"        
#> [4] "R6"
```

</details>

Closes #1130
Closes #1074
Closes #1024
Closes #1036

## Summary by CodeRabbit

* **New Features**
  * Added `buildSimulations()` to prepare scenario simulations without running them.
  * Added support for programmatically supplied populations, which are saved as CSV files with the project.
  * Added population source types for CSV and programmatic populations, including clearer display information.
* **Bug Fixes**
  * Improved scenario and population validation with clearer errors for missing or unknown references.
  * `runScenarios()` can now either stop or continue when a scenario produces no results.


R/populations.R
```diff
@@ -398,6 +491,28 @@ addPopulation <- function(
   # so it is not mistaken for a per-definition field.
   overwrite <- .validateOverwriteFlag(dots[["overwrite"]])
   dots[["overwrite"]] <- NULL
+
+  # Injected-object form: an `ospsuite::Population` (a DotNetWrapper) in the
+  # `species` position. A demographics-spec entry is a plain list also classed
+  # "Population", so dispatch on DotNetWrapper, not on class "Population".
+  if (inherits(species, "DotNetWrapper")) {
```
Felixmil: **Major — the injected-object dispatch accepts any `DotNetWrapper`, not just an `ospsuite::Population`.**

```r
if (inherits(species, "DotNetWrapper")) {
```

Per the PR's own reprex, `class(sim)` for an `ospsuite::Simulation` is `"Simulation" "ObjectBase" "DotNetWrapper" "NetObject" "R6"` — `Simulation`, `Individual`, and other .NET-backed ospsuite objects all share the `DotNetWrapper` base. `addPopulation(project, "x", someSimulationObject)` is silently accepted today, gets stored in `private$.programmaticPopulations`, and only fails later — obscurely, inside `ospsuite::exportPopulationToCSV()` on `saveProject()` (~line 980), or inside `runSimulations(..., population = ...)` at run time — instead of with a clear error at `addPopulation()` call time.

Suggest adding `inherits(species, "Population")` alongside the `DotNetWrapper` check (the comment above already notes a *demographics-spec entry* is a plain list also classed `"Population"`, so this needs to check the injected-object branch specifically, not fall back to dispatching on class name for the spec form). No test currently exercises a wrong-typed injected object.

---

R/populations.R
```diff
@@ -732,3 +906,83 @@ setPopulation <- function(project, id, ...) {
...
+    populations,
+    function(e) if (is.null(e$file)) NA_character_ else basename(e$file),
+    character(1)
+  )
+  collisions <- intersect(
+    fileNames,
+    existingFiles[setdiff(names(populations), toPersist)]
+  )
+  if (length(collisions) > 0) {
+    cli::cli_abort(messages$populationPersistIdCollision(collisions))
```
Felixmil, line 978: **Minor — test-coverage gap.** `messages$populationPersistIdCollision` (raised here in `.persistProgrammaticPopulations()` when an injected population's generated `<id>.csv` collides with an existing population file's basename) has no covering test — no reference to it anywhere in `tests/testthat/test-populations.R`. Given how central the freeze-to-CSV path is to this PR's population-injection feature (#1024), worth adding before merge or as an immediate follow-up.

---

Felixmil (commented):
## Summary

Reviewed against `origin/json-based-project` (merge-base `de2c9eb`, head `8531b56`). This PR extracts `.buildScenarioSimulations()`, adds exported `buildSimulations()`, adds `stopIfFails` to `runScenarios()` (documented breaking default), and adds session-injected `Population` support to `addPopulation()` mirroring the existing `observedData` programmatic pattern (runtime store → `{type: "programmatic"}` sentinel → freeze-to-CSV on `saveProject()` → orphan-sentinel-serializes-verbatim-and-only-errors-on-use).

**What checks out:** the build-cache extraction is correctly scoped per-call (no cross-call leakage), the full population-injection lifecycle (add → save → reload → run-path resolution → `removePopulation`/`reloadProject` cleanup) is correct with no leaks, `stopIfFails` is threaded consistently and documented as breaking in `NEWS.md`, the three-layer authoring chain / `.call` threading / sibling-`_impl` conventions are followed, path handling for the population CSV reuses the existing root-escape guard, the "do not touch" sensitivity files are untouched, `REFACTORING-LOG.md` has one correctly-formatted bullet per commit, and there's no conflicting in-flight PR against this base.

**Requesting changes for one correctness gap** (inline comment on `R/populations.R:498`): the injected-object dispatch checks `inherits(species, "DotNetWrapper")` only, which is the shared base class for *all* ospsuite .NET-backed objects (`Simulation`, `Individual`, etc.), not just `Population`. A wrong-typed object is silently accepted at `addPopulation()` and only fails later, obscurely, at save or run time.

Also flagging as minor/non-blocking:
- Three new `cli::cli_abort()` calls in the same function (`R/populations.R:511-514, 563-566, 570-575`) use inline literal strings rather than the `messages` catalog, unlike the rest of this PR's new user-facing errors.
- No test covers the `populationPersistIdCollision` path (inline comment on `R/populations.R:972`) or the overwrite semantics of a type transition (`programmatic`/`csv`/spec) via `addPopulation(..., overwrite = TRUE)`.
- Two small nits: `.resolveCsvPopulation()` always attributes path-escape errors to `"populationId"` even when the value came from the entry's own `file` field; the `project` type guard is duplicated verbatim between `runScenarios()` and `buildSimulations()` (pre-existing pattern, not introduced by this PR).

Given the type-check gap is cheap to fix and central to the feature this PR closes (#1024), I'd bundle it — and ideally the collision test — into this PR rather than a follow-up.

---

R/populations.R
Felixmil: Fixed in 4bb300db. The injection dispatch now requires `inherits(species, "Population") && inherits(species, "DotNetWrapper")`, so a real `ospsuite::Population` (classed both) is accepted while a plain demographics-spec list (classed only `"Population"`) and any other .NET object (a `Simulation`, an `Individual`) are excluded. Added a test that a `Simulation` passed in the `species` position is rejected up front and never stored in the runtime store.

R/populations.R
```diff
@@ -398,6 +491,28 @@ addPopulation <- function(
...
+      overwrite = overwrite,
+      .call = .call
+    ))
+  }
+  if (is.null(numberOfIndividuals)) {
+    cli::cli_abort(
+      "{.arg numberOfIndividuals} is required when adding a demographics-spec \
+       population."
+    )
+  }
```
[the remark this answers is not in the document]
Felixmil: Addressed in 4bb300db. To be precise: the single-backslash line continuation is not a parse error (R parses `"a \<newline>b"` fine, and this file passed `R CMD check`), but it did embed a literal newline plus the leading indentation into the message for cli to reflow. I collapsed all three inline authoring aborts to single-line strings, dropping the backslash-newline entirely, so cli does the wrapping. No stray whitespace in the rendered message now.

R/populations.R
Felixmil: Added in 4bb300db. New test `saveProject aborts when a frozen population CSV would collide with an existing file`: an injected population `injected` (freezes to `injected.csv`) plus a separate `type:csv` entry already pointing at `injected.csv` triggers `populationPersistIdCollision` on save.

PR #1159, opened 2026-07-23
Authoring API ergonomics: empty PI task seeding, outputPath id-or-path, runScenarios ids

### LINKED ISSUE #1157: Authoring API ergonomics: PI task seeding, outputPath id naming, runScenarios ids
## Purpose

While rebuilding a project from scratch with the authoring API (`initProject()` + `add*()` + `saveProject()`), three ergonomic rough edges made the API harder to use than it needs to be. None is a correctness bug and each has a workaround, but each violates a reasonable expectation set by the rest of the API or by the docs. Grouping them since they share a locus (the authoring surface) and a theme (surprising constraints on otherwise-uniform `add*` verbs).

## `addPITask()` cannot be created empty, so the incremental adders can't seed a task

Natural first attempt (create the task, then fill it):

```r
addPITask(project, "task1", scenarios = "s1", parameters = list(), outputMappings = list())
#> Error in `PITask()`:
#> ! Field "parameters" on PITask "task1" must contain at least one entry.
```

So every `PIParameter(...)` / `PIOutputMapping(...)` must be built up front and passed in the constructor call; the incremental `addPIParameter()` / `addPIOutputMapping()` only work on a task that already has entries. This inverts the usual `add*` model (create, then add) and is a chicken-and-egg for a task's first parameter. Either allow an empty task at creation, or document that the incremental adders cannot seed one.

## `PIOutputMapping(outputPath = )` is an output-path id, but reads like a model path

`outputPath` is canonicalized and looked up as an **output-path id**, so passing a literal model path (as the legacy PI configuration used) aborts with "outputPath '...' not found in project$definitions$outputPaths". The argument name suggests a path; it is a reference. Consider renaming to `outputPathId`, or accepting a literal path and resolving it to the matching output-path id.

## `runScenarios(scenarios = )` needs canonical ids, not the names you authored with

Authoring with a mixed-case id stores it canonicalized (lowercased). Passing the original name back to the run API fails:

```r
addScenario(project, "MyScenario", modelFile = "m.pkml")
runScenarios(project, scenarios = "MyScenario")
#> Error: Unknown scenario names: "MyScenario".
runScenarios(project, scenarios = "myscenario")   # works
```

Documented behavior, but an easy trap when porting existing driver scripts that used mixed-case names. Either canonicalize the `scenarios` argument on the way in (so a caller can pass the name they authored with), or make the error suggest the canonical id.

## Checklist

- [ ] Allow `addPITask()` with empty `parameters`/`outputMappings`, or document the ordering constraint
- [ ] Clarify `PIOutputMapping(outputPath=)` is an id (rename or accept a literal path)
- [ ] Canonicalize the `runScenarios(scenarios=)` argument, or add a "did you mean" to the unknown-name error

Felixmil:
Three ergonomic rough edges on the `add*` authoring surface, surfaced while rebuilding a project from scratch with `initProject()` + `add*()` + `saveProject()`. None was a correctness bug, but each violated an expectation the rest of the authoring API sets. Grouped here because they share a locus (parameter-identification / scenario authoring) and a theme: surprising constraints on otherwise-uniform `add*` verbs.

## Empty PI task at creation

`addPITask()` previously required at least one parameter and one output mapping up front, so the incremental `addPIParameter()` / `addPIOutputMapping()` could never seed a task's first entry (a chicken-and-egg that inverted the usual create-then-add flow). `PITask()` / `addPITask()` now accept empty `parameters` / `outputMappings`, so a task can be created and then seeded. A task left with no parameters or no output mappings cannot run, so `validateProject()` reports it as a critical error, which gates `runPI()`. A non-list is still rejected.

## `outputPath` accepts an id or a literal model path

`PIOutputMapping(outputPath = )` is an output-path reference, but the argument name reads like a model path, and passing a literal model path (as the legacy PI configuration used) aborted. It now accepts either form: the output-path id (a key of `outputPaths`) or the literal model path of a defined output path, resolving both to the id. The name stays `outputPath`, matching the bare-name convention of every other id-reference argument (`individual`, `population`, `observedData`, `scenarios`); the roxygen now states plainly that it is a reference to a defined output path. A value matching no defined output path aborts with a "did you mean" hint, and the two previously divergent inline aborts are unified into one message-catalog entry.

## `runScenarios(scenarios = )` matches case-insensitively

Authoring a scenario with a mixed-case id stores it canonicalized (lowercased), so passing the original name back to `runScenarios()` failed with "Unknown scenario names". The `scenarios` argument is now canonicalized on the way in (in the shared `.buildScenarioSimulations()`, so both `runScenarios()` and `buildSimulations()` are covered), so a caller can pass the name they authored with. A genuinely unknown name still aborts, now naming the canonical form.

## Example

``` r
library(esqlabsR)

# A throwaway copy of the bundled example project.
dir <- withr::local_tempdir()
initProject(dir, type = "example", createExcel = FALSE, overwrite = TRUE)

## 1. Create an empty PI task, then seed it with the incremental adders.
addPITask(
  project,
  "fit",
  scenarios = "aciclovir_iv",
  parameters = list(),
  outputMappings = list()
)
addPIParameter(
  project,
  task = "fit",
  scenarios = "aciclovir_iv",
  path = "Aciclovir|Lipophilicity",
  minValue = -2,
  maxValue = 2,
  startValue = 0
)

## 2. Reference the output path by its literal model path (not its id).
addPIOutputMapping(
  project,
  task = "fit",
  scenarios = "aciclovir_iv",
  outputPath = "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)",
  observedData = "someDataset"
)
# It is stored as the output-path id:
project$definitions$parameterIdentification$fit$outputMappings[[1]]$outputPathId
#> [1] "aciclovir_pvb"

## 3. runScenarios() / buildSimulations() accept the mixed-case name you
##    authored with (ids are stored canonicalized / lowercased).
names(project$definitions$scenarios)
#> [1] "aciclovir_iv"             "aciclovir_iv_population" 
#> [3] "aciclovir_iv_steadystate"
# The example scenario is filed as "aciclovir_iv"; passing "Aciclovir_IV"
# now resolves instead of raising "Unknown scenario names".
built <- buildSimulations(project, scenarios = "Aciclovir_IV")
names(built)
#> [1] "aciclovir_iv"
```

</details>

Closes #1157

## Summary by CodeRabbit

* **New Features**
  * PI tasks can now be created without initial parameters or output mappings and completed later.
  * Output paths accept either their identifier or literal model path.
  * Scenario names are matched case-insensitively when running scenarios or building simulations.
* **Bug Fixes**
  * Improved output-path validation messages, including suggestions for likely matches.
* **Documentation**
  * Updated guidance for PI task creation, output paths, and scenario matching.


R/scenario-execution.R
```diff
@@ -537,6 +537,12 @@
   allScenarios <- project$definitions$scenarios
   if (is.null(scenarioNames)) {
     scenarioNames <- names(allScenarios)
+  } else {
+    # Match against the canonical ids scenarios were filed under, so a caller
+    # can pass the (possibly mixed-case) name they authored with. Silent
+    # because this only resolves a lookup; nothing is stored, so the
+    # "canonicalized to a safe form" heads-up would be noise on every run.
+    scenarioNames <- suppressWarnings(.canonicalizeIdRef(scenarioNames))
```
Felixmil: `.canonicalizeIdRef()` here does the same full canonicalization as definition-side ids (character replacement, trimming, reserved-name suffixing), not just case-folding as the NEWS entry / roxygen for `runScenarios()`/`buildSimulations()` describe. A mistyped name can therefore resolve to an unrelated, real scenario instead of raising "unknown scenario name".

Example: a project has scenarios `"test"` and `"test_2"`. `runScenarios(project, scenarios = "Test?2")` (built from a mistyped label) canonicalizes to `"test_2"` — a real, different scenario — and silently runs it. The `suppressWarnings()` also hides the "canonicalized to a safe form" heads-up that would otherwise flag the rewrite. Before this diff, the match was a bare case-sensitive `setdiff` with no character substitution, so this exact typo would have hit "unknown scenario name" instead.

Worth at least tightening the docs to mention the full canonicalization (not just case), and maybe reconsidering whether `suppressWarnings` should apply here given it can mask a silent wrong-scenario match.

---

R/parameter-identification.R
```diff
@@ -1527,12 +1607,9 @@ addPIOutputMapping <- function(
...
   }
-  outputPath <- .canonicalizeIdRef(outputPath)
-  if (!(outputPath %in% names(self$definitions$outputPaths))) {
-    cli::cli_abort(
-      "outputPath {.val {outputPath}} not found in project$definitions$outputPaths"
-    )
-  }
+  # Accept either an output-path id or the literal model path of a defined
+  # output path; store the resolved id.
+  outputPath <- .resolveOutputPathRef(outputPath, self)
```
Felixmil, line 1621: `.resolveOutputPathRef()` is called on the raw, unvalidated `outputPath` argument here, before the `PIOutputMapping()` constructor runs (constructor call is later, ~line 1630). That violates `.matchOutputPathRef()`'s own doc comment: "`value` must already be a non-empty scalar string, the `PIOutputMapping` constructor guarantees this."

Concretely: `addPIOutputMapping(project, task = "t", outputPath = NULL, observedData = "L", scenarios = "s")` reaches `.matchOutputPathRef()`'s `if (canonical %in% ids)` with `canonical <- .canonicalizeIdRef(NULL)` → `NULL`, so `NULL %in% ids` is `logical(0)`, and `if (logical(0))` raises R's raw "argument is of length zero" instead of the clean `messages$outputPathRefNotFound()` this PR introduces. Same for `outputPath = character(0)`, and a length-2 `outputPath` hits "the condition has length > 1" instead.

This gap is carried forward from the prior code (`.canonicalizeIdRef(outputPath); if (!(outputPath %in% ...))` had the same zero/multi-length exposure), so it's not a regression, but since this PR is specifically introducing the typed `outputPathRefNotFound` message, it'd be worth guarding the scalar-string precondition in `.matchOutputPathRef()`/`.resolveOutputPathRef()` itself (or calling `PIOutputMapping()`'s type check before resolving) rather than relying on the constructor call order.

---

REFACTORING-LOG.md
Felixmil, line 1: Commit `86b485b` ("Fix buildSimulations unknown-name test to expect the canonical form") changes tracked source (`tests/testthat/test-scenarios.R`) but has no corresponding bullet here, while the other commits in this branch's range each added exactly one (per `REFACTORING.md`'s "Log every commit" rule). `git show 86b485b -- REFACTORING-LOG.md` is empty for that commit, and the last bullet added (16:37, from `b8434b2`) predates `86b485b`'s 16:41 timestamp with nothing appended after it.

Minor/process nit — not a functional issue — but worth a follow-up bullet so the log stays a complete record of this branch's commits.

---

Felixmil (commented):
Automated code review on this authoring-ergonomics PR. Verified each finding against the diff before posting — two are real (if edge-case) correctness gaps in the new canonicalization/resolution logic, one is a process nit (missing `REFACTORING-LOG.md` bullet for a commit in this branch's range). Nothing blocking; leaving as comments rather than change requests.

---

R/scenario-execution.R
Felixmil: Fixed in f440ab79. Dropped the `suppressWarnings()` so a rewritten name now emits the "canonicalized to a safe form" warning naming the id it resolved to, which surfaces exactly the wrong-scenario case you describe (`Test?2` -> `test_2`) instead of running it silently. Also corrected the roxygen on `runScenarios()`/`buildSimulations()` to say the argument is fully canonicalized (lowercasing + character substitution), not merely case-insensitive. Kept the canonicalization itself, since it matches how the PI scenario references already resolve.

R/parameter-identification.R
Felixmil: Fixed in f440ab79. Added a scalar-string guard to `.matchOutputPathRef()`: a `NULL` / `character(0)` / length-2 `outputPath` now returns `NULL`, so `.resolveOutputPathRef()` raises the typed `outputPathRefNotFound` abort instead of a raw zero/multi-length condition error. Also hardened `.nearestMatch()` to suggest nothing for a non-scalar query (it was indexing `adist(...)[1, ]`). Added a regression test for `outputPath = NULL` and a length-2 vector.

REFACTORING-LOG.md
Felixmil: Fixed in f440ab79. Added the missing `REFACTORING-LOG.md` bullet for the earlier test-fix commit.

man/addPITask.Rd
```diff
@@ -23,9 +23,11 @@ existing task id.}
...
 in \code{names(project$definitions$scenarios)}.}
 
-\item{parameters}{Non-empty list of \code{PIParameter} records.}
+\item{parameters}{List of \code{PIParameter} records. May be empty (\code{list()})
+to create a task seeded later with \code{\link[=addPIParameter]{addPIParameter()}}.}
 
-\item{outputMappings}{Non-empty list of \code{PIOutputMapping} records.
+\item{outputMappings}{List of \code{PIOutputMapping} records. May be empty
+(\code{list()}) to create a task seeded later with \code{\link[=addPIOutputMapping]{addPIOutputMapping()}}.
 Each \code{outputPath} must exist in \code{names(project$definitions$outputPaths)}.}
```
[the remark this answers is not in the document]
Felixmil: Fixed in f440ab79. Updated the `addPITask()` roxygen so `outputMappings` states each `outputPath` is either an output-path id or the literal model path of a defined output path.

NEWS.md
```diff
@@ -80,12 +81,14 @@ Deprecation intensity follows the lifecycle policy: forwarding shims that still
...
 - `loadObservedData()` now aborts when two observed-data sources resolve to the same data-set name (previously the later source silently shadowed the earlier), and a `script` source that returns a list of data sets is keyed by each data set's own name. (#1055) It also warns once per session that a `script` source executes arbitrary R when it is resolved (only resolve observed data from a project you trust), and warns by name when a `programmatic` source resolves to no data (a sentinel read from disk with no matching in-session `DataSet`) instead of dropping it silently. (#1033, #1081)
+- `PIOutputMapping()` (and the `outputPath` argument of `addPITask()` / `addPIOutputMapping()`) now accepts either an output-path id or the literal model path of a defined output path, resolving both to the id; an `outputPath` matching no defined output path aborts with a "did you mean" hint. (#1157)
```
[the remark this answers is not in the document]
Felixmil: Fixed in f440ab79. Reworded the NEWS bullet: resolution to the canonical id happens at add time via `addPITask()` / `addPIOutputMapping()`, and `PIOutputMapping()` on its own stores the supplied `outputPath` verbatim.

REFACTORING-LOG.md
```diff
@@ -40,3 +40,7 @@ One bullet per commit, newest at the bottom: `- YYYY-MM-DD HH:MM: plain-language
 - 2026-07-23 14:31: ponytail cleanup: inline single-use `.asPopulation`, shrink `.populationEntryRequiredFields`, fold `.resolveProgrammaticPopulation` into its switch arm (#1130).
 - 2026-07-23 14:45: Made the esqlabsRSettingNames test order-insensitive (set-equality on the names plus an enum-shape check) instead of asserting a fixed list order, since the names derive from an environment whose order is not guaranteed. (#1059)
 - 2026-07-23 15:06: PR review: narrow injected-population dispatch to a real `Population` object; clean up inline abort strings (#1024).
+- 2026-07-23 16:17: runScenarios()/buildSimulations() now match the `scenarios` argument case-insensitively, so a scenario runs under the mixed-case name it was authored with. (#1157)
```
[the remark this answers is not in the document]
Felixmil: Fixed in f440ab79. Reworded the log entry to say the argument is canonicalized and matched to the canonical id, dropping the "runs under the mixed-case name" phrasing. Also removed the dead `is.list()` guard and gave the aggregated `addPITask()` outputPath-not-found error the same actionable guidance as the standalone message.

PR #1160, opened 2026-07-23
Migrate legacy 5.x Excel projects (applications, PI, columns, ids, round-trip)

### LINKED ISSUE #1158: Excel importer: legacy 5.x layout not migrated (applications, PI, columns, ids, round-trip)
## Purpose

`importProjectFromExcel()` cannot faithfully migrate a real pre-6.0 Excel project written in the standard 5.x layout. Depending on the workbook it either aborts outright or silently drops whole sections, and the result fails validation and cannot run. Several of these reproduce against the package's own bundled fixture `tests/testthat/data/TestProjectExcel/`, so no external project is needed to see them.

This continues the Excel-bridge work: the correctness pass landed, but migrating a genuine legacy project surfaces a further set of layout-compatibility and round-trip gaps. Grouping them here since they share one fix locus (`R/project-excel.R` parsers) and one theme (the importer assumes a newer Excel shape than 5.x projects were written in).

## Reproduce (bundled fixture, no external data)

```r
library(esqlabsR)
src <- system.file("..", "tests", "testthat", "data", "TestProjectExcel", package = "esqlabsR")
# (or the in-repo path tests/testthat/data/TestProjectExcel/)
out <- tempfile("fix"); dir.create(out)
file.copy(list.files(src, full.names = TRUE), out, recursive = TRUE)
jp <- importProjectFromExcel(file.path(out, "ProjectConfiguration.xlsx"), outputDir = out, overwrite = TRUE)
p  <- loadProject(jp)
length(p$definitions$applications)             #> 0   (expected 3)
length(p$definitions$parameterIdentification)  #> 0   (expected 1)
length(p$definitions$observedData)             #> 0
```

## Importer drops all Applications (legacy one-sheet-per-protocol layout)

- The Applications parser (`R/project-excel.R:272`) only builds application definitions from a sheet literally named `ApplicationProtocols`; every other sheet is absorbed into `parameterSets`. A 5.x `Applications.xlsx` has one sheet per protocol and no `ApplicationProtocols` sheet (the bundled fixture is exactly this shape), so zero applications are created and the protocol values land only as parameter sets.
- Consequence: every scenario that names an application by id now has a dangling reference; the imported project fails validation with one critical error per scenario. The fix is to create, per protocol sheet, both a parameter set and an `Application` wrapping it (the reified `{id, parameterSets}` shape).

## Importer drops the parameter-identification task (legacy PI layout)

- `.parseExcelParameterIdentification()` (`R/project-excel.R:1665`) returns nothing unless a sheet named `PITasks` exists. The 5.x layout keys tasks by a `PITaskName` column across `PIOutputMappings` / `PIParameters` / `PIConfiguration` / `AlgorithmOptions` / `CIOptions`, with no `PITasks` sheet, so the whole PI configuration is silently dropped.

## Importer hard-aborts on a Scenarios sheet missing `OverwriteFormulasInSS`

- `.scenarioSheetRequiredColumns` (`R/project-excel.R:1244`) lists `OverwriteFormulasInSS` as hard-required and `.parseExcelScenarios()` aborts if it is absent. That column is newer than 5.x, so any real pre-6.0 `Scenarios.xlsx` is rejected before anything is read:

  ```
  Error in `.parseExcelScenarios()`:
  ! The Scenarios sheet is missing required column: "OverwriteFormulasInSS".
  ```
- This is inconsistent with the sibling `InitialConditions` column a few lines below, which is explicitly guarded (`"a newer column; an older Scenarios sheet omits it, so guard the lookup rather than abort"`). `OverwriteFormulasInSS` should be defaulted to `FALSE` the same way rather than blocking migration.

## Observed data is never imported

- The importer has no parse step for the `dataFile` / `dataImporterConfigurationFile` recorded in the project configuration (parse sections are `R/project-excel.R:208-311`), so `observedData` is always empty after import and any plot or PI mapping that references observed data has nothing to resolve against. If this is intentional (observed data is authored via `addObservedData()` in 6.0.0), the importer should at least warn that a configured data file was skipped.

## Individual parameter-set link is dropped

- An individual with an individual-specific parameter sheet gets the parameter set created, but the imported individual definition has no `parameterSets` field linking to it, so the override is orphaned and never applied when that individual is simulated.

## Parameter-set id is not canonicalized to a filename-safe form

- A parameter-set whose name contains characters like a comma or spaces is written to disk with those characters intact (e.g. a `parameter-sets/` file named with a literal comma and spaces), contradicting the documented "lowercased and made filename-safe" id contract. Canonicalization should replace filename-unsafe characters, not only lowercase.

## Quoted multi-value parameter-set cell is mis-parsed

- A `ModelParameterSheets` cell using the quoted-CSV convention (`"A", "B", "C, with comma"`) is parsed with the quotes kept as literal characters and the embedded comma still treated as a separator, yielding broken tokens that resolve to nothing. Quotes should be stripped and quoted commas protected.

## Round-trip encoding disagreements (importer vs `add*()`)

The importer and the authoring API represent the same value differently, so a project round-tripped through the authoring API is not byte-identical to the imported one:

- A parameter entry with no unit: importer writes JSON `null`; `addParameterEntry()` rejects `NA` and requires `""`.
- `proteinOntogenies` on a population: the importer writes it, but `addPopulation()` / `setPopulation()` reject it as an unknown field, so an imported population's ontogenies cannot be re-authored or edited. (`addIndividual()` accepts it, so the asymmetry is population-specific.)
- A scenario with `steadyState = FALSE`: importer writes `steadyStateTime: null` / `steadyStateTimeUnit: null`; the authoring API writes the formal defaults (`1000` / `"min"`). Harmless functionally, but shows up as `git diff` noise after any authoring round-trip.

## Checklist

- [ ] Applications: build an `Application` per protocol sheet (one-sheet-per-protocol layout)
- [ ] PI: parse the multi-sheet `PITaskName`-keyed layout, not only a `PITasks` sheet
- [ ] Scenarios: default `OverwriteFormulasInSS` when the column is absent (match `InitialConditions`)
- [ ] Observed data: import the configured data file, or warn that it was skipped
- [ ] Link an individual to its individual-specific parameter set on import
- [ ] Canonicalize parameter-set ids to filename-safe (not only lowercase)
- [ ] Strip quotes and protect quoted commas in multi-value cells
- [ ] Reconcile no-unit / `proteinOntogenies` / steady-state-off encoding between importer and authoring API

Felixmil:
`importProjectFromExcel()` could not faithfully migrate a real pre-6.0 Excel project written in the standard 5.x layout: depending on the workbook it aborted outright or silently dropped whole sections, and the result failed validation and could not run. All of these reproduced against the bundled fixture `tests/testthat/data/TestProjectExcel/`. This PR makes that fixture (and a genuine legacy project shaped like it) import, load, and validate clean, and makes an imported project round-trip through the authoring API without diff noise. The fixes share one locus (the `R/project-excel.R` parsers) plus two small reconciliations in the id canonicalizer and the population authoring API.

## Applications

With no `ApplicationProtocols` sheet (the 5.x one-sheet-per-protocol layout), the importer now builds one `Application` per protocol sheet, each wrapping the same-named parameter set, so a scenario naming a protocol by id resolves instead of dangling. The newer single-sheet layout is unchanged.

## Parameter identification

The importer now parses the 5.x `PITaskName`-keyed multi-sheet layout (`PIParameters` / `PIOutputMappings` / `PIConfiguration` / `AlgorithmOptions` / `CIOptions`, no `PITasks` sheet), reassembling each task's parameters (joining `Container Path` + `Parameter Name` into the flat path), output mappings, and nested configuration into the same shape the newer layout produces. An output mapping that names its output path by full OSPS path is rewritten to the id of the matching output-path definition so the reference resolves. The newer single-sheet layout is unchanged.

## Scenarios

A `Scenarios.xlsx` lacking the newer `OverwriteFormulasInSS` column no longer aborts the import; the column defaults to `FALSE`, matching the sibling `InitialConditions` handling. A multi-value cell using the legacy quoted-CSV convention (`"A", "B", "C, with comma"`) now parses correctly (quotes stripped, a quoted comma kept inside one token) alongside the existing backslash-escape convention.

## Observed data

The importer reifies the configured `dataFile` / `dataImporterConfigurationFile` as one `excel` observed-data definition listing the workbook's sheets, so a plot or PI mapping referencing observed data has something to resolve against. A configured-but-missing data file warns and is skipped rather than silently producing nothing.

## Individuals

A parameter-set sheet named after an individual is linked to that individual on import (matched on the canonical id), so an individual-specific override is applied rather than orphaned.

## Ids and round-trip

Definition id canonicalization (the single shared chokepoint for the whole authoring API) now replaces commas and interior whitespace with `_`, so an id is a safe single filename segment, fixing a literal-comma parameter-set filename. Two encoding disagreements between the importer and the authoring API are reconciled so an imported project round-trips without diff noise: `addPopulation()` / `setPopulation()` accept `proteinOntogenies` (individuals already did), and a non-steady-state scenario's steady-state time/unit default to the same `1000` / `"min"` the authoring API writes rather than `null`. The no-unit parameter-entry case needed no change: the importer and `addParameterEntry("")` already store the same absent-`units` shape.

## Known gap (pre-existing, out of scope)

The Excel export writes the newer single-sheet PI layout, and its `config.*` flattening does not round-trip a task's nested `algorithmOptions` / `ciOptions`. Loading the 5.x import is clean; only a re-export drops those nested options. This is a writer-side limitation the importer work surfaced, not introduced.

## Example

Migrating the bundled legacy 5.x fixture, which previously produced 0 applications, 0 PI tasks, and 0 observed-data entries:

``` r
devtools::load_all("esqlabsR", quiet = TRUE)
# Migrate the bundled legacy 5.x Excel fixture into a JSON project.
src <- system.file("..", "tests", "testthat", "data", "TestProjectExcel", package = "esqlabsR")
out <- withr::local_tempdir()
file.copy(list.files(src, full.names = TRUE), out, recursive = TRUE)
#> [1] TRUE TRUE TRUE TRUE TRUE TRUE
jp <- importProjectFromExcel(file.path(out, "ProjectConfiguration.xlsx"), outputDir = out, overwrite = TRUE, silent = TRUE)
p <- loadProject(jp)
# Sections that were previously dropped are now populated:
length(p$definitions$applications)            # was 0
#> [1] 3
length(p$definitions$parameterIdentification) # was 0
#> [1] 3
length(p$definitions$observedData)            # was 0
#> [1] 1
# The comma/space parameter-set id is now filename-safe:
"sheet__with_comma" %in% names(p$definitions$parameterSets)
#> [1] TRUE
```

Closes #1158

## Summary by CodeRabbit

* **New Features**
  * Improved Excel project import compatibility with legacy 5.x files, including applications, parameter identification, observed data, scenarios, and individual parameter overrides.
  * Added support for quoted and escaped comma-separated Excel values.
  * Added support for population protein ontogeny settings.
* **Bug Fixes**
  * Improved identifier safety by normalizing spaces, commas, and other unsafe characters.
  * Added clearer notifications when configured observed-data files are missing.
  * Improved round-trip consistency and default handling for imported scenarios.


R/project-excel.R
```diff
@@ -251,6 +251,25 @@ importProjectFromExcel <- function(
...
+          # override; link it so the override is applied. The match is on the
+          # canonical id, so `Indiv1` sheet links to the `Indiv1` individual
+          # regardless of case, and the sheet's set id is added to the
+          # individual's `parameterSets` (deduplicated, keeping any set the
+          # `ParameterSets` column already named).
+          jsonData$individuals <- lapply(jsonData$individuals, function(indiv) {
+            match <- paramSheetNames[
+              .canonicalizeOneId(indiv$individualId) ==
+                vapply(paramSheetNames, .canonicalizeOneId, character(1))
+            ]
```
Felixmil: A blank/NA `IndividualId` row makes this inject `NA` into every individual's `parameterSets`.

If `IndividualBiometrics` has a row with a blank/NA `IndividualId` (a common trailing blank row) and there's at least one non-biometrics sheet, `.canonicalizeOneId(indiv$individualId)` returns `NA_character_`. Comparing that to `vapply(paramSheetNames, .canonicalizeOneId, character(1))` produces an all-`NA` logical vector, and indexing `paramSheetNames[<all-NA logical>]` in R returns a vector of `NA`s of `length(paramSheetNames)` — not an empty vector. So `length(match) > 0L` is `TRUE`, and `unique(c(unlist(indiv$parameterSets), match))` inserts `NA_character_` into that individual's `parameterSets`, corrupting the imported project with a dangling reference.

Worth guarding, e.g. `match <- match[!is.na(match)]` or filtering blank `individualId`s out before this loop.

---

R/project-excel.R
```diff
@@ -1301,6 +1372,54 @@ projectStatus <- function(project, silent = FALSE) {
...
+  dataFilePath <- .resolveProjectPath(dataFile, dataFolder, "dataFile")
+  if (!file.exists(dataFilePath)) {
+    cli::cli_warn(messages$importSkippedObservedData(dataFile))
+    return(jsonData)
+  }
+
+  importerConfig <- prop("dataImporterConfigurationFile")
+  entry <- list(
+    type = "excel",
+    file = basename(dataFile),
```
Felixmil: Storing `basename(dataFile)` here discards any subfolder component, even though the existence check just above (`.resolveProjectPath(dataFile, dataFolder, ...)` at line 1399) resolves the full relative path.

If a legacy `Project.xlsx`'s `dataFile` names a file in a subfolder relative to `dataFolder` (e.g. `"Subdir/Data.xlsx"`), the import-time check succeeds because it uses the full relative value, but `entry$file <- basename(dataFile)` stores only `"Data.xlsx"`. On load, `.loadObservedExcel()` → `.resolveDataPath(entry$file, dataFolder)` (`R/observed-data.R:700-714`) looks for `<dataFolder>/Data.xlsx`, which doesn't exist — so observed-data resolution fails post-import even though the import itself reported success.

Every sibling Excel-bridge file property elsewhere in this importer is carried through unresolved/untruncated, so this is also inconsistent with the rest of the parser. Consider storing `dataFile` (and `importerConfig`) as given (or the path relative to `dataFolder`) rather than truncating to basename.

---

R/project-excel.R
```diff
@@ -1729,6 +1853,290 @@ projectStatus <- function(project, silent = FALSE) {
...
+  }
+  options <- list()
+  for (i in seq_len(nrow(rows))) {
+    name <- as.character(rows[["OptionName"]][[i]])
+    if (is.na(name) || name == "") {
+      next
+    }
+    raw <- rows[["OptionValue"]][[i]]
+    numeric <- suppressWarnings(as.numeric(raw))
+    options[[name]] <- if (!is.na(numeric)) numeric else as.character(raw)
```
Felixmil: A boolean-looking `OptionValue` (e.g. `"TRUE"`/`"FALSE"`, a common encoding for algorithm-option flags in 5.x sheets) fails `as.numeric()` → `NA`, so this falls back to `as.character(raw)` and stores the literal string `"TRUE"` rather than the logical `TRUE`.

If a downstream `PIConfiguration` consumer (e.g. the ospsuite bridge) expects a real logical for that option, this silently passes a string where a boolean was expected instead of erroring or coercing correctly. Might be worth trying a logical coercion (e.g. via the existing `.toLogical` used by sibling scalar fields) before falling through to numeric/character.

---

R/project-excel.R
```diff
@@ -1729,6 +1853,290 @@ projectStatus <- function(project, silent = FALSE) {
...
+#
+# @keywords internal
+# @noRd
+.pi5xPath <- function(containerPath, parameterName) {
+  container <- as.character(containerPath)
+  parameter <- as.character(parameterName)
+  if (is.na(container) || container == "") {
+    return(parameter)
+  }
+  paste(container, parameter, sep = "|")
```
Felixmil, line 2117: A blank/NA `Parameter Name` cell with a filled `Container Path` produces the literal string `"...|NA"` instead of a null/invalid path.

`parameter <- as.character(parameterName)` stays `NA_character_` when the cell is blank, `is.na(container)` is `FALSE` so the early return above is skipped, and `paste(container, parameter, sep = "|")` stringifies the `NA` (base R's `paste()` renders `NA` as `"NA"`), producing e.g. `"Aciclovir|NA"` as the parameter's `path` — a value that looks valid but silently references nothing, rather than something validation would catch.

---

R/project-excel.R
```diff
@@ -1729,6 +1853,290 @@ projectStatus <- function(project, silent = FALSE) {
...
+  }
+  configuration
+}
+
+# The rows of a 5.x PI sheet belonging to one task (empty data frame when the
+# sheet is absent or has no matching rows).
+#
+# @keywords internal
+# @noRd
+.pi5xTaskRows <- function(df, task) {
```
Felixmil, line 2051: Nit (not a correctness issue): `.pi5xTaskRows` re-filters the full sheet for every task, and `.parseExcelPI5x`'s `lapply(taskNames, ...)` (line 1890) calls it independently for params/mappings/config/algOptions/ciOptions — 5 separate linear scans per task across the 5 PI sheets, instead of grouping each sheet by `PITaskName` once up front. Harmless at fixture scale, but a needless repeated-scan cost that grows with a larger legacy project's PI sheets. Not blocking — just flagging for a possible follow-up if large PI workbooks turn out to be slow to import.

---

Felixmil (commented):
High-effort code review of the 5.x Excel migration parsers in `R/project-excel.R`. Found 4 correctness bugs (all verified against the current head, `9b8c9f76`) plus 1 non-blocking efficiency nit — inline comments below. The bugs share a shape: several 5.x-layout code paths don't guard against blank/NA cells the way the rest of the importer does, so a blank `IndividualId`, `Parameter Name`, or boolean-string `OptionValue` in a real legacy workbook produces a silently-wrong value (an injected `NA`, a `"...|NA"` path, or a stringified `"TRUE"`) instead of erroring or being skipped, and observed-data files in a subfolder lose their path on import. None of these are covered by the PR's own summary of fixes, and none overlap with existing open issues.

---

R/project-excel.R
Felixmil: Fixed in 7628427d. Guard the blank/NA `individualId` explicitly (`is.na(indivCanonical)` early-returns) so an all-NA logical index can no longer inject `NA` into `parameterSets`. Regression test: "the individuals import skips a blank IndividualId row rather than injecting NA".

R/project-excel.R
Felixmil: Fixed in 7628427d. `file` and `importerConfiguration` are now stored as given (relative to `dataFolder`), not truncated to a basename, so a subfolder file still resolves via `.resolveDataPath()` on load. Only the section key is reduced to a basename (an id cannot hold a path separator). Regression test: ".parseExcelObservedData keeps a subfolder path rather than truncating to basename".

R/project-excel.R
Felixmil: Fixed in 7628427d. An explicit `"TRUE"`/`"FALSE"` `OptionValue` now coerces to a real logical; a numeric value stays numeric (so a `1`/`0` option is not misread as a boolean), everything else stays a string. Regression test: ".pi5xOptionRows coerces a boolean-string option but keeps numbers numeric".

R/project-excel.R
Felixmil: Fixed in 7628427d. `.pi5xPath()` returns `NULL` for a blank/NA parameter name, so the caller drops the `path` field (validation then flags the incomplete parameter) instead of building a `"<container>|NA"` path. Regression test: ".pi5xPath drops a blank parameter name instead of building a \`|NA\` path".

R/project-excel.R
Felixmil: Acknowledged, leaving as-is for now. At fixture and realistic legacy-project PI scale the repeated per-task scans are negligible; grouping each sheet by `PITaskName` once up front is a worthwhile follow-up only if a large PI workbook turns out to be slow to import, so deferring rather than adding the indexing complexity pre-emptively.

R/project-excel.R
```diff
@@ -1729,6 +1869,306 @@ projectStatus <- function(project, silent = FALSE) {
...
+  if (is.null(df) || !("PITaskName" %in% names(df)) || nrow(df) == 0L) {
+    return(df[0, , drop = FALSE] %||% data.frame())
+  }
+  df[
+    !is.na(df$PITaskName) & as.character(df$PITaskName) == task,
+    ,
+    drop = FALSE
+  ]
+}
```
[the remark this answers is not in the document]
Felixmil: Fixed in 20bf8106. `.pi5xTaskRows()` now returns `data.frame()` for `NULL` input before evaluating `df[0, , drop = FALSE]`, so importing a legacy 5.x workbook missing optional PI sheets no longer aborts.

PR #1161, opened 2026-07-23
Auto-upgrade a previous-version project snapshot on restoreProject()

### LINKED ISSUE #1156: Auto-upgrade a previous-version project to the 6.0 format on load
## Problem

A user with an esqlabsR 5.x project (Excel-first: `Project.xlsx` + `Configurations/`) has no single-step way to open it in 6.0 and get a definitions-tree project. The only path today is `importProjectFromExcel()`, a separate manual step that the user has to know to reach for. `loadProject()` and `restoreProject()` (the natural "open my project" entrypoints) do not accept a previous-version project at all: `restoreProject()` reads only a v6 `.esqlabsR` file or a legacy single-file `Project.json`, and errors on anything older.

So a previous-version snapshot is importable only if you already know it must go through the Excel bridge, and the result feels like a conversion rather than opening the project.

## Ask

A previous-version esqlabsR project should be openable through the ordinary load path and auto-upgraded to the 6.0 definitions-tree format on read, no separate command required.

Concretely:

- `loadProject()` / `restoreProject()` detect that the source is a 5.x project (Excel-first workbooks) rather than a v6 project, and upgrade it in place to the current definitions tree instead of erroring.
- The upgrade reuses the existing Excel-import logic (`importProjectFromExcel()`), it does not reimplement parsing. This issue is about wiring that behind the load path and making it automatic, not about a new parser.
- On a successful auto-upgrade, inform the user (a one-line `cli::cli_inform` via the messages catalog) that a previous-version project was detected and upgraded, so the behaviour is visible rather than silent.

## Out of scope

- PK-Sim project snapshots (`.pksim5` / PK-Sim-generated JSON) as a project source. No esqlabsR code reads those as a project today; if wanted, a separate issue.
- Upgrading an older esqlabsR JSON `schemaVersion` (< 2.0). The current fixtures are all 2.0; if a pre-2.0 JSON snapshot ever needs upgrading on read, track it separately.

## Notes

- `importProjectFromExcel()` already carries the one documented caveat that observed data does not travel through Excel (it stays in the definitions tree). Any auto-upgrade path inherits that caveat; call it out in the upgrade message or docs.
- Keep Excel as a supported secondary entrypoint; this only removes the need to reach for it manually when the intent is "open my old project".

Felixmil:
This makes `restoreProject()` open a project snapshot written by a previous esqlabsR version. Such a snapshot (the single JSON `snapshotProjectConfiguration()` produced) is a mirror of the project's Excel workbooks, not a v6 `Project.json`, so today it fails the load. `restoreProject()` now detects one and upgrades it to the current definitions-tree format on read, so a previous-version project opens through the ordinary restore path without a separate manual import step. `loadProject()` stays strict (`Project.json` only), and the Excel `.xlsx` path stays with the dedicated `importProjectFromExcel()` bridge.

The PR targets `1158-excel-legacy-migration` because the upgrade reuses that branch's importer fixes (applications, PI, observed data, ids); basing here keeps the diff to just this change. It should retarget to `json-based-project` once #1158 merges.

## Detection

A previous-version snapshot is recognized by the absent `schemaVersion` (every v6 snapshot carries `"2.0"`) together with the mirror's `projectConfiguration` sheet and at least one workbook section, so neither a v6 snapshot nor an unrelated JSON is mistaken for one. This top-level shape has been stable across every released `snapshotProjectConfiguration()`.

## Upgrade

The upgrade rebuilds the mirrored Excel workbooks into a scratch directory (reversing the sheet-to-JSON dump and restoring per-column types) and runs them through `importProjectFromExcel()`, writing the definitions tree into the restore `dir`. The rebuilt root workbook is named `Project.xlsx` so the container lands as the canonical `Project.json`. A one-line `cli_inform` announces the upgrade and notes that observed data does not travel in a snapshot (the same caveat the Excel bridge documents).

## Example

```r
# A snapshot written by a previous esqlabsR version opens directly:
project <- restoreProject("ProjectConfiguration.json", "my-upgraded-project")
#> i Detected a previous-version project snapshot and upgraded it to the current project format.
#> ! Observed data does not travel in a snapshot; add it with `addObservedData()` if a plot or parameter identification needs it.
project$definitions$scenarios
#> scenarios: 8 definitions
```

Closes #1156

## Summary by CodeRabbit

* **New Features**
  * `restoreProject()` now accepts monolithic project snapshots created by earlier esqlabsR versions.
  * Legacy snapshots are automatically upgraded to the current project format during restoration.

* **Bug Fixes**
  * Added clear errors for malformed legacy snapshot data.
  * Improved compatibility while preserving restoration of current-format snapshots.

* **Documentation**
  * Clarified supported snapshot formats and upgrade behavior.
  * Documented that observed data is not included and may need to be added separately.


R/import-legacy-snapshot.R
```diff
@@ -0,0 +1,201 @@
...
+
+# Rebuild the Excel workbooks a monolithic snapshot mirrors, into `dir`. Writes
+# the root `ProjectConfiguration.xlsx` (path/property sheet), one workbook per
+# mirrored section under `Configurations/`, and any populations CSVs under
+# `Configurations/PopulationsCSV/`. Returns the path to the root
+# `ProjectConfiguration.xlsx`, the entry point `importProjectFromExcel()` reads.
+#
+# @keywords internal
+# @noRd
+.materializeLegacySnapshot <- function(jsonData, dir) {
```
Felixmil, line 111: `.materializeLegacySnapshot()` never recreates the observed-data workbook/folder (`dataFile`/`dataFolder`), but it does copy the `projectConfiguration` property sheet verbatim, so those properties still point at the (unmaterialized) data file.

Confirmed against the bundled fixture: `tests/testthat/data/TestProjectExcel/ProjectConfiguration.json` sets `dataFile = "TestProject_TimeValuesData.xlsx"` and `dataFolder = "Data/"`. On import, `.parseExcelObservedData()` (`R/project-excel.R:1402-1413`) does `file.exists(dataFilePath)` and, since the workbook was never written into the scratch dir, always hits the `!file.exists` branch and calls `cli::cli_warn(messages$importSkippedObservedData(...))`. That warning is **not** gated by `silent` — `silent` only guards the `interactive()` prompt at `R/project-excel.R:448` — so it fires on every upgrade of a snapshot that had observed data configured, in addition to the intended single `cli_inform` in `.upgradeLegacySnapshot()`.

Concretely, the new test `restoreProject upgrades a previous-version snapshot to a v6 tree` (`tests/testthat/test-import-legacy-snapshot.R:55-77`) only wraps the call in `expect_message()`, not `suppressWarnings()` (unlike the analogous calls in `test-project-excel.R`). Since `Config/testthat/edition: 3` is set (`DESCRIPTION:62`), and 3e promotes uncaught warnings to test failures, this looks like it will fail CI for this exact fixture/test.

Worth either suppressing/collapsing that warning during the upgrade (it's redundant with the `upgradedLegacySnapshot()` notice, which already tells the user observed data doesn't travel), or wrapping the test call appropriately if the warning is intentional.

---

R/import-legacy-snapshot.R
```diff
@@ -0,0 +1,201 @@
...
+  # `Project.json` container, the canonical name a restored project carries
+  # (matching the v6-snapshot restore path), rather than one named after the
+  # previous version's `ProjectConfiguration` workbook.
+  pcPath <- file.path(dir, "Project.xlsx")
+  .writeExcel(.legacySheetToDf(jsonData$projectConfiguration), pcPath)
+
+  # The snapshot keys each configuration workbook by the section name the
+  # exporter used; map each back to its conventional workbook filename (the
+  # default names the bridge expects when the property is a bare filename).
+  workbookFiles <- c(
```
Felixmil: The rebuilt workbook filenames are hardcoded here (`workbookFiles`) instead of read from the legacy snapshot's own `projectConfiguration` sheet (`modelParamsFile`, `individualsFile`, `populationsFile`, `scenariosFile`, `applicationsFile`, `plotsFile`, `parameterIdentificationFile`), even though those properties are documented, user-customizable filenames (see the fixture's own property descriptions, e.g. "Name of the excel file with global model parametrization... Must be located in the `configurationsFolder`").

Since `.materializeLegacySnapshot()` copies the `projectConfiguration` sheet verbatim, a legacy project that customized one of these filenames will have its `projectConfiguration` record the custom name, while the workbook is actually written under the hardcoded conventional name instead. `importProjectFromExcel()`'s section loop resolves the path from the (customized) property, finds no file there, and silently skips that whole section (`if (!is.null(file) && file.exists(file))` — no warning). The bundled fixture happens to use the default names for every property, so this doesn't currently surface as a test failure, but it will silently drop a section for any real-world legacy project that renamed a workbook.

Consider looking up each workbook's actual filename from `jsonData$projectConfiguration` (matching on the corresponding property, e.g. `modelParamsFile`) rather than the hardcoded map.

---

R/import-legacy-snapshot.R
```diff
@@ -0,0 +1,201 @@
...
+
+# Restore a column's type after the mirror flattened every cell to a string: a
+# column whose non-blank values are all `TRUE`/`FALSE` becomes logical, one that
+# parses fully as numeric becomes numeric, everything else stays character. This
+# matches what the previous version wrote, so the rebuilt workbook reads back
+# the same values it was exported from.
+#
+# @keywords internal
+# @noRd
+.restoreColumnTypes <- function(df) {
```
Felixmil, line 254: `.restoreColumnTypes()` numeric-coerces any column whose non-blank values all parse via `as.numeric()`, with no exemption for id-like columns. A zero-padded id (e.g. an `IndividualId`/`Scenario_name`/population-id column with values like `"01"`, `"02"`, `"010"`) parses cleanly as numeric, so it gets coerced to `1`, `2`, `10`. `writexl` then serializes those as numeric Excel cells, and re-parsing on the way back through `.parseExcelIndividuals()`/`.parseExcelScenarios()` stringifies them as `"1"`, `"2"`, `"10"` — silently dropping the zero-padding and breaking any foreign-key reference that still uses the original padded form (a scenario's `IndividualId`, a parameter-set sheet named after the individual, etc.).

This is a pre-existing-data edge case rather than something the bundled fixture exercises, but since this function is specifically restoring types for id-bearing columns coming out of a previous-version project, it might be worth excluding known id columns (or any column feeding into a canonicalized id) from the numeric-coercion branch.

---

R/project-snapshot.R
```diff
@@ -167,6 +173,15 @@ restoreProject <- function(snapshot, dir = ".", overwrite = FALSE) {
     dir.create(dir, recursive = TRUE, showWarnings = FALSE)
   }
 
+  jsonData <- jsonlite::fromJSON(snapshot, simplifyVector = FALSE)
+
+  # A previous-version monolithic snapshot (Excel workbooks dumped to one JSON by
+  # an older `snapshotProjectConfiguration()`) is not a v6 `Project.json`; upgrade
+  # it through the Excel bridge rather than trying to parse it as a v6 snapshot.
+  if (.isLegacySnapshot(jsonData)) {
```
Felixmil, line 181: This early return for a legacy snapshot skips the stale-tree-handle warning that the v6-snapshot restore path emits on overwrite. `replacedExistingTree <- isProjectInitialized(dir)` is computed above (line 171), and further down (lines 226-227) there's `if (replacedExistingTree && overwrite) cli::cli_warn(messages$restoreOverwroteTree(dir))` — but the legacy-snapshot branch returns at line 182, before ever reaching that check.

So `restoreProject()` with a legacy snapshot, `overwrite = TRUE`, and a `dir` that already holds an initialized project tree replaces that tree exactly like the v6 path does, but the user gets no warning that any previously-loaded `Project` object bound to `dir` is now stale. Since `.upgradeLegacySnapshot()` also forwards `overwrite` into `importProjectFromExcel()` and does the actual overwrite, it'd be worth surfacing the same warning here (e.g. after the upgrade returns, or by threading `replacedExistingTree` through).

---

Felixmil (commented):
Automated code review (high effort) of this PR's actual diff (base `1158-excel-legacy-migration`...head `b258992c`, the 7 files listed on the PR: `R/import-legacy-snapshot.R`, `R/messages.R`, `R/project-snapshot.R`, plus docs/log).

4 findings, all verified against the current diff and the bundled `TestProjectExcel` fixture:

1. `.materializeLegacySnapshot()` doesn't recreate the observed-data workbook, so every upgrade of a snapshot with `dataFile` configured (the bundled fixture included) triggers an uncaught `cli_warn` — likely a CI failure given `Config/testthat/edition: 3` and the new test only wrapping the call in `expect_message()`.
2. `.materializeLegacySnapshot()` hardcodes workbook filenames instead of reading them from the snapshot's own `projectConfiguration` properties, so a legacy project with a customized workbook name would silently lose that whole section.
3. `.restoreColumnTypes()` numeric-coerces any all-numeric-looking string column with no id exemption, which would silently strip zero-padding from an id column if one existed in the source data.
4. The new legacy-snapshot early return in `restoreProject()` bypasses the stale-tree-handle warning the v6 path emits on overwrite.

---

R/import-legacy-snapshot.R
Felixmil: Fixed in 2b1e2125. The upgrade now muffles this warning via `withCallingHandlers` on the classed condition `esqlabsR_importSkippedObservedData` (classed at its source in `R/project-excel.R`), since it is redundant with the observed-data caveat the `upgradedLegacySnapshot()` inform already carries. New test asserts `expect_no_warning()` on the fixture upgrade, so the 3e failure is covered.

R/import-legacy-snapshot.R
Felixmil: Fixed in 2b1e2125. `.materializeLegacySnapshot()` now reads each workbook filename from the snapshot's own `projectConfiguration` via `.legacyConfigProperties()` (mapping section -> property, e.g. `scenariosFile`), falling back to the conventional default only when the property is absent, and basenames it so it stays under `configDir`. New test renames `scenariosFile` and asserts the workbook lands under the custom name (and not the hardcoded one).

R/import-legacy-snapshot.R
Felixmil: Fixed in 2b1e2125. `.restoreColumnTypes()` now skips numeric coercion for a column with any leading-zero value (`^0[0-9]`, e.g. `01`, `007`), keeping it as text so the padding and any reference to it survive; a lone `0` or a decimal stays coercible. New test covers a zero-padded id column staying character while a genuine numeric column still coerces.

R/project-snapshot.R
Felixmil: Fixed in 2b1e2125. `replacedExistingTree` is now threaded into `.upgradeLegacySnapshot()`, which emits the same `restoreOverwroteTree` warning the v6 path does when an overwriting restore replaced an initialized tree. New test asserts the warning on an overwriting re-restore.

R/import-legacy-snapshot.R
```diff
@@ -0,0 +1,201 @@
...
+# value that is not a `{column_names, rows}` object (a corrupt or hand-edited
+# snapshot) aborts with a clear message rather than a cryptic downstream error.
+#
+# @keywords internal
+# @noRd
+.legacySheetToDf <- function(sheet) {
+  if (!is.list(sheet) || is.null(sheet$column_names)) {
+    cli::cli_abort(messages$legacySnapshotMalformedSheet())
+  }
+  columnNames <- unlist(sheet$column_names)
```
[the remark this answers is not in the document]
Felixmil: Fixed in c1f825b6. `.legacySheetToDf()` now rejects a present-but-malformed `rows` (a scalar, or a non-list row record) with `legacySnapshotMalformedSheet()`, while an absent `rows` stays a valid empty sheet. Added tests for empty rows, scalar rows, and a non-list row record.

Issue #1146, opened 2026-07-22
DataCombined section: shape & validation


Felixmil:
Umbrella for the `dataCombined` section of the v6 project model. Part of #1088.

Covers the on-disk/JSON shape of DataCombined entries, the authoring surface, and its validation path. These share the `R/data-combined.R` surface and the DataCombined JSON spec, so they are settled together: the shape is decided first, then the validator is rewritten against the final shape.

Issue #1001, opened 2026-04-17
`DataCombined` section in JSON - one entry per DC


PavelBal:
Currently, `dataCombined` section of the project is structures as it was done in Excel - one entry per one observed or simulated data. I think this can be improved, by having sub-entries for simulated results and observed data:

Current state:

```
      {
        "DataCombinedName": "Aciclovir_individual",
        "dataType": "simulated",
        "label": "Aciclovir simulated (male)",
        "scenario": "Aciclovir_iv",
        "path": "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)",
        "dataSet": null,
        "group": "Aciclovir PVB",
        "xOffsets": null,
        "xOffsetsUnits": null,
        "yOffsets": null,
        "yOffsetsUnits": null,
        "xScaleFactors": null,
        "yScaleFactors": null
      },
      {
        "DataCombinedName": "Aciclovir_individual",
        "dataType": "simulated",
        "label": "Aciclovir simulated (female)",
        "scenario": "Aciclovir_iv_female",
        "path": "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)",
        "dataSet": null,
        "group": "Aciclovir PVB",
        "xOffsets": null,
        "xOffsetsUnits": null,
        "yOffsets": null,
        "yOffsetsUnits": null,
        "xScaleFactors": null,
        "yScaleFactors": null
      },
```

Felixmil:
100% right, what about something like this:
```
    "dataCombined": [
      {
        "name": "Aciclovir_individual",
        "simulated": [
          {
            "label": "Aciclovir simulated (male)",
            "scenario": "Aciclovir_iv",
            "path": "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)",
            "group": "Aciclovir PVB",
            "xOffsets": null,
            "xOffsetsUnits": null,
            "yOffsets": null,
            "yOffsetsUnits": null,
            "xScaleFactors": null,
            "yScaleFactors": null
          },
          {
            "label": "Aciclovir simulated (female)",
            "scenario": "Aciclovir_iv_female",
            "path": "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)",
            "group": "Aciclovir PVB",
            "xOffsets": null,
            "xOffsetsUnits": null,
            "yOffsets": null,
[...]

Issue #940, opened 2026-02-10
Plots -> DataCombined: rename columns


PavelBal:
cf https://github.com/esqLABS/esqapp-desktop/issues/26

Felixmil:
This now needs a decision rather than an implementation, because the codebase went a different way.

#1065 unified the JSON identifiers on camelCase (`plotID` → `plotId`, `DataCombinedName` → `dataCombinedName`), and the definition files use `dataCombinedId`, `plotId`, `plotGridId`. The human-readable headers requested here (`Name`, `Data Type`, `Output path`, `X axis offset`, ...) were not adopted, and the Excel bridge still reads and writes `DataCombinedName` and `dataType`. The counterpart esqLABS/esqapp-desktop#26 was closed on 2026-06-29.

Two ways forward:

1. Display-only labels in the app, over the current keys. No package change, no migration.
2. A real rename in the Excel bridge and the JSON definitions. That is a schema migration, so it belongs in the 6.0.0 window rather than after it.

Leaving open on that choice.

Issue #907, opened 2025-11-14
DataCombined rework - define specifications for grouping and data mapping


PavelBal:
## Problem

Currently, mapping of observed data to simulated results is done via specifying the groups in `DataCombined`. This is a shortcoming, as a) mapping cannot be reused, and b) sometimes data must be grouped separately.

## Proposal

Proposed is the extension of the current mapping/grouping concept:

- Separate mapping of observed data to a simulation result (n to 1 relationship)
- Separate grouping of data (simulated and/or observed)

### Simulated and observed data mapping

- [ ] Extend the class `DataSet` from `{ospsuite}` with a property `simulationPath`. In simulated results, this property is automatically set to the output entity's path. For observed data, this property can be set and indicates that the data set is mapped to a simulation result via the path (as defined in the "Observed Data" tab in PK-Sim/MoBi). Should this be a field only in the R class, or should we also add this new property in .NET? Then the user could pre-define the mapping in the data set, and it would be applied as a default to all simulations that use that data set. @msevestre @Yuri05
- [ ] The property can be pre-calculated during observed data import based on metadata, if sufficient data are available.
- [ ] In the `{esqlabsR}` workflow, add a separate sheet/section in the configuration to map observed data to simulation outputs. `simulationPath` setting can be overwritten, or even removed. If the entry for a data set is not available, a default entry (populated during import) will be used.

This way, mapping of observed data and simulation outputs during plotting can be performed automatically in most cases.

### Grouping of data sets

All kinds of datasets, observed or simulated, can be grouped in the `DataCombined` as is done now.

### Plotting and calculating residuals

- [ ] When converting a `DataCombined` to a data frame, add the column `simulationPath` that will be extracted from the `DataSet` objects. This way, an aesthetics grouping based on observed data <-> simulation results mapping is already included!

It has to be specified how the default behavior for the different use cases should be, considering both, sim results <-> obs data mapping, as well as grouping of data sets.

1. One simulated result `S1`, no grouping
    a.  **Individual time profile**: one curve, one legend entry with the label `S1`
    b. **Simulated-vs-observed**: NA
    c. **Calculate residuals**: NA

2. One observed data `O1`, no grouping
    a.  **Individual time profile**: One symbol set, one legend entry with the label `O1`
    b. **Simulated-vs-observed**: NA
    c. **Calculate residuals**: NA

3. Two simulated results `S1` and `S2`, no grouping
[...]

PR #1176, opened 2026-07-24
Validate dataCombined on JSON spec; accept output-path id in addDataCombined

### LINKED ISSUE #1026: Validators operate on JSON specs directly (drop legacy DataFrame reshape)
Part of the JSON-primary rework (#1088). This issue has been narrowed to the piece that is still outstanding on the current `Project` model; the plot side has already been addressed.

## What is already done

The plot-configuration and plot-grid validators no longer go through a legacy flat-DataFrame reshape. They now operate on the JSON specs directly (`.assertPlotConfigurationsBuildable()` and `.assertPlotGridsBuildable()` in `R/create-plots.R`, plus the lazy `.validatePlots()` in `R/plots.R` wired into `validateProject()`). The old `.createPlotGridsFromDataFrames()` idempotency guard and the `.validatePlotConfigurationFromExcel()` / `.validatePlotGridsFromExcel()` helpers are gone.

## What remains

The `dataCombined` validation path still reshapes the JSON spec into the legacy flat tibble before validating, so the transitional code is still present for this one section:

- `.specsToDataCombinedDataFrame()` (`R/data-combined.R`) reshapes the JSON `dataCombined` spec into the flat tibble the legacy Excel-driven path expects.
- `.specEntryToRow()` (`R/data-combined.R`) hardcodes the 13 legacy DataFrame columns (`dataCombinedName`, `dataType`, `label`, `scenario`, `path`, `dataSet`, `group`, `xOffsets`, `xOffsetsUnits`, `yOffsets`, `yOffsetsUnits`, `xScaleFactors`, `yScaleFactors`).
- `.validateDataCombinedFromExcel()` (`R/data-combined.R`, called from `.createDataCombinedFromProcessedDF()`) still validates against that reshaped DataFrame rather than the JSON spec. There is no `dataCombined` entry in the `.validationAdapters` registry in `R/validation.R`; only cross-reference checks read the JSON spec directly.

## Resolution

Rewrite the `dataCombined` validation to operate on the JSON spec directly, then remove the three transitional pieces above and drop the `FromExcel` suffix now that the path is no longer Excel-specific.

### LINKED ISSUE #1042: Support output-path ids in addDataCombined() simulated entries
`addDataCombined()` currently requires each simulated entry's `path` to be a literal model quantity path (matched against the simulation results' `allQuantityPaths`). It does not accept an **output-path id** (a key of `project$outputPaths`), so users and agents must repeat the full literal path even when the project already names it.

This is inconsistent with `addScenario()`, which already resolves `outputPathIds` against the project's `outputPaths` map (`id -> literal path`, see `R/scenarios.R`). The same id-to-path resolution should be available when building a `DataCombined`.

## Why it matters

- Reduces duplication: the literal path is declared once via `addOutputPath()` and referenced by id everywhere else.
- Keeps the API coherent: a single, named output path can be referenced the same way from scenarios and from plots.
- Helps agents and the companion Skill: referencing a short, project-defined id is far less error-prone than re-typing a long `Organism|...|Concentration in container` string.

## Fix sketch

In the simulated-entry handling of `addDataCombined()` (`R/plots.R`), when `path` matches a key of `project$outputPaths`, resolve it to the literal path before storing (or store the id and resolve at plot/run time). Reuse the existing `outputPathIds -> outputPaths` resolution helper used by `addScenario()` rather than duplicating the lookup. Keep accepting literal paths for backward compatibility (treat a value that is not a known id as a literal path).

## Scope

`addDataCombined()` and `.checkDataCombinedEntry()` in `R/plots.R`. Should mirror the resolution semantics already used for scenario `outputPathIds`.

## Verification

- Checked branches `excel-deprecation-and-reorg` (#1038) and the end-state `json-as-primary-input-v2`: neither resolves output-path ids in `addDataCombined()`; `path` is stored verbatim.
- No existing open issue covers this (closest are #1025 on `outputPathIds` round-trip lossiness and #940 / #1001 on `DataCombined` structure, none of which address id resolution in `addDataCombined()`).

Part of #1030.

Felixmil:
Settles the shape/validation scope of the DataCombined umbrella (#1146) against the JSON-primary project model. The entry shape (one entry per DataCombined, with nested `simulated[]` / `observed[]`) was already in place (#1001); this PR rewrites the validation onto that JSON spec and lets a simulated entry reference an output path by id. The four remaining DataCombined feature requests (#907 mapping/grouping rework, #760 color/symbol, #871 autofill-from-pkml, #940 column rename) are out of scope and stay open.

## dataCombined validation on the JSON spec (#1026)

`dataCombined` is now its own validated section. A `dataCombined` adapter (`.dataCombinedValidatorAdapter` / `.validateDataCombined`) checks the JSON spec directly, registered in `.validationAdapters` alongside the other sections. `createDataCombined()` builds DataCombined objects straight from the nested `simulated[]` / `observed[]` entry lists, so the transitional flat-DataFrame reshape is gone: `.specsToDataCombinedDataFrame()`, `.specEntryToRow()`, and `.validateDataCombinedFromExcel()` are removed, and the six catalog messages only that validator used are dropped.

Entry-shape validation previously lived in the `plots` adapter, which validated three sections at once. That block moved to the new `dataCombined` adapter, so a missing required field is reported once rather than twice; the `plots` adapter now validates only the plot list and grids and uses `dataCombined` only as the set of known ids a plot may reference. The required-field table is shared by the load-time validator and the write-time gate via `.requiredDataCombinedFields()`, so the two can never disagree.

## Output-path id in addDataCombined() simulated entries (#1042)

A simulated entry's `path` may now be an output-path id (a key of the project's `outputPaths`) as well as a literal model path, mirroring how `addScenario()` and `addPIOutputMapping()` accept ids. The id is resolved to its literal path when the DataCombined is built by `createDataCombined()`; the id is stored verbatim in the JSON, so it round-trips and a single output path declared once is referenced by id everywhere. A value matching no known id is treated as a literal path, as before.

While rewriting the build path, a missing-dataSet observed entry dropped under `stopIfNotFound = FALSE` now also skips the offset/scale transform step (it previously reached the transform against an absent, all-NA row); a regression test covers it.

## Example

```r
devtools::load_all(quiet = TRUE)
# `aciclovir_pvb` is an output-path id declared once in the project.
project$definitions$outputPaths$aciclovir_pvb
#> [1] "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)"
# Reference it by id in a DataCombined simulated entry, instead of retyping the literal path.
addDataCombined(project, "dc", simulated = list(list(label = "sim", scenario = "aciclovir_iv", path = "aciclovir_pvb")))
# The id is stored verbatim in the spec ...
project$definitions$dataCombined$dc$simulated[[1]]$path
#> [1] "aciclovir_pvb"
# ... and resolves to the literal model path when the DataCombined is built.
res <- createDataCombined(project, dataCombined = "dc", scenarioResults = runScenarios(project, scenarios = "aciclovir_iv"))
unique(res$dc$toDataFrame()$dataType)
#> [1] "simulated"
```

Closes #1026
Closes #1042

## Summary by CodeRabbit

* **New Features**
  * `addDataCombined()` and `createDataCombined()` now accept output-path IDs or literal model paths.
  * Output-path IDs are resolved consistently when building combined data.
  * Added dedicated validation for combined-data definitions, including clearer handling of missing or invalid references.
  * Missing observed datasets can be skipped when configured, without applying their transformations.

* **Documentation**
  * Updated guidance describing supported paths and validation behavior.


R/data-combined.R
```diff
@@ -164,275 +130,261 @@ createDataCombinedFromExcel <- function(...) {
...
-            label = simulated[j, ]$label
-          )
-        }
+  for (entry in spec$simulated %||% list()) {
+    scenarioName <- entry$scenario
+    scenarioResult <- scenarioResults[[scenarioName]]
+    results <- scenarioResult$results
+    # Resolve an output-path id to its literal path; a value that is not a known
+    # id is used verbatim as a literal model path.
+    path <- outputPaths[[entry$path]] %||% entry$path
```
Felixmil, line 178: **Correctness:** `.buildDataCombined()` has no guard against a missing/NULL `entry$scenario` (line 159-160) or `entry$path` (this line) — the removed `.validateDataCombinedFromExcel()` used to hard-abort on both with a clean message before any entry was processed.

- `scenarioResults[[scenarioName]]` with `scenarioName == NULL` throws a raw "subscript out of bounds" (no `call =`, so it doesn't read as `Error in \`createDataCombined()\``).
- `outputPaths[[entry$path]]` with `entry$path == NULL` errors the same way.

This is reachable whenever a `dataCombined` entry loaded from JSON is missing a required field and the caller's pre-flight validation didn't request the `dataCombined` section (see the `createPlots()` gap noted in the review summary) — or via `createDataCombined()` called directly with `validate`-independent semantics. Worth restoring an explicit guard here (or making sure the section is always validated first) so this fails with a clear message instead of a cryptic low-level error.

---

R/data-combined.R
```diff
@@ -164,275 +130,261 @@ createDataCombinedFromExcel <- function(...) {
...
+          scenarioName = scenarioName,
+          path = path
+        )
+      } else {
+        messages$wrongOutputPath(
+          dataCombinedName = name,
+          scenarioName = scenarioName,
+          path = path
+        )
       }
```
Felixmil, line 209: **Correctness (minor):** when `scenarioName` doesn't exist in `scenarioResults` at all (e.g. a typo, or `runScenarios()` was called on a subset), `scenarioResult` is `NULL` here, so this falls into the `messages$wrongOutputPath()` branch rather than a message naming the actual problem — the emitted text blames the output `path` ("has not been simulated") when the real issue is an unresolvable `scenario` reference. Might be worth a distinct message for "scenario not present in scenarioResults" so the abort/warning points at the right field.

---

R/data-combined.R
```diff
@@ -164,275 +130,261 @@ createDataCombinedFromExcel <- function(...) {
...
-    observed <- dplyr::filter(
-      dfDataCombined,
-      dataCombinedName == name,
-      dataType == "observed"
-    )
-    if (nrow(observed) > 0) {
-      dataSets <- observedData[observed$dataSet]
+  observedEntries <- spec$observed %||% list()
+  if (length(observedEntries) > 0) {
+    dataSetIds <- vapply(observedEntries, function(e) e$dataSet, character(1))
```
Felixmil, line 220: **Correctness:** `vapply(observedEntries, function(e) e$dataSet, character(1))` (and the analogous `e$label` extraction below, at what's now line 223) will error with a generic "values must be length 1" if any observed entry is missing the required field — the field is NULL, so the closure returns `character(0)`, breaking `vapply`'s length-1 contract. Same class of issue as the simulated-entry guards above: this used to be caught by `.validateDataCombinedFromExcel()` with a clean "no data set is defined" / "missing label" message before any processing happened.

---

R/data-combined.R
```diff
@@ -164,275 +130,261 @@ createDataCombinedFromExcel <- function(...) {
...
+# @noRd
+.applyDataCombinedTransformations <- function(
+  dataCombined,
+  name,
+  entries,
+  skippedLabels
+) {
+  df <- dataCombined$toDataFrame()
+  for (entry in entries) {
+    if (entry$label %in% skippedLabels) {
```
Felixmil, line 279: **Correctness, two issues on this line:**

1. `entry$label %in% skippedLabels` throws "argument is of length zero" when `entry$label` is `NULL` (missing required field) — `NULL %in% x` is `logical(0)`, and `if (logical(0))` errors. Same missing-field-guard gap as above.
2. Matching is by `label` alone, not `(label, dataType)`. `skippedLabels` accumulates from both the simulated-skip branch (line 193) and the observed-drop branch (line ~213-216), and `entries` at the call site is `c(spec$simulated, spec$observed)` — so if a skipped *simulated* entry happens to share a `label` with an unrelated, successfully-added *observed* entry (or vice versa), the observed entry's transform is silently skipped too, even though its data was added to the `DataCombined`. Worth scoping the skip-check to the same `(label, dataType)` pair the removed Excel-era code used (it filtered on `dataType == "simulated"` for this drop).

---

R/plots.R
```diff
@@ -1087,7 +1043,10 @@ removePlotGrid <- function(project, id) {
 #' @param simulated For a single DataCombined (`id` length 1), a list of
 #'   named lists, each including `label`, `scenario`, and `path` (optional
 #'   `group`, `xOffsets`, `xOffsetsUnits`, `yOffsets`, `yOffsetsUnits`,
-#'   `xScaleFactors`, `yScaleFactors`). The `scenario` reference is
+#'   `xScaleFactors`, `yScaleFactors`). `path` may be either a literal model
+#'   quantity path or an output-path id (a key of the project's `outputPaths`
+#'   definitions); an id is resolved to its literal path when the DataCombined
+#'   is built by [createDataCombined()]. The `scenario` reference is
```
Felixmil, line 1049: **Consistency gap** (not introduced by this diff, but this doc update advertises the new id-or-literal-path behavior, so worth flagging here): `.addDataCombined_impl()` only canonicalizes the `scenario` reference at write time (a few lines down); it never resolves or validates a simulated entry's `path` against `outputPaths` ids, unlike `.addPIOutputMapping_impl()` (`R/parameter-identification.R`) which calls `.resolveOutputPathRef(outputPath, self)` and aborts immediately with a "did you mean" hint on an unresolvable value.

Concretely, `addDataCombined(project, "dc1", simulated = list(list(label = "l", scenario = "s", path = "acclovir_pvb")))` (typo) succeeds silently and stores the bad string verbatim; the mistake only surfaces much later at `createDataCombined()` time — as a generic "output path has not been simulated" warning/abort, not a "did you mean" hint — instead of failing fast at authoring time like the equivalent PI call does. `.validateCrossReferences()` also has no check for this reference (it checks the scenario `outputPaths` and PI `outputMapping$outputPathId` against `outputPathKeys`, but not a dataCombined simulated entry's `path`), so `validateProject()` can't catch this class of typo either.

Might be worth reusing `.matchOutputPathRef()`/`.resolveOutputPathRef()` here for parity with the PI path.

---

R/data-combined.R
```diff
@@ -164,275 +130,261 @@ createDataCombinedFromExcel <- function(...) {
...
+#' declare `label` and `dataSet`. An empty section is valid. Cross-section
+#' reference resolution (a simulated entry's `scenario` against defined
+#' scenarios) is handled by the `crossReferences` phase, not here; runtime
+#' resolution against scenario results and observed data happens in
+#' [createDataCombined()].
 #'
-#' @returns Processed `dfDataCombined`
+#' @param dataCombined Named list of DataCombined specs (from `dataCombined`
+#'   definitions), each with nested `simulated` / `observed` entry lists.
+#' @return validationResult.
```
Felixmil: **Convention:** `@return validationResult.` should be `@returns` (with the s) — the house tag per `REFACTORING.md`'s conventions section.

---

REFACTORING-LOG.md
```diff
@@ -62,3 +62,4 @@ One bullet per commit, newest at the bottom: `- YYYY-MM-DD HH:MM: plain-language
...
 - 2026-07-24 00:34: PR #1161 CodeRabbit: tighten .legacySheetToDf() sheet validation, a present-but-malformed `rows` (scalar or non-list record) now raises the clear malformed-sheet message instead of a cryptic indexing error; an absent `rows` stays a valid empty sheet. (#1156)
 - 2026-07-24 07:30: PR #1160 review: guard `NULL` before subsetting in `.pi5xTaskRows()`, so importing a legacy 5.x workbook missing optional PI sheets returns an empty frame instead of aborting. (#1158)
+- 2026-07-24 14:19: dataCombined validation now runs on the JSON spec directly via a `dataCombined` validation adapter (removed the legacy DataFrame reshape and the `FromExcel` validator), and `addDataCombined()` simulated entries accept an output-path id for `path` (resolved to the literal path when the DataCombined is built). Moved dataCombined shape validation out of the plots adapter so an error is no longer reported twice, and dropping a missing-dataSet observed entry now also skips its transform. (#1026, #1042)
```
Felixmil: **Convention:** this bullet cites two issue numbers, `(#1026, #1042)`. Every other bullet in this file ends with exactly one `(#N)`; `REFACTORING.md`'s logging rule is written for the single-issue case. Since this commit closes two issues, consider two bullets (one per issue) to keep the established format, or confirm the multi-issue form is intentionally fine here.

---

tests/testthat/test-validation.R
```diff
@@ -939,7 +902,7 @@ test_that("validateProject() flags a scenario referencing a removed outputPath",
 
 # Cross-reference: dataCombined -> observed dataSet ----
 
-test_that(".validatePlots flags an empty observed dataSet reference", {
+test_that(".validateDataCombined flags an empty observed dataSet reference", {
```
Felixmil: **Convention:** this test now exercises `.dataCombinedValidatorAdapter()` directly (a few lines down), which is owned by `R/data-combined.R` per this same diff's "plural domain file owns its section, 1:1 with its test" convention. The rest of the dataCombined-adapter tests moved to `test-data-combined.R` — consider moving this one there too rather than leaving it in `test-validation.R`.

---

Felixmil (commented):
Automated code review (high effort, 8 independent finder passes + cross-verification against source).

**Most significant finding — not inline because `R/create-plots.R` isn't part of this diff, but directly caused by it:** `createPlots(validate = TRUE)`'s pre-flight check (`R/create-plots.R:71`) calls `project$ensureValid(sections = c("plots", "scenarios", "observedData", "crossReferences"))`. This PR moves `dataCombined` entry-shape validation (required `label`/`scenario`/`path`/`dataSet`) out of the `plots` adapter into its own new `dataCombined` adapter (`.dataCombinedValidatorAdapter`/`.validateDataCombined()`), but that call site was never updated to request the new `"dataCombined"` section. `.validateCrossReferences()` (run via `"crossReferences"`) re-derives some dataCombined data for its own reference checks, but never surfaces a shape problem as a critical error. Net effect: a `dataCombined` entry missing a required field (very reachable — JSON is the primary authoring surface per the project's migration model, and a hand-edited `definitions/data-combined/*.json` file bypasses the `addDataCombined()` write-gate entirely) now sails through `createPlots()`'s default pre-flight check and crashes later with a raw R error instead of the clean abort `validate = TRUE` is supposed to guarantee. Combined with the missing-field guards noted inline below, this is the main correctness gap in this PR.

Inline comments cover the specific crash sites (`R/data-combined.R:164,188,199,258`), a related consistency gap in `addDataCombined()`'s write-time gate (`R/plots.R:1049`), and a few convention nits (`@returns` house tag, multi-issue log citation, a misplaced test).

Lower-priority simplification/efficiency notes not included as inline comments (happy to add if useful): the x/y offset blocks in `.applyDataCombinedTransformations()` are near-duplicates; `skippedLabels` is grown with repeated `c()` in a loop; `.checkDataCombinedEntryFields()` (`R/data-combined.R`) duplicates the same required-field loop as `.checkDataCombinedEntry()` (`R/plots.R`), differing only in how the failure is reported.

---

R/data-combined.R
Felixmil: Fixed in 91f4b43d. Root cause was the removal of the up-front `.validateDataCombinedFromExcel()` abort. `createDataCombined()` now takes `validate = TRUE` and pre-flights the `dataCombined` section via `project$ensureValid(sections = "dataCombined")`, so a definition missing a required field (`label`/`scenario`/`path`/`dataSet`) aborts cleanly before any entry is built, rather than crashing on a NULL field. Covered by a new test ("aborts cleanly on a dataCombined missing a required field").

R/data-combined.R
Felixmil: Fixed in 91f4b43d together with the sibling missing-field crash: the `validate = TRUE` pre-flight in `createDataCombined()` rejects an observed entry with a missing `dataSet`/`label` before the `vapply` runs, so the length-1 contract is never violated at build time.

R/data-combined.R
Felixmil: Fixed in 81c64fe9. The unresolved-simulated-data branch now splits into three distinct cases: scenario absent from `scenarioResults` -> new `messages$scenarioNotInResults()` (names the scenario reference), scenario present but `results = NULL` -> `scenarioRunFailed()`, and path not in `allQuantityPaths` -> `wrongOutputPath()`. New test: "reports a scenario absent from results distinctly".

R/data-combined.R
Felixmil: Point 1 (the `NULL %in%` crash on a missing `label`) is fixed by the `validate = TRUE` pre-flight in 91f4b43d: an entry missing `label` aborts before the transform step runs. Point 2 (matching by `label` alone rather than `(label, dataType)`): kept as-is by design. A `DataCombined` keys every curve by its `name` (= `label`) and both `addSimulationResults()` and `addDataSets()` register under that name, so labels are already unique within one DataCombined; the transform also matches `df$name == entry$label`. Adding `dataType` to the skip match would be defensively redundant given the whole build keys on label. Happy to add it if you prefer the extra precision.

R/data-combined.R
Felixmil: Fixed in 91f4b43d: `@return` -> `@returns`.

REFACTORING-LOG.md
Felixmil: Fixed in 91f4b43d: split into one bullet per issue, `(#1026)` and `(#1042)`.

tests/testthat/test-validation.R
Felixmil: Fixed in 91f4b43d: moved the empty-observed-dataSet test to `test-data-combined.R` (it exercises `.dataCombinedValidatorAdapter()`, owned by `R/data-combined.R`).

R/plots.R
Felixmil: Acknowledged, and deliberately left as-is. Unlike PI's `outputPath` (always an id reference), a DataCombined simulated `path` is dual-purpose: an output-path id **or** a literal model quantity path. A literal path cannot be verified at `addDataCombined()` time (no model/simulation is loaded then, and there is no list of valid quantity paths in the project definitions), and the shipped Example project itself uses a literal path. A `.resolveOutputPathRef()`-style fail-fast would therefore reject every legitimate literal path.

Resolution stays lenient (id -> literal, unknown -> treated as literal) and the unresolvable-path abort stays at build time in `createDataCombined()` (`stopIfNotFound = TRUE` by default), now with distinct messages per failure cause. Making `path` an id-only reference would break the example and existing projects and contradict #1042 ("treat a value that is not a known id as a literal path"), so it is out of scope here. Leaving this thread open for visibility if we later add an explicit id-marker to the shape.

PR #1178, opened 2026-07-24
Run animal (non-human) individuals: optional gender, age/height-less run, build-time stopIfFails skip

### LINKED ISSUE #1162: Animal (non-human) individuals cannot be run: gender required + age/height crash
## Summary

A migration test of the 6.0.0 JSON model against seven legacy Excel projects found that no animal (non-human) project can be run on either the importer path or the hand-authoring path. Two independent defects combine: `gender` is treated as a mandatory non-empty field even for a species whose only valid PK-Sim gender is `UNKNOWN`, and an individual that legitimately carries no `age`/`height` crashes `runScenarios()` with an uninformative error. Both were reproduced identically on two separate cross-species projects. Tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49` (branch `json-based-project`, esqlabsR 6.0.0.9001), ospsuite 12.4.4 / PK-Sim 12.3.173.

These are new findings, not covered by the migration issues #1118 or #1157.

## Defect 1: `gender` is unconditionally required but cannot be set for an `UNKNOWN`-only species

An animal population whose only valid PK-Sim gender is `UNKNOWN` cannot be represented. The importer writes such individuals with no `gender` field; `validateProject()` then flags each as a critical error, and `addIndividual()` refuses to create one at all. Inventing a gender to get past authoring is rejected downstream by PK-Sim at build time.

Messages observed:

- Validator: `[Missing Fields] Required field 'gender' is missing or empty in individual '<id>'`
- Authoring: `Cannot add individual "<id>": gender must be a non-empty string`
- PK-Sim at build time (after inventing a gender): `Could not find gender 'MALE' for population '<animal population>'. Available genders are: UNKNOWN`

Root cause: `R/individuals.R:73` lists `gender` in `requiredFields` unconditionally, and `R/individuals.R:259-268` rejects a gender-less individual at add time. `addIndividual(gender = "UNKNOWN")` is already accepted (`UNKNOWN` is a valid entry in `GenderInt`), so the gap is twofold: the importer does not default an absent gender to `UNKNOWN`, and the validator treats an absent gender as critical rather than defaulting it.

Suggested fix: treat `gender` as optional, defaulting to `UNKNOWN` when absent (in both the importer output and the validator), rather than requiring a non-empty value.

## Defect 2: an `age`-less / `height`-less individual crashes `runScenarios()`

The one fully-runnable scenario of each animal project fails identically on both the imported and the hand-authored tree:

```
Error in if (is.null(value) || is.na(value)) { : missing value where TRUE/FALSE needed
```

Call path: `runScenarios()` -> `.prepareScenario()` -> `ospsuite::createIndividualCharacteristics()` -> `.createSnapshotParameter(age, ageUnit)`.

Root cause: `R/scenario-execution.R:390-392` passes `weight`/`height`/`age` through `as.double(indivData$...)`. For an animal individual that carries only weight, `age` and `height` are `NULL`, so `as.double(NULL)` is `numeric(0)`. ospsuite's `.createSnapshotParameter` does `if (is.null(value) || is.na(value))`, and `is.na(numeric(0))` is `logical(0)`, so the `if` aborts. Reproduced in isolation: `if (is.null(as.double(NULL)) || is.na(as.double(NULL)))` raises the same message.

Suggested fix: coerce a missing biometric to `NA_real_` (length 1) before handing it to ospsuite, or omit the argument entirely when the field is absent, so an individual that legitimately has no age/height yields a clean run rather than a length-zero crash.

## Impact

Together these block every non-human individual from running. Defect 1 blocks validation and authoring; even once past it (by supplying `gender = "UNKNOWN"`), defect 2 blocks the actual simulation. A migrated animal project is therefore not runnable end to end without hand edits to the individuals and a code fix.

## Reproduction

Both defects reproduce on esqlabsR's own bundled fixture pattern by adding an individual with `species = "Dog"`, `population = "Beagle"`, only a `weight` set, and no `gender`/`age`/`height` (the standard shape for a PK-Sim animal individual). No source-project data is needed. Root-cause lines above are all in `R/` at the tested commit.

Related: #666 (species scaling robustness) touches the same species/individual area but is about applying species, not the gender/age/height run blocker.

### LINKED ISSUE #1170: runScenarios(stopIfFails=FALSE) does not skip a build-time scenario failure
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). `runScenarios(..., stopIfFails = FALSE)` does not skip a scenario that fails at build time (e.g. a missing model parameter path); the whole batch aborts. A user expecting "run the ones that work, skip the broken one" instead gets a full-batch abort.

Root cause: `stopIfFails` only guards result collection (`R/scenario-execution.R:589-600`); the scenario build runs unguarded earlier at `R/scenario-execution.R:669`, so a build-time failure escapes the `stopIfFails` handling entirely.

Suggested fix: wrap the build step in the same `stopIfFails` handling as result collection, so a build-time failure of one scenario is surfaced-and-skipped (not fatal) when `stopIfFails = FALSE`.

Felixmil:
Migration testing of the 6.0.0 JSON model against legacy Excel projects surfaced three defects that together block every animal (non-human) project from running. This fixes all three: an individual whose only valid PK-Sim gender is `UNKNOWN` can now be represented and run, an individual that legitimately carries no age/height no longer crashes the run, and `runScenarios(stopIfFails = FALSE)` now also skips a scenario that fails at build time instead of aborting the whole batch.

## Gender is optional, defaulting to UNKNOWN (#1162)

`gender` was unconditionally required on an individual, so an animal species whose only valid PK-Sim gender is `UNKNOWN` could not be represented: the importer wrote such individuals with no gender, `validateProject()` flagged each as a critical error, and `addIndividual()` refused to create one. `gender` is now optional and defaults to `UNKNOWN` when absent. The Excel importer defaults a blank Gender cell to `UNKNOWN`, `addIndividual()` defaults an absent gender (a supplied gender must still be a valid `Gender` token), and `validateProject()` no longer treats a gender-less individual as an error.

## An age-less / height-less individual no longer crashes the run (#1162)

An animal individual legitimately carries only a weight. `runScenarios()` passed `age`/`height`/`weight` through `as.double()`, so an absent field became a length-zero `numeric(0)`, which defeated ospsuite's own `is.null()` guard in `createIndividualCharacteristics()` and aborted with `missing value where TRUE/FALSE needed`. The biometrics are now passed through as-is; an absent one stays `NULL` (which ospsuite defaults), so a weight-only individual runs cleanly.

## stopIfFails also skips a build-time failure (#1170)

`stopIfFails` only guarded result collection; a scenario that failed at build time (e.g. a missing model parameter path) escaped it and aborted the whole batch. The build step is now wrapped in the same handling: with `stopIfFails = FALSE` a build failure warns, that scenario is skipped, and the rest still build, run, and return. `buildSimulations()` keeps aborting on a build failure. A skipped scenario keeps an explicit `NULL` slot (single-bracket assignment) so the positional run loops are unaffected when a non-last scenario fails, and result collection no longer touches a skipped scenario's `NULL` simulation.

## Example

Behaviour change for a gender-less animal individual (previously `addIndividual()` aborted with "gender must be a non-empty string" and `validateProject()` reported a critical error):

``` r
devtools::load_all("<repo>", quiet = TRUE)
projectPath <- system.file("extdata/projects/Example/Project.json", package = "esqlabsR")
project <- loadProject(projectPath)
# An animal individual whose only valid PK-Sim gender is UNKNOWN carries only a
# weight: no gender, no age, no height. addIndividual() now accepts it, defaulting
# gender to UNKNOWN (previously it aborted: "gender must be a non-empty string").
addIndividual(project, "beagle", species = "Dog", population = "Beagle", weight = 10)
project$definitions$individuals[["beagle"]]
#> <Individual>
#>   • Species: Dog
#>   • Population: Beagle
#>   • Gender: UNKNOWN
#>   • Weight: 10
#>   • Height: <empty string>
#>   • Age: <empty string>
#>   • Parameter Sets: <empty string>
# validateProject() no longer reports the gender-less individual as a critical error.
length(validateProject(project)$individuals$critical_errors)
#> [1] 0
```

## Drive-by: the `paramSheets` deprecation never warned

Not part of either issue, but the branch could not go green without it: `tests/testthat/test-scenario-from-pkml.R` ("paramSheets argument is soft-deprecated") was failing on `json-based-project` before this PR branched.

`lifecycle::deprecate_soft()` gates every warning path on `is_direct()`, computed from `user_env = caller_env(2)`. The deprecation was raised inside `.createScenariosFromPKML_impl()`, three layers below the exported function (free function -> `Project` method -> `private$.impl()` -> `_impl`), so `caller_env(2)` resolved to an internal esqlabsR frame, the call was judged indirect, and nothing was ever signalled. Setting `lifecycle_verbosity = "warning"` only sets `always = TRUE`; it does not bypass that gate, which is why the test's `withr::local_options()` could not opt back in.

`createScenariosFromPKML()` now warns and translates `paramSheets` before forwarding. The `_impl` still accepts and translates it, so the matching `Project` method behaves exactly as before. The warning now fires on every call rather than depending on once-per-session dedup.

Closes #1162
Closes #1170

## Summary by CodeRabbit

* **New Features**
  * Individuals can now omit gender; it defaults to `UNKNOWN`.
  * Excel imports apply the same default for empty gender values.
  * Scenario runs can skip scenarios that fail during setup when `stopIfFails = FALSE`, allowing other scenarios to continue.

* **Bug Fixes**
  * Fixed crashes when individual age or height is missing.
  * Skipped scenarios now correctly report no results and retain `NULL` output values.

* **Documentation**
  * Clarified scenario failure handling and default gender behavior.


R/scenario-execution.R
```diff
@@ -383,13 +383,19 @@
...
+        # weight/height/age are stored as doubles (or absent). Pass them through
+        # as-is: an absent biometric stays NULL, which
+        # `createIndividualCharacteristics()` defaults. Coercing with
+        # `as.double()` would turn a NULL into `numeric(0)`, which defeats
+        # ospsuite's own `is.null()` guard and crashes an animal individual that
+        # legitimately carries only a weight.
         individualCharacteristics <- ospsuite::createIndividualCharacteristics(
           species = indivData$species,
           population = indivData$population,
           gender = indivData$gender,
```
Felixmil, line 395: **The `gender = "UNKNOWN"` default only exists on the authoring/Excel paths, not the JSON-load path.**

`addIndividual()` (`.buildIndividualEntry`) and the Excel importer both explicitly default an absent gender to `"UNKNOWN"` at write time. `.parseIndividuals()` (R/individuals.R:12-41) does not: it loops over the raw entry's fields and simply skips any field whose value is `NULL`/absent, so an individual hand-authored in `definitions/individuals/*.json` without a `gender` key keeps `gender` unset forever. That reaches this line as `indivData$gender == NULL`, which is forwarded directly into `ospsuite::createIndividualCharacteristics(gender = NULL, ...)` — unlike `weight`/`height`/`age`, which the comment above documents ospsuite as NULL-safe for, gender has no such documented/tested NULL handling.

Since JSON is the primary source of truth for this project model (not `addIndividual()`), this is exactly the "animal individual with no gender" scenario PR #1178 is meant to unblock, but it's only fixed for the authoring API — the JSON-first path (`loadProject()` → `runScenarios()`) still hits the old crash. `.validateIndividuals` now accepts a gender-less individual (test: "does not flag an absent gender"), so this passes validation and only fails deep inside ospsuite at run time.

Suggest defaulting `gender` to `"UNKNOWN"` in `.parseIndividuals()` itself (mirroring `.buildIndividualEntry`), and adding a test that loads a gender-less individual straight from JSON through `runScenarios()`.

---

R/scenario-execution.R
```diff
@@ -560,15 +567,36 @@
...
+    result <- tryCatch(
+      .prepareScenario(
+        scenario = allScenarios[[name]],
+        project = project,
+        customParams = customParams,
+        cache = cache,
+        simulationRunOptions = simulationRunOptions,
+        stopIfParameterNotFound = stopIfParameterNotFound
+      ),
+      error = function(e) {
```
Felixmil, line 585: **This `tryCatch` catches every error type, not just build-time data problems.**

`error = function(e)` matches on the base `error` condition class — there's no more specific condition class distinguishing "legitimate build failure" (missing model parameter path, bad individual data) from an unrelated bug elsewhere in `.prepareScenario()`'s call chain (file resolution, `ospsuite::loadSimulation`, population resolution, etc.). With `stopIfFails = FALSE` — which the PR/NEWS position for production batch runs — any such bug for one scenario is now silently downgraded to a per-scenario warning and the batch continues, rather than surfacing as the real failure it is.

Worth scoping this catch to a dedicated condition class raised at the specific "this scenario's data is bad" sites, so a genuine defect still aborts loudly even under `stopIfFails = FALSE`.

---

R/individuals.R
```diff
@@ -70,7 +71,7 @@
     return(result)
   }
 
-  requiredFields <- c("species", "gender")
+  requiredFields <- c("species")
```
Felixmil, line 81: **`.validateIndividuals()` no longer checks that a *present* `gender` is a valid token.**

Dropping `gender` from `requiredFields` correctly stops flagging an *absent* gender, but there's no replacement check for a gender that's present-but-invalid (e.g. `"gender": ""` or `"gender": "foo"` in a hand-authored JSON file). `addIndividual()`/`.buildIndividualEntry` validate against the `GenderInt` enum on the authoring path, but `.validateIndividuals` — the check that runs against hand-edited/imported JSON via `.individualsValidatorAdapter` — has no equivalent, so such a value passes `validateProject()` cleanly and only surfaces as an opaque ospsuite error at `runScenarios()` time.

Suggest adding an enum-membership check here for a present `gender`, mirroring what `.buildIndividualEntry` already does.

---

R/scenario-execution.R
```diff
@@ -589,7 +617,10 @@
   stopIfFails = TRUE
 ) {
   outputQuantities <- NULL
-  if (!is.null(scenario$outputPaths)) {
+  # `simulation` is NULL for a scenario skipped at build time (stopIfFails =
+  # FALSE); there is nothing to resolve output quantities against, so leave
+  # them NULL and let the no-results branch below record the skip.
+  if (!is.null(simulation) && !is.null(scenario$outputPaths)) {
```
Felixmil, line 624: **A build-time skip now produces two warnings for the same event.**

When a scenario fails to *build* under `stopIfFails = FALSE`, `.buildScenarioSimulations()` already warns via `messages$scenarioBuildFailed()` (line 589) and leaves that scenario's `prepared` entry `NULL`. That flows through to this function with `simulation = NULL, results = NULL`; a few lines below (the `if (is.null(results))` branch), `stopIfFails = FALSE` triggers a *second* warning — `messages$missingResultsForScenario()` — for a scenario that never built, not one that ran and produced no results.

A caller running `runScenarios(project, stopIfFails = FALSE)` over a batch with one bad scenario sees two differently-worded warnings for what looks like two different root causes. Consider skipping the "no results" warning here when `simulation` is also `NULL` (i.e. this scenario already warned at build time).

---

Felixmil (commented):
Automated high-effort review (multi-angle pass). One finding stood out as corroborated independently across every review angle and isn't caught by the existing CodeRabbit comments: the `gender = "UNKNOWN"` default added for `addIndividual()`/the Excel importer never made it into `.parseIndividuals()`, so a hand-authored `definitions/individuals/*.json` file (the primary, JSON-first authoring path this project's migration is built around) omitting `gender` still reaches `ospsuite::createIndividualCharacteristics(gender = NULL, ...)` untested — the exact crash this PR is meant to fix, just reachable from a different entrypoint. The other three comments (double warning on a build-time skip, an overly broad `tryCatch`, and a validator gap for a present-but-invalid `gender`) are lower-severity but were each flagged by 2+ independent review passes.

---

R/scenario-execution.R
Felixmil: Fixed in 89c05724. `.parseIndividuals()` now defaults an absent `gender` to `"UNKNOWN"`, mirroring `.buildIndividualEntry()` and the Excel importer, so every stored individual carries a concrete gender regardless of which entrypoint created it. Added a test covering the JSON-load path directly.

One correction to the diagnosis: I checked, and `ospsuite::createIndividualCharacteristics(gender = NULL, ...)` does **not** crash — it returns an object with an empty gender. So this was a consistency gap (the JSON-first path, the primary one, behaved differently from the other two) rather than the run-time crash described. The fix stands either way.

R/individuals.R
Felixmil: Fixed in 89c05724. `.validateIndividuals()` now reports a present-but-invalid `gender` as a critical error, mirroring the `GenderInt` check in `.buildIndividualEntry()`. An absent gender stays valid (it is defaulted to `UNKNOWN` on every load path). Regression test covers both `"banana"` and `""`.

R/scenario-execution.R
Felixmil: Fixed in 89c05724. `.collectScenarioResult()` skips the "no results" warning when `simulation` is `NULL`, which is exactly the build-time-skip case that already warned via `scenarioBuildFailed()`. A scenario that built but produced no results still warns as before. Added tests for both the single-warning skip and the still-warns-on-empty-results case.

R/project-excel.R
```diff
@@ -1485,7 +1485,9 @@ projectStatus <- function(project, silent = FALSE) {
       individualId = as.character(row$IndividualId),
       species = as.character(row$Species),
       population = as.character(row$Population),
-      gender = as.character(row$Gender),
+      # An empty Gender cell defaults to UNKNOWN (the only valid PK-Sim gender
+      # for some animal species), rather than an NA the validator would flag.
+      gender = .naToDefault(as.character(row$Gender), "UNKNOWN"),
```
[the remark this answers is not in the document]
Felixmil: Fixed in 89c05724. Added a `.blankToDefault()` helper (`.naToDefault()` only handled `NA`) and used it for the Gender cell, so an empty-string or whitespace-only cell imports as `UNKNOWN` rather than an invalid empty gender. Test covers both `""` and `"   "`.

man/runScenarios.Rd
```diff
@@ -43,10 +43,11 @@ already validated the project).}
...
 
-\item{stopIfFails}{Logical. If \code{TRUE} (default), a scenario whose
-simulation produced no results aborts the run with an error. Set to
-\code{FALSE} to instead warn and leave that scenario's \code{outputValues}
-\code{NULL} while the other scenarios are still returned.}
+\item{stopIfFails}{Logical. If \code{TRUE} (default), a scenario that fails to
+build (e.g. a missing model parameter path) or whose simulation produced
+no results aborts the run with an error. Set to \code{FALSE} to instead warn
+and leave that scenario's \code{outputValues} \code{NULL} while the other scenarios
+are still built, run, and returned.}
```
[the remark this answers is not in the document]
Felixmil: Fixed in 89c05724. The roxygen `@returns` in `R/scenarios.R` now states that with `stopIfFails = FALSE` a scenario skipped at build time is still returned with `simulation`, `results`, and `outputValues` all `NULL`; the existing wording for simulation failures is retained. `man/runScenarios.Rd` regenerated.

tests/testthat/test-scenario-execution.R
```diff
@@ -153,33 +165,106 @@ test_that("runScenarios aborts by default when a scenario simulation produces no
...
+  )
+  local_mocked_bindings(
+    runSimulations = .mockNoResults
+  )
+  out <- suppressWarnings(
+    runScenarios(project, stopIfFails = FALSE)
+  )
+  expect_true(all(c("testscenario", "second") %in% names(out)))
+  expect_null(out$second$outputValues)
+})
```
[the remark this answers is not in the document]
Felixmil: Fixed in 89c05724. Both tests now assert `expect_s3_class(out$testscenario$simulation, "Simulation")`, so an unexpected `.prepareScenario()` failure can no longer pass as the same `NULL`-results shape. The build-skip test additionally asserts `out$second$simulation` is `NULL`, pinning both sides of the distinction.

R/scenario-execution.R
Felixmil: Not changing this here, deliberately.

The concern is fair in principle, but the failure is not silent: the caught error's own message is threaded verbatim into the warning via `messages$scenarioBuildFailed(conditionMessage(e))`, so an unrelated bug in `.prepareScenario()`'s call chain still surfaces its real message, named against the scenario it happened on. And `stopIfFails = FALSE` is an explicit opt-in to best-effort batch behaviour: the caller has asked for "run what you can, tell me what you couldn't".

Introducing a dedicated condition class would mean raising it at every "this scenario's data is bad" site across file resolution, `loadSimulation`, parameter application, and population resolution, and deciding for each existing abort whether it qualifies. That is a design change well beyond the two bug fixes this PR carries (#1162, #1170), and it would touch code paths neither issue is about.

Worth its own issue if we want the stricter contract.

R/individuals.R
```diff
@@ -252,14 +256,14 @@ addIndividual <- function(project, id, species, ...) {
...
+  # valid PK-Sim gender for some animal species). A supplied gender must still
+  # be a valid `GenderInt` token.
   gender <- fields$gender
-  if (
-    is.null(gender) ||
-      !is.character(gender) ||
+  if (is.null(gender)) {
+    gender <- "UNKNOWN"
+  } else if (
+    !is.character(gender) ||
```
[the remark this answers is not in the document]
Felixmil: Not applying this one: as of 89c05724 it is unreachable.

All three entrypoints that can put an individual into `project$definitions$individuals` now materialize a concrete gender: `.parseIndividuals()` (JSON load), `.buildIndividualEntry()` (authoring API), and `.parseExcelIndividuals()` (Excel import) each default an absent gender to `"UNKNOWN"`. `.setOneIndividual()` only touches gender when the caller supplies one, and rejects an empty value. So `ic$gender` is never `NULL` by the time `.individualsToExcelDf()` sees it, and `as.character(ic$gender %||% "UNKNOWN")` would be dead defensive code.

The gap the comment identifies was real when it was written (the default existed only on the authoring and Excel paths); it is closed at the source instead, which also fixes the same gap for `runScenarios()` and `validateProject()`.

PR #1179, opened 2026-07-27
Excel import: keep colliding parameter sets, copy referenced assets, report the result

### LINKED ISSUE #1163: Importer: parameter-set id collision silently drops an individual's own parameter set
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). When a `ModelParameters` sheet and an `Individuals` sheet canonicalize to the same parameter-set id, the importer gives the id to the ModelParameters set, renames the individual-specific set to `<id>.1`, and leaves `<id>.1` referenced by nobody. The individual's own parametrization is therefore silently dropped from every scenario that uses that individual. No warning is emitted about the collision.

Unique sheet names link correctly (this is the individual->parameter-set link tracked as fixed under #1118 / regression from #1118's list). The gap is specifically the name-collision case.

Suggested fix: detect the collision at import time and either disambiguate both ids deterministically while preserving both links, or warn clearly that a rename happened and re-point the individual's `parameterSets` reference to the renamed set.

Reproducible with two sheets whose names canonicalize to the same id (e.g. a ModelParameters sheet and an Individuals sheet both resolving to `rat`). Confidential-safe: no source data needed.

### LINKED ISSUE #1164: Importer does not copy model/data assets: output project is not self-contained
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). `importProjectFromExcel()` writes only the JSON definition tree; it does not copy or re-point the model `.pkml` files or the data files referenced by the source Excel project. Every imported project therefore fails file-existence validation with File-Not-Found warnings until the user hand-copies models into `Models/Simulations/` and data into `Data/`. The importer output is not a runnable, self-contained project.

Root cause: `importProjectFromExcel` (`R/project-excel.R`) emits definitions only; the missing files are flagged later by the validators at `R/scenarios.R:373-377` and `R/observed-data.R:165,189`.

Suggested fix: copy (or offer to copy) the referenced `.pkml` and data assets into the new project's `Models/Simulations/` and `Data/` folders as part of the import, or at minimum emit a clear post-import message listing the assets the user must place.

### LINKED ISSUE #1165: Importer canonicalizes a dataCombined's simulated scenario ref but not its observed ref
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). Within a single imported `dataCombined` file, `simulated[].scenario` is canonicalized (lowercased/filename-safe) while `observed[].scenario` keeps the original Excel case. The same file then carries two different casings for the same scenario. This is consistent across all imported dataCombined files.

It did not raise a validation error in testing, but any cross-reference check keyed on the canonical id will fail to match the observed block.

Root cause: the observed-block scenario reference is not routed through `.canonicalizeId()` the way the simulated-block reference is, in the importer's dataCombined path (`R/project-excel.R`).

Suggested fix: canonicalize the `observed[].scenario` reference identically to `simulated[].scenario`.

### LINKED ISSUE #1166: importProjectFromExcel() runs silently: no progress or summary output
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). `importProjectFromExcel()` runs with no progress, summary, or success output. The user cannot tell what was imported, how many definitions of each section were produced, or that the call succeeded at all, without separately loading and inspecting the result.

Root cause: the importer (`R/project-excel.R`) emits no `cli` summary.

Suggested fix: emit a short `cli` summary at the end of import (per-section definition counts, the output path, and any skipped/absent assets), consistent with how the rest of the package reports outcomes.

Felixmil:
Four defects found during migration testing of the 6.0.0 JSON model, all in the Excel -> JSON importer. Two of them lose or mis-resolve data silently; the other two are about the importer's output being usable. Fixed in that order, one commit each, plus a review follow-up commit.

## Parameter-set id collisions (#1163)

Before 6.0.0 the model-parameters, individuals, and applications workbooks were three separate parameter-set namespaces, so a legacy project may legitimately use one sheet name in two of them. They now share one `parameterSets` namespace. The importer never noticed the clash: it handed both sets the same key and `jsonlite::toJSON()` silently uniquified the second to `<id>.1`, which nothing then referenced. The individual's own parametrization disappeared from every scenario using it, with no warning.

- Every workbook now appends through one seam, `.appendParameterSets()`. It uniquifies on the *canonical* id, so `Rat` and `rat` in two workbooks clash exactly as `Rat` twice does; both map to one definition filename.
- The workbook parsed first keeps the plain id; the later sheet is renamed, keeping its readable spelling (`Indiv1` -> `Indiv1_1`).
- `.applyIdRenames()` re-points the references that workbook makes: an individual's own sheet link and its `Individual Parameter Sets` column, an application's `ParameterSets` column, and the 5.x per-sheet application wrapper.
- A warning names each `old -> new` pair and the workbook it came from.

A scenario's `ModelParameterSheets` needs no re-pointing: that column only names ModelParameters sheets, and that workbook is parsed first, so its ids are never renamed.

## dataCombined observed scenario reference (#1165)

An imported data combination carried the canonical scenario id in its simulated block and the original Excel casing in its observed block, so a single definition file held two spellings of one scenario and any check keyed on the canonical id missed the observed entries. The observed entries now run through the same reference transform as the simulated ones.

## Self-contained imports (#1164)

`importProjectFromExcel(outputDir = ...)` wrote only the definition tree. Because a definition names its model or data file by a path relative to the project folder, every such reference dangled and the import ended in File-Not-Found warnings until the assets were hand-copied.

The referenced input folders (simulations, data, populations) are now copied to the same relative location under `outputDir`. Whole folders travel, so an asset nothing names statically (a data-importer configuration, a population csv) comes along too.

Not copied, each for its own reason: the output folder (results the project writes, not inputs it reads); a folder given as an absolute path or through a `${VAR}`, which deliberately sits outside the project and resolves from anywhere; a `../`-climbing value naming a location the project does not own; and nothing at all on an in-place import. A folder that could not be copied is reported.

The containment check came out of review. Without it a `modelFolder` of `"../Shared"` read the parent of the Excel project and wrote to the parent of `outputDir`, overwriting whatever sat beside it. It now uses `.pathEscapesRoot()`, the same predicate every other author-controlled path in this file goes through. The regression test fails without the guard.

## Import summary (#1166)

The summary was gated on `interactive()`, so an import run from a script finished with no sign of what had been written, or that it had succeeded. It is now gated on `silent` alone and reports the project written, the per-section definition counts, and the folders copied or not copied. The counts are rendered by the project's own definitions block, so the summary and `print(project)` cannot disagree about labels.

## Example

Both blocks run the same script: build a legacy Excel project whose model-parameters and individuals workbooks each carry an `adult_male` sheet, then import it into a fresh folder.

``` r
devtools::load_all("/tmp/esqlabsR-base", quiet = TRUE)

# A legacy Excel project. Before 6.0.0 the model-parameters and individuals
# workbooks were separate parameter-set namespaces, so both may carry a sheet
# named after the individual.
excelDir <- file.path(tempdir(), "legacy")
dir.create(excelDir)
suppressMessages(initProject(excelDir, type = "example", createExcel = TRUE))

params <- data.frame(
  "Container Path" = "Organism|Liver",
  "Parameter Name" = "Volume",
  Value = 2.5,
  Units = "l",
  check.names = FALSE
)
addSheet <- function(file, name) {
  sheets <- readxl::excel_sheets(file)
  kept <- lapply(sheets, function(s) readxl::read_excel(file, sheet = s))
  writexl::write_xlsx(c(setNames(kept, sheets), setNames(list(params), name)), file)
}
addSheet(file.path(excelDir, "Configurations", "ModelParameters.xlsx"), "adult_male")
addSheet(file.path(excelDir, "Configurations", "Individuals.xlsx"), "adult_male")

# Import it into a fresh folder.
jsonDir <- file.path(tempdir(), "imported")
importProjectFromExcel(file.path(excelDir, "Project.xlsx"), outputDir = jsonDir)

# Did both parameter sets survive, and does the individual still reach its own?
project <- loadProject(file.path(jsonDir, "Project.json"))
names(project$definitions$parameterSets)
#> [1] "aciclovir"                  "aciclovir_iv_250mg_default"
#> [3] "adult_male.1"               "adult_male"                
#> [5] "adult_male_default"         "global"
unlist(project$definitions$individuals[["adult_male"]]$parameterSets)
#> [1] "adult_male_default" "adult_male"

# Is the imported project self-contained?
list.files(jsonDir)
#> [1] "definitions"  "Project.json"
file.exists(file.path(jsonDir, "Models", "Simulations", "Aciclovir.pkml"))
#> [1] FALSE
```

</details>

Both bugs are visible: `adult_male.1` is an orphan nobody references, the individual resolves to the model-parameters set instead of its own, the call prints nothing, and the output holds no `Models/`.

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/issue-1146-reevaluation-feb95a", quiet = TRUE)

# A legacy Excel project. Before 6.0.0 the model-parameters and individuals
# workbooks were separate parameter-set namespaces, so both may carry a sheet
# named after the individual.
excelDir <- file.path(tempdir(), "legacy")
dir.create(excelDir)
suppressMessages(initProject(excelDir, type = "example", createExcel = TRUE))

params <- data.frame(
  "Container Path" = "Organism|Liver",
  "Parameter Name" = "Volume",
  Value = 2.5,
  Units = "l",
  check.names = FALSE
)
addSheet <- function(file, name) {
  sheets <- readxl::excel_sheets(file)
  kept <- lapply(sheets, function(s) readxl::read_excel(file, sheet = s))
  writexl::write_xlsx(c(setNames(kept, sheets), setNames(list(params), name)), file)
}
addSheet(file.path(excelDir, "Configurations", "ModelParameters.xlsx"), "adult_male")
addSheet(file.path(excelDir, "Configurations", "Individuals.xlsx"), "adult_male")

# Import it into a fresh folder.
jsonDir <- file.path(tempdir(), "imported")
importProjectFromExcel(file.path(excelDir, "Project.xlsx"), outputDir = jsonDir)
#> Warning: ! 1 parameter set in 'Individuals.xlsx' reuses an id an earlier workbook
#>   already defined, so it was renamed:
#> • "adult_male" -> "adult_male_1"
#> ℹ The three former parameter-set kinds now share one parameterSets namespace,
#>   so one sheet name cannot serve two sets. References made in
#>   'Individuals.xlsx' point at the renamed set; rename the sheet in Excel to
#>   choose the id yourself.
#> Imported
#> '/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpD2wXXj/legacy/Project.xlsx'
#> into the JSON project
#> '/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpD2wXXj/imported/Project.json'.
#> ── Definitions ─────────────────────────────────────────────────────────────────
#>   • Scenarios: 3
#>   • Individuals: 1
#>   • Populations: 1
#>   • Parameter Sets: 6
#>   • Initial Conditions: 1
#>   • Applications: 1
#>   • Output Paths: 2
#>   • Data Combined: 1
#>   • Plots: 1
#>   • Plot Grids: 1
#>   • Parameter Identification: 1
#> 
#> Copied 3 referenced folders into the new project: 'Models/Simulations/',
#> 'Data/', and 'Populations/'.

# Did both parameter sets survive, and does the individual still reach its own?
project <- loadProject(file.path(jsonDir, "Project.json"))
names(project$definitions$parameterSets)
#> [1] "aciclovir"                  "aciclovir_iv_250mg_default"
#> [3] "adult_male"                 "adult_male_1"              
#> [5] "adult_male_default"         "global"
unlist(project$definitions$individuals[["adult_male"]]$parameterSets)
#> [1] "adult_male_default" "adult_male_1"

# Is the imported project self-contained?
list.files(jsonDir)
#> [1] "Data"         "definitions"  "Models"       "Populations"  "Project.json"
file.exists(file.path(jsonDir, "Models", "Simulations", "Aciclovir.pkml"))
#> [1] TRUE
```

</details>

Closes #1163
Closes #1165
Closes #1164
Closes #1166

## Summary by CodeRabbit

* **New Features**
  * Improved Excel project imports by merging duplicate parameter sets, renaming conflicts, and updating references automatically.
  * Copies referenced model, data, and population folders into the imported project when available.
  * Canonicalizes scenario references for consistent imported data.
  * Added a clear import summary showing generated files, definition counts, and copied or missing folders.

* **Bug Fixes**
  * Prevented unsafe or invalid folders from being copied during imports.
  * Added warnings for renamed parameter sets and missing referenced folders.

* **Documentation**
  * Clarified that `silent = TRUE` suppresses the complete import summary.


R/project-excel.R
```diff
@@ -445,11 +494,37 @@ importProjectFromExcel <- function(
...
 
-  if (interactive() && !silent) {
-    inputFile <- fs::path_rel(projectConfigPath, start = getwd())
-    outputFile <- fs::path_rel(outputPath, start = getwd())
-    msg <- messages$createdFileSnapshot(inputFile, outputFile)
+  # The definitions reference models, data, and population files by a path
+  # relative to the project folder, so importing into a different folder would
+  # leave every one of those references dangling. Bring the referenced input
+  # folders along, so the imported project runs where it was written.
+  assets <- .copyExcelProjectAssets(filePathProps, pcDir, outputDir)
```
Felixmil: **`projectStatus()` now recursively copies the whole asset tree on every call.**

The copy is unconditional for every caller of `importProjectFromExcel()`, and one of those callers is internal: `.compareJsonToExcel()` (line 793) imports the side-car workbook into a `tempfile()` dir to diff against. That runs on every `projectStatus()` / `project$status` read, so a read-only status check now recursively copies `Models/Simulations/`, `Data/`, and `Populations/` into a temp directory and deletes them again — for a project with large pkml/csv assets that turns a cheap query into a full tree copy.

It also has a correctness edge: the copy can fail (permissions, a full disk, a file locked by another process) and the failure propagates out of `.compareJsonToExcel()`, where `projectStatus()`'s `tryCatch` at line 1029 turns it into `excel_in_sync = NA` — "cannot compare the Excel files", which misattributes a copy failure to an unreadable workbook.

The asset copy exists to make a *user-facing* import self-contained; the sync snapshot only needs the definition tree, since it is serialized and thrown away. Suggest gating it — an internal `copyAssets` argument defaulting to `TRUE`, passed `FALSE` from `.compareJsonToExcel()` — rather than letting `silent` do double duty.

---

R/project-excel.R
```diff
@@ -1197,6 +1281,200 @@ projectStatus <- function(project, silent = FALSE) {
...
+    if (from == fs::path_norm(sourceDir)) {
+      next
+    }
+    if (!fs::dir_exists(from)) {
+      result$notCopied <- c(result$notCopied, value)
+      next
+    }
+    to <- fs::path(outputDir, value)
+    fs::dir_create(fs::path_dir(to))
+    fs::dir_copy(from, to, overwrite = TRUE)
```
Felixmil, line 1424: **The asset copy sits outside the `overwrite` guard, but copies with `overwrite = TRUE`.**

The import's guard (line 453) covers exactly two things: the `Project.json` container and a non-empty `definitions/` tree.

```r
if (!overwrite && (file.exists(outputPath) || hasDefinitionTree)) {
```

Nothing in it looks at the asset folders, so `importProjectFromExcel(excel, outputDir = d)` with the default `overwrite = FALSE` proceeds whenever `d` holds no JSON project — and then unconditionally copies into `d/Data`, `d/Models/Simulations`, `d/Populations` with `overwrite = TRUE`. Any pre-existing same-named file the user had under those folders is replaced, with no warning and no way to decline: the argument the user set to *not* overwrite anything does not reach this call.

The realistic case is a user who prepares an output folder (drops in a corrected pkml, or a data file they curated) and then imports the legacy workbook beside it. Passing `overwrite` through to this function and skipping — or at minimum warning about — an existing non-empty target folder would make the flag mean what it says.

---

R/project-excel.R
```diff
@@ -1197,6 +1281,200 @@ projectStatus <- function(project, silent = FALSE) {
...
+#' @returns `list(copied, notCopied)`: the folder values copied, and those a
+#'   definition may reference but that could not be copied (absent from the
+#'   source project, or naming a location outside it). Both empty when the
+#'   project was imported in place (nothing to copy).
+#' @keywords internal
+#' @noRd
+.copyExcelProjectAssets <- function(filePathProps, sourceDir, outputDir) {
+  result <- list(copied = character(), notCopied = character())
+  # Imported in place: the folders are already where the definitions expect them.
+  if (fs::path_norm(sourceDir) == fs::path_norm(fs::path_abs(outputDir))) {
```
Felixmil: **The imported-in-place check is a purely lexical path comparison, so it can miss.**

`fs::path_norm()` and `fs::path_abs()` are both lexical — they collapse `.`/`..` and prepend the working directory, but they never resolve a symlink and never case-fold. `sourceDir` arrives as `dirname(fs::path_abs(projectConfigPath))` (line 65), equally lexical. Contrast the containment code at line 129, which deliberately finishes with `normalizePath()`.

So the same directory reached by two spellings compares unequal and the early return is skipped:

- `outputDir = "C:/proj"` against a `projectConfigPath` under `c:/proj` (Windows is case-insensitive; `path_norm` is not)
- `outputDir = "/tmp/proj"` where `/tmp` is a symlink to `/private/tmp` (macOS) and the workbook path resolved through the target
- any bind mount or symlinked project root

When it is skipped, `from` and `to` name the same directory and every asset folder is copied onto itself with `overwrite = TRUE`. That is at best pointless work on a large tree and at worst a copy of a directory into its own subtree.

`normalizePath(..., mustWork = FALSE)` on both sides — matching what line 129 already does for the configurations folder — would make the check agree with the rest of the file.

---

R/project-excel.R
```diff
@@ -1197,6 +1281,200 @@ projectStatus <- function(project, silent = FALSE) {
...
+.appendParameterSets <- function(existing, incoming, source) {
+  if (length(incoming) == 0L) {
+    return(list(sets = existing, renames = character()))
+  }
+  ids <- c(names(existing), names(incoming))
+  canonical <- vapply(ids, .canonicalizeOneId, character(1), USE.NAMES = FALSE)
+  # `make.unique()` leaves each first occurrence untouched and suffixes the
+  # later ones, and `existing` comes first, so an already-accumulated id is
+  # never renamed out from under a reference that resolved to it.
+  uniqued <- make.unique(canonical, sep = "_")
```
Felixmil, line 1475: **`make.unique()` spans one workbook's own sheets too, and the warning then misattributes the clash.**

`ids` is `c(names(existing), names(incoming))`, so the uniquifying does not only compare across workbooks — two sheets *within* `incoming` that canonicalize to one id (`Set 1` and `Set_1`, or `Rat` and `rat`, in the same workbook) also collide here, and the second is renamed to `Set_1_1`.

The rename itself is defensible, and the references that workbook makes do get re-pointed, so this is not data loss. The problem is what the user is told: the warning at line 1442 says the id was one *"an earlier workbook already defined"* and advises that references *"made in `<this workbook>`"* now point at the renamed set. For a same-workbook clash there is no earlier workbook — both sheets are in the file being named — so the message sends the user looking in the wrong place for a sheet they should rename in the file they are already holding.

Worth either splitting the message (a within-workbook variant naming both sheet names), or wording the existing one so it does not assert an earlier workbook.

Separately, a within-workbook collision is exactly the shape `.canonicalizeId()` treats as real ambiguity and aborts on (`R/definition-id.R:65-91`, *"Two distinct ids that canonicalize to the same id are ambiguous"*). Silently renaming here where the id layer would refuse is a deliberate-looking divergence — if it is intended, it's worth a comment saying so, since the next reader will expect the abort.

---

R/messages.R
```diff
@@ -194,6 +194,37 @@ messages$importWouldOverwriteProject <- function(outputDir) {
...
+# Raised by `.appendParameterSets()` when a workbook's sheet reuses a
+# parameter-set id an earlier workbook already took. `bullets` are the
+# pre-bound `old -> new` templates from `.canonicalizedIdBullets()`; the caller
+# binds `sourceLabel` and `renamedCount` into the same environment and passes it
+# as `.envir`, so the whole message is glue-parsed exactly once and a sheet name
+# containing `{` is never evaluated.
+messages$importRenamedDuplicateParameterSets <- function(bullets) {
+  c(
+    "!" = "{renamedCount} parameter set{?s} in {.file {sourceLabel}} reuse{?s/} \\
+    an id an earlier workbook already defined, so {?it was/they were} renamed:",
```
Felixmil: **The `"!"` bullet's pluralization is also stuck on singular** — same root cause as the `"i"` bullet flagged below, but two more markers.

cli binds each `{?}` to the quantity of the *nearest preceding* substitution. `{.file {sourceLabel}}` is a substitution of length 1, so it resets the quantity to 1 for everything after it in this bullet:

```
"{renamedCount} parameter set{?s} in {.file {sourceLabel}} reuse{?s/} \
an id an earlier workbook already defined, so {?it was/they were} renamed:"
```

Only `set{?s}`, which sits directly after `{renamedCount}`, gets the real count. `reuse{?s/}` and `{?it was/they were}` both see qty 1, so with 3 renamed sets the warning reads:

> ! 3 parameter sets in 'Individuals.xlsx' **reuses** an id an earlier workbook already defined, so **it was** renamed:

Adding `{cli::qty(renamedCount)}` before `reuse{?s/}` fixes both markers in this bullet (the quantity carries forward from there). The PR description's example output happens to have `renamedCount == 1`, which is why it looks right there — the bug only shows with two or more.

---

Felixmil (commented):
Automated review of the 5-commit range `c3d4942...91f7d7c` (#1163, #1164, #1165, #1166). Six findings; four are inline below, one extends an existing thread, one was already reported.

**Two worth acting on before merge, both in the new asset copy:**

1. **`projectStatus()` now copies the asset tree on every call.** The copy is unconditional for every caller of `importProjectFromExcel()`, and `.compareJsonToExcel()` is one — so a read-only status query recursively copies `Models/`, `Data/`, `Populations/` into a temp dir and deletes them. A copy failure there also surfaces as `excel_in_sync = NA`, blaming the workbook.
2. **The copy ignores `overwrite`.** The import's guard covers only `Project.json` and `definitions/`, while `fs::dir_copy()` runs with `overwrite = TRUE` regardless — so `overwrite = FALSE` still replaces pre-existing user files under the target's asset folders.

**Three minor:** the imported-in-place check is a lexical path comparison that a symlinked or case-differing `outputDir` slips past (line 1344); `make.unique()` also catches within-workbook collisions, where the warning's "an earlier workbook" is wrong (line 1424); and the rename warning's `"!"` bullet is stuck on singular pluralization (`R/messages.R:219`).

**Already reported:** `.applyIdRenames()` matching raw strings while clashes are detected canonically — CodeRabbit's [thread](https://github.com/esqLABS/esqlabsR/pull/1179#discussion_r3657204306) has it, and I agree with the diagnosis and the fix. Of the six, that one is the most consequential: it can silently re-point a case-differing reference at the wrong workbook's set, which is the class of bug #1163 set out to fix.

The four inline findings are new. Nothing to flag in the observed-`dataCombined` canonicalization (#1165), the summary's `capture.output` + `cli_verbatim` mechanism, or the summary's `silent` gating — internal callers all pass `silent = TRUE`.

Caveat on method: this environment has no R runtime, so every finding is from reading code, not from executing it. The `fs::dir_copy` merge semantics in finding 2 and the cli quantity-scoping in the pluralization finding were taken from those packages' documented behaviour; the rest is local control flow. Worth a quick check against a live session before changing anything.

---

R/project-excel.R
Felixmil: Fixed in afebc85d. Confirmed first: `.compareJsonToExcel()` imports into a `tempfile()` dir on every `projectStatus()` read, so the copy did turn a read-only query into a recursive copy of the whole asset tree.

Took the suggested shape: `importProjectFromExcel()` gains `copyAssets` (default `TRUE`), and `.compareJsonToExcel()` passes `FALSE`. That also removes the misattribution you noted, since no copy runs on that path to fail.

One deviation: `copyAssets` is a documented argument rather than an internal one. An undocumented argument on an exported function is an `R CMD check` warning, so a truly internal one would have needed a separate unexported wrapper. Documented as the definitions-only escape hatch.

Test asserts `projectStatus()` performs zero `fs::dir_copy()` calls.

R/project-excel.R
Felixmil: Fixed in afebc85d. `overwrite` now reaches `.copyExcelProjectAssets()`: a target asset folder that already holds files is left untouched and reported, instead of being replaced with `overwrite = TRUE` regardless of what the caller asked for.

Test covers the case you described: a hand-curated file placed in the output folder survives `overwrite = FALSE` (and the workbook's own data does not land on it), then travels once `overwrite = TRUE`.

R/project-excel.R
Felixmil: Fixed in afebc85d. Verified the gap with real directories before changing anything: on macOS, `/tmp` symlinked to `/private/tmp`, the lexical compare returns `FALSE` where `normalizePath()` correctly returns `TRUE`.

Both sides now use `normalizePath(..., mustWork = FALSE)`, matching the containment code above. My original in-place test only passed because both paths came from the same string; there is now a test that hands `outputDir` a symlink to the project folder and asserts nothing was duplicated into its own subtree.

R/project-excel.R
Felixmil: Fixed in afebc85d. Reproduced the within-workbook case first: `Set 1` and `Set_1` in one workbook do get uniquified to `Set_1_1`, so the warning was indeed sending the user to look for a nonexistent earlier workbook.

The message no longer asserts one: it now reads "reuse(s) an id that is already taken", which is true whether the clash is across workbooks or inside one.

On the divergence from `.canonicalizeId()`: intended, and now documented on `.appendParameterSets()`. A legacy workbook cannot be hand-edited retroactively, so an import renames and says so rather than refusing the whole migration. The comment says that explicitly, so the next reader is not surprised by the missing abort.

R/messages.R
Felixmil: Fixed in afebc85d. Your diagnosis was exactly right, including why it looked fine in the PR description: my snapshot test had `renamedCount == 1`. With 3 it read

> ! 3 parameter sets in 'Individuals.xlsx' **reuses** an id ... so **it was** renamed:

`cli::qty(renamedCount)` now re-arms the quantity in both bullets, so the `"i"` bullet's `renamed set{?s}` pluralizes too. Verified at 1 and 3.

R/messages.R
```diff
@@ -194,6 +194,37 @@ messages$importWouldOverwriteProject <- function(outputDir) {
...
+    bullets,
+    "i" = "The three former parameter-set kinds now share one \\
+    {.field parameterSets} namespace, so one sheet name cannot serve two sets. \\
+    References made in {.file {sourceLabel}} point at the renamed set{?s}; \\
+    rename the sheet in Excel to choose the id yourself."
+  )
+}
```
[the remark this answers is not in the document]
Felixmil: Fixed in afebc85d. `cli::qty(renamedCount)` added to the `"i"` bullet, and to the `"!"` bullet where two more markers had the same problem.

R/project-excel.R
```diff
@@ -1197,6 +1281,200 @@ projectStatus <- function(project, silent = FALSE) {
...
+#' @keywords internal
+#' @noRd
+.applyIdRenames <- function(ids, renames) {
+  if (length(renames) == 0L || length(ids) == 0L) {
+    return(ids)
+  }
+  ids <- as.character(ids)
+  mapped <- unname(renames[ids])
+  ifelse(is.na(mapped), ids, mapped)
+}
```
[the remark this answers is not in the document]
Felixmil: Fixed in afebc85d. Confirmed the gap: with a `Rat` sheet in one workbook and `rat` in the next, a reference spelled `Rat` stayed `Rat` and resolved to the earlier workbook's set.

Took the anchor suggestion: the map is keyed by canonical id and `.applyIdRenames()` canonicalizes its lookups, so `Rat`, `rat`, and `RAT` all re-point to `rat_1` while an unrelated reference passes through. The sibling site needed no change, as you said. Warning bullets still show the raw sheet name so the user can find the sheet.

Integration test added: a lower-case `Individual Parameter Sets` cell naming an `Indiv1` sheet now follows the rename.

PR #1180, opened 2026-07-27
Fix the authoring friction found in 6.0.0 migration testing

### LINKED ISSUE #1167: addParameterEntry(units=) rejects NA/NULL for "no unit"; NULL message is misleading
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). An empty units cell comes back from `readxl` as `NA`, but `addParameterEntry(units = NA)` aborts, and `units = NULL` aborts with a message that does not name the real fix. A modeler scripting `add*()` from a data frame hits this on every unit-less row.

Messages observed:

- `units = NA`: `Invalid parameter entry: units must be a string scalar (use "" for none)` (clear).
- `units = NULL`: the length guard fires first with a misleading message: `\`containerPath\`, \`parameterName\`, \`value\`, and \`units\` must be vectors of the same length. Got lengths 1, 1, 1, and 0.`

Root cause: the length assertion in `R/parameters.R` runs before the helpful hint in `.validateParameterEntry()` (`R/parameters.R:1015-1018`). The stored value is fine (both the importer and `units = ""` serialize to JSON `"units": null`); only the authoring input contract is stricter than the data a user reads from Excel.

Suggested fix: accept `NA`/`NULL` as "no unit" (coerce to `""`), so the authoring input matches what the importer tolerates and what `readxl` returns.

### LINKED ISSUE #1168: addScenario(simulationTime=) rejects the numeric time grid and has contradictory roxygen
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). A scenario's simulation-time grid is a start/end/resolution triple. `addScenario()` accepts only the raw string form (e.g. `"0, 42, 48"`); passing the natural parsed numeric vector `c(0, 42, 48)` aborts:

`\`simulationTime\` must be length 1 or length 1 (the number of ids). It is length 3.`

The two constructors also document the field with contradictory types: `addScenario()`'s roxygen (`R/scenarios.R:774`) describes `simulationTime` as a Character `"start, end, resolution"`, while the lower-level `Scenario()` constructor (`R/scenarios.R:66`) describes it as a length-3 numeric vector `c(start, end, resolution)`.

Root cause: the vectorization guard treats the length-3 numeric vector as a bad recycle (`R/authoring-vectorize.R:194`); the error message ("length 1 or length 1") does not name the real requirement.

Suggested fix: accept a length-3 numeric vector (or a list of them) in `addScenario()`, and align the two roxygen blocks to one documented type.

### LINKED ISSUE #1171: addObservedData(entry=) list shape is undiscoverable; paths are relative to dataFolder
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). `addObservedData(project, entry, overwrite)` gives no hint from its signature that `entry` must be a list carrying a `type` field (`excel`/`pkml`/`script`/`programmatic`) plus per-type required fields. The first natural attempt fails with `observedData entry must include a type field`, and the accepted shape is only discoverable by reading `R/observed-data.R:63-71`.

Additionally, `file` and `importerConfiguration` are resolved relative to `paths$dataFolder` (`R/observed-data.R:162`), so prefixing them with `Data/` yields a `Data/Data/...` not-found, which is easy to hit and not documented at the call site.

Suggested fix: document the `entry` list contract in the `addObservedData()` roxygen (with a per-type field table), and clarify that `file`/`importerConfiguration` are relative to `dataFolder`. A typed constructor helper for the entry would remove the guesswork entirely.

### LINKED ISSUE #1172: Observed-data definition file is named from the file basename, not the entry id
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). Authoring an observed-data entry with `id = "obs_demo"` and `file = "<name>.xlsx"` writes the definition file as `definitions/observed-data/<name>.xlsx.json` (a doubled `.xlsx.json` extension), not `obs_demo.json`. It is consistent across both the importer and authoring paths and is functionally correct, but the on-disk filename not matching the definition id is surprising given every other section files by id.

Root cause: the observed-data serializer in `R/definition-files.R` uses the `file` basename as the filename stem rather than the entry `id`.

Suggested fix: name the observed-data definition file from the entry `id` (like every other section), so the on-disk layout is uniform.

### LINKED ISSUE #1173: initProject() aborts instead of creating a missing destination folder
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). `initProject("authored", type = "minimal", createExcel = FALSE)` aborts if the destination folder does not already exist:

`The specified destination folder does not exist.`

This is the very first call in the hand-authoring workflow, so a new user's natural `initProject("myproject")` in an empty parent fails until they `dir.create()` first.

Root cause: the destination guard in `R/project-lifecycle.R` (around `:263-320`), message at `R/messages.R:246-248`.

Suggested fix: create the destination folder if it is absent (it is about to be populated by `initProject()` anyway), or make the missing-folder message name the one-line fix.

Felixmil:
Fixes the authoring friction found while migration-testing a legacy Excel project against the 6.0.0 JSON model. Each fix is a place where the authoring API was stricter, or less discoverable, than the data a modeller actually has in hand: a unit cell that Excel returns as `NA`, a time grid held as three numbers, a project folder that does not exist yet, an observed-data declaration that could not be named. One commit per fix.

Merges after #1179, which owns the Excel-importer half of the same migration test.

## Authoring API

- `addParameterEntry()` reads `NA` and `NULL` as "no unit", so a parameter table read from Excel can be handed straight to it; `units` gained a `NULL` default and can be omitted for a set of base-unit entries.
- `addScenario()`, `setScenario()`, and `createScenariosFromPKML()` accept a scenario's `simulationTime` as the numeric grid `c(start, end, resolution)`, not only as the `"start, end, resolution"` string. A list gives one grid per id; several intervals are still written as one string. The conversion runs ahead of the vectorization step in each of the three, so a length-3 numeric is one interval rather than three definitions.
- Clearing `simulationTime` with `setScenario(project, id, simulationTime = NULL)` no longer drops the slot from the record, where a later read of the field partial-matched `simulationTimeUnit` and returned the unit string.
- `initProject()` creates the `destination` folder when it does not exist, instead of aborting. It is the first call of the authoring workflow, so `initProject("myProject")` in an empty parent folder now works without a preceding `dir.create()`.
- The recycling engine no longer emits "must be length 1 or length 1" when a single id was given.

## Observed data

- An `observedData` declaration can carry an `id`. It names the declaration's file (`definitions/observed-data/<id>.json`) and is the key `removeObservedData()` matches it by, so observed data is filed by id like every other section rather than under its data file's name (which produced the doubled `Aciclovir_TimeValuesData.xlsx.json`). A declaration without an `id` derives one from its source exactly as before, so existing projects are unaffected.
- Consequence worth a reviewer's eye: `removeObservedData()` now also matches a programmatic sentinel that has no live `DataSet` behind it, where it previously reported the id as not found.
- `addObservedData()` documents the `entry` list it takes: a per-type field table, and the fact that `file` and `importerConfiguration` resolve under the project's data folder (so they carry no `Data/` prefix). The observed-data article gained the same for `id`.

## Known gaps

- `validateProject()` does not flag an unsafe or duplicate observed-data id; only `saveProject()` aborts on one. That matches the existing behaviour for file basenames, so it is a gap rather than a regression.
- The clean-schema RFC (#1069) wants `simulationTime` stored as a real array. This PR only accepts the numeric form as *input*; the stored form is still the string.

## Example

Before:

``` r
devtools::load_all("/path/to/esqlabsR", quiet = TRUE)
dir <- file.path(tempdir(), "before")

# 1. initProject() into a folder that does not exist yet
initProject(dir, type = "minimal", createExcel = FALSE)
#> Error in `initProject()`:
#> ! The specified destination folder does not exist.
#>   ('/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpA4rdTx/before')
dir.create(dir)
initProject(dir, type = "minimal", createExcel = FALSE)

# 2. an empty Units cell, which readxl gives as NA
addParameterEntry(project, "liver", "Organism|Liver", "Volume", 1.8, NA)
#> Created parameter set "liver" on demand to hold the new entry.
#> Error in `addParameterEntry()`:
#> ! Invalid parameter entry:
#> ✖ units must be a string scalar (use "" for none)

# 3. the time grid as numbers
addScenario(project, "iv", modelFile = "Aciclovir.pkml", simulationTime = c(0, 42, 48))
#> Error in `addScenario()`:
#> ! `simulationTime` must be length 1 or length 1 (the number of ids).
#> ✖ It is length 3.
```

After:

``` r
devtools::load_all("/path/to/esqlabsR", quiet = TRUE)
dir <- file.path(tempdir(), "after")

# 1. initProject() into a folder that does not exist yet
initProject(dir, type = "minimal", createExcel = FALSE)

# 2. an empty Units cell, which readxl gives as NA
addParameterEntry(project, "liver", "Organism|Liver", "Volume", 1.8, NA)
#> Created parameter set "liver" on demand to hold the new entry.
project$definitions$parameterSets$liver[[1]]$units
#> NULL

# 3. the time grid as numbers
addScenario(project, "iv", modelFile = "Aciclovir.pkml", simulationTime = c(0, 42, 48))
project$definitions$scenarios$iv$simulationTime
#> [[1]]
#> [1]  0 42 48
```

</details>

Closes #1167
Closes #1168
Closes #1171
Closes #1172
Closes #1173

## Summary by CodeRabbit

* **New Features**
  * Observed-data entries can use explicit IDs for saving, replacing, and removal.
  * Numeric simulation-time grids are supported for scenarios, including per-scenario grids.
  * Project initialization now creates missing destination folders.
  * Parameter units are optional; `NULL`, `NA`, and empty values represent no unit.
  * Session-provided observed datasets are preserved when projects are saved and reloaded.

* **Bug Fixes**
  * Clearing scenario simulation time no longer leaves an outdated time unit.
  * Improved validation messages for invalid simulation times, observed-data IDs, and field lengths.

* **Documentation**
  * Expanded guidance for observed-data formats, identifiers, simulation times, and parameter units.


R/definition-files.R
```diff
@@ -885,34 +887,53 @@
...
   id
 }
 
+# The same id, `NA` when the declaration carries nothing to derive one from.
+# Used where a declaration without an id is a normal case to skip past (matching
+# an id against the section) rather than a failure.
+#
+# @keywords internal
+# @noRd
+.observedDataEntryIdOrNA <- function(entry) {
```
Felixmil, line 962: **The declared `id` is added to the write side without the load-side guard every other keyed section gets.**

`observedData` is now a keyed section whose key lives *inside* the file, but `.parseObservedDataTree()` (L858-874) only strips `.definitionFile` and re-stamps the class. Every other keyed kind runs its record's inner id through `.keyedTreeRecordId()` (L1171), which enforces two things this section now needs and doesn't get:

1. **id vs. filename stem.** `definitions/observed-data/foo.json` containing `"id": "bar"` loads without complaint; the next `saveProject()` writes `bar.json` and reconciles `foo.json` away. A project that was only *opened* comes back with a renamed definition file. Worse, two files whose inner `id`s coincide both load fine and then make `saveProject()` abort in `.serializeObservedDataSet()` — an unsavable project produced purely by reading one.
2. **type and length.** A JSON number, bool, or array reaches `entry[["id"]]` verbatim. `nzchar(2024)` is `TRUE`, so `.observedDataEntryIdOrNA()` hands back a *double*, and the failure lands somewhere else entirely: `vapply(..., .observedDataEntryIdOrNA, character(1))` in `.addObservedData_impl()` dies with `values must be type 'character'`, and `saveProject()` dies inside `basename()` with `a character vector argument expected`. An array id errors in the `||` on this line's `if`, on a length-2 condition.

Wiring `observedData` into `.keyedTreeRecordId()` (or an id-specific equivalent that also honours the derive-from-`file` fallback) closes both, and would subsume the blank/`NA` case CodeRabbit raised a few lines down.

---

R/observed-data.R
```diff
@@ -500,28 +527,29 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
...
       ))
     }
     private$.observedDataNamesCache <- NULL
     observedData <- private$.getSection("observedData")
-    # On overwrite, replace the existing entry sharing this basename in place;
-    # otherwise append (a non-overwrite basename clash aborted above).
-    replaceIdx <- if (overwrite) which(existingFiles == fileBase) else integer()
+    # On overwrite, replace the existing entry carrying this id in place;
+    # otherwise append (a non-overwrite id clash aborted above).
+    replaceIdx <- if (overwrite) which(existingIds == entryId) else integer()
```
Felixmil, line 571: **`overwrite = TRUE` can replace a programmatic sentinel here and orphan the session's `DataSet`.**

`existingIds` is now computed with `.observedDataEntryIdOrNA()` over the *whole* section, so it includes `programmatic` sentinels (whose id is their `name`). A config declaration whose `id` — or `file` basename — equals a live `DataSet` name therefore matches, and `observedData[[replaceIdx[[1]]]]` swaps the sentinel out in place. `private$.programmaticDataSets[[name]]` is never cleared.

```r
addObservedData(project, dataSet)                                      # name "obs"
addObservedData(project, list(type = "pkml", file = "x.pkml", id = "obs"),
                overwrite = TRUE)                                      # sentinel gone, DataSet not
```

Three consequences, all silent:

- `saveProject()` → `.persistProgrammaticObservedData()` only collects entries that are still `type = "programmatic"`, so the `DataSet` is never written to PKML. It's simply gone after a reload — and the `.unresolvedProgrammaticNames()` warning that normally makes such a drop loud can't fire, because the sentinel it keys on no longer exists.
- `removeObservedData(project, "obs")` takes the `one %in% names(private$.programmaticDataSets)` branch first (L618), finds no matching programmatic entry, so `dropIdx` stays empty: it deletes the orphaned `DataSet` and leaves the pkml declaration in place, without the not-found warning.
- If the new source happens to resolve a data set also named `"obs"`, `.mergeObservedDataSets()` aborts on the collision on *every* subsequent `loadObservedData()`.

The `DataSet` branch already refuses the mirror image of this (L486-496: a name colliding with a file-based source aborts and tells you to remove it first). The config branch reaching the same crossing should refuse it symmetrically, or clear `private$.programmaticDataSets[[entryId]]` alongside the replacement.

---

R/observed-data.R
```diff
@@ -500,28 +527,29 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
...
-    fileBase <- basename(entry[["file"]])
-    existingFiles <- vapply(
+    # Config entries are keyed by the id `removeObservedData()` matches on: the
+    # entry's own `id` when it declares one, else its `file` basename. That id
+    # also names the definition file, so reject an unsafe one here rather than
+    # at save. Abort on a duplicate to match the other mutators' convention,
+    # unless overwriting.
+    entryId <- .observedDataEntryId(entry)
+    .validateObservedDataId(entryId)
+    existingIds <- vapply(
```
Felixmil: **Add-time collision detection is now asymmetric: this branch keys on declaration ids, the `DataSet` branch keys on resolved data-set names.**

The change makes the config branch id-aware, which is right. But the `DataSet` branch still tests `name %in% existingNames` (L457-458), where `existingNames` comes from `.getObservedDataNames_impl()` — *resolved* data-set names, not declaration ids. A declaration carrying `id = "obs"` whose file resolves data sets named something else is invisible to that check:

```r
addObservedData(project, list(type = "pkml", file = "x.pkml", id = "obs"))
addObservedData(project, dataSet)   # dataSet$name == "obs" — no clash reported
```

The sentinel is appended, and the section now holds two entries with the on-disk id `"obs"`. Nothing surfaces until `saveProject()`, which aborts in `.serializeObservedDataSet()` (or the duplicate check in `.persistProgrammaticObservedData()`) — the "surviving collision aborts saveProject()" path, reached this time without the user ever forcing an overwrite. And while the `DataSet` is live, `removeObservedData(project, "obs")` hits the programmatic branch first, so the declaration can't be reached by its own id.

Since the id resolver is the same for both kinds now, the `DataSet` branch could test against `existingIds` too (in addition to `existingNames`, which it still needs for the resolved-name case).

---

R/observed-data.R
```diff
@@ -500,28 +527,29 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
...
-    # overwriting.
-    fileBase <- basename(entry[["file"]])
-    existingFiles <- vapply(
+    # Config entries are keyed by the id `removeObservedData()` matches on: the
+    # entry's own `id` when it declares one, else its `file` basename. That id
+    # also names the definition file, so reject an unsafe one here rather than
+    # at save. Abort on a duplicate to match the other mutators' convention,
+    # unless overwriting.
+    entryId <- .observedDataEntryId(entry)
+    .validateObservedDataId(entryId)
```
Felixmil: **Error attribution: an unsafe declared `id` aborts as `` Error in `.validateObservedDataId()` ``, not `` Error in `addObservedData()` ``.**

Neither `.validateObservedDataId()` nor `.observedDataEntryId()` takes a `call`, so their `cli_abort()`s default to their own frame — the `rlang::local_error_call(.call)` at the top of `.addObservedData_impl()` doesn't reach into a callee. `tests/testthat/_snaps/definition-files.md:127` captures the internal attribution, so this is the live behaviour.

Reaching it is now easy in a way it wasn't: `addObservedData(project, list(type = "pkml", file = "a.pkml", id = "../escape"))` is a plain user-facing call whose whole point is the new `id` field, and it names an internal helper in the error. Per the project's error-attribution convention, thread `.call` through both helpers (`call = rlang::caller_env()` defaulted, as `.simulationTimeIntervalString()` in this same PR does) and pass it from the `_impl`.

Worth noting L455 already had this wart before the PR, so it's an existing pattern being extended rather than a regression — but the same one-line fix covers both call sites.

---

Felixmil (commented):
Reviewed `c3d4942..58eee7e` (29 files, +546/−132). Five findings, all in the observed-data `id` feature; the other four fixes look clean to me.

### Verified clean

- **`simulationTime` numeric grid.** `.asSimulationTimeString()` is correctly placed *before* `.alignAuthoringArgs()` in `addScenario()`/`setScenario()` and before the scenario count in `createScenariosFromPKML()`, so a length-3 numeric can't be misread as three per-id values. The string form passes through untouched, so multi-interval strings and length-n character vectors keep working.
- **`sc["simulationTime"] <- list(NULL)`.** A real fix, and necessary: `$` on a list partial-matches, so once the slot is dropped `sc$simulationTime` returns the `simulationTimeUnit` string. The `dots$simulationTime` read at L1235 is safe because the exact name is guaranteed present by the guard above it. The sibling prefix pair `steadyStateTime`/`steadyStateTimeUnit` is also safe — the only path that drops that slot implies the unit is `NULL`. Serialization is fine too (`.writeDefinitionJson` uses `null = "null"`, so no `{}` for `.assertNoEmptyObjectFields` to reject later).
- **`.normalizeParameterUnits()`.** Correct for `NULL`, a logical `NA` scalar, and `NA` elements inside a character vector — including the all-empty Excel Units column, which `readxl` gives as a length-n logical `NA` vector. Runs before the parallel-vector length check as the comment claims; a non-character value still falls through to `.validateParameterEntryArgs()` to be rejected by type.
- **`initProject()`.** `fs::dir_create()` is a no-op on an existing folder and still aborts if the path is a regular file; `messages$pathNotFound` has no remaining callers.
- **`.recycleField()` message.** No existing snapshot captures the `n == 1` wording, so nothing breaks.

### Findings

Four are inline. The fifth has no anchor in the diff:

**`.persistProgrammaticObservedData()` drops a declared `id` (`R/observed-data.R:851-856`).** The rebuilt entry is `list(type = "pkml", file = fileNames[[i]])` — no `id` carried over. `.observedDataEntryIdOrNA()` honours `id` for *every* type, including `programmatic`, so a hand-authored `{"type": "programmatic", "name": "obs", "id": "myobs"}` loses its remove handle at save and its definition file moves from `myobs.json` to `obs.pkml.json`. Same at L485, where an overwrite replaces a sentinel with a freshly built one. Only reachable by hand-authoring JSON, so low severity — but the PR makes `id` the removal handle, and this is the one path that silently revokes it. Either preserve `entry[["id"]]` in the rebuild, or reject `id` on a `programmatic` declaration so the doc table (`excel`/`pkml`/`script` only) is the actual contract.

### Root cause

The first two inline findings share one: `id` was added to the write and match sides without the load-side guard the other keyed sections get from `.keyedTreeRecordId()`. Wiring `observedData` into that helper closes both, and subsumes CodeRabbit's blank/`NA` point at `R/definition-files.R:928`.

One correction to that CodeRabbit comment while it's open: `nzchar(NA)` is `TRUE`, so `id = NA_character_` does *not* reach `if (NA)` — it returns `NA_character_` and aborts with the intended "no id to name its definition file" message. Only the `id = ""` half of that finding stands (an empty string suppresses the `file` fallback instead of being treated as absent).

Caveat on method: no R in this environment, so everything above is from reading the code and the committed snapshots, not from running it. The failure scenarios are worth a quick reprex before acting on them.

---

man/addObservedData.Rd
```diff
@@ -9,22 +9,47 @@ addObservedData(project, entry, overwrite = FALSE)
...
+\item{entry}{An \code{\link[ospsuite:DataSet]{ospsuite::DataSet}} object, or a configuration list
+carrying a \code{type} (\code{"excel"}, \code{"pkml"}, or \code{"script"}) and that type's
+fields, as described above.}
 
 \item{overwrite}{Logical scalar. When \code{FALSE} (default), a source whose id
-already exists (a \code{DataSet} name, or a config \code{file} basename) aborts.
-When \code{TRUE}, the existing source with that id is replaced (last-write-wins).}
+already exists (a \code{DataSet} name, or a configuration list's \code{id} or \code{file}
+basename) aborts. When \code{TRUE}, the existing source with that id is replaced
+(last-write-wins).}
```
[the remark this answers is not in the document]
Felixmil: Documented in 26efc180. The `overwrite` param now states that a replacement only happens between sources of the same kind, and that a cross-kind one is refused in either direction (it would either strand the session's `DataSet` or leave two sources resolving to one name), pointing at `removeObservedData()` first. The refusal is now symmetric in code too: 515e2b85 added the missing direction, a configuration entry overwriting a live programmatic source.

man/addObservedData.Rd
```diff
@@ -9,22 +9,47 @@ addObservedData(project, entry, overwrite = FALSE)
...
+
+\code{file} and \code{importerConfiguration} are paths \emph{relative to the project's data
+folder} (\code{project$paths$dataFolder}). A file sitting directly in that folder
+is \code{"observed.xlsx"}, not \code{"Data/observed.xlsx"}; one in a subfolder of it is
+\code{"subfolder/observed.xlsx"}. \code{sheets} lists the Excel sheet names to import.
+A \code{script} source runs the R file it names (see the Security section of
+\code{\link[=loadObservedData]{loadObservedData()}}).
+
+\code{id} names the declaration itself: it becomes the declaration's file in
+\verb{definitions/observed-data/} and is the id \code{\link[=removeObservedData]{removeObservedData()}} matches on.
```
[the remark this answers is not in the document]
Felixmil: Fixed in 26efc180 at all three sites. The roxygen now says the declaration's file lives under the project's definitions folder, with `definitions/observed-data/<id>.json` given as the default layout; NEWS.md and the vignette sentence carry the same qualification.

R/definition-files.R
```diff
@@ -885,34 +887,53 @@
...
+      entry[["name"]]
+    } else {
+      basename(entry[["file"]] %||% "")
+    }
+  }
+  if (is.null(id) || !nzchar(id)) {
+    return(NA_character_)
+  }
+  id
+}
```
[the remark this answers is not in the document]
Felixmil: Fixed in 515e2b85, and generalised. `.usableObservedDataId()` now gates every id source: absent, `""`, `NA`, or a non-string (a hand-edited JSON number or array) all count as no id, so the resolver falls back to `file` / `name` and returns `NA` only when nothing usable remains. That also stops a non-string id reaching `basename()` or a `character(1)` `vapply()`. At the authoring boundary the same value is rejected outright rather than silently ignored: `addObservedData(project, list(id = "", ...))` aborts with "An observedData entry's id must be a single non-empty string", since quietly filing under the file basename would hide the mistake.

R/messages.R
```diff
@@ -488,6 +482,15 @@ messages$unknownScenarioNames <- function(unknownNames) {
   )
 }
 
+messages$invalidSimulationTimeArgument <- function() {
+  cliFormat(
+    "{.arg simulationTime} must be a length-3 numeric vector \\
+    {.code c(start, end, resolution)}, or the same grid as a string \\
+    {.val 0, 42, 48} (several intervals separated by {.val ;}). To give a \\
+    different grid per id, pass a list with one element per id."
+  )
```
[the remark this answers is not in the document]
Felixmil: Not changed, and here is the reasoning so you can overrule it. The backslash-newline continuation is the established idiom in this file: every multi-line entry in `R/messages.R` is written that way, and `cli` joins such a continuation without inserting the newline or the leading indentation. The rendered abort is a single line:

```
! `simulationTime` must be a length-3 numeric vector `c(start, end, resolution)`, or the same grid as a string "0, 42, 48" (several intervals separated by ";"). To give a different grid per id, pass a list with one element per id.
```

Collapsing the literal onto one source line would put it well past the file's width and make this one entry the odd one out. Happy to change it if the house rule is actually the other way round.

R/scenario-from-pkml.R
```diff
@@ -53,14 +53,15 @@
...
-#'   scenarios) or a vector with the same length as `pkmlFilePaths`.
+#' @param simulationTime Optional simulation time to use for scenarios. One
+#'   interval is a length-3 numeric vector `c(start, end, resolution)` or the
+#'   same triplet as a string `"start, end, resolution"`, where `resolution` is
+#'   the number of simulated points per time unit defined in the
+#'   `simulationTimeUnit`; several intervals go in one string separated by
+#'   `';'`. If `NULL` (default), simulation time will be extracted from the PKML
+#'   files' output schema intervals. Can be a single value (recycled for all
+#'   scenarios) or, as a list or character vector, one value per entry of
+#'   `pkmlFilePaths`.
```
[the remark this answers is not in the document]
Felixmil: Fixed in 26efc180. The `@param` now says `simulationTimeUnit` is both the unit `simulationTime` is given in and, when the intervals come from the PKML output schema instead, the unit their bounds are converted to. The Rd was regenerated with `devtools::document()`, not hand-edited.

vignettes/observed-data.Rmd
```diff
@@ -53,11 +53,25 @@ project$definitions$observedData[[1]]
 
 Two strings attach to every source, and they are not always equal, so keep them distinct:
 
-- **The source id** is the remove handle that `removeObservedData()` matches. It is taken verbatim from the source (not canonicalized like other definitions): the file basename for an `excel`, `pkml`, or `script` source (for example `Aciclovir_TimeValuesData.xlsx`), or the `DataSet` name for a `programmatic` source.
+- **The source id** is the remove handle that `removeObservedData()` matches. It is taken verbatim (not canonicalized like other definitions): the declaration's own `id` field when it sets one, and otherwise the file basename for an `excel`, `pkml`, or `script` source (for example `Aciclovir_TimeValuesData.xlsx`), or the `DataSet` name for a `programmatic` source.
 - **The DataSet name** is the reference handle that a DataCombined uses and `getObservedDataNames()` lists. It comes from the data itself: the Excel importer expands a sheet into a long, fully qualified name; a `pkml` or `script` source uses the `DataSet$name`.
```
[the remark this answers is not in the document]
Felixmil: Fixed in 26efc180. The "Removing a source" paragraph now leads with the declaration's own `id` and falls back to the file basename or the DataSet name, matching the ids section above it.

R/definition-files.R
Felixmil: Both holes closed in 515e2b85, with an observedData-specific guard rather than `.keyedTreeRecordId()` itself, because this section's id is only *sometimes* stored and the shared helper insists on a stored one.

`.assertObservedDataTreeRecordId()` runs per record in `.parseObservedDataTree()` and, for a record that came from a tree file:

1. rejects a stored `id` that is not a single non-empty string, so a JSON number, bool, or array fails at the file that holds it instead of later in `basename()` or a `character(1)` `vapply()`;
2. compares the *effective* id (declared, else derived from `file` / `name`) with the filename stem and aborts naming both, so `foo.json` holding `"id": "bar"` no longer loads only to be re-filed and reconciled away by the next save.

A record with no derivable id at all still passes: reporting an under-specified source stays `validateProject()`'s job. Inline snapshot records have no filename to check against and are skipped. The abort is attributed to the guard itself, matching how `.keyedTreeRecordId()` reads in the existing snapshots. Three tests added in `test-definition-files.R` (mismatch, non-string id, and a declared id that does match).

R/observed-data.R
Felixmil: Fixed in 515e2b85, by refusing the replacement rather than by cleaning up after it. Deleting the displaced `DataSet` would still silently discard session-only data the user never asked to lose, so the config branch now aborts when the entry it would replace is a `programmatic` sentinel, mirroring the refusal that already guarded the opposite direction:

```
observedData entry with id "obs2" is a programmatic source holding a <DataSet> in this session,
so it cannot be overwritten with a configuration entry.
i Remove it first with `removeObservedData()`, then add the configuration entry.
```

Covered by a new test in `test-observed-data.R` that also asserts the `DataSet` still resolves afterwards. The `overwrite` roxygen documents the symmetry (26efc180).

R/observed-data.R
Felixmil: Fixed in 515e2b85, but scoped to *declared* ids rather than every derived one, because the wider check broke a documented behaviour.

The `DataSet` branch now clashes against `.observedDataDeclaredIds()` (the `id` fields the declarations state outright) as well as `existingNames`, which covers your case exactly: a config entry carrying `id = "obs"` is now seen by `addObservedData(project, dataSet)` when `dataSet$name == "obs"`.

Checking *derived* ids too was my first attempt, and the suite caught it: `test-observed-data.R` locks in that a programmatic name equal to an existing file basename saves cleanly, because the sentinel is rewritten to `<name>.pkml` at save and the two never share a definition file. A derived id is transient in a way a declared one is not, so only the declared ones belong in this check.

R/observed-data.R
Felixmil: Fixed in 515e2b85. `.observedDataEntryId()` and `.validateObservedDataId()` both take `call = rlang::caller_env()` now, and `.addObservedData_impl()` passes its `.call` to both, so the unsafe-id abort reads:

```
Error in `addObservedData()`: observedData id "../escape" is not a single safe filename segment.
```

That covers the pre-existing `DataSet`-branch call site you noted as well, since it goes through the same helper.

PR #1188, opened 2026-07-28
Excel import: four migration blockers (stray sheets, out-of-reach data, blank PI cells, blank rows)

### LINKED ISSUE #1181: Excel import: a non-parameter sheet in a parameter workbook aborts the import with a bare R error
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). A parameter workbook that carries a sheet without the standard parameter columns aborts the whole import with a bare R error that names neither the file nor the sheet.

`.parseExcelParameterSheets()` treats every sheet in a parameter workbook as a parameter sheet. Real legacy projects routinely keep a notes, scratch, organ-list, or fit-bounds sheet alongside the parameter sheets, and any one of them stops the migration dead.

## Error

```
Error in if (is.na(df[["Units"]][[i]]) || df[["Units"]][[i]] == "") :
  missing value where TRUE/FALSE needed
```

## Root cause

`R/project-excel.R:1573`:

```r
units = if (is.na(df[["Units"]][[i]]) || df[["Units"]][[i]] == "") {
```

A sheet with no `Units` column makes `df[["Units"]]` `NULL`, so `df[["Units"]][[i]]` is `NULL`, `is.na(NULL)` is `logical(0)`, the whole condition evaluates to `NA`, and `if (NA)` aborts.

## Reproducible example

Uses the bundled example project only.

```r
base <- file.path(tempdir(), "straysheet")
dir.create(base)
initProject(base, type = "example", createExcel = TRUE)

mp <- file.path(base, "Configurations", "ModelParameters.xlsx")
sheets <- lapply(readxl::excel_sheets(mp), function(s) readxl::read_excel(mp, sheet = s))
names(sheets) <- readxl::excel_sheets(mp)

# A notes sheet of the kind legacy projects keep beside their parameter sheets.
sheets[["Organ notes"]] <- data.frame(Organ = c("Liver", "Kidney"), Comment = c("see ref", "TBD"))
writexl::write_xlsx(sheets, mp)

importProjectFromExcel(file.path(base, "Project.xlsx"),
                       outputDir = file.path(base, "imported"), overwrite = TRUE)
#> Error: missing value where TRUE/FALSE needed
```

## Impact

Across 19 legacy projects tested, this blocked 3 migrations outright: 1 Excel import and both of the 2 snapshot auto-upgrade failures, since `restoreProject()` runs the same bridge. It is the single largest cause of "this project cannot be migrated at all" in the set.

A related sheet in the same family reaches the good message instead, because there the `Value` column existed and carried text:

```
Cannot interpret the Value cell as a number.
Sheet "RefConc_fit", row 1: "lower".
```

## Suggested fix

Two parts, the first sufficient to unblock:

1. Guard the `Units` branch and attribute the failure to file and sheet. Line 1567 immediately above already passes `sheet =` and `row =` into `.parseNumericCell()` and produces exactly that attributed message, so the pattern is present in the same expression and simply is not used one field over.
2. Decide what a non-parameter sheet in a parameter workbook means. Skipping it with a warning naming the sheet would migrate these projects unattended; aborting with a clear message would at least tell the user which sheet to remove. Either beats the current bare error.

Part of #1088.

### LINKED ISSUE #1182: Excel import: an absolute `dataFolder` aborts the import instead of skipping observed data
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). A legacy project whose `dataFolder` is an absolute path cannot be imported at all: the containment check aborts the whole import rather than skipping the observed data.

## Error

```
✖ dataFolder "<absolute path to a synced-drive folder>" resolves outside the project folder.
ℹ It must stay under '<project root>'. A project file cannot reference a path outside the project.
```

## Why this matters

An absolute `dataFolder` pointing at a synced drive is common in real 5.x projects, because the observed data was shared through the drive rather than committed. The containment rule itself is right, but applying it as a hard abort at import time means the project cannot be migrated without hand-editing the source workbook first, which is exactly the manual step the bridge exists to remove.

The importer already has the gentler behaviour one line of behaviour away. When the configured data file is merely missing it warns and continues:

```
! The configured data file '<name>' was not found, so no observed data was imported.
ℹ Any plot or parameter-identification mapping that references observed data will not resolve until the data file is present.
```

An out-of-project `dataFolder` is the same situation from the user's point of view: the data is not available to the new project. #1179 already settled the analogous question for asset copying, where a folder given as an absolute path is deliberately not copied and is reported rather than treated as fatal.

## Suggested fix

Downgrade an out-of-project `dataFolder` from abort to warn-and-skip, leaving observed data unimported and saying so, consistent with the missing-data-file path and with the #1179 asset-copy policy. Keep the abort for a path that would be *written* to.

Minor, but worth folding in: the message echoes the absolute path verbatim, which in practice embeds a user account name.

Part of #1088.

### LINKED ISSUE #1190: Snapshot restore: .pi5xUniqueId() aborts on an empty OutputPath cell (same bare error as #1181, second site)
Found while re-testing the 19 legacy projects against PR #1188 (build `ed8a1b77998e726486d7f3c2d2dc1aa39f4e7e70`). One legacy snapshot still fails to restore, with the same bare error #1181 fixed, but raised from a different function.

```
Error in if (is.na(base) || base == "") : missing value where TRUE/FALSE needed
Calls: ... -> .parseExcelPI5xMappings -> .pi5xUniqueId
```

## Root cause

`.pi5xUniqueId()`, reached from `.parseExcelPI5xMappings()` while parsing a 5.x parameter-identification output-mapping sheet:

```r
id <- .pi5xUniqueId(sub(".*\\|", "", outputPath), ids)
```

When a mapping row's `OutputPath` cell is empty, `outputPath` is `NA`/empty, the derived `base` is not a usable string, and `is.na(base) || base == ""` evaluates to `NA`, so `if (NA)` aborts. Exactly the `if (is.na(x) || x == "")` shape as `R/project-excel.R:1573`, at a second site that #1188 does not touch.

## Why this is worth filing separately

I previously attributed both snapshot-restore failures to #1181, because both surfaced the identical message. That was wrong, and re-testing against the PR build shows it: one snapshot now restores cleanly, the other fails unchanged. Same symptom, different site.

Worth a grep for the remaining instances of this idiom rather than fixing only this one, since the pattern has now produced two independent hard failures and gives no clue where it came from either time.

## Suggested fix

Guard the empty case in `.pi5xUniqueId()` and attribute the failure to the PI task, sheet and row, so an unusable mapping row names itself. A mapping row with no `OutputPath` has nothing to map, so skipping it with a warning is likely the right behaviour, consistent with the direction #1188 took.

Part of #1088.

### LINKED ISSUE #1191: Excel import: trailing blank rows are treated as definitions and abort the import
Found while re-testing the 19 legacy projects against PR #1188 (build `ed8a1b77998e726486d7f3c2d2dc1aa39f4e7e70`). With the absolute-`dataFolder` abort fixed (#1182), one project reveals a second blocker underneath and still cannot be imported:

```
A definition file for kind population has no usable populationId.
✖ populationId must be a single non-empty string (it names the definition and its file).
```

## Root cause

The `Demographics` sheet of that project's populations workbook carries 6 real population rows followed by 18 entirely blank rows. `readxl` returns them, because stray cell formatting extends the sheet's used range past the last real row, so `PopulationName` is `NA` for rows 7 to 24:

```
rows: 24
rows with blank/NA PopulationName: 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24
```

The importer treats each of those as a population definition and aborts on the first one.

Trailing blank rows are an ordinary artefact of an Excel file that has been edited over time. They carry no information and no user would consider them part of the project.

## Scope

Of the 10 legacy projects tested that have a `Demographics` sheet, 1 has this problem (18 blank rows) and 9 have none, so it is narrower than the other import blockers. It is still a hard failure for that project, the fix is small, and the hazard is generic enough to recur in any workbook.

## Suggested fix

Skip all-blank rows when parsing a sheet into definitions, rather than treating them as records with missing required fields. Worth applying at the shared row-parsing seam rather than in the populations parser alone, since any sheet can pick up trailing blank rows the same way.

Part of #1088.

Felixmil:
Four migration blockers found while testing real 5.x projects against the 6.0.0 JSON model, all in the Excel bridge, all aborting an import or restore that has nothing else wrong with it. #1190 and #1191 were found by re-testing the same legacy projects against this PR's own first build, each surfacing once the blocker above it was cleared. Each fix is its own commit.

## A non-parameter sheet no longer aborts the import

`.parseExcelParameterSheets()` treated every sheet of a parameter workbook as a parameter sheet, and real projects routinely keep a notes, scratch, organ-list or fit-bounds sheet beside the real ones. Such a sheet made `df[["Units"]]` `NULL`, so `if (NA)` aborted with a bare R error naming neither the file nor the sheet.

The check is at the sheet, not the cell. Guarding only the `Units` branch would have been worse than the abort: the other three columns fail silently (`as.character(NULL)` is a zero-length vector, not an error), so the row loop would have run to completion and written parameter entries with empty paths and no value.

- a sheet is a parameter sheet when it has `Container Path`, `Parameter Name`, `Value` and `Units`, the same four `readParametersFromXLS()` requires and `exportProjectToExcel()` writes, now one shared list rather than a copy on each side
- one that lacks them is skipped, with one warning per workbook naming the file and the sheets
- a skipped sheet contributes nothing else: it does not become an `Application` in the 5.x one-sheet-per-protocol layout, it is not linked as an individual's own override, and a `ParameterSets` / `Individual Parameter Sets` cell naming it has that reference dropped
- a header-only sheet carrying the four columns is still an empty parameter set, so the round trip of an empty set is unchanged

## An out-of-project `dataFolder` skips observed data instead of aborting

An absolute `dataFolder` on a synced drive is ordinary in 5.x projects, because the observed data was shared through the drive rather than committed. The containment rule is right, but enforcing it as a hard abort at import time meant the project could not be migrated without hand-editing the source workbook first, which is the manual step the bridge exists to remove.

- an out-of-project `dataFolder`, or a `dataFile` that climbs out of it, now warns and imports no observed data, matching what already happens when the configured data file is simply missing and the #1179 policy for asset copying
- the two boundaries get their own message: a `dataFile` escaping `dataFolder` is usually still inside the project, where "outside the project folder" would be untrue and its remedy a no-op
- neither message echoes the path, which is absolute in the `dataFolder` case and in practice carries the user's account name
- the value stays in `filePaths` exactly as the workbook spells it, and `print(project)` now shows a folder the loader refuses rather than aborting on it, so the migrated project can still be inspected. Reading it through `project$paths$dataFolder` still refuses, so the containment guard is unchanged
- a `${VAR}` `dataFolder`, the supported way to keep data outside the project, was not expanded here at all: it was read as a literal folder name, so the data vanished behind a misleading "file not found". It is now expanded, and a relative expansion is anchored to the project file at import just as the loader anchors it, with the folder copied along so both find the same place

## A blank cell in a 5.x parameter-identification sheet

`if (is.na(x) || x == "")` is only safe on a cell that exists. On an absent column it evaluates to `NA` and `if (NA)` aborts with `missing value where TRUE/FALSE needed`, naming nothing. `.pi5xUniqueId()` was the site that fired, on a snapshot whose `OutputPath` cell was blank; `.pi5xPath()` and `.pi5xOptionRows()` carried the same latent crash waiting on a different blank cell. All four now share one `.isBlankCell()` test, plus a `.cellValue()` read that yields `NA` for a column the sheet does not have instead of aborting on the index. An unusable cell yields an absent field and a coined mapping id, which validation reports.

## Trailing blank rows are no longer definitions

Stray cell formatting extends a sheet's used range past its last real row, so `readxl` returns trailing all-`NA` records. One project's populations sheet had 6 real rows and 18 empty ones, and the import aborted on the first empty one for having no `populationId`. `readExcel()` now drops rows blank in every column, at the single point every sheet is read, so it covers every parser rather than the populations one alone. A row with a value in any column is untouched, and a header-only sheet is unchanged.

## Containment: one predicate, and `$PATH`

The rule "a path naming an environment variable may point outside the project" was the same regex written out at each of the three checks that apply it. It is now `.declaresEnvVarPath()` in `R/file-utils.R`, sharing its pattern with the `.replaceEnvVarPath()` expander.

Consolidating it surfaced a real bypass: `.replaceEnvVarPath()` deliberately never expands `$PATH`, so a value like `$PATH/../../etc` earned the exemption while staying literal, and then resolved two levels above the root with no check having run. The predicate now grants the exemption only to a reference that will actually be expanded.

## Example

One project carrying the first two problems: a notes sheet beside the parameter sheets, and observed data on a synced drive.

Before:

``` r
devtools::load_all("esqlabsR", quiet = TRUE) # at 00e3df73, this PR's base

# A legacy project with the two things real 5.x projects carry: a notes
# sheet beside the parameter sheets, and observed data on a synced drive.
base <- file.path(tempdir(), "legacy")
dir.create(base)
initProject(base, type = "example", createExcel = TRUE)

params <- file.path(base, "Configurations", "ModelParameters.xlsx")
sheetNames <- readxl::excel_sheets(params)
sheets <- lapply(sheetNames, function(s) readxl::read_excel(params, s))
names(sheets) <- sheetNames
sheets[["Organ notes"]] <- data.frame(
  Organ = c("Liver", "Kidney"),
  Comment = c("see ref", "TBD")
)
writexl::write_xlsx(sheets, params)

synced <- file.path(tempdir(), "SyncedDrive")
dir.create(synced)
file.copy(list.files(file.path(base, "Data"), full.names = TRUE), synced)
#> [1] TRUE TRUE TRUE
config <- file.path(base, "Project.xlsx")
props <- readxl::read_excel(config)
props$Value[props$Property == "dataFolder"] <- synced
props <- rbind(props, data.frame(
  Property = "dataFile",
  Value = "Aciclovir_TimeValuesData.xlsx",
  Description = NA_character_
))
writexl::write_xlsx(props, config)

importProjectFromExcel(
  config,
  outputDir = file.path(base, "imported"),
  overwrite = TRUE,
  silent = TRUE
)
#> Error in `if (is.na(df[["Units"]][[i]]) || df[["Units"]][[i]] == "") ...`:
#> ! missing value where TRUE/FALSE needed
```

After:

``` r
devtools::load_all("esqlabsR", quiet = TRUE)

# A legacy project with the two things real 5.x projects carry: a notes
# sheet beside the parameter sheets, and observed data on a synced drive.
base <- file.path(tempdir(), "legacy")
dir.create(base)
initProject(base, type = "example", createExcel = TRUE)

params <- file.path(base, "Configurations", "ModelParameters.xlsx")
sheetNames <- readxl::excel_sheets(params)
sheets <- lapply(sheetNames, function(s) readxl::read_excel(params, s))
names(sheets) <- sheetNames
sheets[["Organ notes"]] <- data.frame(
  Organ = c("Liver", "Kidney"),
  Comment = c("see ref", "TBD")
)
writexl::write_xlsx(sheets, params)

synced <- file.path(tempdir(), "SyncedDrive")
dir.create(synced)
file.copy(list.files(file.path(base, "Data"), full.names = TRUE), synced)
#> [1] TRUE TRUE TRUE
config <- file.path(base, "Project.xlsx")
props <- readxl::read_excel(config)
props$Value[props$Property == "dataFolder"] <- synced
props <- rbind(props, data.frame(
  Property = "dataFile",
  Value = "Aciclovir_TimeValuesData.xlsx",
  Description = NA_character_
))
writexl::write_xlsx(props, config)

importProjectFromExcel(
  config,
  outputDir = file.path(base, "imported"),
  overwrite = TRUE,
  silent = TRUE
)
#> Warning: ! Skipped sheet "Organ notes" in
#>   '/private/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpDG4LPu/legacy/Configurations/ModelParameters.xlsx':
#>   not a parameter sheet.
#> ℹ A parameter sheet carries the columns Container Path, Parameter Name, Value,
#>   and Units.
#> Warning: ! dataFolder points outside the project folder, so no observed data was
#>   imported.
#> ℹ Data kept outside the project (a synced drive shared between projects) is
#>   named with a `${VAR}` environment variable, which resolves from wherever the
#>   project is opened. Set one, or copy the data under the project folder, then
#>   import again.
#> ℹ Any plot or parameter-identification mapping that references observed data
#>   will not resolve until then.
```

</details>

Closes #1181
Closes #1182
Closes #1190
Closes #1191


R/messages.R
```diff
@@ -243,6 +243,40 @@ messages$importSkippedObservedData <- function(dataFile) {
...
+
+messages$importSkippedNonParameterSheets <- function(
+  filePath,
+  sheets,
+  columns
+) {
+  # Interpolated here rather than left for `cli_warn` to glue, because the
+  # values live in this function's frame and not in the caller's.
+  c(
+    "!" = cli::format_inline(paste0(
```
Felixmil: **A sheet name containing `{}` aborts the import — the exact failure #1181 removes.**

These two bullets are pre-rendered with `cli::format_inline()`, and the result is handed to `cli::cli_warn()` at `R/project-excel.R:1631`. `cli_warn()` glue-parses each element of the vector it receives, so the *already-rendered* text goes through a second glue pass. The rendered text still contains the sheet name's literal `{`/`}` characters, so those get evaluated as R expressions.

A workbook with a sheet named `Notes {draft}` — the kind of scratch sheet this PR exists to tolerate — makes the second pass evaluate `draft`, and the resulting `object 'draft' not found` propagates out of `cli_warn` and aborts the whole import. Sheet names are free text, so `{`/`}` in them is entirely reachable: `Fit {old}`, `PK {2019}`, `Params {v2}`.

This isn't a hypothetical — the repo already documents it twice, in both directions:

- `R/definition-id.R:104-113` states it outright: *"Pre-rendering each bullet with `cli::format_inline()` and handing the rendered strings to a second `cli_warn()` call is not actually safe: the rendered text still contains the value's literal `{`/`}` characters, and cli glue-parses that text again."*
- The sibling message six lines above (`importSkippedOutOfProjectData`, line 246) carries a comment explaining that it deliberately *doesn't* pre-format, for this very reason.

The comment here ("Interpolated here rather than left for `cli_warn` to glue, because the values live in this function's frame and not in the caller's") identifies a real constraint, but `format_inline()` is the wrong remedy for it. `.canonicalizedIdBullets()` solves exactly this shape: bind the values in an environment and pass the still-unglued templates plus `.envir` to a single `cli_warn()`, so the templates are glue-parsed exactly once and the values are never parsed at all.

Worth noting `{.val {sheets}}` and `{.field {columns}}` are the vector-valued cases, so whatever form the fix takes needs to keep the `cli::qty()` pluralization working.

---

R/messages.R
```diff
@@ -243,6 +243,40 @@ messages$importSkippedObservedData <- function(dataFile) {
   )
 }
 
+messages$importSkippedOutOfProjectData <- function(fieldName) {
+  # Left for `cli_warn` to glue in `.skipOutOfProjectObservedData()`, whose
+  # `fieldName` argument is this one: pre-formatting here would put a literal
+  # `${VAR}` in the result, which the emitting `cli_warn` would then read as an
+  # expression to evaluate.
+  c(
+    "!" = "{.field {fieldName}} points outside the project folder, so no \\
```
Felixmil: **This message is factually wrong for the `dataFile` caller, and its remedy can't fix that case.**

One message serves both `.skipOutOfProjectObservedData()` call sites, but the two conditions are different:

- `R/project-excel.R:1849` — `.pathEscapesRoot(dataFolderRaw, pcDir)`, i.e. the folder escapes the *project*. "points outside the project folder" is accurate here.
- `R/project-excel.R:1854` — `.pathEscapesRoot(dataFile, dataFolder)`, i.e. the file escapes the *data folder*, which is a different and much smaller boundary.

In the second case the path is very often still inside the project. With `dataFolder = "Data"` and `dataFile = "../Values.xlsx"`, the file resolves to `<project>/Values.xlsx` — comfortably within the project root — yet the user is told it "points outside the project folder."

The follow-up bullet compounds it: *"Copy the data under the project folder […] then import again"* describes a state that already holds, so acting on the advice changes nothing and the import silently skips the data again. The `${VAR}` suggestion is also inapplicable — `dataFile` is never env-var-expanded on this path, only `dataFolder` is (line 1832).

Since `.skipOutOfProjectObservedData()` already takes `fieldName`, the boundary could be named per-caller ("points outside the data folder" vs. "outside the project folder") along with the matching remedy — move the file under `dataFolder`, rather than copy it under the project.

---

R/project-excel.R
```diff
@@ -1750,8 +1829,34 @@ projectStatus <- function(project, silent = FALSE) {
...
-  dataFilePath <- .resolveProjectPath(dataFile, dataFolder, "dataFile")
+  dataFolder <- if (.declaresEnvVarPath(dataFolderRaw)) {
+    # A `${VAR}` is the sanctioned way to keep the data outside the project (a
+    # synced drive shared between projects), so it is expanded and exempt from
+    # containment, the same order `configurationsFolder` above uses.
+    expanded <- .replaceEnvVarPath(dataFolderRaw)
+    if (fs::is_absolute_path(expanded)) {
+      expanded
+    } else {
+      file.path(pcDir, expanded)
```
Felixmil: **A `${VAR}` that expands to a relative path is anchored to a different base at import than at load.** (Narrower than the other findings — flagging as plausible rather than certain.)

Here a relative expansion is joined onto `pcDir`, the *source* project-configuration directory. At load time the same raw value goes through `.resolveWorkingFolder()` (`R/project.R:1305-1332`), which resolves it against `projectDir` — the *output* directory. When `outputDir != pcDir`, which is the normal case for an import, the two disagree and the entry resolves somewhere the data isn't.

The absolute-expansion branch is fine; it's specifically the `else` here that picks up a base.

What makes it silent rather than merely wrong: `.copyExcelProjectAssets()` skips env-var folders at `R/project-excel.R:1400` with a bare `next`, so unlike the `.pathEscapesRoot()` branch just below it (which appends to `result$notCopied`), nothing is added to `notCopied` and the skip is never reported. The observed-data entry is written into the imported project pointing at a folder that was neither copied nor reachable, and the user is told nothing.

The comment says this follows "the same order `configurationsFolder` above uses" — worth checking that claim, since `configurationsFolder` at line 125 exempts the env-var case from containment but I don't think it does this `pcDir` join.

If the intent is that a relative `${VAR}` is anchored to the project, then import and load need to agree on *which* project; if the intent is that `${VAR}` always names an out-of-project location, the relative case is arguably worth rejecting outright rather than silently anchoring it.

---

Felixmil (commented):
Reviewed the three commits (`37e26ff`, `24c3c23`, `ed8a1b7`) against base `00e3df7`.

The shape of both fixes is right. Checking at the sheet rather than the cell is the correct call, and the PR description's reasoning for it is sound — guarding only the `Units` branch really would have been worse than the abort, since `as.character(NULL)` is a zero-length vector rather than an error and the row loop would have written parameter entries with empty paths. The `.declaresEnvVarPath()` extraction is behaviour-preserving and actually fixes latent `logical(0)`/`NA` crashes at the two call sites that previously lacked the length and `is.character` guards.

Three inline findings, one of which is a genuine blocker:

**`R/messages.R:269` — the double-glue.** `importSkippedNonParameterSheets()` pre-renders with `cli::format_inline()` and the result is glue-parsed a second time by the emitting `cli_warn()`. A sheet named `Notes {draft}` aborts the entire import — reintroducing exactly the failure mode #1181 removes, for exactly the kind of scratch sheet this PR exists to tolerate. `R/definition-id.R:104-113` documents this hazard explicitly and `.canonicalizedIdBullets()` gives the fix. This one I'd want resolved before merge.

The other two are the `dataFile` skip reusing a message that is factually wrong for it, and a narrower question about relative `${VAR}` expansion being anchored differently at import and at load.

**Already covered elsewhere.** CodeRabbit's thread on `R/project-excel.R:1859` reaches the same conclusion I did about the retained out-of-project `dataFolder`, so I haven't reposted it. Adding one detail to that thread: it isn't only `print(project)` — `validateProject()` reads `paths$dataFolder` too, so both inspection paths abort on a project that imported "successfully." `loadProject()` survives only because it runs just `scenarios` + `crossReferences`. The abort moved one step later rather than going away, and the new test never reads the path, so it doesn't catch this. CodeRabbit's separate `$PATH` finding on `.declaresEnvVarPath()` also looks real to me and is worth a response — the regex matches `$PATH`, but `.replaceEnvVarPath()` deliberately never expands it, so the exemption is granted to a token that stays literal.

**Two things I didn't file inline:**

- `readParametersFromXLS()` (`R/parameters.R:74`) still hardcodes its own copy of the four column names rather than using the new `.parameterSheetColumns` constant (`R/project-excel.R:1565`), whose documentation claims to be the shared source of truth. They agree today; nothing stops them drifting, and the constant's doc comment would then be wrong.
- In the `ApplicationProtocols` branch (`R/project-excel.R:349-353`), an application whose `ParameterSets` column names a skipped sheet still points at a set that was never created. The rename-following logic below is guarded on `length(appended$renames) > 0L` and handles renames, not skips. The NEWS/PR claim that "no definition is left pointing at a set that was never created" holds for the individual-override path but is narrower than stated — worth either covering the applications case or softening the claim.

---

R/messages.R
Felixmil: Fixed in 399b3df4. Confirmed first: a sheet named `Notes {draft}` aborted the import with `object 'draft' not found`, exactly as described.

`importSkippedNonParameterSheets()` now returns the templates still unglued together with an environment binding `filePath` / `sheets` / `columns`, and `.parseExcelParameterSheets()` passes both to one `cli_warn()` via `.envir`. That is the `.canonicalizedIdBullets()` shape you pointed at, so each template is glue-parsed exactly once and a value is only ever reached through a variable. The `cli::qty()` pluralization is kept (it needed a second `qty()` marker before the trailing plural, which was wrong independently: it read "not a parameter sheet" for a list of several).

Regression test added: a workbook whose only sheet is `Notes {draft}` warns, quoting the name, and returns no sets.

R/messages.R
Felixmil: Fixed in 399b3df4. Verified the case you described: with `dataFolder = "Data"` and `dataFile = "../Values.xlsx"` the file resolves to `<project>/Values.xlsx`, inside the project, while the message claimed otherwise.

Rather than parameterise one wording, there are now two catalog entries, since the remedies differ as much as the boundaries do:

- `importSkippedOutOfProjectDataFolder()` keeps the project-boundary wording.
- `importSkippedOutOfProjectDataFile()` says it points outside `dataFolder`, and its remedy is to move the file under that folder or repoint `dataFolder` at the folder holding it. No `${VAR}` suggestion, since you are right that `dataFile` is never expanded on this path.

`.skipOutOfProjectObservedData()` selects on its `fieldName`. Test asserts the `dataFile` warning names the `dataFolder` boundary and does not contain "outside the project folder".

R/file-utils.R
```diff
@@ -137,6 +137,21 @@ readExcel <- function(path, sheet = NULL, ...) {
...
+# repeated at each check, so the exemption cannot come to mean different things
+# in different places.
+# @keywords internal
+# @noRd
+.declaresEnvVarPath <- function(path) {
+  is.character(path) &&
+    length(path) == 1L &&
+    !is.na(path) &&
+    grepl("\\$\\{?[A-Za-z_]", path)
+}
```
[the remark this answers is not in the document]
Felixmil: Valid, and fixed in d069bf05. Confirmed the bypass: `.declaresEnvVarPath("$PATH/../../etc")` returned `TRUE` while `.replaceEnvVarPath()` left the token literal, so the value skipped containment and then resolved two levels above the root.

Took the first of the two options. `.envVarPathPattern` and `.envVarNamesInPath()` are now shared by the expander and the predicate, and `.declaresEnvVarPath()` grants the exemption only when at least one referenced variable is not `PATH`. A path mixing the two (`${MYDATA}/$PATH`) still qualifies, since `MYDATA` really is expanded.

Tests cover which references earn the exemption and that a `$PATH`-only folder is still rejected by `.resolveProjectPath()`.

R/project-excel.R
```diff
@@ -1750,8 +1829,34 @@ projectStatus <- function(project, silent = FALSE) {
...
+  } else {
+    .resolveProjectPath(dataFolderRaw, pcDir, "dataFolder")
+  }
+  if (.pathEscapesRoot(dataFile, dataFolder)) {
+    return(.skipOutOfProjectObservedData("dataFile", jsonData))
+  }
+  dataFilePath <- .absoluteAgainstRoot(
+    dataFile,
+    as.character(fs::path_abs(dataFolder))
+  )
```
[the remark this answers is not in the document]
Felixmil: Real, and fixed in 7138347e, though not by dropping the value.

Verified each reader named in this thread and its follow-up. `print(project)` and `project$paths$dataFolder` did abort. `validateProject()` did not: the adapter passes `project$paths$dataFolder` as a lazy argument that `.validateObservedData()` only forces inside its per-entry loop, and the skip leaves that section empty, so the promise is never forced. `projectStatus()` and `saveProject()` were also unaffected.

So the gap was inspection, not validation. `.printPathsBlock()` now resolves each folder through a `tryCatch` and falls back to the raw stored value, since printing reads no files and there is nothing to guard against there. `project$paths$dataFolder` still aborts, so the containment boundary is untouched.

The value stays in `filePaths` deliberately: an out-of-project data folder is a legitimate setup (a SharePoint-synced folder is common), and the supported spelling for it is a `${VAR}`, which the warning now points at instead of treating the arrangement as a mistake. An unset `${VAR}` needs no special case, since it survives expansion untouched and prints as the variable.

Tests: an out-of-project folder prints while the getter still refuses, and an unset `${VAR}` prints as the variable.

R/project-excel.R
Felixmil: Confirmed and fixed in 863ff80c. The divergence was real: the import joined a relative expansion onto `pcDir`, the loader joins it onto the output project directory.

One detail in your note is off: `configurationsFolder` at line 125 *does* do the `pcDir` join, so the comment's claim was accurate. It just does not matter there, because that value is only used to find the workbooks during the import and is never stored and re-resolved.

Resolved by the rule that a relative path is resolved against `Project.json`, so both ends agree: `.parseExcelObservedData()` takes the output `projectDir` and anchors a relative expansion there, and `importProjectFromExcel()` settles `outputDir` before parsing observed data. An absolute expansion is unchanged.

That alone would have left the folder consistent but unreachable, so `.copyExcelProjectAssets()` was the other half: a `${VAR}` folder that expands to a relative path now travels with the project, copied under the expanded name the loader will look for and reported under the raw one. It is skipped only when the expansion is absolute, which does resolve from anywhere. An unset variable matches no folder and is reported in `notCopied` rather than passed over with a bare `next`, which was the silence you flagged.

PR #1193, opened 2026-07-28
Excel import: three records the importer could not complete (unparseable Value, dataSet-less observed curve, OutputPath-less PI mappings)

### LINKED ISSUE #1183: Excel import: a missing observed-data file produces a project that fails its own validateProject()
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). When the configured observed-data file is absent, the importer warns and continues, but still writes `dataCombined` observed entries that carry no `dataSet`. The project it produces then fails its own `validateProject()`.

## What happens

The import warns, correctly:

```
! The configured data file '<name>' was not found, so no observed data was imported.
ℹ Any plot or parameter-identification mapping that references observed data will not resolve until the data file is present.
```

and then the freshly imported project reports:

```
DataCombined '<id>' has a observed entry missing required field: dataSet
```

once per affected entry. In the project where this surfaced that was 12 critical errors, so `validateProject()` on a just-imported, never-edited tree comes back invalid.

Worth stating explicitly: this is not a test artefact. That repository contains no data folder and no such file, so the absence is genuine and is exactly the situation the warning is designed for.

## Why this matters

The warning tells the user the observed data did not come along, which is fine and expected. What is not fine is that the importer then emits half-populated `dataCombined` records describing observed data it knows it does not have, so the very next call a user makes reports the project as broken. The import is internally inconsistent with its own warning.

## Suggested fix

When observed data was not imported, either drop the observed half of the affected `dataCombined` entries (keeping the simulated half, which is valid), or keep them and have the import warning say plainly that the resulting project will not validate until the data file is placed. The first keeps "a fresh import validates clean" true, which is the more useful invariant.

Part of #1088, and belongs with the observed-data contract in #1131.

### LINKED ISSUE #1189: Excel import: a non-numeric Value cell in a parameter sheet aborts the whole import
Found while re-testing the 19 legacy projects against PR #1188 (build `ed8a1b77998e726486d7f3c2d2dc1aa39f4e7e70`). One project still cannot be imported, and #1188 deliberately does not cover this case.

A genuine parameter sheet, one that carries all four required columns, holds fit bounds rather than numbers in its `Value` column:

| Container Path | Parameter Name | Value | Units |
|---|---|---|---|
| Target | Reference concentration | lower | nmol/l |

The import aborts:

```
Cannot interpret the Value cell as a number.
✖ Sheet "RefConc_fit", row 1: "lower".
ℹ A blank cell is allowed; a non-blank cell must be numeric (use "." as the decimal separator).
```

## Why #1188 does not cover it

#1188 classifies a sheet as a parameter sheet by the presence of `Container Path`, `Parameter Name`, `Value` and `Units`. This sheet has all four, so it is correctly classified as a parameter sheet and its rows are parsed, and then `.parseNumericCell()` rejects `"lower"`. Verified against the PR build: the abort is unchanged.

The message itself is good, precise and attributed. The problem is the consequence: one unparseable cell in one sheet stops the migration of an otherwise healthy project, and the sheet in question is a working note about a fitting range rather than model parametrization.

## Suggested fix

Decide what an unparseable `Value` means at the row level rather than the import level. Skipping the offending row with a warning that names the sheet, row and cell, and continuing, would migrate the project while losing nothing real, and matches the direction #1188 took for sheets. Aborting is defensible for a value that was *meant* to be a number, but the import currently cannot tell the difference and pays the maximum price either way.

Note the interaction with #1188: as the sheet predicate becomes stricter, sheets like this one fall on the "real parameter sheet" side precisely because they were authored by copying a real one. So the row-level decision is the one that actually governs whether these projects migrate.

Part of #1088.

### LINKED ISSUE #1192: Snapshot restore: an older PIOutputMappings layout without an OutputPath column cannot be imported
Found re-testing all 19 legacy projects against the merged `json-based-project` HEAD (`132297c8eb728383164109a656b2dc51fe950a8d`, that is with #1188 merged). This is the last snapshot that still fails to restore, and it is a different defect from #1190, which #1188 fixed.

```
Required field "outputPath" is missing or empty on PIOutputMapping "item".
```

## Root cause: an older `PIOutputMappings` schema, not a blank cell

#1190 was about a blank *cell* in an otherwise-present column. Here the column does not exist. Comparing the two 5.x layouts found in the tested estate:

| Layout | `PIOutputMappings` columns |
|---|---|
| imports fine | `PITaskName, Scenarios, **OutputPath**, ObservedDataSheet, DataSet, Scaling, xOffset, yOffset, xFactor, yFactor, Weight` |
| still blocked | `PITaskName, Scenarios, ObservedDataSheet, DataSet, Scaling, xOffset, yOffset, xFactor, yFactor, Weight` |

The parser reads `row[["OutputPath"]]`, which is `NULL` for every row of the second layout, so every mapping in the workbook is incomplete and the restore stops. All 4 mapping rows are affected.

`PIParameters` differs between the same two layouts as well, the blocked one carrying an extra `Value` column, so this looks like a genuine older revision of the parameter-identification workbook rather than a one-off hand edit.

After #1190 the failure is at least attributed and readable, which is the improvement that fix delivered. The remaining problem is that the layout cannot be imported at all.

## Open question for whoever picks this up

Whether the output path is recoverable from this layout, or genuinely absent from it. `ObservedDataSheet` sits where `OutputPath` sits in the newer layout, which may mean the column was renamed rather than added, or that this revision identified the output differently. Worth checking against an esqlabsR release that wrote this variant before deciding whether to map the old column, derive the path elsewhere, or skip such mappings with a warning.

If the path is genuinely not present, skipping the incomplete mappings with a warning naming the task and sheet would let the rest of the project migrate, which matters because parameter identification is one section of an otherwise healthy project.

Part of #1088.

Felixmil:
Migration testing of the 6.0.0 JSON model against 19 legacy 5.x projects left the same question asked in three places: what should the Excel importer do when a record cannot be completed from the source workbook? It answered three different ways, so an unusable parameter row aborted the whole import (#1189), an observed curve with no data set produced a project that failed its own `validateProject()` with nothing explaining why (#1183), and a `PIOutputMappings` sheet from before the `OutputPath` column existed could not be restored at all (#1192). #1188 already settled this for whole sheets ("skip a non-parameter sheet instead of aborting"); this pushes the decision down to rows and records, behind one shared predicate and one shared warning emitter. #1189 and #1192 were the last two hard blockers in the tested estate.

The policy that came out of it: a row that names nothing real is dropped, and a record that is merely incomplete is kept and reported. A parameter row whose `Value` is text describes no parameter, so it goes. An observed curve or a PI mapping with a field missing still describes something the user can finish, so it loads and `validateProject()` names it.

## Shared pieces

- `.isUnusableNumericCell()` states the numeric-cell rule once (a blank cell is allowed; a non-blank cell must be a number), next to the existing `.isBlankCell()`.
- `.warnFormatted()` owns the unglued `bullets`/`envir` emit that every warning naming modeller-written text needs, so a sheet name or definition id containing `{` or `}` survives. `.parseExcelParameterSheets()`'s existing warning now goes through it too.

## A non-numeric parameter Value (#1189)

A fit-bounds sheet authored by copying a real parameter sheet carries all four parameter columns, so #1188's sheet-level test correctly classifies it as a parameter sheet and its rows are read. A `Value` of `lower` then stopped the migration of a project with nothing else wrong with it. Such a row is now skipped, with one warning per workbook naming each skipped row's sheet, row and cell, so a genuine typo in a real parameter is still visible. A blank `Value` is unchanged: still an allowed absent value, not a skipped row. `.parseNumericCell()` is removed, its abort having been its only reason to exist.

The reported row is the workbook row. Blank rows are dropped when a sheet is read, so a parsed-frame index is off by the header plus every blank row above the offender, which on a workbook edited over years is routinely several. `.dropBlankRows()` is the one place rows disappear, so it now records the kept rows' original positions in a `sheetRow` attribute for a parser that reports a row to the user. It is an attribute rather than a column so nothing looping over `names(df)` sees it.

A sheet that had rows but kept none describes no parameter at all, so it becomes no definition rather than an empty parameter set (and, in the 5.x applications layout, an `Application` wrapping one). Both callers derive the sheets they may link from the names of what the parser returns, so omitting it also unlinks references to it. A header-only sheet is a different thing, a real set that happens to be empty, and is unaffected.

## An observed curve with no data set (#1183)

Worth recording, since it changes what the fix is: there is no code path linking a missing observed-data file to an absent `dataSet`. `.parseExcelObservedData()` returns early and leaves `dataCombined` untouched, and an observed entry's `dataSet` comes only from the workbook cell. So the reported critical errors mean that project's `dataSet` cells are blank, and rewording the missing-file warning alone would not have addressed them.

Nothing is dropped, since a user may still be able to supply the missing piece. Instead the import names the affected data combinations and states the consequence plainly: `validateProject()` reports each one as a critical error until the cell is filled and the project imported again, or the curve completed with `addDataCombined()`. The check runs after id canonicalization, so the ids it names are the ones the tree and validation use. The three warnings about observed data that could not be imported at all (a missing data file, a `dataFolder` outside the project, a `dataFile` outside its `dataFolder`) gained the same consequence, since "will not resolve" understated a project that reports critical errors.

## A PIOutputMappings sheet with no OutputPath column (#1192)

The issue asked whether the path was recoverable. It is, and `ObservedDataSheet` is not a rename of `OutputPath`. Two commits inside PR #928 bracket the two layouts exactly as the issue describes them: `fix: add OutputPath column to PIOutputMappings` and `fix: remove unused Value column from PIParameters sheet`. The blocked workbook was therefore authored against a pre-release build of the 5.x PI workflow, before either landed. In the code before that commit, `runPI()` took the output paths from the referenced scenario (`scenarioConfiguration$outputPaths`, the `OutputPathsIds` cell of `Scenarios.xlsx`) and built one mapping per output path, all sharing the row's `DataSet`.

The importer now reproduces exactly that. The column's absence is read as the older layout, and each row becomes one mapping per output path the scenarios it names declare, carrying the row's data set, scaling, offsets, factors and weight; a path two scenarios share becomes one mapping naming both rather than two mappings. The derived value is already an output-path id, which is what the JSON model wants. Scenario names are matched on the canonical id, as every other cross-sheet reference in the import is, since two hand-maintained sheets routinely differ in case or spacing; matching raw text would resolve nothing and then blame the scenario for having no output path. A row whose scenarios declare no output path still yields a mapping, without one, and is named in a warning saying which cell to fill. `restoreProject()` gets all of this for free, since the legacy-snapshot upgrade rebuilds the workbooks and runs them through the same importer.

Only the column's absence is read as the older layout. A blank cell in a column that IS present is an authoring gap in the newer layout and is left alone.

## Loading an incomplete PI output mapping (#1192)

Keeping an incomplete mapping was not possible before this: `PIOutputMapping()` aborts on an absent `outputPath` or `observedData`, and `importProjectFromExcel()` reaches it eagerly through `Project$new()`, so a workbook with one such blank cell could not be opened and therefore could not be fixed. The abort also arrived naming a coined mapping id, with nothing pointing back at the sheet, which is the `Required field "outputPath" is missing or empty on PIOutputMapping "item"` from the issue.

`.validateCrossReferences()` was already written for the lenient shape: it guards `outputPathId` for `NULL` and reports "does not define an outputPath", a branch the strict constructor made unreachable. So `.piOutputMappingRecord()` is split out of `PIOutputMapping()`, which keeps every guard, and `.parsePIOutputMappings()` builds through it. An incomplete mapping now loads and validation names it; an absent `observedData` is reported next to the absent `outputPath` that already was. Authoring one through `PIOutputMapping()` or `addPIOutputMapping()` is unchanged and still rejected. `runPI()` calls `project$ensureValid(sections = c(..., "crossReferences"))`, so an incomplete mapping still cannot reach a run.

The same applies to parameters. `PIParameter()` aborts identically on a blank `MinValue` / `StartValue`, which a hand-maintained 5.x `PIParameters` sheet produces, so `.piParameterRecord()` is split out the same way and `.parsePIParameters()` builds through it. An absent bound stays absent rather than becoming `NA`, so the validator can tell "no value" from "an unusable one", and the bounds comparison only runs once all three are present (`NULL <= NULL` is `logical(0)`, which `if` cannot branch on).

Both records' missing-field reports come from the `parameterIdentification` section validator rather than the cross-reference phase. That phase documents itself as skipping entirely once any section has a critical error, so a workbook with both a blank bound and a blank `OutputPath` reported only the parameter, and the mapping would have been discovered on the next run. A missing required field is not a reference to resolve, so it moved; the phase now resolves an `outputPathId` that is present and leaves its absence alone. Every gap in a task is listed in one report, and `runPI()` refuses naming all of them.

## Tests

`devtools::test()`: `[ FAIL 0 | WARN 47 | SKIP 2 | PASS 2959 ]`. The 47 warnings and 2 skips are pre-existing (id-canonicalization notices and two environment-gated tests). Note that the suite needs a UTF-8 locale; under `LC_CTYPE=C` three unit round-trip tests fail and `test-sensitivity-time-profiles.R` will not parse, both on the base commit as well.

One regression test per fix, all built on the bundled `TestProjectExcel` fixture, plus unit tests for the derivation, the canonical-id match, the lenient load against the strict constructor for both record kinds, every gap reported at once, and the two new warnings.

`devtools::check()` locally: `1 WARNING, 7 NOTEs`, unchanged from the base. The warning is undocumented `...` in `createDataCombined.Rd` and `createPlots.Rd`, neither of whose sources this branch touches. The former `.parseExcelParameterSheets` abort snapshot is replaced; the new warning names the workbook, whose path is a temp file under test, so that one asserts on class and message rather than recording a machine-specific snapshot.

## Example

The `#1189` blocker, before and after, on the bundled example project.

``` r
devtools::load_all("/private/tmp/claude-501/-Users-felix-Code-esqlabsR--claude-worktrees-migration-test-th-model-check-589822/939dbb5c-0edb-43a2-b3f6-baee844f4cf7/scratchpad/base", quiet = TRUE)
dir <- withr::local_tempdir()
initProject(dir, type = "example", createExcel = TRUE)

# A fit-bounds sheet, authored by copying a real parameter sheet: it carries
# all four parameter columns, so it IS a parameter sheet, and its Value cells
# hold the words a modeller writes when noting a fitting range.
paramFile <- file.path(dir, "Configurations", "ModelParameters.xlsx")
sheets <- sapply(
  readxl::excel_sheets(paramFile),
  \(s) readxl::read_excel(paramFile, sheet = s),
  simplify = FALSE
)
sheets[["RefConc_fit"]] <- data.frame(
  `Container Path` = "Target",
  `Parameter Name` = c("Reference concentration", "Kd"),
  Value = c("lower", "2.5"),
  Units = "nmol/l",
  check.names = FALSE
)
writexl::write_xlsx(sheets, paramFile)

importProjectFromExcel(
  file.path(dir, "Project.xlsx"),
  outputDir = withr::local_tempdir(),
  silent = TRUE
)
#> Error in `.parseNumericCell()` at base/R/project-excel.R:1653:9:
#> ! Cannot interpret the Value cell as a number.
#> ✖ Sheet "RefConc_fit", row 1: "lower".
#> ℹ A blank cell is allowed; a non-blank cell must be numeric (use "." as the
#>   decimal separator).
```

</details>

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/migration-test-th-model-check-589822", quiet = TRUE)
dir <- withr::local_tempdir()
initProject(dir, type = "example", createExcel = TRUE)

# A fit-bounds sheet, authored by copying a real parameter sheet: it carries
# all four parameter columns, so it IS a parameter sheet, and its Value cells
# hold the words a modeller writes when noting a fitting range.
paramFile <- file.path(dir, "Configurations", "ModelParameters.xlsx")
sheets <- sapply(
  readxl::excel_sheets(paramFile),
  \(s) readxl::read_excel(paramFile, sheet = s),
  simplify = FALSE
)
sheets[["RefConc_fit"]] <- data.frame(
  `Container Path` = "Target",
  `Parameter Name` = c("Reference concentration", "Kd"),
  Value = c("lower", "2.5"),
  Units = "nmol/l",
  check.names = FALSE
)
writexl::write_xlsx(sheets, paramFile)

out <- withr::local_tempdir()
jsonPath <- importProjectFromExcel(
  file.path(dir, "Project.xlsx"),
  outputDir = out,
  silent = TRUE
)
#> Warning: ! Skipped 1 row in
#>   '/private/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpEPHiIL/file5c624460a7a3/Configurations/ModelParameters.xlsx':
#>   the Value cell is not a number.
#> ✖ Sheet "RefConc_fit", row 2: "lower".
#> ℹ A blank cell is allowed; a non-blank cell must be numeric (use "." as the
#>   decimal separator).

# The sheet's real parameter survived; only the unparseable row was dropped.
loadProject(jsonPath)$definitions$parameterSets[["refconc_fit"]]
#> <ParameterSet>
#>   • Number of Entries: 1
#>   • Target|Kd = 2.5 [nmol/l]
```

</details>

Closes #1183
Closes #1189
Closes #1192

GitHub will not auto-close these three, because this PR's base is `json-based-project` rather than the default branch. They need closing by hand when it merges.


R/project-excel.R
```diff
@@ -2467,10 +2545,21 @@ projectStatus <- function(project, silent = FALSE) {
 # a per-mapping `id` from the output path's last segment (de-duplicated within
 # the task) and maps the offset/factor/weight columns.
 #
+# The sheet's oldest revision has no `OutputPath` column at all: it predates the
+# column, and identified a mapping's output through the scenario, one mapping per
+# output path the scenario declares. `.pi5xDerivedMappingRows()` reproduces that,
+# so such a workbook restores instead of failing on every mapping. A blank cell
+# in a column that IS present is a different thing (an authoring gap in the newer
+# layout) and still yields a mapping with no `outputPath` for validation to
```
Felixmil: **A blank cell in a *present* `OutputPath` column does not reach validation — it aborts the import.**

This comment (and the matching NEWS paragraph, and #1190's) says such a cell "still yields a mapping with no `outputPath` for validation to report". Tracing it:

`importProjectFromExcel()` bootstraps the imported data through `Project$new()` (`R/project-excel.R:533`), which parses `parameterIdentification` eagerly (`R/project.R:1447-1450`) → `.parsePITasks()` → `.parsePIOutputMappings()`, which hands `raw[["outputPath"]]` — `NULL`, since `.naToNull()` made it `NULL` and `.dropNulls()` dropped the field — straight to `PIOutputMapping()`:

```r
if (!is.character(outputPath) || length(outputPath) != 1L || is.na(outputPath) || nchar(outputPath) == 0) {
  cli::cli_abort(messages$PIRequiredField("outputPath", "PIOutputMapping", id))
}
```
(`R/parameter-identification.R:188-199`). Nothing in between drops an incomplete mapping: `.resolvePIOutputPathRefs()` passes a `NULL` `outputPath` through untouched.

So a 5.x workbook that *has* the column but leaves one cell blank still fails to migrate, with exactly the class of hard error this PR removes for the column-less case. Contrast the #1183 half of this PR, where the parse-leniently-then-validate contract genuinely holds — a `dataCombined` observed entry survives a missing `dataSet` and `validateProject()` reports it, as your own test asserts. PI mappings don't have that leniency.

Two knock-ons worth noting:

- A blank `DataSet` cell aborts the same way (`observedData` is required too, `R/parameter-identification.R:200-211`). The new derivation path fills `OutputPath` but never `DataSet`, so an `OutputPath`-less sheet with a blank `DataSet` cell now gets *past* the derivation and aborts immediately after it.
- This isn't introduced here — it's already true on `json-based-project`, and #1190's NEWS entry ("an unusable cell yields an absent field … that validation reports, instead of stopping the migration") makes the same claim. But since this PR is the one deciding what happens to a record that can't be completed, it seems like the place to settle it: either make `.parsePIOutputMappings()` tolerate an absent `outputPath` / `observedData` and leave the report to `.validateParameterIdentification()` (what both NEWS entries and this comment promise), or drop the claim from the comment and NEWS.

---

R/project-excel.R
```diff
@@ -2492,6 +2581,77 @@ projectStatus <- function(project, silent = FALSE) {
...
+# @param scenarios The parsed `scenarios` section (an unnamed list of records
+#   carrying `name` and `outputPaths`), or NULL when the project has none.
+# @returns A data frame with the same columns plus `OutputPath`, one row per
+#   (row, output path) pair. Zero rows when nothing could be derived.
+# @keywords internal
+# @noRd
+.pi5xDerivedMappingRows <- function(rows, task, scenarios) {
+  outputPathsOf <- function(name) {
+    for (scenario in scenarios %||% list()) {
+      if (identical(as.character(scenario$name), name)) {
```
Felixmil: **The scenario lookup is exact-match, where every other reference in the importer is canonicalized.**

Both sides of this `identical()` are raw workbook text at this point: `scenario$name` comes verbatim from the `Scenario_name` cell (`.parseExcelScenarios()`, `R/project-excel.R:1812`), and `name` comes verbatim from the row's `Scenarios` cell. `.canonicalizeProjectJsonIds()` — which lowercases and maps spaces/commas to `_` on both a definition and every reference to it, so a spelling difference resolves anyway — doesn't run until line 508, well after this parse.

Net effect: a `PIOutputMappings` row whose `Scenarios` cell reads `aciclovir_iv` against a `Scenarios` sheet spelling it `Aciclovir_IV` derives nothing, the mapping is dropped, and the warning tells the user to "give those scenarios an output path" when they already have one — the one wrong thing to tell them. Everywhere else in the same import (the task's own `scenarios`, the parameter rows' `scenarios`, a scenario's `individual` / `population` / `parameterSets`) that difference is absorbed.

`.canonicalizeOneId()` on both sides of the comparison would make this consistent with the rest of the importer:

```r
if (identical(.canonicalizeOneId(as.character(scenario$name)), .canonicalizeOneId(name))) {
```

Given that the derived `OutputPath` values get canonicalized downstream anyway, the comparison is the only place the raw spelling still matters.

---

R/project-excel.R
```diff
@@ -1650,36 +1672,48 @@ projectStatus <- function(project, silent = FALSE) {
     entries <- list()
     if (nrow(df) > 0) {
       for (i in seq_len(nrow(df))) {
+        value <- df[["Value"]][[i]]
+        if (.isUnusableNumericCell(value)) {
+          badSheets <- c(badSheets, sheet)
+          badRows <- c(badRows, i)
```
Felixmil: **The row number in the warning isn't the workbook row, and this warning is now the only record of the skip.**

`i` indexes the parsed frame, so the Excel row is `i + 1` at minimum (the header), and drifts further for every row `readExcel()`'s `.dropBlankRows()` removed above it — which, per #1191, is routinely a lot on a workbook edited over the years. `messages$importSkippedNonNumericRows()` states the intent explicitly ("a row has to be findable in the workbook to be fixed"), and unlike the abort it replaces, the row is now silently gone from the imported project if the user can't locate it.

`i + 1` is closer but still wrong once blank rows are dropped. The two honest fixes: carry the pre-`.dropBlankRows()` row index (e.g. stash it as an attribute or a column in `readExcel()`), or reword the bullet to say "data row" so the number isn't read as a spreadsheet row. Either beats a number that looks authoritative and points one or more rows off.

---

R/project-excel.R
```diff
@@ -1650,36 +1672,48 @@ projectStatus <- function(project, silent = FALSE) {
...
             NULL
           } else {
             as.character(df[["Units"]][[i]])
           }
         )
-        entries[[i]] <- entry
+        entries[[length(entries) + 1L]] <- entry
       }
     }
     result[[sheet]] <- entries
```
Felixmil, line 1710: Minor: a sheet whose *every* row is skipped still becomes a parameter set — an empty one.

`entries` is `list()`, and `result[[sheet]] <- list()` keeps the key (only `NULL` would drop it), so the sheet reaches `.appendParameterSets()` and gets a `definitions/parameter-sets/<sheet>.json` with no entries. In the 5.x `Applications.xlsx` layout it also gets an `Application` wrapping it. `.validateParameterSets()` is quiet about it (`if (length(set) == 0) next`, `R/validation.R:688-690`), so the fit-bounds sheet from the reported case migrates as a definition corresponding to nothing at all.

A zero-row sheet already behaved this way, so this isn't new machinery — but the new skip makes the case reachable for a sheet that *does* have rows, which is the case where "this sheet became a definition" is misleading. Two options, both cheap: drop the sheet when nothing survived (it is then indistinguishable from a non-parameter sheet, and `.dropSkippedSheetRefs()` already handles unlinking references to a sheet that produced no set), or mention it in the skip warning so it's clear an empty definition landed.

---

Felixmil (commented):
Reviewed the four commits against `json-based-project` (132297c). The shape of the change is right — one predicate for "this cell must be a number", one emitter for the unglued warnings, and the skip/keep decision pushed down from the sheet to the row — and the #1189 and #1183 halves do what they claim.

Four findings, most severe first, as inline comments:

1. **`R/project-excel.R:2553`** — the comment (and NEWS for #1192 and #1190) says a blank cell in a *present* `OutputPath` column yields a mapping with no `outputPath` "for validation to report". It doesn't: `PIOutputMapping()` aborts on an absent `outputPath`, and `importProjectFromExcel()` reaches it eagerly via `Project$new()`. That workbook still fails to migrate, with the same class of hard error this PR removes for the column-less case. Blank `DataSet` aborts identically, and the derivation never fills it — so an `OutputPath`-less sheet with a blank `DataSet` now gets past the derivation and dies right after.
2. **`R/project-excel.R:2606`** — the derivation's scenario lookup is `identical()` on raw workbook text, while every other reference in the same import resolves after `.canonicalizeOneId()`. A case/whitespace difference between the `Scenarios` cell and `Scenario_name` silently drops the mapping and produces a warning that misdirects.
3. **`R/project-excel.R:1678`** — the skipped-row warning's row number is the parsed-frame index, not the workbook row (off by the header, plus every blank row `.dropBlankRows()` removed). Findability is the bullet's stated purpose, and it's now the only record of the skip.
4. **`R/project-excel.R:1696`** — a sheet whose every row is skipped still becomes an (empty) parameter set definition, unremarked by validation.

Checked and clean: `.parseNumericCell()` has no other callers and its snapshot is replaced; the `entries[[length(entries) + 1L]]` change correctly closes the index hole the `next` would otherwise leave; `.warnIncompleteObservedCurves()` is NULL/zero-length-safe (via `.isBlankCell()`), reads the right field name, and runs after canonicalization as documented; `.warnFormatted()` preserves the unglued `bullets`/`envir` contract; the derived-row fan-out keeps a shared path as one mapping naming both scenarios, and the derived `OutputPath` is already an id, so `.resolvePIOutputPathRefs()` passes it through; a fully-derived-away task still builds, since `PITask()` accepts empty `outputMappings`; no snapshot or test matches the reworded observed-data messages.

One maintainability note, not a bug: inside `.parseExcelPI5x()`'s `lapply`, the local `scenarios <- unique(unlist(...))` shadows the formal that was just passed to `.parseExcelPI5xMappings()`. It resolves correctly today only because the promise is forced during that call, before the reassignment executes. Renaming one of the two would remove a trap for whoever next moves a line in that function.

---

R/project-excel.R
Felixmil: Confirmed, and settled the way you suggested rather than by dropping the claim. `PIOutputMapping()` does abort on an absent `outputPath`, `Project$new()` reaches it eagerly (`R/project.R:1448`), and nothing in between drops the mapping.

What decided it: `.validateCrossReferences()` was *already* written for the lenient shape. `R/validation.R:1186` guards `is.null(m$outputPathId) || is.na(...)` and reports "does not define an outputPath" — a branch the strict constructor made unreachable. So the contract you describe was the intended design; only the parse seam contradicted it.

`.piOutputMappingRecord()` is now split out of `PIOutputMapping()`, which keeps every guard. `.parsePIOutputMappings()` builds through it, so an incomplete mapping loads and validation names it. Authoring one through the constructor or `addPIOutputMapping()` is rejected exactly as before. The added `observedData` check sits next to the existing `outputPath` one, since only the latter had been written.

Safe end to end because `runPI()` calls `project$ensureValid(sections = c(..., "crossReferences"))`, so an incomplete mapping never reaches `project$definitions$outputPaths[[m$outputPathId]]`.

Both knock-ons handled. The blank-`DataSet` case no longer needs the skip I had added: the row loads and is reported. That also removed the reason-per-row wording, and the whole drop-with-warning branch of the derivation went with it — an underivable row is now kept as an `NA`-path mapping and warned about, which is the same keep-and-report policy as the observed curve with no data set. The comment and both NEWS entries now describe that.

One thing deliberately left: `PIParameter()` aborts the same way on a blank `MinValue`/`StartValue`, so a 5.x `PIParameters` sheet with such a cell is still a blocker. Extending the leniency there means `.validatePI()`'s bounds check (`p$minValue <= p$startValue`, `R/parameter-identification.R:588`) has to stop assuming those are numbers, which is a wider change than this thread. Not tracked yet — say if you want an issue for it.

R/project-excel.R
Felixmil: Fixed. Both sides now go through `.canonicalizeOneId()`, and the scenario names are canonicalized once outside the row loop rather than per lookup.

Your point about the warning misdirecting is the part that made this worth more than consistency: telling the user to give a scenario an output path it already has is worse than saying nothing.

Test added: a `Scenarios` cell reading `aciclovir iv` against a `Scenario_name` of `Aciclovir_IV` now derives the path. That covers whitespace and case in one case, which is the pair that actually differs between two hand-maintained sheets.

CodeRabbit raised the same finding independently.

R/project-excel.R
Felixmil: Fixed, and with the truthful row rather than the reworded bullet, since as you say findability is the whole point and the row is now the only record of the skip.

`.dropBlankRows()` is the single place rows disappear, so it records the kept rows' original positions in a `sheetRow` attribute; `.parseExcelParameterSheets()` reads it and adds 1 for the header. An attribute rather than a column, so no parser looping over `names(df)` picks it up — which also means it does not survive subsetting the frame, and that constraint is written down where it is set.

Test: a sheet with one data row, then two blank rows, then the offender, reports `row 5`. Under the old arithmetic that was `row 2`.

This also shifted the two assertions in the earlier unit test from rows 1 and 2 to rows 2 and 3.

R/project-excel.R
Felixmil: Fixed, taking the first option: a sheet that had rows but kept none is left out of the result entirely.

Your note that `.dropSkippedSheetRefs()` already handles this turned out to be more than a convenience — both callers compute what they may link from `setdiff(sheetNames, names(parsedSets))` (`R/project-excel.R:280`, and the applications path), so omitting the sheet unlinks the references and stops the 5.x applications layout wrapping it in an `Application`, with no extra wiring.

Gated on `nrow(df) > 0 && length(entries) == 0L`, so the header-only sheet stays what it is: a real set that happens to be empty, covered by the existing test.

Two tests: the all-skipped sheet leaves only the good one in `parsed`, and the header-only case is unchanged.

Issue #1148, opened 2026-07-22
v6 mutation API: policy & consistency


Felixmil:
Umbrella for the programmatic `add*`/`set*`/`remove*` authoring API on the v6 `Project` model. Part of #1088.

Tracks cross-family behavioural consistency of the mutation API (duplicate-entry handling, auto-cleanup, error vs overwrite) so the authoring surface behaves predictably regardless of which section is edited.

Felixmil:
Reopening: migration testing of the 6.0.0 JSON model surfaced new findings on this umbrella's surface, attached as sub-issues below (tested at commit 677f9c8e).

PR #1206, opened 2026-07-30
Accept the empty reference lists the importer writes, and settle the eager-versus-deferred reference policy

### LINKED ISSUE #1169: Authoring validates references eagerly while the importer defers to validateProject()
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). `addScenario()` cross-checks the referenced `individual`/`population` against existing definitions immediately and rejects a dangling reference at authoring time, whereas `importProjectFromExcel()` accepts the same scenario and leaves the gap for `validateProject()` to report.

Observed authoring aborts:

- `Cannot add scenario "<id>": individual '<indiv>' not found in individuals`
- `Cannot add scenario "<id>": population '<pop>' not found in populations`

The eager authoring error is arguably a feature, but the asymmetry is worth settling as a deliberate policy on the mutation API: the importer will produce a project that only `validateProject()` flags, while the same construction is blocked up front when hand-authoring. Either document the difference as intended, or make the two entry points agree on when a dangling reference is an error.

Root cause: `addScenario()` validates references at add time; the importer parses sections independently and defers to the cross-reference validation phase.

### LINKED ISSUE #1187: addScenario() rejects the empty outputPaths/parameterSets the importer writes
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). The definition tree stores an absent `outputPaths` or `parameterSets` as an empty array. Handing that value straight back to `addScenario()` aborts, so the authoring API cannot express a scenario the importer routinely writes.

## Errors

```
Cannot add scenario "<id>":
✖ outputPaths must be a non-empty character vector with no NA or empty entries
```

```
Cannot add scenario "<id>":
✖ parameterSets must be a non-empty character vector with no NA or empty entries
```

## Why this matters

The importer writes such scenarios happily, they validate clean, and they run. So "import a legacy project, then adjust it programmatically" breaks on the round trip: reading a scenario record out of the tree and passing its own fields back into the authoring API is rejected for a value the importer itself produced. In the project used for the end-to-end test, 6 of 8 scenarios carried an empty `outputPaths` and 1 an empty `parameterSets`.

## Measured impact

Rebuilding one imported project by hand through the `add*()` API, in dependency order, reached **7 of 8 scenarios and 1 of 3 parameter-identification tasks**. Re-expressing just those two empty lists as `NULL`, with no other change, took the same script to **8 of 8 scenarios and 3 of 3 PI tasks**, matching the imported tree on every section (scenarios, individuals, populations, applications, parameter sets, output paths, observed data, PI, plots, plot grids, data combined).

The two blocked PI tasks were downstream casualties rather than separate defects: both reference the single scenario that could not be added.

With that one change both trees also run identically: 7 of 8 scenarios, failing on the same deliberately-broken scenario, which each path reports the same way.

## Suggested fix

Treat a zero-length `outputPaths` / `parameterSets` as `NULL` (meaning "none") rather than aborting.

## Relationship to #1169

#1169 tracks the same `addScenario()` importer-versus-authoring asymmetry, but for *dangling* references (an individual or population that does not exist), and frames it as a policy question about eager versus deferred validation. This is a different failure mode: an empty list is not a dangling reference, it is a valid "there are none" state that the importer writes and the authoring API refuses. Filed separately because it has a concrete fix that does not depend on settling that policy, though both belong to the mutation-API consistency work in #1148.

Part of #1088.

Felixmil:
Two findings from 6.0.0 migration testing, both about the gap between what the importer writes and what the authoring API accepts.

## Empty reference lists (#1187)

A definition file always writes `[]` for a reference list a scenario has none of, never `null` and never an absent key. `addScenario()` and `setScenario()` refused that value, so reading a scenario's own written fields and handing them back aborted on a value the write path had just produced. `character(0)` (and the empty list a JSON reader yields for `[]`) now means "there are none", exactly as `NULL` does.

The same helper closes the populated case. The package reads a definition file with `jsonlite::fromJSON(simplifyVector = FALSE)`, so a scenario's two output paths come back as `list("a", "b")`, which the reference validators rejected as non-character; an all-strings list now flattens to the character vector the reference list is. A list holding anything else still aborts. So a scenario round-trips through its own file whether it has references or not.

`.canonicalizeVectorIdRef()` holds both rules for the three vector-valued reference fields, and the record ends up in the shape `.parseScenarios()` builds from the same file. The NA and empty-string rejections are untouched. A zero-length *scalar* reference (`individual = character(0)`) still aborts: no write path produces one, and it is far more likely a slip than an intent. A plot grid's `plots` also stays non-empty, since a grid with no plots has nothing to render.

Against the bundled legacy Excel fixture the importer writes 15 empty reference lists across its 8 scenarios; rebuilding all 8 from their own written fields now succeeds.

## When a reference is checked (#1169)

The asymmetry the issue describes is deliberate and is now written down rather than left to be rediscovered. Authoring resolves a reference at the call that makes it, so a typo surfaces where it was typed; reading a project defers, so a legacy project with a real gap still loads and imports and `validateProject()` reports every gap at once. The policy is one paragraph in the shared `vectorizedAuthoring` details block that all 19 authoring topics inherit, plus a subsection in `vignette("validate-project")` that also names `validate = FALSE` as the documented way past the pre-run check.

The one part of the split that was not deliberate is closed for scenarios: the deferred path named the closest existing id ("did you mean 'adult_male'?") and the eager path did not. `addScenario()` and `setScenario()` now use the same `.suggestSuffix()` / `.suggestSuffixMulti()` helpers, so the message reads alike whichever path finds the typo. The other sections' eager reference errors still name only the missing id; the inherited doc block says so rather than promising the hint everywhere, and extending it belongs with the rest of the mutation-API consistency work in #1148.

## Example

Before:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/base-json-based-project", quiet = TRUE)
dir <- withr::local_tempdir()
initProject(destination = dir, type = "example", createExcel = FALSE)

# the value a definition file carries for a scenario with no output paths
addScenario(
  project,
  id = "no_outputs",
  modelFile = "Aciclovir.pkml",
  outputPaths = character(0)
)
#> Error in `addScenario()`:
#> ! Cannot add scenario "no_outputs":
#> ✖ outputPaths must be a non-empty character vector with no NA or empty entries
"no_outputs" %in% names(project$definitions$scenarios)
#> [1] FALSE

# a mistyped reference
addScenario(
  project,
  id = "typo",
  modelFile = "Aciclovir.pkml",
  individual = "adult_mal"
)
#> Error in `addScenario()`:
#> ! Cannot add scenario "typo":
#> ✖ individual 'adult_mal' not found in individuals
```

After:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/feature-dev-1187-1169-539da5", quiet = TRUE)
dir <- withr::local_tempdir()
initProject(destination = dir, type = "example", createExcel = FALSE)

# the value a definition file carries for a scenario with no output paths
addScenario(
  project,
  id = "no_outputs",
  modelFile = "Aciclovir.pkml",
  outputPaths = character(0)
)
"no_outputs" %in% names(project$definitions$scenarios)
#> [1] TRUE

# a mistyped reference
addScenario(
  project,
  id = "typo",
  modelFile = "Aciclovir.pkml",
  individual = "adult_mal"
)
#> Error in `addScenario()`:
#> ! Cannot add scenario "typo":
#> ✖ individual 'adult_mal' not found in individuals (did you mean 'adult_male'?)
```

</details>

Closes #1187
Closes #1169


R/authoring-vectorize.R
```diff
@@ -55,10 +55,30 @@
 #' different multi-valued list per definition, pass a list of the same length
 #' as the id vector (one vector per definition).
 #'
+#' A zero-length reference list means "there are none", the same as `NULL`. A
```
Felixmil: The zero-length rule is documented here, in the shared block, but implemented only for scenarios.

`.canonicalizeVectorIdRef()` is called from `addScenario()` / `setScenario()` only. The other vector-valued reference field this same block names — a grid's `plots` — still goes through `.requireNonEmptyStringVector()` (`R/plots.R:961`, rejecting `length(x) == 0L`), so:

```r
addPlotGrid(project, "g", plots = character(0))
#> Error: `plots` must be a non-empty character vector
```

on the very `addPlotGrid.Rd` page that now inherits "a zero-length reference list means 'there are none'". An individual's / application's `parameterSets` lands in a third place again: `character(0)` is accepted but stored verbatim as `character(0)` (`R/individuals.R:344-362`), so the record shape differs from the `NULL` `.parseIndividuals()` builds from the same `[]` — the shape-equality argument in `.canonicalizeVectorIdRef()`'s comment doesn't hold there.

Either narrow the paragraph to the scenario fields it actually covers, or route the other vector reference args through the same helper.

---

R/authoring-vectorize.R
```diff
@@ -55,10 +55,30 @@
...
+#' definition file writes `[]` for a reference list a definition has none of, so
+#' `character(0)` (or the empty list `jsonlite::fromJSON()` yields for `[]`)
+#' reads a definition's own written fields straight back into an authoring call.
+#'
 #' The call is all-or-nothing: every definition is validated first, and if any
 #' fails the whole call aborts and writes nothing. On success all definitions
 #' are folded into the section and persisted in a single write-through.
 #'
+#' An authoring function resolves every reference at the call that makes it: an
```
Felixmil, line 71: "aborts naming the missing id (and the closest existing one)" is true only for the two scenario FK helpers this PR touched.

`.suggestSuffix()` / `.suggestSuffixMulti()` are now wired into `.checkScalarScenarioFK()` and `.checkVectorScenarioFK()` (`R/scenarios.R:1751`, `:1786`). Every other eager reference check still aborts bare:

- `addIndividual(parameterSets = "globl")` / `setIndividual()` — `R/individuals.R:352-361`
- `addApplication()` / `setApplicationParameterSets()` — `R/applications.R:216`
- `addPlotGrid()`'s unknown `plotId` path — `R/plots.R:964-969`
- `addPlot(dataCombined = ...)` — `R/plots.R:780`

`validateProject()` does suggest on those same references (`R/validation.R:1075`, `:1095`), so the asymmetry this PR set out to close is still there for everything but scenarios. Since the paragraph is inherited by all 19 authoring topics, it currently promises the hint on pages where it isn't given. Worth either extending the two helpers to those call sites (they're each a one-line `paste0`/`cli` change) or scoping the sentence.

---

vignettes/validate-project.Rmd
```diff
@@ -47,6 +47,16 @@ Validation has exactly **two** levels, not three:
...
 
+## When a reference is checked
+
+Authoring and reading check references at different moments, on purpose.
+
+An authoring function checks immediately: `addScenario(project, "s", modelFile = "m.pkml", individual = "adlt_male")` aborts on the spot, naming the missing id and the closest existing one, because at that moment you are writing one reference by hand and the typo is right in front of you.
+
+Reading a project defers to `validateProject()`. `loadProject()` and `importProjectFromExcel()` accept a project whose references do not all resolve: each section is read on its own, so a legacy project with a real gap still loads and imports, and the full report tells you about every gap at once instead of stopping at the first. That is what the worked example below shows.
+
+So a project can hold a dangling reference that `addScenario()` would have refused. That is the deferred half doing its job, not an inconsistency: nothing runs until the reference resolves either way, since `runScenarios()` and `createPlots()` validate before they run.
```
Felixmil: "nothing runs until the reference resolves either way" isn't quite right — both functions expose a documented opt-out.

`runScenarios()` and `createPlots()` take `validate = TRUE` (`R/scenarios.R:653`, `:724`; `R/create-plots.R:66`), and the section validators only run under `if (isTRUE(validate))` (`R/scenario-execution.R:508`; `R/create-plots.R:67`). With `validate = FALSE` a dangling `individual` reaches the build, which warns (`messages$noIndividualCharacteristics`, `R/scenario-execution.R:372-377`) and simulates on with the PKML's built-in individual — so the scenario does run with an unresolved reference.

That's a sound escape hatch, but the sentence as written tells a reader the guarantee is unconditional. Suggest qualifying it, e.g. "…since `runScenarios()` and `createPlots()` validate before they run unless you pass `validate = FALSE`."

---

R/definition-id.R
```diff
@@ -183,6 +183,25 @@
...
 }
 
+# Canonicalize a vector-valued foreign-key argument (a scenario's
+# `parameterSets` / `initialConditions` / `outputPaths`), treating a zero-length
+# value as `NULL`. A definition file carries `[]` for a scenario that references
+# none, so reading that field back out and handing it to an authoring function
+# arrives as `character(0)` (or the `list()` `jsonlite` yields), which means
+# exactly what `NULL` means here: there are none. Normalizing before validation
+# keeps the record shape identical to the one `.parseScenarios()` builds from
+# the same `[]`, so the file -> authoring -> file round trip closes.
```
Felixmil: "the file -> authoring -> file round trip closes" over-claims: only the *empty* case closes.

The package reads definition files with `jsonlite::fromJSON(simplifyVector = FALSE)` (`R/definition-files.R:424`, `R/project-lifecycle.R:333`, `R/project.R:1357`), which yields `list("a", "b")` for a non-empty array. `.canonicalizeIdRef()` returns a non-character value untouched (`R/definition-id.R:147-149`), `.canonicalizeVectorIdRef()` passes it through on `length != 0`, and `.checkVectorScenarioFK()`'s `!is.character(value)` branch then rejects it:

```
outputPaths must be a non-empty character vector with no NA or empty entries
```

So a scenario with two output paths still can't be rebuilt from its own written fields as the package itself reads them. (And when the list length happens to equal the id count, `.wholeField()` treats it as the documented per-definition form and splits one ref per scenario instead — correct per the rule, but a silent change of meaning for someone doing the round trip.)

The new test at `tests/testthat/test-scenarios.R:557` doesn't catch this because it uses `jsonlite::fromJSON()` with default simplification, not the reader the package uses. Two options: add an `unlist()`/`as.character()` for an all-scalar-strings list in `.canonicalizeVectorIdRef()` and assert the non-empty case with `simplifyVector = FALSE`, or narrow the comment (and the PR description's round-trip claim) to the empty case that's actually handled.

---

NEWS.md
```diff
@@ -34,7 +34,7 @@
...
...validateProject(project)` validates a parsed project and reports per-section critical errors and warnings. A dangling cross-reference (a scenario, individual, application, plot, or PI task pointing at an id with no matching definition) now suggests the closest existing ids ("did you mean '...'?") in its message, and so does the reference error `addScenario()` / `setScenario()` raise when they catch the same dangling reference at authoring time. A DataCombined entry that is missing its required `label` is now reported here, so a label-less record is caught up front instead of passing validation and then failing later at plotting with a misleading message. Removing a definition that is still referenced warns at removal time as well; `removeScenario()` now warns when a `dataCombined` definition still points at the removed scenario, matching the existing warnings from `removeIndividual()` and the other removers. The returned results object now prints a readable summary grouped by definition type (a cross marks each critical error, a `!` marks each warning, with overall counts in the header), while the structured object stays fully indexable (`results$scenarios$critical_errors`). (#908)
```
Felixmil, line 37: The NEWS edit covers the "did you mean" half (#1169) but there's no entry for the zero-length half (#1187), which is the user-visible behaviour change of the two.

`setScenario(project, "s", outputPaths = character(0))` goes from aborting to silently clearing the field, and `addScenario(outputPaths = character(0))` from abort to accepted. Someone who relied on the abort as a guard gets no signal from the changelog. Worth a sentence on the existing vectorized-authoring bullet (line 40), where the "vector-valued-per-definition field" rule is already described.

---

Felixmil (commented):
Reviewed the three commits on this branch (`47792fe...e73dd7c`) at effort `high`.

**The code changes themselves are sound.** I traced `.canonicalizeVectorIdRef()` through `.wholeField()` / `.alignAuthoringArgs()` on both the add and set paths (`character(0)`, `list()`, `NA`, `NULL`, batch and per-definition list forms all behave as described), confirmed the `supplied <- names(fields)` snapshot in `setScenario()` survives the `fields$x <- NULL` element deletion, verified `.scenarioToJson()` still emits `[]`, and hand-checked `adist`/threshold against the fixture ids so the existing bare-message snapshots (`ghost`, `ghostset`, `doesnotexist`, `nosuchprotocol`) stay bare while the three new ones (`indiv1`, `aciclovir_pvb`, `adult_male`) are correct. No R interpreter in this environment, so the suite was not executed — `R-CMD-Check` was still in progress on all three platforms when this review was written.

**Every finding below is in the prose the PR adds**, which is most of what it adds. Four are places where an inherited doc block or a comment promises more than the code delivers; the fifth is a missing changelog entry. None of them is a defect in the new behaviour.

| # | Where | Issue |
|---|---|---|
| 1 | `R/authoring-vectorize.R:58` | Zero-length rule documented for all 17 inheriting pages, implemented for scenarios only |
| 2 | `R/authoring-vectorize.R:67` | "closest existing id" hint promised everywhere, wired into scenario FK helpers only |
| 3 | `vignettes/validate-project.Rmd:58` | "nothing runs until the reference resolves" ignores `validate = FALSE` |
| 4 | `R/definition-id.R:193` | Round-trip claim holds for the empty case only; the package's own reader yields a list |
| 5 | `NEWS.md:37` | No entry for the zero-length change, the more visible of the two behaviour changes |

Findings 1 and 2 are the ones I'd want addressed before merge, since the doc block is inherited by 19 authoring topics and currently documents behaviour that only `addScenario()` / `setScenario()` have. Either extending the coverage or narrowing the sentences resolves them; the narrowing is a two-line change if the broader wiring is better left to a follow-up on #1088.

---

R/authoring-vectorize.R
Felixmil: Right, and narrowed rather than widened: the scope decision on this PR was scenario reference fields only, so the paragraph now names them and says a plot grid's `plots` stays non-empty (a grid with no plots has nothing to render, which is a real requirement rather than the same rule missing).

The individuals/applications point stands as you describe it: `character(0)` is accepted and stored verbatim there, so the shape does differ from the `NULL` the parser builds. No behaviour change on that path here; routing it through the same helper belongs with the rest of the mutation-API consistency work in #1148.

Fixed in 27874dff.

R/authoring-vectorize.R
Felixmil: Confirmed, all four sites still abort bare. Scoped the sentence rather than wiring them up: the paragraph now attributes the closest-existing-id hint to a scenario's reference errors and says the other sections name only the missing id.

Extending it to individuals, applications and the two plot paths is worth doing, but it is four more sites and their snapshots on a PR that is meant to be the two named issues, so it fits better with #1148.

Fixed in 27874dff.

vignettes/validate-project.Rmd
Felixmil: Correct, the guarantee is conditional. The sentence now says the two functions validate before they run and names `validate = FALSE` as the way past it, with what happens then (the scenario builds on whatever the model file already carries).

Fixed in 27874dff.

NEWS.md
Felixmil: Added, on the vectorized-authoring bullet where the vector-valued rule already lives: an empty reference list means "there are none", `addScenario(outputPaths = character(0))` adds a scenario with no output paths, `setScenario(outputPaths = character(0))` clears them, and a plot grid's `plots` stays non-empty.

Fixed in 27874dff.

R/definition-id.R
Felixmil: Reproduced, and took the first option rather than narrowing the claim. A two-element list read with `simplifyVector = FALSE` did abort:

```
Cannot add scenario "rt2":
✖ outputPaths must be a non-empty character vector with no NA or empty entries
```

Worth noting the length-1 case passed by accident, not by design: `.wholeField()` sees `list("a")` against one id, treats it as the per-definition form, and hands the element through. That is exactly the silent change of meaning you flag.

`.canonicalizeVectorIdRef()` now flattens an all-strings list to the character vector the reference list is, before canonicalizing; anything else in the list is left for the validators to reject (new snapshot for `list("aciclovir_pvb", 1)`). The round-trip test reads with `simplifyVector = FALSE` and covers a populated list as well as `[]`. The comment and the PR description now describe both cases instead of claiming the whole round trip from the empty one.

The `.wholeField()` ambiguity when the list length equals the id count is untouched: it is the documented per-definition form, so there is nothing to disambiguate it against.

Fixed in cf93b3e1.

PR #1207, opened 2026-07-30
Import a quoted legacy plotIDs cell without breaking the grid's plot references

### LINKED ISSUE #1184: Excel import: plot grids reference plot ids the plots section does not define
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). Two of the tested projects import into a tree whose plot-grid and plot-configuration sections reference plot ids that the plots section does not define, so a freshly imported project reports critical errors.

## Errors

```
plotGrids references unknown plotIds: _<scenario>_, _<scenario>_<outputTag>_, ...   (35 ids in one project)
```

```
plotConfiguration references unknown dataCombinedId: <id>, <id>_mean
```

## Why this looks like an id-canonicalization mismatch, not bad source data

The referenced ids carry leading and trailing underscores that the defined ids do not. That pattern points at the referencing side and the defining side disagreeing on how a plot id is derived from its source cell, rather than at the workbook naming a plot that genuinely does not exist. The `<id>` versus `<id>_mean` pair in the second message has the same shape: a suffix that survives on one side of the reference and not the other.

## Suggested fix

Run plot ids through the same `.canonicalizeId()` seam on both the defining and the referencing side so a grid and its plots cannot end up with different spellings of the same id. Worth a regression test that imports a workbook whose plot names need canonicalizing and asserts the grid still resolves.

Part of #1088.

Felixmil:
A legacy Excel project whose `plotGrids` sheet writes its membership in the 5.x quoted form (`"P1", "P2"`) imported into a tree that immediately failed its own `validateProject()`, reporting every grid member as an unknown plot id (35 of them in one project). The quotes were the cause. This also brings the two plot reference checks onto the same canonicalization seam the rest of the validator already uses.

## Excel import

The `plotIDs` cell is a multi-value cell, and the 5.x convention wraps each value in `""` so a plot id may itself contain a comma; v5 read it with `scan(sep = ",")`, which strips the quotes. The importer instead decoded it with `.splitPlotIDs()`, the in-memory codec, which only understands backslash escaping. The quotes survived, and id canonicalization then replaced each `"` with `_`, so `"P1", "P2"` became a grid referencing `_p1_` and `_p2_` while the plots section defined `p1` and `p2`.

`.parseExcelPlotGridSheet()` now decodes the cell with `.parseCommaListToArray()`, the helper every other multi-value Excel cell already uses (it honours both the quoted form and the backslash escaping this package writes), and re-encodes it with `.joinPlotIDs()` into the single escaped string the `plots` JSON key holds. Nothing downstream changes: `.canonicalizeProjectJsonIds()`, `.validatePlots()`, and `createPlots()` keep reading it with `.splitPlotIDs()`. A workbook written by `exportProjectToExcel()` round-trips unchanged, since `.parseCommaListToArray()` is a strict superset of `.splitPlotIDs()`.

## Validation

`.validatePlots()` compared both of its reference sets with a bare `setdiff()` and printed a raw id list. It now uses `.danglingRefs()`, so a plot's `dataCombined` reference and a grid's member ids are canonicalized on both sides before comparing (a hand-edited `P1` resolves to the plot `p1`), and appends `.suggestSuffixMulti()` so a reference that genuinely does not resolve names the closest existing id:

```
plotConfiguration references unknown dataCombinedId: aciclovir_pbv (did you mean 'aciclovir_pvb'?)
plotGrids references unknown plotIds: p2 (did you mean 'p1'?)
```

The issue reports a second symptom as well, a `plotConfiguration references unknown dataCombinedId` on a different project. That one is not reproducible from the code: a plot's `dataCombined` reference and the `dataCombined` section's own id go through the identical canonicalization call, and v5 compared those two cells verbatim, so a workbook that was valid under v5 cannot diverge there. The suggestion hint above is what will make the next run on that project say whether it is a spelling mismatch or genuinely absent data.

Closes #1184


R/plots.R
```diff
@@ -346,22 +353,26 @@ print.DataCombined <- function(x, ...) {
...
       function(p) p$plotId %||% NA_character_,
       character(1)
     )
+    knownPlotIds <- knownPlotIds[!is.na(knownPlotIds)]
     allGridIds <- unique(unlist(lapply(
       plotGrids,
       function(g) .splitPlotIDs(g$plotIds)
     )))
-    invalidGridRefs <- setdiff(allGridIds, knownPlotIds)
+    invalidGridRefs <- .danglingRefs(allGridIds, knownPlotIds)
```
Felixmil: The validator now accepts grid references the builder still rejects, so `validateProject()` can call a project clean and `createPlots()` then hard-aborts on it.

`.assertPlotGridsBuildable()` (`R/create-plots.R:487`) still compares with a bare `setdiff(referencedPlotIds, plotIDs)`. Two inputs now pass here and fail there:

- **Case-only reference.** Exactly the fixture added in `test-plots.R` ("resolves a reference that matches its definition canonically"): a grid with `"plots": "P1"` against a plot `p1`. `loadProject()` reads references verbatim — `.canonicalizeProjectJsonIds()` only runs on the Excel-import and snapshot paths — so a hand-edited tree really can carry `P1`. Validation is clean; `createPlots()` aborts with `invalidPlotID`.
- **Empty-string member.** `.danglingRefs()` drops `NA` and `""` (`R/validation.R:832-834`); `setdiff` did not. A grid written `"plots": "p1, "` yields `c("p1", "")` from `.splitPlotIDs()`, so validation goes from reporting it to passing it, while `create-plots.R:487` still aborts on the `""`.

`validate = FALSE` is no escape hatch either — these asserts run unconditionally, ahead of the build.

Suggest pointing the two build-time asserts at `.danglingRefs()` too, so the gate and the thing it gates agree.

---

R/plots.R
```diff
@@ -319,16 +324,18 @@ print.DataCombined <- function(x, ...) {
       plotConfig,
       function(p) p$dataCombinedId
     ))
-    invalidDataCombinedRefs <- setdiff(
+    dataCombinedKeys <- names(dataCombined %||% list())
+    invalidDataCombinedRefs <- .danglingRefs(
```
Felixmil: Same validator/builder split on the `dataCombinedId` side: `.assertPlotConfigurationsBuildable()` (`R/create-plots.R:444`) still uses `setdiff(unlist(dataCombinedIds), dataCombinedNames)`, so `dataCombined: "Aciclovir_PVB"` against the definition `aciclovir_pvb` (the second half of the new canonical-resolution test) validates clean and then aborts in `createPlots()` with `invalidDataCombinedName`.

The empty-string case is worse here than on the grid side, because nothing else catches it. `dataCombined: ""` already slips the missing-field loop at `R/plots.R:293` (`is.null(p[[field]])` — `""` is not `NULL`), and `.danglingRefs()` now drops it as well. So an empty `dataCombined` reference has no remaining validation error at all, and `create-plots.R:444` is the first thing to notice.

Worth keeping the empty-string reference reportable — either by leaving it to a presence check here (`!nzchar()` alongside the `is.null()` loop) or by not filtering `""` at these two call sites.

---

R/plots.R
```diff
@@ -319,16 +324,18 @@ print.DataCombined <- function(x, ...) {
...
-      names(dataCombined %||% list())
+      dataCombinedKeys
     )
     if (length(invalidDataCombinedRefs) > 0) {
       result$addCriticalError(
         "Invalid Reference",
         paste0(
           "plotConfiguration references unknown dataCombinedId: ",
-          paste(invalidDataCombinedRefs, collapse = ", ")
+          paste(invalidDataCombinedRefs, collapse = ", "),
```
Felixmil, line 341: Minor: an unknown `dataCombinedId` is now listed once per referencing plot. `referencedDataCombined` (line 323) is not `unique()`d, and `.danglingRefs()` preserves duplicates where the `setdiff()` it replaced deduplicated its output — so ten plots sharing one typo render the id ten times in the message. The `.suggestSuffixMulti()` hint is unaffected (it `unique()`s internally).

`allGridIds` at line 365 is already deduped, so only this site is affected. A `unique()` on line 323 (or on `invalidDataCombinedRefs`) covers it.

---

Felixmil (commented):
Reviewed at high effort. The Excel half is solid; the validation half loosens the validator without loosening the builder it is meant to gate.

**Excel import (`9d00b35`) — clean.** `.parseCommaListToArray()` → `.joinPlotIDs()` is identity for anything `exportProjectToExcel()` writes, since `.canonicalizeOneId()` maps `,`, `"` and `\` to `_` and canonical ids therefore cannot contain the delimiters; unquoted legacy cells behave as before. The fix and its regression test look right.

**Validation (`c7eacbf`) — 2 real findings + 1 cosmetic**, commented inline. The headline one: `.validatePlots()` now compares canonically and drops empty references, but `.assertPlotConfigurationsBuildable()` / `.assertPlotGridsBuildable()` (`R/create-plots.R:444`, `:487`) still use a bare `setdiff`. The exact fixture added in `test-plots.R` — a grid referencing `P1` against a plot `p1` — passes `validateProject()` and then hard-aborts inside `createPlots()`, and `validate = FALSE` does not skip those asserts. Bringing the two build-time asserts onto `.danglingRefs()` closes the gap; the empty-reference case wants a `nzchar()` presence check so it stays reportable somewhere.

Also checked and found clean: `.suggestSuffix*` on the plot messages does not invalidate the un-updated scenario snapshots, and dropping the `NA` `plotId` from the suggestion candidates is correct (a missing `plotId` already raises its own critical error above).

Note: R is not available in this environment, so the suite was not run locally — CI is green on all 12 checks for `c7eacbf`, and the analysis above is static.

---

R/plots.R
Felixmil: Fixed in 7dcd75ba, but in the opposite direction to the suggestion: `.validatePlots()` goes back to `setdiff()`, rather than the two build-time asserts moving to `.danglingRefs()`.

Pointing the asserts at `.danglingRefs()` would only move the crash, because the build resolves by exact lookup *after* the assert. On the grid side `plotList[intersect(.splitPlotIDs(entry$plotIds), configPlotIds)]` (`R/create-plots.R:641`) silently drops a tolerated `P1`, so the grid comes out empty instead of aborting; on the plot side `dataCombinedList[[entry$dataCombinedId]]` (`:573`) hands `NULL` to `plotIndividualTimeProfile()`. Making tolerance actually work needs canonical resolution in the builder, not just in its gate.

That is out of scope here. #1184 asked for the canonicalization seam at import, which is the other commit, and canonical resolution in `createPlots()` would make plots the only section resolving that way at run time: every other lookup is exact (`R/definition-files.R:534` and siblings). So the gate and the thing it gates agree again, by tightening the gate.

The case-only test that motivated the finding is gone. The `.suggestSuffixMulti()` hints stay, which is the part that makes the second symptom in #1184 diagnosable.

Worth noting separately: `.danglingRefs()` in `.validateCrossReferences()` (`R/validation.R:831`) has this same validator/builder split package-wide, and that is a #1088 policy question rather than something to settle in this PR.

R/plots.R
Felixmil: Fixed in 7dcd75ba by the same revert to `setdiff()`, which restores the empty-string reference as a reported error on both sides.

You are right that nothing else catches it: `""` slips the required-field loop at `R/plots.R:293` because it is not `NULL`, so this check is the only place it surfaces before `create-plots.R:444` aborts on it.

Pinned by a new test, `.validatePlots reports an empty reference, which the missing-field loop lets through`, covering both `dataCombinedId: ""` and a grid written `"p1, "`.

R/plots.R
Felixmil: Fixed in 7dcd75ba. Reverting to `setdiff()` restores the deduplication, so a typo shared by ten plots renders once again.

PR #1210, opened 2026-07-30
Write Project.json from the Excel import, with a settable name

### LINKED ISSUE #1174: Importer writes ProjectConfiguration.json but loadProject() defaults to Project.json
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). The importer writes the project manifest as `<outputDir>/ProjectConfiguration.json`, but `loadProject()` and `initProject()` default to `Project.json`. So `loadProject("imported")` (the natural next call after import) fails, and passing the directory gives an opaque error:

`Failed to parse '<dir>' as JSON. Caused by error in \`open.connection()\`: cannot open the connection`

The caller must pass the exact `<dir>/ProjectConfiguration.json`. The manifest filename is inconsistent across the three entry points (`R/project-lifecycle.R:48`).

Suggested fix: have the importer also write `Project.json` (or write it under that name), or make `loadProject()` accept a directory and resolve the manifest inside it. Either removes the mismatch and the opaque directory-path error.

Felixmil:
The Excel import named the project file after the workbook it read, so a legacy `ProjectConfiguration.xlsx` produced a `ProjectConfiguration.json` while `loadProject()` and `initProject()` both default to `Project.json`. The obvious next call after an import therefore failed, and passing the folder instead (the natural second guess) gave an opaque `Failed to parse '<dir>' as JSON`, because a directory slips past the file-exists guard in the JSON reader. This makes the three entry points agree on one name, and keeps a way to choose a different one.

## Import

- `importProjectFromExcel()` writes `Project.json` whatever the source workbook is called, so an imported project opens exactly like one created by `initProject()`.
- The new `projectFileName` argument (default `"Project.json"`) sets a different name, to label the project file after the study rather than by the generic name. A `.json` extension is *appended* when the name does not already end in one, rather than set: setting it would read the `.v1` of a dotted stem as an extension and collapse `trial.v1` and `trial.v2` onto one file.
- This names the project file only. The `definitions/` tree beside it, and the copied `Models/` and `Data/` folders, are shared by whatever sits in `outputDir`, so each project still needs its own `outputDir`.
- The name must be a plain filename. One carrying a path separator (or `.` / `..`) would write outside `outputDir` and over whatever sits there, so it aborts.
- Saving is unchanged: a project always writes back to the file it was loaded from, so a name chosen through `projectFileName` survives an edit rather than forking a stray `Project.json`.

## Excel side-car pairing

A project and its workbook are paired by name, but the two halves used to derive that pairing differently: `projectStatus()` swapped the project file's extension, while `exportProjectToExcel()` hardcoded `Project.xlsx`. For any container not called `Project.json` the loop never closed — following the status check's own advice wrote a workbook it would then keep ignoring, leaving `excel_in_sync` at `NA` forever. That dead end already existed for a legacy `ProjectConfiguration.json`.

Both halves now derive the name in one place (`.excelSideCarName()`), so a project loaded from `MyStudy.json` exports `MyStudy.xlsx` and the status check finds it. The "no Excel file" message names the file it looked for instead of a fixed `Project.xlsx`, and a test pins that exporting actually satisfies the advice the message gives.

## Shared filename guard

`importProjectFromExcel()`'s guard was condition-for-condition identical to `snapshotProject()`'s, so both now use one `.validateFilenameSegment()` in `R/file-utils.R`. It takes the message and a `call`, so the abort reads `Error in \`importProjectFromExcel()\`:` rather than exposing a dot-prefixed internal — which also fixes the same attribution wart in `snapshotProject()`.

One nit deliberately left alone: the `x` bullet in both messages claims the name "contains a path separator or is `.` / `..`" for every rejection, which misdescribes an empty string, `NA`, or a non-character. Pre-existing in both, and worth a follow-up rather than more churn here.

## Docs

The changelog entry for `importProjectFromExcel()` is edited in place rather than given a sibling, since the function is introduced in this same development cycle. The migration vignette, the function's own help, and the comments that still described the old stem-derived naming are updated alongside.

Closes #1174


R/project-excel.R
```diff
@@ -477,8 +484,10 @@ importProjectFromExcel <- function(
   # `file` / `importerConfiguration` basenames against `dataFolder`.
   jsonData <- .parseExcelObservedData(jsonData, prop, pcDir, outputDir)
 
-  outputFileName <- sub("\\.xlsx$", ".json", basename(projectConfigPath))
-  outputPath <- file.path(outputDir, outputFileName)
+  outputPath <- file.path(
+    outputDir,
+    fs::path_ext_set(projectFileName, "json")
```
Felixmil: **`projectFileName` cannot actually keep several projects side by side in one folder.**

The container name is now settable, but the definitions tree that container owns is not. Three lines below, the guard is hardcoded:

```r
definitionsDir <- file.path(outputDir, "definitions")
```

and `.writeProjectTree()` resolves the tree from the container's `definitionsFolder` field, which the import never sets — so it falls back to `"definitions"` (`R/project.R:926`). Both imports therefore write into the *same* `<outputDir>/definitions`:

1. `importProjectFromExcel(a.xlsx, outputDir = d, projectFileName = "StudyA")` → `d/StudyA.json` + `d/definitions/`.
2. `importProjectFromExcel(b.xlsx, outputDir = d, projectFileName = "StudyB")` → `hasDefinitionTree` is `TRUE`, so it aborts with `importWouldOverwriteProject()` — whose own advice is *"import into a different `outputDir`"*, the opposite of what the new docs suggest.
3. With `overwrite = TRUE` it gets worse rather than better: `.writeDefinitionTree()` owns each `definitions/<kind>/` directory and deletes every `.json` not in the set it just wrote (`R/definition-files.R:475-486`). StudyB's import reconciles the shared tree to StudyB alone, so StudyA's definition files are removed and `StudyA.json` is left pointing at a tree that no longer describes it. `.copyExcelProjectAssets()` collides on `Models/`, `Data/` etc. in the same folder too.

The side-by-side workflow is promised in three places, so this reads as a supported use case rather than an edge: `@param projectFileName` above ("to keep several projects side by side in one folder"), the NEWS entry, and `vignettes/migrate-from-excel.Rmd:83` ("To migrate several Excel projects into one folder, give each import its own `projectFileName`").

Two ways out, either fine:

- Derive `definitionsFolder` from the project file stem when `projectFileName` is non-default (e.g. `StudyA.json` → `StudyA-definitions/`) and set it on the imported container, so the tree travels with its container. The container field already supports this; only the import's hardcoded `"definitions"` and the `hasDefinitionTree` guard need to follow it.
- Or keep one tree per folder and drop the side-by-side claim from the three doc sites, describing `projectFileName` as what it reliably is: a way to choose the container's name.

---

R/messages.R
```diff
@@ -531,11 +542,20 @@ messages$restoredProjectConfiguration <- function(inputFile, outputFile) {
...
+# folder on disk has no name to derive one from, hence the pathless variant.
+messages$syncNoExcel <- function(excelPath = NULL) {
+  if (is.null(excelPath)) {
+    return(cli::format_inline(
+      "No Excel configuration file found; nothing to compare."
+    ))
+  }
   cli::format_inline(
-    "No Excel configuration file ({.file Project.xlsx}) found; nothing to compare."
+    "No Excel configuration file ({.file {fs::path_file(excelPath)}}) found \\
```
Felixmil, line 560: **For a container that isn't `Project.json`, this advice can never satisfy the check.**

Naming the file that was looked for is a clear improvement over the fixed `Project.xlsx`, but the `exportProjectToExcel()` suggestion pairs badly with it. The two sides derive the side-car differently:

- `projectStatus()` swaps the container's extension: `MyStudy.json` → looks for `MyStudy.xlsx` (`R/project-excel.R:1081`).
- `exportProjectToExcel()` hardcodes the name: `projConfigPath <- file.path(outputDir, "Project.xlsx")` (`R/project-excel.R:686`).

So for a project imported with `projectFileName = "MyStudy"`, following the advice writes `Project.xlsx`, the status check keeps looking for `MyStudy.xlsx`, and `excel_in_sync` stays `NA` however many times the user runs it — with the message repeating the same instruction each time. The new test in `test-project-excel.R` asserts exactly that message against a `MyStudy` project, so it cements the dead end rather than catching it.

Worth fixing on the export side, since that's where the asymmetry is: have `exportProjectToExcel()` derive its workbook name from `project$info$projectFilePath` (extension swapped to `.xlsx`), falling back to `Project.xlsx` for a project with no file on disk. That makes the round trip `import(projectFileName = "MyStudy")` → `exportProjectToExcel()` → `projectStatus()` close, and it also fixes the same dead end for a legacy `ProjectConfiguration.json` container, which has it today. If you'd rather not touch the export in this PR, dropping the `exportProjectToExcel()` sentence when the derived name isn't `Project.xlsx` at least avoids advising something that cannot work.

---

R/project-excel.R
```diff
@@ -35,20 +35,27 @@
...
 #'   assets would be wasted work, as when the import feeds a throwaway
 #'   comparison snapshot.
-#'
-#' @return Invisibly returns the path to the created project file (the
-#'   `Project.json`).
+#' @param projectFileName Name of the project file the import writes in
+#'   `outputDir`. Defaults to `"Project.json"`, the same name [initProject()]
+#'   and [loadProject()] use, so an imported project opens like any other. Pass
+#'   another name (for example `"MyStudy"`, or `"MyStudy.json"`) to keep several
+#'   projects side by side in one folder; the `.json` extension is added when it
```
Felixmil: **"added when it is missing" isn't what `fs::path_ext_set()` does — it replaces an existing extension.**

`fs::path_ext_set(projectFileName, "json")` on line 489 rewrites whatever extension the name already has, and a version-ish or dated stem reads as an extension to `fs::path_ext()`:

| `projectFileName` | documented | actual |
| --- | --- | --- |
| `"MyStudy"` | `MyStudy.json` | `MyStudy.json` ✓ |
| `"MyStudy.json"` | `MyStudy.json` | `MyStudy.json` ✓ |
| `"study.2024"` | `study.2024.json` | `study.json` |
| `"trial.v1"` | `trial.v1.json` | `trial.json` |
| `"trial.v2"` | `trial.v2.json` | `trial.json` |

The last two are the bad case: two names the caller chose to be distinct collapse onto one container, which with `overwrite = TRUE` means the second import quietly replaces the first. `.validateProjectFileName()` passes all of these, since it only rejects path separators and `.` / `..`.

Appending only when the extension isn't already `json` matches the documented behaviour:

```r
if (!identical(tolower(fs::path_ext(projectFileName)), "json")) {
  projectFileName <- paste0(projectFileName, ".json")
}
outputPath <- file.path(outputDir, projectFileName)
```

Same wording to correct in the `man/` page and the NEWS entry ("the `.json` extension is added when missing"), and a test for a dotted stem would pin it.

---

R/project-excel.R
```diff
@@ -1338,6 +1347,30 @@ projectStatus <- function(project, silent = FALSE) {
...
+#'
+#' The name becomes a path via `file.path(outputDir, ...)`, so one carrying a
+#' path separator or a `..` segment would write the project file outside the
+#' folder the caller named, over whatever sits there.
+#'
+#' @param name The `projectFileName` the caller passed.
+#' @returns `NULL`, invisibly; called for the abort.
+#' @keywords internal
+#' @noRd
+.validateProjectFileName <- function(name) {
```
Felixmil: Minor: this is condition-for-condition the same guard as `.validateSnapshotStem()` (`R/project-snapshot.R:248`) — same five checks in the same order, only the message differs. The PR body says it mirrors `snapshotProject()`, and it does, exactly; a shared helper taking the message would collapse the duplication:

```r
.validateFilenameSegment <- function(name, message) {
  if (!is.character(name) || length(name) != 1L || is.na(name) ||
      !nzchar(name) || grepl("[/\\]", name) || name %in% c(".", "..")) {
    cli::cli_abort(message(name))
  }
  invisible(NULL)
}
```

Two things it would be a good moment to fix in both, if you do extract it:

- Neither takes `call = rlang::caller_env()`, so the abort is attributed to the private helper — the new snapshot records `Error in \`.validateProjectFileName()\`:`, exposing a dot-prefixed internal in a user-facing error. Threading `call` through to `cli_abort()` (the `.assertIdVector()` / `.validateObservedDataId()` pattern) would make it read `Error in \`importProjectFromExcel()\`:`.
- The `x` bullet claims the name "contains a path separator or is `.` / `..`" for every rejection, including `""`, `NA_character_`, a non-character, and a length-2 vector, where it misdescribes the problem.

Both are pre-existing in `.validateSnapshotStem()`, so entirely reasonable to leave for a follow-up.

---

Felixmil (commented):
Reviewed the two commits on this branch (`c7eacbf..5a1e2dc`) against the surrounding code in `R/project-excel.R`, `R/project-lifecycle.R`, `R/definition-files.R`, `R/project-snapshot.R` and `R/messages.R`, plus the new tests, the `man/` page and the vignette.

The core of the change holds up. Making the container name canonical whatever the workbook was called is the right call, the `saveProject()` side genuinely preserves a chosen name (a project writes back to `project$info$projectFilePath`, and `.writeProjectTree()` is passed that path), the path-separator guard closes a real write-outside-`outputDir` hole, and the comment cleanups in `definition-files.R` / `project-lifecycle.R` / `import-legacy-snapshot.R` correctly follow the naming change through. Dropping the `Project.xlsx` rename from the canonicalization-collision test is a nice consequence — that test now passes because the name is canonical, rather than because the fixture was renamed to make it so.

Four findings, two of them worth acting on before merge:

1. **`projectFileName` doesn't deliver the side-by-side workflow it's documented for** (`R/project-excel.R:489`). The container name is settable but the `definitions/` tree is hardcoded, so a second import into the same folder aborts — and with `overwrite = TRUE` deletes the first project's definition files. Promised in the `@param`, in NEWS, and in `vignettes/migrate-from-excel.Rmd:83`.
2. **The new `projectStatus()` advice can't be satisfied for a non-`Project.json` container** (`R/messages.R:556`). The status check derives `<stem>.xlsx`; `exportProjectToExcel()` hardcodes `Project.xlsx`. `excel_in_sync` stays `NA` however many times the user follows the message, and the new test asserts that message.
3. **`fs::path_ext_set()` replaces rather than appends** (`R/project-excel.R:42`), so `"study.2024"` becomes `study.json` and `"trial.v1"` / `"trial.v2"` collapse onto the same file — contradicting the documented "added when it is missing".
4. Minor: `.validateProjectFileName()` duplicates `.validateSnapshotStem()` exactly (`R/project-excel.R:1360`); notes there on error attribution and message accuracy, both pre-existing.

One small doc-drift the commit otherwise cleaned up thoroughly: the comment at `R/project-excel.R:545` still describes what gets written as "the slim (`containerOnly`) `Project.json` container", which is no longer necessarily its name — the sibling comments in `definition-files.R:1329` and `project-lifecycle.R:134` were updated for exactly this.

Note I couldn't run `testthat` in this environment (no R toolchain available), so everything above is from reading the source rather than from a failing test.

---

R/project-excel.R
Felixmil: Confirmed by reproducing it: a second import into the same folder aborts, and the abort's own advice ("import into a different `outputDir`") directly contradicts what the docs promised.

Took the second option — dropped the claim — rather than deriving a per-project `definitionsFolder`. That route wouldn't actually deliver side-by-side either: as you noted, `.copyExcelProjectAssets()` still collides on `Models/`, `Data/` and the population folder in a shared `outputDir`, so it would trade a clear abort for a half-working feature that fails later and less obviously.

`projectFileName` is now described as what it reliably is — a way to label the project file after the study — with the limitation stated explicitly in all three places that had over-promised (`@param`, NEWS, `vignettes/migrate-from-excel.Rmd`), each pointing at the one thing that does work: give each project its own `outputDir`.

Fixed in a3f30567.

R/messages.R
Felixmil: Fixed on the export side, as you suggested — the asymmetry was the bug, not the message.

Both halves now derive the side-car name from one helper, `.excelSideCarName(projectFilePath)`: `exportProjectToExcel()` uses it for the workbook it writes (and for its overwrite guard), and `projectStatus()` uses it for the file it looks for. So a project loaded from `MyStudy.json` exports `MyStudy.xlsx`, the status check finds it, and the round trip closes. A project with no file on disk still falls back to `Project.xlsx`.

As you pointed out, this also fixes the same dead end for a legacy `ProjectConfiguration.json` container, which had it before this PR.

The test no longer just asserts the message — it follows the advice and checks it works: export, then assert `MyStudy.xlsx` exists and `excel_in_sync` is no longer `NA`.

Worth flagging for review: this changes `exportProjectToExcel()`'s output filename for any non-canonical container, which is beyond what #1174 asked for. It's called out in the PR body.

Fixed in a3f30567.

R/project-excel.R
Felixmil: Confirmed — verified your table directly:

```
fs::path_ext_set("study.2024", "json")  #> study.json
fs::path_ext_set("trial.v1",   "json")  #> trial.json
fs::path_ext_set("trial.v2",   "json")  #> trial.json
```

I'd actually noticed the replace-vs-append behaviour while reviewing my own diff and waved it off as "consistent with `snapshotProject()`". That was the wrong call: `snapshotProject()` documents that it *forces* the extension, so it's accurate there, whereas I'd documented "added when it is missing" and then not done that. The `trial.v1` / `trial.v2` collapse under `overwrite = TRUE` is the real hazard.

Now goes through `.withJsonExtension()`, which appends unless the name already ends in `.json` (case-insensitively), keeping everything before it verbatim. Wording corrected in the `@param`, the NEWS entry, and the `i` bullet of `invalidProjectFileName()`; `man/` regenerated. A test pins it by importing `trial.v1` and `trial.v2` into the same folder and asserting two distinct project files.

Fixed in a3f30567.

R/project-excel.R
Felixmil: Extracted, and it turned out to be worth doing beyond the duplication.

`.validateFilenameSegment(name, message, call = rlang::caller_env())` now lives in `R/file-utils.R`; both `.validateProjectFileName()` and `.validateSnapshotStem()` are gone, with `importProjectFromExcel()` and `snapshotProject()` calling the shared guard. Threading `call` fixes the attribution in both — the snapshots now read `Error in \`importProjectFromExcel()\`:` and `Error in \`snapshotProject()\`:` instead of exposing a dot-prefixed internal.

The extraction surfaced a live bug in the process. These `messages$` entries are glue templates expanded at the abort site, and `invalidSnapshotName()` interpolated `{stem}` — a name that no longer exists once the abort is raised from the shared frame. `snapshotProject()` started failing with `cannot coerce type 'closure' to vector of type 'character'` (cli resolving the missing binding). Both entries now interpolate `{name}`, matching what the shared guard binds, with a comment on `invalidSnapshotName()` recording the coupling, since it isn't obvious that the placeholder is tied to the raising frame rather than to the public argument.

Left for a follow-up, as you suggested: the `x` bullet claiming "contains a path separator or is `.` / `..`" for every rejection, which misdescribes `""`, `NA`, a non-character, and a length-2 vector. Pre-existing in both messages, and now in one shared code path, so it's a tidy standalone fix. Noted in the PR body.

Fixed in a3f30567.

PR #1212, opened 2026-07-31
Settle the messages conventions and the three migration-facing message defects

### LINKED ISSUE #1122: Decide on message formatting helper: ospsuite.utils::cliFormat() vs cli::format_message()
`R/messages.R` currently mixes two helpers for building user-facing messages:

- `ospsuite.utils::cliFormat()` — 104 uses (the majority)
- `cli::format_message()` — 20 uses (the most recently added message functions)

Both produce cli-markup-formatted messages, but the file should use one consistently.

The contributor guide being added in #1120 currently states the `cliFormat()` rule (as collected in #986). The team should decide which helper is the standard going forward, adapt `R/messages.R` for consistency if required, and update the rule in `AGENTS.md` if the decision differs.

Came up while curating #986 into the agent/contributor guide (#1120).

### LINKED ISSUE #1149: Messages & code conventions
Umbrella for user-facing message construction and code conventions in the v6 project model. Part of #1088.

Tracks the `R/messages.R` catalog and the shared conventions (message-helper choice, casing, file layout) so new work stays consistent.

### LINKED ISSUE #1175: Authoring emits a canonicalization warning storm (one per id); batch or summarize
Found during migration testing of the 6.0.0 JSON model (tested at commit `677f9c8e546181db27d60295599d0ccf6fedbb49`). Rebuilding a real project with mixed-case legacy ids via the `add*()` API emits dozens of per-call id-canonicalization warnings (roughly 40-60 in a single authoring run), which drown the useful output. Each individual warning is correct and precise (it names the id and its referenced ids); the volume is the friction.

Root cause: per-call canonicalization warnings at `R/definition-id.R:58,179,230`.

Suggested fix: batch or summarize the canonicalization warnings (e.g. one grouped warning per `add*` call, or a single end-of-session summary), rather than one warning per canonicalized id, consistent with the messages conventions this umbrella settles.

### LINKED ISSUE #1185: loadProject() on a previous-version snapshot does not point at restoreProject()
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). Opening a previous-version monolithic snapshot with `loadProject()`, the function whose name a user reaches for first, fails with a message that does not mention the entrypoint that would have worked.

## What happens

```r
loadProject("ProjectConfiguration.json")
#> Unsupported schemaVersion: "<missing>". Expected "2.0".
```

Observed on all 8 legacy snapshots tested. `restoreProject()` opens every one of them and auto-upgrades it (#1156), but nothing in the error says so, so a user with a 5.x snapshot is told their file has the wrong schema version and left to discover `restoreProject()` on their own.

## Why this is cheap to fix

The detection predicate already exists and is already precise. `.isLegacySnapshot()` in `R/import-legacy-snapshot.R` recognizes exactly this file: no `schemaVersion`, a `projectConfiguration` key, and at least two workbook sections. The load path can call it before raising the schema-version error and, when it matches, name the fix.

## Suggested fix

In the `loadProject()` / `restoreProject()` schema-version failure, test `.isLegacySnapshot()` and, on a match, raise a message that says a previous-version snapshot was detected and points at `restoreProject(<file>, dir = ...)`. Keep the current message for a genuinely unsupported version.

This is adjacent to #1174, which is the same class of "the natural next call does not work and the error does not say what does", but for the manifest filename rather than the schema version.

Part of #1088.

### LINKED ISSUE #1186: Importer raises a raw tibble warning for a missing `Protein Ontogenies` column
Found during migration testing of the 6.0.0 JSON model (tested at commit `00e3df73777584cbad0dfd66af5eef23c3190e91`). The importer surfaces a raw tibble warning to the user instead of a catalog message:

```
Unknown or uninitialised column: `Protein Ontogenies`.
```

Raised on 2 of the 19 legacy projects tested, on an individuals workbook whose biometrics sheet has no `Protein Ontogenies` column.

Two problems. It bypasses `R/messages.R`, against the convention this umbrella settles: user-facing text goes through the catalog and is raised via `cli::cli_warn`. And it is not actionable as written, because it names neither the workbook nor the sheet, and its wording ("uninitialised column") describes an internal tibble access rather than anything the modeller can act on. A user reading it cannot tell whether their protein ontogenies were dropped or whether the column was simply optional and absent.

## Suggested fix

Guard the column access, and if the absence is worth reporting at all, report it through the catalog naming the workbook and sheet. If an absent `Protein Ontogenies` column is simply the normal case for an individual without ontogenies, say nothing.

Part of #1088, under the conventions tracked in #1149.

Felixmil:
The #1149 umbrella covers user-facing message construction and the shared conventions in the v6 project model. This closes it together with its three open message defects, all found during migration testing over real legacy projects. The highest-value one is #1185: opening a previous-version snapshot with `loadProject()` is a migrating user's first contact with 6.0.0, and it never named the call that actually works.

## `loadProject()` on a previous-version snapshot (#1185)

`Project$.readJson()` now tests the existing `.isLegacySnapshot()` predicate before raising the schema-version error, and on a match names `restoreProject()` instead. Every entrypoint that opens a project file reads it through that one method, so `loadProject()`, `Project$new(path)`, the deprecated `ProjectConfiguration(path)` and `reloadProject()` are all covered by the single guard. The inline literal moved into the catalog as `messages$unsupportedSchemaVersion()`; the message for a genuinely unsupported version is unchanged, which is why the three existing regex tests still pass untouched.

The method's aborts also no longer attribute to `private$.readJson()`. Everything it fails on is a property of the file, not of the call that opened it, and the internal method name means nothing to the reader.

## Raw tibble warning from the Excel importer (#1186)

`readExcel()` returns a tibble, and a tibble's `$` warns on an unknown column, so a sheet omitting an optional column leaked `Unknown or uninitialised column: 'Protein Ontogenies'.` to the user. An individual with no protein ontogenies is the normal case, so this reports nothing now, matching how every other optional column in that file is already handled.

**Wider than the issue asked, deliberately.** #1186 names `Protein Ontogenies`, but every column on both sheets was read the same way: the new regression test tripped `Unknown or uninitialised column: proportionOfFemales` immediately. Fixing only the named column would leave the identical defect one column over. Both parsers now read every cell with `[[`, which yields `NULL` for an absent column and is already the idiom in `.parseExcelScenarios()` in the same file. Values are byte-identical, including for the required columns, since the defect was warning-only. The round trip is unaffected: export re-creates the column with `NA` cells, so import -> export -> import is a fixed point.

## Canonicalization warning volume (#1175)

Per-call batching of reference canonicalizations already existed, but each `add*()` / `set*()` still emitted two warnings: the collector's, plus one from `.canonicalizeId()` on the definition's own id, which was never batched. That is the 40-60 an authoring run over a migrated project produced.

The collector is now opened once per authoring call at `private$.impl()`, the single dispatch choke point, and `.canonicalizeId()` feeds the same sink. One warning per call lists every id that call rewrote, each `input -> canonical` pair still on its own bullet, so no precision is lost. The 6 existing in-`_impl` collectors stay in place: the depth counter makes a nested one share the outer sink, and they still batch on the reentrant paths that do not route through `.impl()`.

Three consequences worth a look:

- `suppressWarnings(.canonicalize*())` no longer suffices for the four sites that canonicalize only to compare or re-key, because the pair is now recorded and flushed after the suppressing frame has exited. They use an explicit `.silentlyCanonicalized()` instead.
- `.collectCanonicalizedRefs()` now carries the wrapped expression's visibility. Without that, wrapping the dispatch made every `add*()` print the whole project.
- The grouped warning flushes when the call completes, so where a call warns about something else too, the canonicalization warning now comes second. Visible in the `removePITask()` snapshot.

## The message-helper standard (#1122)

`cliFormat()` is `cli::format_inline(paste(..., sep = "\n"))`. It handles interpolation, the inline classes, pluralization and collapsing; what it cannot do is bullets, because it drops the names of a `cli` vector and glues the elements into one run-on string (`c("Top", "i" = "info")` becomes `"Topinfo"`). It also does not process cli's `\` end-of-line continuation, which the issue did not record.

So this is an audit, not a conversion pass. 33 of the 38 `cli::format_message()` entries are multi-bullet and must stay. Five were single-bullet and are now `cliFormat()`; their snapshot diffs show the old output was doubled, printing the entry's own glyph after the one `cli_abort()` / `cli_inform()` already adds. The rule is stated once in the `R/messages.R` header rather than repeated above each of the 33 entries that keep `format_message()`.

On the growth the issue flagged: `format_message()` went from 20 to 38 through legitimate v6 feature work, four structurally identical `observedData*` / `population*` families of three-bullet diagnostics, not drift. The same commits added `cliFormat()` entries for their single-line messages.

Where the rule lives: `AGENTS.md` does not exist on `json-based-project` (it landed on `main` in #1120), so the rule went into `REFACTORING.md`, which `CLAUDE.md` includes on this branch. `AGENTS.md` on `main` needs the same wording.

## Example

Before and after for #1185, on the bundled `tests/testthat/data/TestProjectExcel/ProjectConfiguration.json` fixture (a real previous-version monolithic snapshot).

Before:

``` r
devtools::load_all("/path/to/esqlabsR-base", quiet = TRUE)

# A project snapshot written by a previous esqlabsR version.
snapshot <- "/path/to/esqlabsR-base/tests/testthat/data/TestProjectExcel/ProjectConfiguration.json"

loadProject(snapshot)
#> Error in `private$.readJson()` at base-wt/R/project.R:238:7:
#> ! Unsupported schemaVersion: "<missing>". Expected "2.0".
```

After:

``` r
devtools::load_all("/path/to/esqlabsR", quiet = TRUE)

# A project snapshot written by a previous esqlabsR version.
snapshot <- file.path("/path/to/esqlabsR", "tests/testthat/data/TestProjectExcel/ProjectConfiguration.json")

loadProject(snapshot)
#> Error:
#> ✖ 
#>   '/path/to/esqlabsR/tests/testthat/data/TestProjectExcel/ProjectConfiguration.json'
#>   is a previous-version project snapshot, not a project of the current format.
#> ℹ A previous-version snapshot has to be upgraded before it can be opened.
#> ℹ Upgrade it into a new folder with `restoreProject(<snapshot>, dir =
#>   <newFolder>)`, which returns the upgraded project.

# ... and that call does open it.
project <- restoreProject(snapshot, dir = file.path(tempdir(), "myProject"))
#> ℹ Detected a previous-version project snapshot and upgraded it to the current
#>   project format.
#> ! Observed data does not travel in a snapshot; add it with `addObservedData()`
#>   if a plot or parameter identification needs it.
length(project$definitions$scenarios)
#> [1] 8
```

Absolute paths shortened to `/path/to/...` for readability; the message text is verbatim.

</details>

## Merging

Merging this into `json-based-project` will not auto-close the issues below, so they need closing by hand afterwards.

Closes #1185
Closes #1186
Closes #1175
Closes #1122
Closes #1149


tests/testthat/test-import-legacy-snapshot.R
```diff
@@ -67,6 +67,26 @@ test_that(".legacySheetToDf reads cells by name and rejects a malformed sheet",
...
+  snapshot <- file.path(dir, "ProjectConfiguration.json")
+  file.copy(.legacySnapshotFixture(), snapshot)
+
+  # The message names the absolute path it was handed, which differs per
+  # machine; only the shape of the message is under test.
+  expect_snapshot(
+    loadProject(snapshot),
+    error = TRUE,
+    transform = \(lines) {
+      sub("'/[^']*/ProjectConfiguration\\.json'", "'<path>'", lines)
```
Felixmil: **This is why `R-CMD-Check / windows-latest (release)` is red on this PR** — not a flake, and not pre-existing on the base.

The pattern anchors the path on a leading `/`, so it never matches a drive-prefixed Windows temp path. From the failing job:

```
── Failure ('test-import-legacy-snapshot.R:81:3'): loadProject on a previous-version snapshot points at restoreProject ──
Snapshot of code has changed:
-   x '<path>' is a previous-version project snapshot, not a project of the current format.
+   x 'C:/Users/RUNNER~1/AppData/Local/Temp/Rtmpqs9NlD/file11b86d1012e1/ProjectConfiguration.json' is a previous-version project snapshot, not a project of the current format.
[ FAIL 1 | WARN 42 | SKIP 3 | PASS 3041 ]
```

`tests/testthat/helpers.R` already has `.redactTmpDir()` for exactly this, and its docs call out the case this hand-rolled regex misses ("both a Unix (`/tmp/...`) and a drive-prefixed Windows (`C:/...`) temp path are matched"). It also normalizes backslashes first. It's already used as a `transform` in four other places, so:

```r
  expect_snapshot(
    loadProject(snapshot),
    error = TRUE,
    transform = .redactTmpDir
  )
```

That changes the committed snapshot line to `'<tmp-path>'`, which needs re-accepting.

---

R/messages.R
```diff
@@ -8,6 +8,18 @@ messages <- ospsuite.utils::messages
...
 # base `stop()`/`warning()`/`message()` on an inline literal string.
 #
+# Two helpers build the text, and the rule between them is capability, not
+# taste: use `cliFormat()` by default, and `cli::format_message()` only where
+# `cliFormat()` cannot express the message. `cliFormat()` is
+# `cli::format_inline(paste(..., sep = "\n"))`, so it formats one inline string:
+# it handles interpolation, the inline classes (`{.file}`, `{.fn}`, `{.val}`,
+# ...), pluralization (`{?s}`, `{cli::qty()}`) and collapsing (`{.or}`), but it
+# drops the names of a `cli` bullet vector and glues the elements into one
+# run-on line, and it does not process the `\\` end-of-line continuation. So a
```
Felixmil: The bullet-dropping half of this rule is right and well worth writing down. The clause **"and it does not process the `\\` end-of-line continuation"** is not — `cliFormat()` does process it, and five entries in this same file already rely on that.

`messages$invalidSimulationTimeArgument()` (line ~735) is a `cliFormat()` call whose template is split across four physical lines with `\\`, and its committed output is clean:

```
tests/testthat/_snaps/scenarios.md:57
! `simulationTime` must be a length-3 numeric vector `c(start, end, resolution)`, or the same grid as a string "0, 42, 48" (several intervals separated by ";"). To give a different grid per id, pass a list with one element per id.
```

Same for `cannotPromptNonInteractive()` → `_snaps/project-lifecycle.md:33`, and `PIInvalidNumericField()` / `PIInvalidScaling()` → `_snaps/parameter-identification.md:294,303,312`. No stray backslash, no embedded newline + indent — which is what an unprocessed continuation would leave behind, since `format_inline()` defaults to `keep_whitespace = TRUE`.

Two knock-ons, which is why this is worth fixing rather than leaving as a stray sentence:

1. The trailing instruction it justifies — *"a single-line message uses `cliFormat()` and, when long, keeps its template on one physical line"* — is then unnecessary, and this PR already acts on it: the converted `outputPathAliasIgnored()` at line 544 became a 118-character single line where the `format_message()` version it replaced was wrapped with `\\`. `air.toml` sets `line-width = 80`, and a formatter can't rewrap a string literal, so these have to be wrapped by hand or not at all.
2. It's mirrored verbatim into `REFACTORING.md:34`, so it becomes a project convention every future message follows.

Suggest dropping the clause and the "one physical line" sentence, keeping the bullet-vector reason (which is the real, correct constraint).

Separately, while the rule is being edited: it presents `cliFormat()` and `cli::format_message()` as the only two options, but there is a deliberate third that `legacySnapshotNotLoadable()` in this very PR uses, and `restoreDirNotEmpty()` (line 199) before it — return the raw templated bullet vector, unglued, so the raising `cli_abort()` interpolates it exactly once. Worth naming, or the new entry reads as a violation of the rule it ships with.

---

R/project-excel.R
```diff
@@ -2140,18 +2140,24 @@ projectStatus <- function(project, silent = FALSE) {
...
+    # tibble, whose `$` warns about an unknown column, so a workbook omitting an
+    # optional column (`Protein Ontogenies` on an individual with no ontogenies,
+    # a biometric a legacy sheet never carried) would leak a raw tibble warning
+    # to the user. `[[` yields `NULL` for an absent column, which `.naToNull()`
+    # turns into an absent field. Matches `.parseExcelScenarios()` below.
     indiv <- list(
-      individualId = as.character(row$IndividualId),
-      species = as.character(row$Species),
-      population = as.character(row$Population),
+      individualId = as.character(row[["IndividualId"]]),
```
Felixmil, line 2155: Agreed on widening this past the one column #1186 named. One consequence the PR body doesn't cover, though: the warning that goes away was also the only signal for a missing **required** column.

`individualId` / `species` / `population` here (and `populationId` / `species` / `population` in `.parseExcelPopulations()`) aren't wrapped in `.naToNull()` — they're required, and nothing upstream validates the sheet's columns: `readExcel()` doesn't, and the call site at line 273 hands the frame straight to this parser.

So for a workbook whose `IndividualBiometrics` sheet is missing `IndividualId`:

- **before**: `Unknown or uninitialised column: 'IndividualId'.` — ugly, but it names the problem
- **after**: silence, `individualId = character(0)`, and then line 323 `.canonicalizeOneId(indiv$individualId)` → `if (is.na(indivCanonical))` on a `logical(0)` → `argument is of length zero`, blaming nothing

Values really are byte-identical, as the PR body says; it's the diagnostic that regresses. A required-column check at the top of each parser would fix it and be a strict improvement over the tibble warning:

```r
.requireExcelColumns(indivDf, c("IndividualId", "Species", "Population"), "IndividualBiometrics")
```

Minor, related: line 2178 still reads `row$ParameterSets`. It's safe because `hasParameterSets` guards it, but it does read oddly right under a comment that says cells are read with `[[`, not `$`.

---

R/project.R
```diff
@@ -1172,7 +1172,15 @@ Project <- R6::R6Class(
...
-      fn(self, private, ..., .call = rlang::caller_env(2))
+      # Resolve the attribution call as its own statement, before the collector
+      # wraps the dispatch: `caller_env(2)` is read from this method's frame
+      # either way, and naming it here keeps that independent of when the wrapped
+      # expression is forced.
+      call <- rlang::caller_env(2)
+      # One canonicalization collector per authoring call, so a call over
+      # non-canonical ids reports the definition's own id and every reference it
+      # names in a single warning instead of one per id.
+      .collectCanonicalizedRefs(fn(self, private, ..., .call = call))
```
Felixmil, line 1183: I checked the two mechanically risky parts of this and both hold up: hoisting `call <- rlang::caller_env(2)` out of the argument resolves to the same frame the promise did (it's read from this method's frame either way), and `withVisible(force(expr))` does carry `invisible(self)` through free function → method → `_impl`.

The consequence worth a second look is the flush ordering the PR body flags. On the success path the grouped warning now lands *after* whatever else the call warned about, and `removePITask()` is a case where that reads backwards:

```
tests/testthat/_snaps/parameter-identification.md:229
  Warning:
  PI task "notthere" not found; no-op.
  Warning:
  Canonicalized 1 id to a safe form:
  * "NotThere" -> "notthere"
```

The user typed `"NotThere"`; the rename is the reason the first warning quotes something else, so it's the explanation arriving after the thing it explains. The old order had it right.

Not obviously worth complicating `.collectCanonicalizedRefs()` over — but if it is, flushing before the impl's own body runs isn't possible (the pairs aren't collected yet), so the practical option is flushing eagerly per `.canonicalizeId()` batch once the sink already holds pairs, or keeping the order and accepting it. Happy either way; mostly flagging that this one snapshot is the case that reads worst, rather than a neutral reorder.

---

R/messages.R
```diff
@@ -125,6 +137,10 @@ messages$fileNotFound <- function(filePath) {
   cliFormat("File not found: {.file {filePath}}")
 }
 
+messages$unsupportedSchemaVersion <- function(version) {
+  cliFormat("Unsupported schemaVersion: {.val {version}}. Expected {.val 2.0}.")
```
Felixmil: Small inconsistency with the entry 90 lines below. `legacySnapshotNotLoadable()` deliberately leaves `{jsonPath}` unglued, with a comment explaining that pre-gluing would let a path containing `{` or `}` be glue-parsed a second time by `cli_abort()`. This entry does the pre-gluing: `cliFormat()` interpolates `version` into a finished string, which `cli::cli_abort()` at `R/project.R:1391` then glue-parses again.

The old inline literal didn't have the problem — it was a single-pass `cli_abort()` template.

Reaching it takes a `Project.json` whose `schemaVersion` value contains a brace, so this is a corrupt/hand-edited file rather than anything a user hits normally — and the failure is a confusing glue error instead of the version message. Cheap to make consistent with the neighbour:

```r
messages$unsupportedSchemaVersion <- function() {
  "Unsupported schemaVersion: {.val {version}}. Expected {.val 2.0}."
}
```

...with the value bound under `version` in the raising frame. Or leave it and note why the hazard is acceptable here but not for `jsonPath` — right now the two entries reason oppositely about the same thing.

---

R/definition-id.R
```diff
@@ -220,19 +202,37 @@
...
 .canonRefSink <- new.env(parent = emptyenv())
 .canonRefSink$depth <- 0L
+.canonRefSink$silent <- 0L
 .canonRefSink$inputs <- character()
 .canonRefSink$canonicals <- character()
 
+# Canonicalize without reporting the change. Some callers canonicalize only to
+# compare or to re-key, where the transform is an internal detail the user did
+# not ask for: re-keying a section read off disk, matching a reference against
+# the id it was filed under. `suppressWarnings()` alone does not cover those any
```
Felixmil: "`suppressWarnings()` alone does not cover those **any more**" describes the change rather than the code, which the comment rule in `CLAUDE.md` asks comments not to do ("describe how the code works currently, never its history"). The reason itself is exactly the non-obvious constraint a comment should carry, so it just needs the tense fixed:

> `suppressWarnings()` does not cover those: inside a collector the pair is recorded rather than warned, and would surface later, from the flush, outside the suppressing frame.

---

Felixmil (commented):
Reviewed the four commits against `362775f`, with a clean working tree.

**One blocker: `R-CMD-Check / windows-latest (release)` is failing on this PR**, on the new `loadProject()` snapshot test. The snapshot transform anchors the path on a leading `/`, so it doesn't redact `C:/Users/RUNNER~1/...` and the snapshot mismatches. `.redactTmpDir()` in `tests/testthat/helpers.R` already handles the drive-prefixed form. Details inline. Ubuntu and macOS are green, so this is Windows-only.

The mechanically risky parts of the canonicalization change hold up on inspection: `withVisible(force(expr))` preserves `invisible(self)` through the free-function → method → `_impl` chain, the hoisted `call <- rlang::caller_env(2)` resolves to the same frame the lazy promise did, the six nested in-`_impl` collectors correctly no-op via the depth guard, all four `suppressWarnings()` → `.silentlyCanonicalized()` conversions are the complete set that needed converting (no `suppressWarnings(.canonicalize*)` left in `R/`), and `.naToNull()` / `.blankToDefault()` both handle the `character(0)` an absent `[[` column produces. The grouped-warning behaviour is visible working in the CI log — `addPITask()` emits one warning through the new `.impl()` collector.

Five other findings, none blocking:

| | |
|---|---|
| `R/messages.R:18` | The new rule's claim that `cliFormat()` doesn't process the `\` end-of-line continuation is false — five existing `cliFormat()` entries use it and render clean in committed snapshots. It's now also a convention in `REFACTORING.md:34`, and it's what motivates the 118-char single line at `R/messages.R:544`. |
| `R/project-excel.R:2150` | The `[[` switch also removes the warning for a missing **required** column, and nothing validates the sheet's columns, so a sheet with no `IndividualId` imports silently and dies at line 323 with `argument is of length zero`. |
| `R/messages.R:141` | `unsupportedSchemaVersion()` pre-glues its template, so a braced `schemaVersion` gets glue-parsed twice — the hazard the entry 90 lines below documents and deliberately avoids. |
| `R/project.R:1183` | Deferring the flush puts the canonicalization warning after the warning it explains, and `removePITask()` is the case that reads worst. |
| `R/definition-id.R:224` | Comment describes history ("does not cover those any more"), against the `CLAUDE.md` comment rule. |

Nothing here questions the shape of the change — the `.impl()` choke point is the right seam for the collector, and widening #1186 past the one named column was clearly correct.

---

tests/testthat/test-import-legacy-snapshot.R
Felixmil: Right diagnosis, and fixed in d16d3383 — but not with `.redactTmpDir()`, because that helper does not match this path. I tried it first and it left the path intact.

`.redactTmpDir()` anchors on the escaped `tempdir()` prefix, and on macOS `tempdir()` comes back with a doubled separator:

```
tempdir()                 #> /var/folders/_6/…/T//Rtmp7AhFey
```

The path in the message went through `fs::path_abs()` in `.readJson()`, which normalizes that to a single slash (`…/T/Rtmp…`), so the anchored prefix never appears in the text being redacted. Its backslash normalization does not help, since the mismatch is the `//` in the prefix itself, not the slash direction. That is a pre-existing limitation of the helper for any `fs`-normalized path, and widening it would touch the four assertions already relying on it, so I left it alone.

The fix anchors on the filename instead of a leading `/`, which covers both platforms:

```r
sub("'[^']*ProjectConfiguration\\.json'", "'<path>'", lines)
```

Verified against both forms:

```
windows: x '<path>' is a previous-version project snapshot.
macos  : x '<path>' is a previous-version project snapshot.
```

Committed snapshot stays `'<path>'`.

R/messages.R
Felixmil: Half right, and the half that is right matters — the rule is corrected in d16d3383.

`cliFormat()` genuinely does not resolve the continuation. Rendering the entry you cite directly:

```
> messages$invalidSimulationTimeArgument()
`simulationTime` must be a length-3 numeric vector \
    `c(start, end, resolution)`, or the same grid as a string \
    "0, 42, 48" (several intervals separated by ";"). To give a \
    different grid per id, pass a list with one element per id.
```

The backslash and the newline+indent survive. The committed snapshot is clean because the raising `cli_abort()` glue-parses that finished string a **second** time and resolves the continuation then. So the clean output is real, but it comes from the call site, not from `cliFormat()`.

Your conclusion holds either way: the "one physical line" instruction was unnecessary, so it is gone and `outputPathAliasIgnored()` is wrapped with `\` again. That re-wrap produced **zero** snapshot change, which is the same second pass showing up as evidence.

The rule now states the mechanism rather than the symptom, and your second point is taken: it documents the third form (return the raw unglued template when an interpolated value may contain `{` or `}`) and names `restoreDirNotEmpty()` and `unsupportedSchemaVersion()` as the entries that use it, so `legacySnapshotNotLoadable()` no longer reads as a violation of the rule it ships with.

Worth noting the second pass is exactly the hazard that form exists to avoid — the two findings are the same mechanism seen from opposite ends.

R/project-excel.R
Felixmil: Confirmed and fixed in d16d3383. Reproduced before fixing:

```
> .parseExcelIndividuals(dplyr::tibble(Species = "Human", Population = "…", Gender = "MALE"))
individualId: character(0)
> .canonicalizeOneId(out[[1]]$individualId) …
ERROR: argument is of length zero
```

So the diagnostic regressed exactly as described. `.requireExcelColumns()` now runs at the top of both parsers:

```
Error:
x The "IndividualBiometrics" sheet is missing required columns: IndividualId and Population.
i Add them to the workbook, or re-export the project with `exportProjectToExcel()` to get a
  sheet with the columns this version reads.
```

Required sets are `IndividualId` / `Species` / `Population` and `PopulationName` / `species` / `population` — the fields that are not `.naToNull()`-wrapped, which is the same test you used. Two details beyond the suggestion: the abort is attributed to no function rather than to `.requireExcelColumns()`, whose name means nothing to the reader, and the catalog entry returns its template unglued so `cli_abort()` renders real bullets instead of re-wrapping one pre-formatted string with the glyphs inline.

`row$ParameterSets` is `row[["ParameterSets"]]` now too.

R/project.R
Felixmil: Thanks for checking those two — agreed on both.

Keeping the order as it is. Flushing eagerly once the sink already holds pairs would put the definition's own id back in its own warning and leave the references for the flush, which is two warnings per call again — the exact split #1175 asked to remove. The ordering is the cost of grouping, and `removePITask()` is the case where it reads worst because the rename explains a quote in the *earlier* warning.

Both facts still reach the user in the same output, and the values name the connection, so I would rather pay that than reintroduce the volume.

R/messages.R
Felixmil: Valid, and fixed in d16d3383 — the two entries did reason oppositely about the same double-pass hazard. `unsupportedSchemaVersion()` returns its template unglued now, with `version` bound in the raising frame at `R/project.R`, and a comment pointing at the neighbour so the shared reason is visible from both.

Your wider point drove the rule change in the other thread: this unglued form is a deliberate third option, not an exception, so the catalog header now documents it as such.

R/definition-id.R
Felixmil: Fixed in d16d3383, with your wording.

Issue #1213, opened 2026-08-03
6.0.0 migration testing over 19 legacy projects: 26 findings, 2 regressions against 5.x


Felixmil:
Migration testing of the 6.0.0 JSON project model over 19 real legacy projects at commit `2b4f1a9d3febedb78f53c2a31763e2cc745605be` (the merge of #1212). 18 of the 19 completed: 8 legacy monolithic snapshots exercised through `restoreProject()`, and 10 Excel-first projects imported with `importProjectFromExcel()` and independently rebuilt by hand with the `add*()` API, then diffed and run. Every finding below is confirmed on at least two independent projects. Projects are referred to only by shape.

esqlabsR was installed the way a user installs it (`remotes::install_github("esqLABS/esqlabsR@json-based-project")` into a clean library), and every simulation result quoted here came from a live .NET backend.

## What is already working

Worth stating first, because the list below is long and the migration itself is not the problem.

- **All 8 legacy snapshots upgrade and round-trip losslessly.** Each one auto-upgrades through `restoreProject()`, writes a v6 `definitions/` tree with container `Project.json` and `schemaVersion` `"2.0"`, and a `snapshotProject()` then `restoreProject()` round trip reproduces every definition file byte for byte (578/578, 234/234, 178/178, 161/161, 92/92, 62/62, 36/36, 26/26).
- **The importer and the authoring API converge.** Four Excel projects produced byte-identical trees from the two independent paths (1177 of 1178, 434/434, 25/25, 25/25), and three produced bit-identical simulation results between paths. Across roughly 2300 shared definition files there are only two content-bearing disagreements, both listed below.
- **26 of the 27 regression checks from the two prior runs are verified fixed on real projects**, several of them on evidence the earlier runs could not produce (for example the animal-individual fix confirmed by an actual solve rather than a parse, and the parameter-set collision fix confirmed on a naturally occurring collision).

The defects below are therefore mostly *not* migration defects. They cluster in three places: assets that do not travel, parameter identification, and run-time behaviour that validation does not predict.

### 1. Every protein ontogeny in every legacy workbook is silently dropped, and #1186's fix removed the only signal

`R/project-excel.R:2165` and `:2214` read a single column named `Protein Ontogenies`. Real 5.x workbooks carry a two-column `Protein` + `Ontogeny` pair, so the lookup returns nothing and every ontogeny is discarded.

Three independent demonstrations:

[...]

PR #1214, opened 2026-08-03
Key the run's population cache on how the population is produced, not on its id alone

Felixmil:
The run's population cache was keyed on the population id alone, so two scenarios that share a population id but resolve it differently, one from its demographics spec and one from its csv file, both got whichever population the batch resolved first. Nothing failed and nothing warned: the second scenario silently ran the first one's population, and which one that was depended on the order the scenarios were listed in. This keys the cache on how the population is produced instead of on its id, so every scenario gets the population it names whatever the order, and warns when one id stands for two different populations in a batch.

## Cache key

The key is now the population id, the effective population type, and the declared csv file, so a spec-resolved and a csv-resolved population under one id coexist in one run's cache instead of overwriting each other.

The effective type is the part that had to be included: it is not a property of the population entry alone. An entry that carries an explicit `type` (`csv` or `programmatic`) decides for itself, but a demographics-spec entry falls back to the *scenario's* `readPopulationFromCSV` flag, which is how one population id ends up meaning two different populations within a single run.

The declared csv file is in the key too. It is redundant today, since a section holds one entry per id and the populations folder is fixed for a run, but it means the key names every input that decides the resolved path rather than relying on that.

## Injected populations

The injected path (`addPopulation(project, id, <Population>)`) is covered by the same key and cannot be crossed with a spec or csv population of the same id. A programmatic entry carries an explicit `type`, so a scenario's `readPopulationFromCSV` flag cannot turn it into a csv read, and the populations section holds exactly one entry per id, so the effective type for a given id is fixed. A test asserts that an injected population reaches both a plain population scenario and a `readPopulationFromCSV` scenario unchanged.

## Warning

One population id resolving two ways in a run is safe now, but it is almost always unintended, typically `readPopulationFromCSV` set on one scenario and not on another that shares the population. It therefore warns in its own right, naming the population, the scenario, and how that scenario resolved it. The warning fires once per additional way an id resolves, not once per scenario.

## Ordering experiment

Both orderings run in one batch on the bundled `TestProject` fixture, whose `testpopulation` has a demographics spec set to 7 individuals and a `Populations/testpopulation.csv` holding 2. `populationscenario` resolves it from the spec and should report 7; `populationscenariofromcsv` sets `readPopulationFromCSV` and should report 2.

Before:

```
run order: spec, csv
  populationscenario        (spec, 7 individuals): 7
  populationscenariofromcsv (csv,  2 individuals): 7
run order: csv, spec
  populationscenario        (spec, 7 individuals): 2
  populationscenariofromcsv (csv,  2 individuals): 2
```

Whichever scenario ran first won, so the answer was wrong in both directions and silent in both.

After:

```
run order: spec, csv
  populationscenario        (spec, 7 individuals): 7
  populationscenariofromcsv (csv,  2 individuals): 2
run order: csv, spec
  populationscenario        (spec, 7 individuals): 7
  populationscenariofromcsv (csv,  2 individuals): 2
```

Plus the new warning on each run. A batch with no collision resolves and caches exactly as before: the key gains components, and for an id that resolves one way it still resolves and caches once.

Item 3 of #1213. No closing keyword, since that issue is an umbrella covering many unrelated items.


R/messages.R
```diff
@@ -596,6 +596,21 @@ messages$noPopulationsFolderForCSVPopulation <- function(
   ))
 }
 
+messages$populationIdResolvedTwoWays <- function(
+  populationId,
+  scenarioName,
+  effectiveType
+) {
+  cli::format_message(c(
```
Felixmil, line 604: `populationIdResolvedTwoWays` interpolates ids through `cli::format_message()` and is then raised with `cli::cli_warn()` (`R/scenario-execution.R:258`), so the ids get glue-parsed twice. Both helpers glue eagerly and the raising call parses the finished string again, so a `populationId` or `scenarioName` containing `{` or `}` is evaluated as an R expression on that second pass — the warning itself errors, and under the default `stopIfFails = TRUE` that takes the batch down.

Braces do survive canonicalization: `.canonicalizeOneId()` substitutes `[/\\:*?"<>|,[:space:][:cntrl:]]` (`R/definition-id.R:380`) and `{`/`}` are in none of those classes, so `pop{1}` reaches this message intact.

This file's own header prescribes the fix for exactly this case — return the raw templated vector unglued so the raising call interpolates once from its own frame — and `.canonicalizedIdBullets()` exists for the same reason.

In fairness this isn't a regression the PR introduces: the neighbouring `noPopulationsFolderForCSVPopulation` has the identical shape (`format_message` + `cli_abort`, same two id arguments), so the pattern predates this change and the new entry is locally consistent. Worth fixing both together rather than singling this one out.

---

NEWS.md
```diff
@@ -104,6 +104,7 @@ Deprecation intensity follows the lifecycle policy: forwarding shims that still
...
 - `runScenarios()` and `buildSimulations()` now canonicalize the `scenarios` argument (the same lowercasing and character-substitution scenario ids get) before matching, so the name you authored a scenario with resolves to its canonical id; a name rewritten in the process warns, naming the id it resolved to. (#1157)
+- `runScenarios()` and `buildSimulations()` give each scenario the population it asks for when several scenarios in one batch share a population id but resolve it differently, which happens when only some of them set `readPopulationFromCSV`: the demographics spec is built for the scenarios without the flag and the population csv is loaded for the scenarios with it. Since one population id standing for two different populations is usually unintended, a warning names the id and the scenario. (#1213)
```
Felixmil, line 107: The bullet describes the fix but not the one case where the fix turns a passing run into a failing one, which is the part a user upgrading needs to act on.

A legacy project that shares a population id between a spec scenario and a `readPopulationFromCSV` scenario, with no csv actually present, used to complete: the collision meant the csv scenario was handed the cached spec population and `.resolveCsvPopulation()` never ran. Now each scenario resolves its own source, so that scenario reaches `.resolveCsvPopulation()` (`R/scenario-execution.R:289`) and hard-aborts — on the undeclared-`populationsFolder` guard, or in `loadPopulation()` on the missing file.

That is the correct behaviour; the previous silence was the bug. But it reads as an unexplained new failure unless the bullet says the remedy: add the population csv, or clear `readPopulationFromCSV` on the scenario that shouldn't have it. The collision warning text has the same gap — it explains that each scenario gets what it asks for, which is reassuring right up until one of them can't be resolved at all.

---

Felixmil (commented):
Post-merge review of `b666955` (the PR merged while this was running). Nothing here is a correctness problem with the cache-key change itself — the core fix is sound and I verified the parts most likely to break:

- **The `\r` separator is safe.** `.canonicalizeOneId()` maps every control character to `_` (`R/definition-id.R:380`), so no canonical id can contain `\r`, and the delimiter keeps a prefix id from matching a longer one — `pop1` cannot collide with `pop10\r...`.
- **The key captures everything that decides resolution.** `effectiveType` is the only component that can come from the *scenario* rather than the entry, and `popData$file` matches what `.resolveCsvPopulation()` actually reads, including its `<populationId>.csv` fallback (already covered by the id in the key). Only two keys per id are reachable, so the warning fires at most once per id per batch, as the description claims.
- **Moving the cache lookup after the `popData` fetch is harmless.** `NULL[["x"]]` is `NULL` and `project$definitions` is a cheap lazy field group.
- **Empty-cache edge cases hold.** `names(cache$populations) %||% character()` and `any(startsWith(character(0), ...))` both behave.
- **Test fallout is contained.** The only pre-existing tests that build every fixture scenario are `test-scenario-execution.R:614` (now `suppressWarnings`) and `:252` (already suppressed); the rest mock `.prepareScenario` or pass a single scenario. The bundled `Example` project has one population used one way, so it is unaffected. New snapshot ordering matches run order.

Three follow-ups, none blocking, two inline:

1. **`populationIdResolvedTwoWays` double-glues its ids** (`R/messages.R:604`) — a latent abort on an id containing `{`/`}`. Pre-existing pattern shared with the adjacent entry, so best fixed as a pair.
2. **The NEWS bullet omits the one newly-failing case** (`NEWS.md:107`) — a shared id plus a missing csv now aborts where it previously completed, and neither the bullet nor the warning names the remedy.
3. **Stale header comment on `.buildScenarioSimulations()`** (`R/scenario-execution.R:555`, outside this diff so not inline): it still says two scenarios referencing the same population "build it once", which the per-source key makes false. The sibling comment at `:218` was updated; this one was missed.

---

PR #827, opened 2025-04-28
Protein ontogenies

PavelBal:
Fixes #82
- Protein ontogenies for individuals and populations are now defined in one column
`Protein Ontogenies` in the sheets `IndividualBiometrics` (for individuals) or `Demographics`
(for populations). The columns `Protein` and `Ontogeny` are no longer supported. The new column
`Protein Ontogenies` is a comma-separated list of protein names and ontogeny names pairs. For example: 
`CYP3A4:CYP3A4,CYP2D6:CYP2C8` will create a CYP3A4 ontogeny for the protein CYP3A4 and a CYP2D6 ontogeny for the protein CYP2C8.

PR #1215, opened 2026-08-03
Add a legacy-shaped Excel fixture so the #1213 migration defects become testable

### LINKED ISSUE #1186: Importer raises a raw tibble warning for a missing `Protein Ontogenies` column

Felixmil:
Migration testing over 19 real legacy projects (#1213) found defects that the fixtures in this repository are structurally incapable of reproducing. The clearest case: the importer reads a single `Protein Ontogenies` column, real pre-5.6 workbooks carry a two-column `Protein` + `Ontogeny` pair, and the bundled `tests/testthat/data/TestProjectExcel/` fixture survives that code path only because it uses the newer single-column spelling. That is why no test caught the silent loss, and why #1212 could close #1186 as fixed while removing the one warning that leaked it. This adds a second Excel fixture shaped the way legacy workbooks actually are, plus characterization tests that pin today's behaviour so Waves 1-3 have real coverage to fix against.

**No behaviour changed.** Nothing under `R/` is touched, and neither existing fixture is modified. The diff is one new fixture, one generator, four test helpers and 25 tests.

## The fixture

`tests/testthat/data/TestProjectExcelLegacy/`, a plausible pre-5.6 project. Each trait is one the sibling fixture has in its modern spelling, which is what made these defects unreachable:

- Ontogenies in the two-column `Protein` + `Ontogeny` pair, each cell a comma list zipped pairwise with the other. One individual declares one ontogeny, one declares two, and one population declares two. That is genuinely the pre-5.6 layout, removed in #827.
- No `OverwriteFormulasInSS` column at all, and one scenario with a blank `SimulationTimeUnit` cell.
- Multi-value cells quoted in the 5.x style, one of them holding a value that itself contains a comma, and `DataCombinedName` quoted on both the defining and the referencing side.
- `nsd` stored as text, as a hand-maintained workbook routinely stores it.
- `PopulationsCSV/` under the configurations folder with a mixed-case filename, and a scenario that reads it.
- The 5.x parameter-identification sheet layout, keeping its `Group` column (with two rows sharing a group), its `ObservedDataSheet` column, a legacy `Units` cell, and a `DataSet` naming an observed data set the project does not define.

The 7 MB `Aciclovir.pkml` is **not** duplicated. `localLegacyExcelProject()` injects the sibling fixture's copy into each throwaway project, so the fixture is a complete runnable project when a test uses it and adds 112 KB rather than 7 MB to the repository.

`data-raw/TestProjectExcelLegacy.R` regenerates every workbook with `writexl` and documents each legacy trait, so the committed binaries stay reviewable as text. It is covered by `.Rbuildignore`'s existing `^data-raw$`.

## Helpers

Four, in `tests/testthat/helpers.R`, alongside the existing fixture accessors:

- `localLegacyExcelProject()` copies the fixture to a throwaway directory and injects the model.
- `legacyExcelProjectPath()` names the entry workbook, mirroring `testProjectExcelPath()`.
- `editWorkbookSheets(path, edit)` rewrites one workbook through a function of all its sheets, which is how each variant is derived from the base by changing exactly one dimension.
- `importLegacyExcelProject()` imports and loads, returning the project, the output directory, and every warning the import raised. The warnings are returned rather than suppressed because most of what these tests pin is *silence*, and a test asserts that nothing mentions its own subject.

## What each part unblocks

Base fixture, in `test-project-excel.R`:

| #1213 item | Pinned as |
|---|---|
| 1 — ontogenies dropped | every declared ontogeny imports as absent, with no warning; and a population twin differing in nothing else imports to a byte-identical definition file apart from the id |
| 5 — populations folder does not travel | `PopulationsCSV/` is absent from the imported project under either spelling, the report is suppressed, and validation still calls the project clean |
| 8 — `Group` discarded | two rows that shared a group become `tsspec` and `tsspec_2`, three free parameters where v5 estimated two |
| 9 — no `OverwriteFormulasInSS` | the absent column imports as `FALSE` |
| 13 — dangling mapping `observedData` | the reference is carried through, resolves to nothing, and `validateProject()` reports 0 critical errors |
| 14 — legacy `Units` enforced | the `mg` cell v5 ignored is carried onto the parameter |
| 18 — no plots type contract | a text-typed `nsd` imports as the string `"1.96"` |
| 19 — `simulationTimeUnit` default | a blank cell imports as `null` while `addScenario()` defaults to `"h"` |
| 10 / #1207 residue | a quoted multi-value cell splits correctly, and a consistently quoted single-value cell yields the id `_name_` rather than the authored name |

Variants, each mutating one workbook of a copy:

| #1213 item | Mutation | Pinned as |
|---|---|---|
| 8 — absent workbook | delete `Populations.xlsx` | the section imports as 0, nothing names the missing workbook, and the summary omits the zero-count section |
| 9 — duplicated headers | duplicate a `Value` header | the sheet is reported and that is all; the individual named after it loses its parametrization while its sibling keeps one |
| 9 — no biometrics row | drop one biometrics row | the individual disappears, its parameter set is left ownerless, and the scenario dangles |
| 10 — duplicate plot id | reuse a `plotID` | three rows become two plots, the later row wins, and no duplicate is reported |
| 16 — invisible Unicode | ids with U+200B and U+00A0 | both survive into the ids and the definition filenames, two files render identically, and validation is clean |

Outside the importer:

| #1213 item | Where | Pinned as |
|---|---|---|
| 2 — two ontogenies crash | `test-scenario-execution.R` | `.readOntogeniesFromList()` on a length-2 value aborts (snapshotted); one ontogeny and a comma-joined pair still read, so the abort is confined to the multi-value case |
| 6 — lowercased CSV name | `test-scenario-execution.R` | the derived filename is `csvpop.csv` while the project carries `CsvPop.csv`, asserted on the strings so it states the mismatch on every platform rather than passing on macOS |
| 16 — canonicalization | `test-definition-id.R` | `.canonicalizeOneId()` maps an ordinary space to `_` but leaves U+00A0, U+200B, U+2060, U+200D and U+FEFF intact |
| 25 — `did you mean` cap | `test-definition-id.R` | a `_mean` or per-analyte suffix is over the distance cap, so the hint cannot fire in the shapes that occur, while a one-character typo still is suggested |

## Notes

Every test above passes against today's behaviour and names the finding it covers, so each one flips visibly when its fix lands. No `expect_failure`, no skips.

On local verification: the full suite ran green (`[ FAIL 0 | WARN 42 | SKIP 1 | PASS 3122 ]`) with all four test files in place, and `test-project-excel.R` ran green again (323 passing) after the ontogeny pin was strengthened. The last two commits, which tighten one assertion each, were not re-run locally, so CI is the first full check of the final state.

Three deliberate limits, rather than a quietly narrowed fixture:

- **No scenario is run.** Every item pins at a seam that needs no solve, so a run would add minutes without adding coverage. The .NET backend does work here (`loadSimulation()` and `runSimulations()` verified on the sibling fixture's `Aciclovir.pkml`), so a smoke test is easy to add if wanted.
- **Item 8's v5 merge is pinned as an outcome, not a comparison.** Nothing in this repository can execute v5, so the test asserts the three parameters v6 produces and states in a comment that v5 produced two.
- **Item 26's `exportConfiguration` sheet is present but empty.** The legacy per-grid output filename / width / height sheet has no v6 home and is dropped without a word, but that item is outside this fixture's brief, so the sheet is header-only as in the sibling fixture rather than carrying rows no test asserts on.

Part of #1088. Refs #1213.


tests/testthat/test-project-excel.R
```diff
@@ -3163,3 +3163,481 @@ test_that("a sheet whose every row is skipped becomes no parameter set", {
...
 })
+
+# Legacy Excel fixture: characterization of the #1213 migration findings ----
+#
+# These tests pin what the importer does *today* with the workbook shapes real
+# pre-5.6 projects carry, using the `TestProjectExcelLegacy/` fixture. Several of
+# the behaviours pinned here are the defects reported in #1213: each such test
+# passes against the current behaviour and names the finding it covers, so it
+# flips visibly when that finding is fixed rather than quietly keeping a green
+# suite over a silent data loss.
```
Felixmil, line 3174: The block header no longer describes every test under it.

It promises that each test "passes against the current behaviour and names the finding it covers, so it flips visibly when that finding is fixed". That was true when the block was written, but several of the findings it covers have since been fixed on the base branch, and the tests below were updated to assert the *repaired* behaviour instead:

- `a legacy two-column ontogeny pair imports its ontogenies` (3186) and `two populations differing only in their ontogenies import differently` (3211) — #1213 item 1, fixed by 01eacc0 (#1217). They assert the ontogenies *arrive*, not that they are dropped.
- `the populations CSV folder travels with an imported project` (3398) — #1213 item 5, fixed by ce1e55c (#1218). It asserts the folder *travels*.

So the block now holds two kinds of test — pins of behaviour still considered wrong, and regressions guarding a landed fix — and the header only describes the first. The practical cost is misreading a failure: a maintainer who sees the ontogeny test go red will read the header and conclude "good, someone fixed item 1", when in fact #1217 just regressed.

Suggest splitting the header into the two statements, or tagging each test as `pins` vs `guards`.

---

tests/testthat/test-project-excel.R
```diff
@@ -3163,3 +3163,481 @@ test_that("a sheet whose every row is skipped becomes no parameter set", {
...
+  expect_equal(summary$total_critical_errors, 0)
+})
+
+# #1213 item 19: a blank `SimulationTimeUnit` cell imports as `null`, while
+# `addScenario()` defaults the same absent value to `"h"`. So the same blank cell
+# means a different unit depending on which entrypoint wrote the project.
+#
+# #1213 item 9 in the same fixture: the scenarios sheet has no
+# `OverwriteFormulasInSS` column at all, and the absent column imports as `FALSE`
+# rather than as an absent field.
```
Felixmil, line 3323: This half of the comment describes a divergence that does not exist, and the assertion it justifies (`expect_false(scenario$overwriteFormulasInSS)`, 3342) pins nothing.

The claim is that the absent `OverwriteFormulasInSS` column "imports as `FALSE` rather than as an absent field". The importer does the opposite — `R/project-excel.R:1996-2005` wraps the lookup in `.naToNull()`, so an absent column becomes `NULL`, and `R/project-to-json.R:354` then writes no key at all. The field *is* absent in the JSON, exactly as the comment says it should be. The `FALSE` the test observes is added on the way back in by `R/scenarios.R:258` (`entry[["overwriteFormulasInSS"]] %||% FALSE`), which is the same default a hand-authored scenario gets from `addScenario(overwriteFormulasInSS = FALSE)` (`R/scenarios.R:100`). Both entrypoints agree, so there is nothing here of the kind `simulationTimeUnit` has in the paragraph above.

Also, the `#1213 item 9` tag is wrong: item 9 is "an individual can silently lose its parameter set", which the two tests at 3479 and 3514 already cover correctly. There is no #1213 item about `OverwriteFormulasInSS` (the only mention in the issue is the `setScenario(overwrite = )` partial-match bullet under item 26, which is a different defect).

Suggest dropping the second paragraph and the 3342 assertion, and renaming the test to just the `simulationTimeUnit` subject — item 19 on its own is a real and well-pinned finding.

---

tests/testthat/test-project-excel.R
```diff
@@ -3163,3 +3163,481 @@ test_that("a sheet whose every row is skipped becomes no parameter set", {
...
+    fixed = TRUE
+  )
+  expect_false(any(grepl("proteinOntogenies", withoutOntogenies, fixed = TRUE)))
+
+  # Both files were written, so the comparison above is a real difference.
+  expect_true(any(grepl("adultpop", withOntogenies, fixed = TRUE)))
+  expect_true(any(grepl("adultpopnoonto", withoutOntogenies, fixed = TRUE)))
+})
+
+# #1213 item 8: v5 merged `PIParameters` rows by
```
Felixmil, line 3256: Wrong item number, and it collides with a test that has the right one.

This test is about `PIParameters` rows merged by `(Group, Container Path, Parameter Name)`, which is **#1213 item 4** ("Legacy parameter-identification `Group` is discarded, changing what the identification estimates"). #1213 item 8 is "Unconsumed configuration workbooks are never reported…", which is what the test at 3445 covers — correctly tagged `item 8`. As it stands the issue's most severe PI finding has no test pointing at it, and item 8 appears to have two unrelated tests.

The same mistake is at 3356 (`#1213 item 10 / the residue #1207 recorded`): the quoted-single-value-cell subject is the "#1207 residue" paragraph in the issue's closing section, not item 10. Item 10 is the duplicate plot id, correctly tagged at 3551. Dropping the `item 10 /` prefix there fixes it.

Cross-checked the other tags in this block and in `test-definition-id.R` / `test-scenario-execution.R`; items 1, 5, 6, 9, 13, 14, 16, 18, 19, 25 and 2 all line up.

---

data-raw/TestProjectExcelLegacy.R
```diff
@@ -0,0 +1,493 @@
...
+      stringsAsFactors = FALSE
+    )
+  ),
+  file.path(fixtureDir, "ProjectConfiguration.xlsx")
+)
+
+# Individuals.xlsx ----
+# `Adult` declares one ontogeny and `Child` two, through the legacy column pair.
+# Two ontogenies matter on their own: a repaired importer would hand a length-2
+# value to `.readOntogeniesFromList()`, which cannot take one.
```
Felixmil, line 123: Stale rationale — this is the reason the fixture gives for `Child` carrying two ontogenies, and it no longer holds.

Neither half is true of the current code. `.excelProteinOntogenies()` folds the pair into a single comma-joined string (`R/project-excel.R:3588`, `paste(paste0(proteins, ":", ontogenies), collapse = ",")`), so nothing length-2 ever reaches the reader. And `.readOntogeniesFromList()` (`R/scenario-execution.R:686-704`) splits on the comma and loops over however many parts it gets, so it takes two fine — that was #1213 item 2, fixed in 01eacc0 (#1217).

Worth correcting rather than deleting: two ontogenies still earn their place in the fixture, just for a different reason — they are what exercises the multi-pair fold and the `length(proteins) != length(ontogenies)` guard, and `test-scenario-execution.R:1058` runs them end to end. As written, the next person to trim the fixture reads "a repaired importer would…" against a repaired importer, concludes the note is obsolete, and drops the second pair along with that coverage.

---

tests/testthat/test-project-excel.R
```diff
@@ -3163,3 +3163,481 @@ test_that("a sheet whose every row is skipped becomes no parameter set", {
...
+  # The ontogenies are the whole of the difference, once the id line is set aside.
+  dropId <- function(lines) {
+    grep("populationId", lines, fixed = TRUE, invert = TRUE, value = TRUE)
+  }
+  expect_false(identical(dropId(withOntogenies), dropId(withoutOntogenies)))
+  expect_match(
+    grep("proteinOntogenies", withOntogenies, value = TRUE),
+    "CYP3A4:CYP3A4,CYP2D6:CYP2C8",
+    fixed = TRUE
+  )
```
Felixmil, line 3248: This assertion passes vacuously when it should fail.

`grep(..., value = TRUE)` returns `character(0)` if no line matches, and `expect_match()` defaults to `all = TRUE`, which reduces to `all(logical(0))` — `TRUE`. So if `proteinOntogenies` disappeared from `adultpop.json` altogether, this line would still be green.

That is precisely the regression the test exists to catch. The two lines around it don't cover the gap either: 3243 only says the two files differ (the ontogeny-less twin still differs by nothing else), and 3249 asserts the *absence* of the key from the other file.

One extra line closes it:

```r
  onto <- grep("proteinOntogenies", withOntogenies, value = TRUE)
  expect_length(onto, 1L)
  expect_match(onto, "CYP3A4:CYP3A4,CYP2D6:CYP2C8", fixed = TRUE)
```

---

tests/testthat/test-definition-id.R
```diff
@@ -278,3 +278,71 @@ test_that(".suggestSuffix builds a 'did you mean' suffix for a near id", {
...
+# occur in practice, a per-analyte suffix and a `_mean` sibling, both diverge by
+# more than that, so the hint stays silent in exactly the cases it was added for
+# while firing on a one-character typo.
+test_that(".nearestMatch cannot reach a candidate diverging by a real suffix", {
+  # A `_mean` sibling: 5 characters, over the cap of 3.
+  expect_identical(
+    .nearestMatch("aciclovir_pvb_mean", "aciclovir_pvb"),
+    character(0)
+  )
+  # A per-analyte variant: 4 characters, also over the cap.
```
Felixmil, line 338: `_m1og` is 5 characters, not 4 — same as the `_mean` case two lines above that this one is contrasted with. (#1213 item 25 quotes "a 4-character suffix" for the per-analyte shape, but the id chosen here diverges by 5.) Either fix the count or pick a 4-character suffix so the test demonstrates both distances.

---

tests/testthat/helpers.R
```diff
@@ -93,6 +99,113 @@ testProjectExcelPath <- function() {
...
+  file.copy(
+    testthat::test_path(
+      "data",
+      "TestProjectExcel",
+      "Models",
+      "Simulations",
+      "Aciclovir.pkml"
+    ),
+    file.path(dest, "Models", "Simulations", "Aciclovir.pkml")
+  )
```
Felixmil, line 136: Neither `file.copy()` return value is checked, so a fixture that stops being complete fails a long way from the cause.

`file.copy()` reports failure through its return value, not a condition. If `TestProjectExcel/Models/Simulations/Aciclovir.pkml` is ever moved or renamed — which is a plausible future change, since this helper reaches across into the *sibling* fixture precisely to avoid a second copy of the binary — the copy returns `FALSE`, the helper hands back a project with no model, and every test in the block fails somewhere inside import or validation with a File-Not-Found. Since several of the tests here already assert things about validation criticals, that reads as a real defect rather than a broken helper.

```r
  stopifnot(all(file.copy(...)), file.copy(...))
```

or a `testthat::skip()` — either way the failure names the missing file.

---

tests/testthat/helpers.R
```diff
@@ -93,6 +99,113 @@ testProjectExcelPath <- function() {
...
+#'   - `outputDir`: where the JSON project was written (removed when the calling
+#'     test finishes), for assertions about which asset folders travelled.
+#'   - `warnings`: every warning message the import raised.
+#'
+#' The warnings are collected rather than suppressed because most of what these
+#' tests pin is *silence*: an import that drops something without a word. A test
+#' asserts that no collected warning mentions its own subject, which starts
+#' failing as soon as a fix adds one. Gating on the subject rather than using a
+#' bare `expect_no_warning()` keeps each test's assertion about its own defect,
+#' since a variant may legitimately raise an unrelated warning (a skipped sheet,
```
Felixmil, line 182: Two small drifts in this doc, both of which matter because callers rely on the behaviour it describes:

1. `warnings` is documented as "every warning message the import raised", but the helper deliberately collects `loadProject()`'s warnings too (the second `withCallingHandlers`, 205). That is not incidental — four tests match on load-side cross-reference warnings (`undefined population` at 3477, `undefined individual` at 3548), and those would be unreachable if the field were import-only as documented. Worth saying "the import *and the load*", since it's a load-side warning that carries the only trace of #1213 item 8.

2. "Gating on the subject rather than using a bare `expect_no_warning()`" describes the dominant pattern but not all of it: 3511 and 3643 assert `expect_length(imported$warnings, 1L / 0L)`, i.e. exactly the whole-set gate the paragraph argues against. Both look deliberate to me (each is pinning *total* silence as the finding), so it's the doc that needs the "except where a variant's subject is the absence of any warning at all" clause.

---

Felixmil (commented):
## Code review — `TestProjectExcelLegacy/` fixture and the #1213 characterization tests

Reviewed `ce1e55c..80e4c15`: 1247 added lines, all test-side — the `TestProjectExcelLegacy/` fixture plus its `data-raw/` generator, three new helpers, and ~19 tests across `test-project-excel.R`, `test-definition-id.R`, `test-scenario-execution.R`. No production `R/` changes.

I traced every new assertion against the code it exercises: the importer's section loop, `.parseExcelParameterSheets()`, `.excelProteinOntogenies()`, `.parseExcelPI5x*`, `.canonicalizeOneId()` / `.nearestMatch()`, `.populationCsvFileName()`, `.resolveExcelPopulationsFolder()`, `.warnOnCrossReferenceErrors()`, and the `DefinitionList` / typed-class field shapes and JSON key names the file comparisons rely on. **The fixture is a genuinely valuable addition and the assertions themselves hold up** — this is the first fixture in the suite that can fail in the legacy-workbook shapes, which is the whole reason #1213's item 1 went unnoticed. No test here is wrong about what the code does.

What does not hold up is a layer of *commentary* that has drifted out of sync with the fixes that landed underneath it (#1214, #1216, #1217, #1218), plus one assertion that pins nothing:

**Worth fixing before merge**
- `test-project-excel.R:3169` — the block header says every test below pins current buggy behaviour and "flips visibly when that finding is fixed"; three tests now assert the *repaired* behaviour instead, so a real regression in #1217/#1218 will be misread as the expected flip.
- `test-project-excel.R:3321` — the `OverwriteFormulasInSS` half of the comment describes a divergence that isn't there (the importer writes no key; the `FALSE` is the shared load-side default), and it's tagged with the wrong #1213 item.
- `test-project-excel.R:3244` — `expect_match()` over a possibly-empty `grep(value = TRUE)` passes vacuously, so losing `proteinOntogenies` from `adultpop.json` — the exact regression — would go unreported by this line.
- `test-project-excel.R:3256` — mistagged `item 8`; it covers item 4 (PI `Group` merging), whose test would otherwise be missing. Same at 3356.
- `data-raw/TestProjectExcelLegacy.R:122` — the reason given for the fixture carrying two ontogenies is stale, which risks the second pair being trimmed and the multi-ontogeny coverage going with it.

**Nits** — `helpers.R:126` (unchecked `file.copy()`), `helpers.R:175` (doc drift on `warnings`), `test-definition-id.R:338` (off-by-one in a comment).

All comments are inline. Nothing here is a correctness problem in the fixture or the code under test; it's all comment/traceability accuracy plus the one soft assertion.

---

PR #1216, opened 2026-08-03
Stop applying the bundled species-defaults sheet with a hard stop

Felixmil:
A scenario whose individual is a non-human species could not run under the default settings: the bundled `SpeciesParameters.xlsx` was folded into the user's own parameter layer and then applied strictly, so a single one of its paths that a given model happens not to contain aborted the whole scenario. Migration testing measured this costing 18 of 85 scenarios on one project and 10 of 39 on another, in both the imported and the hand-authored tree, with `runScenarios(..., stopIfParameterNotFound = FALSE)` as the only escape hatch. This restores the 5.x behaviour while keeping a user's own bad parameter path a hard error.

## The v5-vs-v6 call-site comparison

v5.7.0 read the bundled sheet in exactly one place, `R/utilities-simulation.R:59` inside `initializeSimulation()`, and applied it with `stopIfNotFound = FALSE`:

```r
if (any(excelSheets == individualCharacteristics$species)) {
  indivParams <- readParametersFromXLS(...)
  # NOT extending the `additionalParams` structure, otherwise it could
  # overwrite user defined parameters!
  ospsuite::setParameterValuesByPath(..., stopIfNotFound = FALSE)
}
```

v5's `Scenario` class merged three layers into `additionalParams` (the individual's own Excel sheet, the application protocol, `customParams`) and the species sheet was **not** among them.

v6 kept that tolerant call site verbatim (it is byte-identical in `R/simulation.R`) and **added a second one**: `.mergeScenarioParameters()` folded `.getSpeciesParameters()` in as a merge layer, and the merged structure is applied through `initializeSimulation()`'s `additionalParams` with `stopIfNotFound = stopIfParameterNotFound`, `TRUE` by default. So the sheet was applied twice per scenario, once tolerantly and once strictly, and the strict application was fatal.

The double application was directly observable: with `stopIfParameterNotFound = FALSE`, the same missing path produced two identical `Warning - Could not find quantity with path ...` lines per run. It now produces one.

## Which option, and why

Of the three options weighed, this takes "apply the bundled sheet separately and tolerantly before the user layer" -- and reaches it by **deleting the merge layer** rather than adding a second application, because that separate tolerant application already existed.

The two conditions coincide exactly, which is what makes the deletion safe rather than a behaviour change. `.mergeScenarioParameters()` folded the sheet in when the scenario had an `individualId`, the individual resolved, and a sheet matched its species. `initializeSimulation()` applies it when `individualCharacteristics` is non-NULL and a sheet matches its species -- and `.prepareScenario()` builds `individualCharacteristics` under precisely the first two of those conditions, from the same `indivData$species`. Removing the layer therefore removes the duplicate, not the behaviour.

It also restores v5's layer ordering. The species sheet had been layer 2 of five, *above* the scenario's own `modelParameterSets`, so the bundled defaults silently overrode a value the user set on the scenario. Applied before `additionalParams` instead, it is overridable by all four user layers, as in v5.

Rejected: per-row strictness tagging would need a new field threaded through `extendParameterStructure()` and both `setParameterValuesByPath()` calls, to re-derive a separation the code already had. Keeping the two structures separate throughout is what this now does, with no new structure.

The species-sheet read (`.getSpeciesParameters()`) moves to `R/simulation.R` next to its now-single caller, replacing that function's inlined copy of the same file-and-sheet lookup. Net effect in `initializeSimulation()`: `stopIfNotFound` is pinned `FALSE` for the bundled sheet, independent of what the caller asked for their own parameters.

## The two-sided test

Both halves are asserted on the same Rat scenario, since the point is that one run treats the two layers differently:

- **bundled path missing runs**: the Rat sheet carries `Organism|EndogenousIgG|...`, which the human Aciclovir fixture has no container for. `buildSimulations()` under the default `stopIfParameterNotFound = TRUE` must still return a `Simulation`. Needing `FALSE` here is the regression.
- **user path missing aborts**: a `customParams` path the model lacks still aborts, snapshotted. The snapshot scrubs only the call name in the `!` bullet, which ospsuite fills with the outermost caller (so it differs between `test_file()`, `devtools::test()` and `R CMD check`), keeping the `.validateEntitiesExist()` attribution and the offending path asserted.
- **applied once**: a counting stub on `.getSpeciesParameters()` pins the application count at 1 per build, so a re-introduced second layer fails rather than merely warning twice.

Verified against real non-human individual scenarios beyond the fixture: on the bundled example project, all seven species with a bundled sheet (Rat, Rabbit, Monkey, Beagle, Dog, Minipig, Mouse) aborted before this change and build after it, under the default strictness in both cases.

`NEWS.md` gets no entry: the defect existed only inside the unreleased 6.0.0 dev cycle and v5.7.0 was already tolerant, so the net delta for a user upgrading from the last release is zero.

Refs #1213


tests/testthat/test-scenario-execution.R
```diff
@@ -468,6 +488,70 @@ test_that(".mergeScenarioParameters silently skips an unknown application parame
...
+  withr::local_options(lifecycle_verbosity = "quiet")
+  project <- .useRatIndividual(.testProject())
+  built <- buildSimulations(project, scenarios = "testscenario")
+  expect_s3_class(built$testscenario$simulation, "Simulation")
+})
+
+test_that("the species sheet is applied once, not once per layer", {
+  # Each application of a sheet path the model lacks emits one native
+  # "Could not find quantity" warning, so a duplicated application shows up as
+  # a doubled count for the single missing Rat path.
```
Felixmil, line 513: This comment describes a mechanism the test doesn't use. It explains the assertion as counting native `Could not find quantity` warnings for the missing Rat path, but the test stubs `.getSpeciesParameters()` to return `NULL` and counts *invocations* of the stub. Nothing is ever applied, so no native warning is reachable — what's pinned is the number of sheet **reads** per build, not applications.

That's a fine thing to pin (a re-added merge layer that goes through `.getSpeciesParameters()` fails here), and it's more robust than warning-counting, so the assertion should stay as it is — only the comment needs to match it. Something like: "The sheet is read once per build. A stub standing in for the read counts calls; a re-introduced merge layer would read it twice and fail here rather than merely warning twice."

Worth noting the residual gap while the comment is being rewritten: because it counts calls to the helper rather than applications, a second layer reintroduced with an *inlined* file-and-sheet lookup (the shape `initializeSimulation()` carried before this PR) would slip past it.

Per the repo's own convention — a comment earns its place by explaining what the code cannot say itself, and describing how the code works currently.

---

Felixmil (commented):
Reviewed post-merge (this landed before the review ran). **The core change verifies correct** — nothing here argues for a revert; two small follow-ups on code now live on `json-based-project`.

## What I verified

The equivalence argument that makes the deletion safe rather than a behaviour change holds. `.mergeScenarioParameters()`'s old layer-2 guard (`scenario$individualId` set and non-`NA`, `indivData` found) is exactly the condition under which `.prepareScenario()` builds `individualCharacteristics` (`R/scenario-execution.R:376-408`), from the same `indivData$species`; `initializeSimulation()` applies the sheet under `!is.null(individualCharacteristics)` (`R/simulation.R:105-129`). So no scenario shape — including population scenarios, which still build individual characteristics — loses its species defaults. The cached-characteristics branch carries `species` too, so the cache doesn't open a hole.

`.getSpeciesParameters()` is now the single reference to the bundled sheet (grepped `R/`, `tests/`, `man/`, `vignettes/`, `inst/`), and its new `nzchar()`/`file.exists()` guards are strictly more defensive than the `readxl::excel_sheets()` call they replace, which was unguarded.

I also checked the test's fixture premise rather than taking it on trust, since a wrong premise would make the regression test vacuous. Parsing `Simulation/Model/Root` out of the Aciclovir pkml (4522 parameter paths, sanity-checked against `Organism|Muscle|Volume`) and diffing against the Rat sheet's 245 rows: exactly one Rat path — `Organism|EndogenousIgG|Kd (FcRn, endogenous IgG) in endosomal space` — is absent from the built model. The test does reproduce the regression. Worth recording the near-miss for anyone who re-checks this: that parameter *is* present in the pkml's individual building block under `ContainerPath` `Organism|EndogenousIgG`, so a shallower check makes the premise look false — it's the built model that omits the container. Confirmed too that there is no `Human` sheet (Rat, Rabbit, Monkey, Beagle, Dog, Minipig, Mouse), as the new roxygen note on `.getSpeciesParameters()` says.

Two things I'd flagged before reading the PR description, both of which it answers, noted here only so the next reader doesn't re-raise them: the **precedence flip** (defaults were layer 2, above the scenario's own `modelParameterSets`; they now lose to all four user layers) is deliberate and restores v5 ordering, and the **absent NEWS entry** is justified — the defect lived entirely inside the unreleased 6.0.0 cycle and v5.7.0 was already tolerant, so the delta for someone upgrading from the last release is zero.

## Follow-ups

1. **Stale test comment** — inline below. It describes counting native warnings; the test counts stub invocations. Comment only; the assertion is the right one.

2. **`stopIfParameterNotFound` is documented as governing only `customParams`** (`R/scenarios.R:620` on `runScenarios()`, `R/scenarios.R:703` on `buildSimulations()`: "a `customParams` path that matches no parameter in a scenario's simulation aborts the run"). It governs the whole merged structure — the scenario's `modelParameterSets`, the individual's and the application's `parameterSets`, and `customParams` — all four applied under `stopIfNotFound = stopIfParameterNotFound`. Pre-existing and outside this diff, but this PR makes it the *only* strict path, so the wording is now the only place a reader can learn which paths are strict, and it under-describes them. `R/parameter-identification.R:1052` already words it correctly ("a parameter listed in a scenario's parameter sets") and is worth copying.

One thing I could not do: R is not installed in this environment (no `Rscript`), so the suite was not run. Every claim above is static reading plus the pkml/sheet analysis, and none of it depends on test execution — but the new snapshot and the three new tests are unverified here. Codecov reports the patch at 86.67% with 2 uncovered lines in `R/simulation.R`, consistent with the `nzchar()`/`file.exists()` and no-matching-sheet early returns being unreachable from the fixtures.

---

PR #1217, opened 2026-08-03
Import the protein ontogenies of a legacy workbook, and run more than one

Felixmil:
Every protein ontogeny in every legacy 5.x workbook was discarded on import, silently, and the run-time crash that repairing the import exposes was one step behind it. This fixes both, and settles the authoring shape of `proteinOntogenies` so the field can be written by hand the way the importer writes it. It covers items 1 and 2 of #1213.

## The silent drop

`.parseExcelIndividuals()` and `.parseExcelPopulations()` read a single `Protein Ontogenies` cell. A real 5.x workbook declares its ontogenies as a `Protein` + `Ontogeny` column pair, each cell a comma-separated list paired up positionally, so the lookup found nothing and every ontogeny was dropped. Both spellings are now read, through one shared reader (`.excelProteinOntogenies()`), and the pair is folded into the single-cell spelling so both layouts import to the same definition and `import -> export -> import` is a fixed point for the field. A filled single cell is still stored verbatim, so a workbook in the current spelling imports byte-identically.

## The signal #1186 removed

#1186 asked for the raw `Unknown or uninitialised column: 'Protein Ontogenies'` tibble warning to stop reaching users, and #1212 delivered that by switching both parsers from `$` to `[[`. `[[` yields `NULL` for an absent column, so the warning went and the silent drop stayed; #1186 is closed as fixed on the premise that nothing was lost, and that premise was wrong. Two populations authored to differ *only* in their ontogenies imported byte-identically apart from the id, with no message at any point.

The raw tibble warning is not coming back: it was unacceptable, as #1186 said, because it named a column rather than a record and fired on a workbook that had nothing wrong with it. What replaces it is a catalog warning (`messages$excelOntogeniesNotReadable`) raised only when a sheet *declares* ontogenies that cannot be paired, that is only one of the two columns filled, or a differing number of proteins and ontogenies. It names the individual or population, the counts, and how to fix the workbook. A sheet that declares no ontogenies stays silent, in either spelling.

The remaining unparseable case needs no new warning: a single cell that is not a `Protein:Ontogeny` pair (`CYP3A4` on its own) is imported verbatim and reported by the existing `wrongOntogenyStructure` abort when the scenario runs, exactly as in 5.x, so it is visible rather than dropped.

## The crash the fix unmasks

`.readOntogeniesFromList()` tested the whole field for `NA` inside `||` before splitting it, so a two-entry value aborted with `'length = 2' in coercion to 'logical(1)'` before a single ontogeny was read. It was unreachable only because the importer wrote one comma-joined string and `addPopulation()` rejected a two-element vector up front. The reader now flattens the value first (`.splitProteinOntogenies()`), accepting every shape the field takes: one entry per ontogeny as a character vector, a single comma-joined cell, and the list a JSON array parses to.

## The authoring shape

`proteinOntogenies` is a character vector of `"Protein:Ontogeny"` entries, one per ontogeny, or the same pairs as one comma-joined string. It is now documented in the roxygen of `addIndividual()`, `setIndividual()`, `addPopulation()` and `setPopulation()`, where the accepted format was previously undocumented, and it is a vector-valued-per-definition field like `parameterSets`: the vector is applied whole to each definition, and a length-`id` list gives one vector per definition. Previously a two-ontogeny vector was rejected as a length mismatch against the number of ids.

A value that cannot be stored is refused at the call by a shared guard, with a message that describes ontogenies rather than ids. That covers the `ospsuite::MoleculeOntogeny` object, which was accepted (as `list(onto)`, whose length of 1 passed the length check) and then broke `saveProject()` with `No method asJSON S3 class: R6`, leaving the whole project unsaved. The pairs themselves are deliberately not validated at authoring time, because a workbook cell is imported verbatim and the ontogeny name is checked against `ospsuite::StandardOntogeny` when the scenario runs.

## Tests

The bundled `TestProjectExcel` fixture uses the current single-column spelling, which is why no existing test could fail this way. The legacy layout is produced from that fixture at test time (`.excelFixtureWithLegacyOntogenies()` rewrites the two workbooks into the `Protein` + `Ontogeny` pair) rather than checked in as a second binary workbook, and the new tests cover: a legacy two-column sheet importing its ontogenies; two populations differing only in their ontogenies *not* importing identically, which is the assertion the old fixture could never make; the single cell winning when both spellings are filled; the unpairable declaration warning through the catalog; a legacy-spelling workbook importing end to end and surviving an export/re-import round trip; two ontogenies running through `runScenarios()`; and the authoring shape, including the refusal of a `MoleculeOntogeny` object and that the refused call leaves the project saveable.

## Example

``` r

# A real 5.x populations sheet declares its ontogenies as a `Protein` +
# `Ontogeny` column pair. These two populations differ ONLY in them.
sheet <- dplyr::tibble(
  PopulationName = c("PopA", "PopB"),
  species = "Human",
  population = "European_ICRP_2002",
  numberOfIndividuals = 10,
  Protein = c("CYP3A4", "CYP2D6"),
  Ontogeny = c("CYP3A4", "CYP2C8")
)
populations <- .parseExcelPopulations(sheet)
populations[[1]]$proteinOntogenies
#> NULL
populations[[2]]$proteinOntogenies
#> NULL

# Do the two imported definitions differ at all, apart from the id?
identical(populations[[1]][-1], populations[[2]][-1])
#> [1] TRUE

# What the scenario runner makes of two ontogenies on one definition.
try(.readOntogeniesFromList(c("CYP3A4:CYP3A4", "CYP2D6:CYP2C8")))
#> Error in is.null(ontogenyString) || is.na(ontogenyString) : 
#>   'length = 2' in coercion to 'logical(1)'

# Authoring the same two ontogenies.
try(addPopulation(
  project,
  "two_ontogenies",
  species = "Human",
  numberOfIndividuals = 10,
  proteinOntogenies = c("CYP3A4:CYP3A4", "CYP2D6:CYP2C8")
))
#> Error in addPopulation(project, "two_ontogenies", species = "Human", numberOfIndividuals = 10,  : 
#>   `proteinOntogenies` must be length 1 (one id was given).
#> ✖ It is length 2.
project$definitions$populations[["two_ontogenies"]]$proteinOntogenies
#> NULL

# Authoring an `ospsuite::MoleculeOntogeny` object.
onto <- ospsuite::MoleculeOntogeny$new(
  molecule = "CYP3A4",
  ontogeny = ospsuite::StandardOntogeny$CYP3A4
)
try(addPopulation(
  project,
  "r6_ontogeny",
  species = "Human",
  numberOfIndividuals = 10,
  proteinOntogenies = list(onto)
))
# What `saveProject()` would then have to serialize.
try(jsonlite::toJSON(
  project$definitions$populations[["r6_ontogeny"]],
  auto_unbox = TRUE
))
#> Error : No method asJSON S3 class: R6
```

</details>

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/protein-ontogenies-legacy-drop-7c9927", quiet = TRUE)

# A real 5.x populations sheet declares its ontogenies as a `Protein` +
# `Ontogeny` column pair. These two populations differ ONLY in them.
sheet <- dplyr::tibble(
  PopulationName = c("PopA", "PopB"),
  species = "Human",
  population = "European_ICRP_2002",
  numberOfIndividuals = 10,
  Protein = c("CYP3A4", "CYP2D6"),
  Ontogeny = c("CYP3A4", "CYP2C8")
)
populations <- .parseExcelPopulations(sheet)
populations[[1]]$proteinOntogenies
#> [1] "CYP3A4:CYP3A4"
populations[[2]]$proteinOntogenies
#> [1] "CYP2D6:CYP2C8"

# Do the two imported definitions differ at all, apart from the id?
identical(populations[[1]][-1], populations[[2]][-1])
#> [1] FALSE

# What the scenario runner makes of two ontogenies on one definition.
try(.readOntogeniesFromList(c("CYP3A4:CYP3A4", "CYP2D6:CYP2C8")))
#> [[1]]
#> <MoleculeOntogeny>
#>   • Molecule: CYP3A4
#>   • Ontogeny: CYP3A4
#> 
#> [[2]]
#> <MoleculeOntogeny>
#>   • Molecule: CYP2D6
#>   • Ontogeny: CYP2C8

# Authoring the same two ontogenies.
try(addPopulation(
  project,
  "two_ontogenies",
  species = "Human",
  numberOfIndividuals = 10,
  proteinOntogenies = c("CYP3A4:CYP3A4", "CYP2D6:CYP2C8")
))
project$definitions$populations[["two_ontogenies"]]$proteinOntogenies
#> [1] "CYP3A4:CYP3A4" "CYP2D6:CYP2C8"

# Authoring an `ospsuite::MoleculeOntogeny` object.
onto <- ospsuite::MoleculeOntogeny$new(
  molecule = "CYP3A4",
  ontogeny = ospsuite::StandardOntogeny$CYP3A4
)
try(addPopulation(
  project,
  "r6_ontogeny",
  species = "Human",
  numberOfIndividuals = 10,
  proteinOntogenies = list(onto)
))
#> Error in addPopulation(project, "r6_ontogeny", species = "Human", numberOfIndividuals = 10,  : 
#>   `proteinOntogenies` must be a character vector of "Protein:Ontogeny"
#> entries. ✖ It is a <MoleculeOntogeny> object. ℹ One entry per ontogeny, e.g.
#> `c("CYP3A4:CYP3A4", "CYP2D6:CYP2C8")`, or the same pairs as one comma-joined
#> string.
# What `saveProject()` would then have to serialize.
try(jsonlite::toJSON(
  project$definitions$populations[["r6_ontogeny"]],
  auto_unbox = TRUE
))
#> {}
```

</details>


R/individuals.R
```diff
@@ -203,9 +209,14 @@ addIndividual <- function(project, id, species, ...) {
...
   dots[["overwrite"]] <- NULL
-  # `parameterSets` is the one vector-valued-per-definition field; everything else
-  # is scalar-per-definition. `species` is a positional formal, not a `...` field.
-  wholeNames <- intersect("parameterSets", names(dots))
+  # `parameterSets` and `proteinOntogenies` are the vector-valued-per-definition
+  # fields (a whole vector belongs to one individual, never split across ids);
+  # everything else is scalar-per-definition. `species` is a positional formal,
+  # not a `...` field.
+  wholeNames <- intersect(
+    c("parameterSets", "proteinOntogenies"),
```
Felixmil, line 217: Moving `proteinOntogenies` from `scalarDots` into `wholeNames` silently changes what an existing batch call means, in the one case where the old call was valid.

Before this PR the field landed in `scalarFields`, so it went through the recycle/align rule: length 1 recycled, length `N` aligned to the ids **by position**. Now it is applied whole to every definition. So a call whose ontogeny count happens to equal its id count flips meaning with no error and no warning:

```r
addIndividual(project, id = c("a", "b"),
              proteinOntogenies = c("CYP3A4:CYP3A4", "CYP2D6:CYP2C8"))
# before: a -> CYP3A4:CYP3A4,  b -> CYP2D6:CYP2C8   (split positionally)
# after:  a -> both,           b -> both            (applied whole)
```

The new behaviour is the right one — a vector of ontogenies belongs to one definition, same as `parameterSets` — so this is about the migration note, not the design. NEWS.md:75 describes only the case that *used to error* ("a two-ontogeny vector was previously rejected as a length mismatch against the number of ids"). A user who wrote the 2-ids-2-ontogenies form reads that, correctly concludes it never errored for them, and skips the note — while their call now writes different definitions. Worth one clause in that NEWS entry saying a length-`N` vector was split positionally and is now applied whole, with the length-`N` list as the way to keep the old meaning.

Same flip at R/individuals.R:479 (`setIndividual`), R/populations.R:530 and R/populations.R:840.

Minor, same area: the new comment at R/messages.R:143-145 justifies `invalidProteinOntogenies()`'s wording with "the field is aligned across ids like every other one" — this PR is what made that untrue, and per the house rule that comments describe current behaviour it now reads backwards.

---

R/messages.R
```diff
@@ -141,6 +141,36 @@ messages$wrongOntogenyStructure <- function(entry) {
...
+# would be dropped; saying so is the difference between a reported gap and
+# ontogenies vanishing from the imported project.
+messages$excelOntogeniesNotReadable <- function(
+  recordType,
+  recordId,
+  proteins,
+  ontogenies
+) {
+  cli::format_message(c(
+    "!" = "The protein ontogenies of {recordType} {.val {recordId}} are not imported: the {.field Protein} and {.field Ontogeny} columns cannot be paired.",
```
Felixmil, line 168: `recordId` comes from the workbook, and this entry glues it eagerly — so an id containing `{` turns an import warning into an aborted import.

`cli::format_message()` interpolates `recordId` into the returned string, and then the raising `cli::cli_warn()` at R/project-excel.R:3509 glue-parses that finished string a second time. A population named `Pop {A}` reaches the second pass as literal `{A}`, which glue evaluates as an R expression — `object 'A' not found`, thrown from inside `importProjectFromExcel()`. The failure mode is the worst-shaped one available here: the warning exists precisely to stop a malformed sheet from being silently dropped, so a malformed *id* takes down the whole import instead of reporting the gap.

This is exactly the hazard the file header calls out at R/messages.R:22-30, and `unsupportedSchemaVersion()` (R/messages.R:184) is the in-repo example of the fix — return the template unglued and let the raising call interpolate once from its own frame. The wrinkle is that this entry also needs `format_message()` for its `!`/`i` bullets, which is the one thing the unglued form can't carry. Cheapest way out is to pre-escape at the boundary, leaving the bullets intact:

```r
"!" = "The protein ontogenies of {recordType} {.val {.recordId}} are not imported: ..."
```
with `.recordId <- gsub("}", "}}", gsub("{", "{{", recordId, fixed = TRUE), fixed = TRUE)`.

Worth checking whether the same double-parse reaches other catalog entries that interpolate a workbook-supplied id through `format_message()` — the ids in this migration path are user data, and `.canonicalizeId()` maps commas and interior spaces but not braces.

---

R/authoring-vectorize.R
Felixmil, line 1: The shared recycling-rule doc wasn't updated for the new whole-field, so two statements on the same man page now contradict each other.

`proteinOntogenies` became a vector-valued-per-definition field in this PR, but both places that enumerate those fields still list only `parameterSets` / `outputPaths`:

- rule 3 in the file-header comment, R/authoring-vectorize.R:13-14
- the `@details` of `vectorizedAuthoring`, R/authoring-vectorize.R:52-56

That `@details` block is inherited onto every authoring man page, so the rendered result disagrees with itself:

- `man/addIndividual.Rd:41-42` (from the `@param`) — "`parameterSets` and `proteinOntogenies` are vector-valued-per-definition (applied whole …)"
- `man/addIndividual.Rd:56-57` (from these shared details) — "an individual's or application's `parameterSets`, a scenario's `outputPaths` and `parameterSets`", omitting `proteinOntogenies`

Same pair at `man/addPopulation.Rd:47` vs `:69`. A reader who reaches the Details section — which is where the `@param` text points them for the rule — comes away believing `proteinOntogenies` recycles/aligns positionally, which is the wrong assumption in precisely the batch case flagged on `R/individuals.R:217`. Adding the field to both enumerations and re-running `document()` covers it.

---

Felixmil (commented):
Post-merge review (scheduled run — the `ready_for_review` webhook fired, but this landed on `json-based-project` before the review ran, so these are notes on merged code rather than a merge gate). CodeRabbit skipped this PR because the base isn't the default branch, so nothing else reviewed it.

The three core changes hold up. I traced each:

- `.splitProteinOntogenies()` (R/utils.R:151) is correct for every shape it claims: `NULL`, `NA`, `""`, `character(0)`, `list()`, comma-joined cells, and nested lists all collapse safely, and `.readOntogeniesFromList()` no longer runs a scalar `is.na()` over a length-2 field.
- Whole-field routing, `.assertProteinOntogenies()` placement (after each builder's aggregated-error block, before any state write), and the `NULL`-clears contract on the set paths are consistent with the existing `parameterSets` handling. `character(0)` is accepted and stored as a JSON `[]`, which survives `.assertNoEmptyObjectFields()` — that guard keys on `names(value)` being `character(0)`, which only an empty *object* produces. Harmless.
- `.excelProteinOntogenies()` reads absent columns safely: `[[` with an unmatched character subscript returns `NULL`, and `.naToNull()` plus the `length(.splitProteinOntogenies(single)) > 0L` guard correctly make a blank-but-present `Protein Ontogenies` cell fall through to the pair columns rather than shadow them. The `molecule:ontogeny` join order matches what `.readOntogeniesFromList()` consumes.
- Export via `.formatOntogeniesToCell()` also fixes a latent crash: the old `ic$proteinOntogenies %||% NA` would have made `data.frame()` recycle or error on a length-2 field.

Three findings, none of which undermine the fix — one silent behaviour change that the NEWS note doesn't cover, one glue double-parse that can abort an import, and a stale shared doc. Details inline.

One caveat on confidence: there is no R/Rscript in this environment, so I could not run `testthat`. Every conclusion above is from reading the code, not from a green suite.

---

PR #1218, opened 2026-08-03
Make model, data and population folders travel on both entrypoints, and fail validation on a missing population csv

Felixmil:
A project's model, data and population folders are what make an imported or upgraded project runnable, and three separate defects kept them from arriving. #1164 fixed this for the Excel entrypoint only: on the snapshot-upgrade path nothing travelled at all and the report that would have said so was suppressed, and the population csv folder was left behind on both paths while `validateProject()` reported the result as clean. This addresses items 5 and 6 of #1213, each defect separately, and adds the validator that was missing so a project whose population csv did not travel fails validation instead of dying at run time in the simulation backend.

## Defect A: the snapshot-upgrade path copied nothing, silently

`.upgradeLegacySnapshot()` materializes the mirrored workbooks into a scratch directory and hands *that* to the importer, so `.copyExcelProjectAssets()` resolved every asset folder against a directory holding only the rebuilt workbooks. Nothing was ever found, and the "N folders were not copied" report is gated on `if (!silent)` while the upgrade passes `silent = TRUE`, so it was suppressed in precisely the case where it always applied.

- The upgrade now copies the asset folders sitting beside the snapshot file into that scratch project first, reconstructing the Excel project the snapshot describes. The bridge's own asset copy then carries them into the new project, so there is one asset copy rather than a second source directory threaded through it. A folder the snapshot carries itself (the population csv files) is already in the scratch project and wins.
- What did not travel is reported as extra bullets on the upgrade notice the user already reads, computed from the loaded project's own folders, so it reflects the end state rather than what the copy attempted. Keeping it on that one notice is deliberate: the whole upgrade summary stays on one stream, as the surrounding code intends.
- The materializer writes the snapshot's embedded population csv files under the folder name the property sheet records, for the same reason it already writes each workbook under its recorded filename: a hardcoded name leaves them where the bridge does not look.
- `.legacyConfigProperties()` returns a list, so an absent property reads as `NULL` for the `%||%` defaults its callers already applied to it (on a named character vector, `[[` errors instead).

## Defect B: `PopulationsCSV/` was left behind on the Excel path too

An Excel project spells `populationsFolder` as a folder *name* under the configurations folder (v5 resolved it against `configurationsFolder`), while a project resolves every working folder against its own root. Read literally the value named a root-level folder that does not exist, so the folder was neither copied nor resolvable, and an in-place import was just as broken as a copied one.

The import now rewrites the value to the project-root-relative path of the folder that actually holds the csv files. Both layouts are accepted, in the order the value itself suggests: a folder present at the project root is taken as authored, otherwise the one under the configurations folder is adopted. One corrected value then means the same thing in the source project and in the imported one, which is the invariant `.copyExcelProjectAssets()` is built on, so the copy itself needed no change and the working `Models/` and `Data/` behaviour is untouched.

This does mean an imported project can carry a `Configurations/PopulationsCSV/`, so the existing "the Excel side is the source, not an asset" test now asserts what it means: no workbook travelled. The alternative (promoting the folder to the project root) would have made the copy special-case one field and left in-place imports still broken.

## Defect C: the derived csv filename was lowercased

`<canonical-id>.csv` is lowercase while the file keeps the case its author gave it, so the derivation resolved only on a case-insensitive filesystem. One shared helper, `.populationCsvFileName()`, now matches the derived name against the folder listing case-insensitively (an exact match wins) and is used by both the runtime and the new validator, so the file the validator checks is the file the run opens. A `.populationCsvFileName()` unit test asserts the returned name, which fails on any filesystem without the fix.

## Validation and the run-time abort

`.validateScenarios()` takes the populations folder and the populations section, and warns `File Not Found` when the csv file a scenario would read is absent. Which file that is follows the runtime's own rule (`.resolvePopulation()`): a `csv` entry names its own `file`, otherwise the scenario's `readPopulationFromCSV` flag derives one. The populations section is read only for that, not to check that the reference resolves, which stays `crossReferences`' job.

Grading: a warning, matching a missing `modelFile` and a missing observed-data file. A critical error would block every scenario in the project over one absent csv file, which is the over-blocking item 12 of #1213 reports. Separately, `.resolveCsvPopulation()` now aborts naming the file and the folder it was expected in, so `validate = FALSE` no longer ends in a raw `OSPSuiteException` naming neither.

## An absolute `modelFolder` no longer writes an unopenable project

The upgrade wrote the whole definitions tree and then aborted on load with `simulationsFolder "<abs>" resolves outside the project folder`. The containment check now runs before anything is written, on the folder set the loaded project actually resolves.

Scoped to the upgrade path deliberately, not to `importProjectFromExcel()`: #1182 made an absolute `dataFolder` survive an import on purpose (a project whose data sits on a synced drive migrates unattended, and the project still loads because nothing reads that folder), and an existing test expects a `../`-climbing folder to be reported rather than to abort. An upgrade is the case where the value is a guaranteed failure, because it has to hand back a *loaded* project.

## Tests

On the bundled fixtures, since both halves reproduce there: `tests/testthat/data/TestProjectExcel/` holds the legacy snapshot *and* sits beside the `Models/`, `Data/` and `Configurations/PopulationsCSV/` folders its property sheet names, so the upgrade path is covered without new fixture data. Copied and in-place imports, a root-level populations layout, a lone snapshot whose folders are reported missing, the refused out-of-project folder (asserting nothing was written), the recorded csv folder name, the filename-case helper, the new run-time abort, and the three validator cases.

## Note on #1164

#1164 is currently closed as fixed and should be reopened or marked superseded by this PR: its Excel half worked, but the snapshot-upgrade path copied nothing and the population folder failed on both paths. The NEWS bullets for #1164 and for the snapshot upgrade are edited in place rather than added beside, so NEWS describes the behaviour that ships rather than narrating this cycle's churn.

Items 5 and 6 of #1213. Not closing it: it is an umbrella of 26 findings.


R/import-legacy-snapshot.R
```diff
@@ -52,21 +52,49 @@
...
 ) {
+  props <- .legacyConfigProperties(jsonData$projectConfiguration)
+
+  # Refuse a working folder the finished project could not resolve, before
+  # anything is written. An upgrade has to hand back a *loaded* project, so such a
+  # value is a guaranteed failure a few lines below, and letting it get that far
+  # would leave a complete but unopenable project behind. (A plain Excel import is
+  # deliberately more permissive: it hands back a tree the user can still fix, and
+  # a folder placed outside the project on purpose is a supported migration.)
+  .assertLegacySnapshotFoldersContained(props, dir)
```
Felixmil, line 75: **The pre-write containment check is stricter than the rule it says it mirrors, and `outputFolder` is a regression.**

The comment justifies this as "the same rule the loaded project applies in `.resolveWorkingFolder()`, checked up front". But `.resolveWorkingFolder()` is only reached through `project$paths$<name>` (`R/project.R:927`), i.e. lazily, and `.printPathsBlock()` deliberately swallows its abort so an out-of-project folder stays inspectable. So a load does *not* enforce containment on a folder nothing reads.

`.assertLegacySnapshotFoldersContained()` iterates all of `.containedWorkingFolders`, which includes `outputFolder` (`R/project.R:33`). `outputFolder` is a real legacy property — the bundled fixture spells it `Results/` — and nothing on the upgrade path ever resolves it: `.missingUpgradedAssetFolders()` reads only `simulationsFolder` / `dataFolder` / `populationsFolder`, which is the same set `.excelProjectAssetFolders` documents as "`outputFolder` is deliberately absent".

Net effect: a 5.x snapshot whose `outputFolder` is absolute or `../`-climbing (results on a shared drive — a very ordinary 5.x setup) upgraded fine before this PR and now aborts `restoreProject()`. And because the check runs before anything is written, the user gets no tree at all — so there is nowhere to correct the value, which is the opposite of the "hands back a tree the user can still fix" rationale given for the Excel path.

`dataFolder` is worth a second look for the same reason: #1182 made a bare-absolute `dataFolder` survive an import on purpose, and this makes the upgrade path refuse it. That one is at least *consistent* now, since `.missingUpgradedAssetFolders()` newly reads `project$paths$dataFolder` and would abort post-write anyway — but that knock-on is a behaviour change this PR doesn't call out, and the pre-write abort is what makes it total.

Suggestion: restrict the assertion to the folders the upgrade actually resolves (drop `outputFolder`), and decide `dataFolder` deliberately against #1182 rather than by inheritance from `.containedWorkingFolders`.

---

R/import-legacy-snapshot.R
```diff
@@ -97,7 +134,27 @@
...
+
+# The input folders an upgraded project declares but that are not on disk under
+# `dir`, as paths relative to the project folder (what the user has to place
+# there). `outputFolder` is deliberately absent: the project writes results
+# there, it does not read them, and the runtime creates it on demand.
+#
+# @keywords internal
+# @noRd
+.missingUpgradedAssetFolders <- function(project, dir) {
+  folders <- c(
```
Felixmil, line 148: **False "referenced folder is missing" notice on the common case: a declared folder that nothing references.**

This keys on the folders the project *declares*, not the ones its definitions *reference*. A legacy `ProjectConfiguration.xlsx` declares `populationsFolder` as a matter of template boilerplate — the bundled fixture does (`PopulationsCSV`) — whether or not any population is read from csv.

For a project that declares it but uses only demographics specs:

- `.materializeLegacySnapshot()` creates the csv folder only under `if (!is.null(jsonData$populationsCSV) && length(...) > 0L)`, so nothing is written;
- nothing sits beside the snapshot to copy;
- `.resolveExcelPopulationsFolder()` finds the folder at neither location and leaves the value as authored;
- `dir/PopulationsCSV` therefore does not exist, and this reports it missing.

The user then reads "1 referenced folder is missing from the upgraded project" and "Nothing that reads from one of them will resolve until you place the folder in the project" — for a folder nothing reads. `dataFolder` has the same shape for a project with no observed data.

That is a false alarm on the majority of upgrades, and it lands on the one notice the user is guaranteed to read, which will teach them to ignore it in the cases where it is true.

Suggestion: report a folder only when some definition actually needs it — `populationsFolder` when a scenario resolves a csv population, `dataFolder` when an `observedData` entry exists, `simulationsFolder` unconditionally (every scenario has a `modelFile`).

---

R/import-legacy-snapshot.R
```diff
@@ -52,21 +52,49 @@
...
+  # workbooks: reconstructing the Excel project the snapshot describes is what
+  # lets the bridge's own asset copy carry them into `dir`, exactly as it does
+  # for an Excel project imported into another folder. Folders the snapshot
+  # carries itself (the population csv files) are already in `scratch` and win,
+  # since `overwrite = FALSE` leaves a target that already holds files alone.
+  # An asset folder is therefore copied twice on its way to `dir`, which is the
+  # price of reusing the one asset copy rather than teaching it a second source
+  # directory to resolve every folder against.
+  if (!is.null(snapshotDir)) {
+    .copyExcelProjectAssets(props, snapshotDir, scratch, overwrite = FALSE)
```
Felixmil, line 95: **The `overwrite = FALSE` protection the comment relies on cannot fire, and a stray folder beside the snapshot silently shadows the csvs the snapshot carries.**

The comment above says: "Folders the snapshot carries itself (the population csv files) are already in `scratch` and win, since `overwrite = FALSE` leaves a target that already holds files alone."

But the two writers use different paths. `.materializeLegacySnapshot()` writes the embedded csvs to `file.path(configDir, basename(populationsFolder))` — i.e. `scratch/Configurations/PopulationsCSV`. This call passes the *raw* legacy `props`, so `.copyExcelProjectAssets()` resolves `populationsFolder` against `sourceDir` and copies `snapshotDir/PopulationsCSV` → `scratch/PopulationsCSV`, at the root. The paths never collide, so the "target already holds files" branch is never reached and the protection is illusory.

The consequence is worse than a stale comment. If a `PopulationsCSV/` folder does sit beside the snapshot file, the copy lands it at `scratch/PopulationsCSV`, and `.resolveExcelPopulationsFolder()` then takes the root-level layout as authored (`fs::dir_exists(fs::path(root, value))` wins first) and returns the value unrewritten. The upgraded project points at that folder, and the population tables the snapshot itself embedded — sitting untouched at `scratch/Configurations/PopulationsCSV` — are dropped without a word. Populations silently coming from the wrong source is the shape of item 3 in #1213.

This is not covered by the new tests: the fixture keeps its `PopulationsCSV/` under `Configurations/`, so `snapshotDir/PopulationsCSV` does not exist and the copy reports it as not-copied (a result this call discards). A test with a root-level `PopulationsCSV/` beside a snapshot that also carries `populationsCSV` would fail.

Suggestion: skip `populationsFolder` in this scratch copy entirely — the snapshot is authoritative for it — or resolve both writers to the same target so `overwrite = FALSE` actually decides.

---

R/messages.R
```diff
@@ -511,13 +511,27 @@ messages$legacySnapshotMalformedSheet <- function() {
   )
 }
 
-messages$upgradedLegacySnapshot <- function() {
-  c(
+messages$upgradedLegacySnapshot <- function(missingFolders = character()) {
+  notice <- c(
     "i" = "Detected a previous-version project snapshot and upgraded it to the \\
     current project format.",
     "!" = "Observed data does not travel in a snapshot; add it with \\
```
Felixmil, line 518: **This caveat is now wrong on the happy path this PR creates — and the handler that leans on it over-muffles.**

The premise ("Observed data does not travel in a snapshot") was true when the scratch project held only the rebuilt workbooks. Now that `.copyExcelProjectAssets()` brings the snapshot's neighbouring folders into `scratch`, `dataFolder` arrives too, so `.importObservedData()` resolves `dataFolder/dataFile` and *does* import the observed data. The bundled fixture is exactly this case: `Data/` sits beside the snapshot and holds both `TestProject_TimeValuesData.xlsx` (the recorded `dataFile`) and the importer configuration.

So on a normal upgrade the user is now told to `addObservedData()` data they already have. Three places carry the stale premise:

- this bullet, unconditionally;
- the muffle-handler comment at `R/import-legacy-snapshot.R:101-106` ("A snapshot never carries the observed-data workbook itself ... so the bridge's warning fires on every such upgrade") — it now fires only when the data folder genuinely did not travel, which is precisely when the user needs to hear it;
- `.skipOutOfProjectObservedData()`'s roxygen at `R/project-excel.R:2057` ("the legacy-snapshot upgrade, which never ships the data workbook").

The muffle is the actionable part: since `esqlabsR_importSkippedObservedData` is shared by the genuine "configured data file was not found" warning *and* by `.skipOutOfProjectObservedData("dataFile", ...)`, an upgrade whose data really is absent, or whose `dataFile` escapes its `dataFolder`, is now silenced. It was a safe blanket muffle when the condition was unconditional; it is a real signal now.

Suggestion: drop the blanket muffle and let the warning through (it is accurate in the case it now fires), and make this bullet conditional on observed data actually not having been imported. If the muffle stays for the third-edition-testthat reason, it wants to be conditional on the same fact.

---

R/scenarios.R
```diff
@@ -427,6 +443,52 @@ print.Scenario <- function(x, ...) {
...
+          result$addCriticalError(
+            "Path Containment",
+            paste0(
+              "Scenario '",
+              name,
+              "' reads a population file outside the project folder: ",
+              fileName
+            )
+          )
+        } else if (!file.exists(file.path(populationsFolder, fileName))) {
```
Felixmil, line 477: **Minor / edge case: `file.path()` here is not the join the runtime uses, so an absolute `file` inside the populations folder gets a false `File Not Found`.**

The stated invariant of the new validator is "the file the validator checks is the file the run opens". The runtime joins with `.resolveProjectPath()` → `.absoluteAgainstRoot()`, which uses `path` verbatim when `fs::is_absolute_path(path)`. This line joins with `file.path()`, which concatenates unconditionally.

The containment branch just above already handles the interesting half: an absolute `file` *outside* `populationsFolder` is a critical error here and an abort at run time, so the two agree. The divergence is the remaining case — a `csv` entry whose `file` is absolute but points *inside* the populations folder. `.pathEscapesRoot()` returns `FALSE`, so control reaches here and tests `<folder>//abs/path/to/pop.csv`, which never exists, while `.resolveCsvPopulation()` resolves and loads the file fine. A clean project reports a spurious warning.

Low severity — an absolute `file` on a csv population is unusual — but it is a one-token fix that also makes the join match the documented invariant by construction:

```r
} else if (!file.exists(.absoluteAgainstRoot(fileName, fs::path_abs(populationsFolder)))) {
```

(For contrast, the `modelFile` branch at `:376` avoids the analogous problem by short-circuiting on `fs::is_absolute_path()` before the containment test and skipping the existence check for absolute values altogether.)

---

Felixmil (commented):
Post-merge review (the PR merged into `json-based-project` before this review landed, so these are findings against the base branch now, not merge blockers).

Reviewed `5aa9d0b..fb9c72d` — `R/import-legacy-snapshot.R`, `R/populations.R`, `R/project-excel.R`, `R/project.R`, `R/scenario-execution.R`, `R/scenarios.R`, `R/messages.R` and tests. R is not installed in this environment, so nothing was executed; every finding below is read off the code and its callers, and each names the path that reaches it.

The three defects the PR sets out to fix look correctly fixed, and the reasoning in the PR body holds up against the code. What the review turned up is fallout at the edges of the new asset-travel path — four issues plus one minor one, none of them duplicating an item already in #1213:

1. **`R/import-legacy-snapshot.R:75`** — the pre-write containment check covers `outputFolder`, which nothing on the upgrade path resolves. A 5.x snapshot writing results to an absolute path upgraded fine before and now aborts with nothing written. Highest-impact of the five.
2. **`R/import-legacy-snapshot.R:148`** — "referenced folder is missing" keys on *declared* folders, so a project that declares the boilerplate `populationsFolder` without using csv populations gets a false alarm on the one notice every upgrade prints.
3. **`R/import-legacy-snapshot.R:95`** — the scratch copy and the materializer write `populationsFolder` to different paths, so the `overwrite = FALSE` precedence the comment relies on cannot fire; a stray `PopulationsCSV/` beside the snapshot silently shadows the csvs the snapshot itself carries. Not covered by the new tests.
4. **`R/messages.R:518`** — with `Data/` now travelling, observed data *does* import, so the "add it with `addObservedData()`" caveat is wrong on the fixture's own path; more importantly the shared-class muffle at `:101-106` now silences a warning that has become accurate.
5. **`R/scenarios.R:477`** — minor: `file.path()` rather than the runtime's `.absoluteAgainstRoot()`, so an absolute `file` pointing inside the populations folder gets a spurious `File Not Found`.

Items 1 and 3 are the ones worth acting on soonest: 1 turns a previously-working upgrade into a hard failure with no artifact to repair, and 3 can substitute the wrong population data without a message.

Two notes on the PR body rather than the diff. The observation that **#1164 needs reopening or marking superseded** is right and still outstanding. And the `dataFolder` half of the containment discussion has a knock-on the body doesn't mention: `.missingUpgradedAssetFolders()` newly reads `project$paths$dataFolder`, so an absolute `dataFolder` would abort on the upgrade path even without the new assertion — worth deciding deliberately against #1182 rather than inheriting from `.containedWorkingFolders`.

---

PR #1223, opened 2026-08-04
Reject a definitionsFolder that escapes the project directory

Felixmil:
A review comment on #1017 noted that `definitionsFolder` is the only project folder field read out of `Project.json` without a containment check, while its four siblings (`simulationsFolder`, `dataFolder`, `outputFolder`, `populationsFolder`) all resolve through `.resolveWorkingFolder()` and are required to stay under the project directory. That gap matters because `definitionsFolder` is the field that feeds the two places in the package that delete files: `.writeDefinitionTree()` owns its `<kind>/` subfolders and removes every `.json` it did not just write, and `.clearProjectArtifacts()` unlinks the folder recursively. A project file is untrusted input, since projects normally travel as shared or downloaded folders, so a value such as `"../.."` would have made a plain `saveProject()` delete unrelated `.json` files outside the project, and `initProject(destination, overwrite = TRUE)` recursively delete the grandparent tree.

## Guard

The field names a single folder directly under the project directory, so it is rejected outright rather than resolved and contained; there is also no `${VAR}` opt-out to honour, unlike the working folders. The check reuses `.validateFilenameSegment()` (`R/file-utils.R`), which already exists for exactly this and rejects path separators plus `.` / `..`, and is already used for `projectFileName` and snapshot stems. The new `.validateDefinitionsFolder()` in `R/definition-files.R` wraps it and lets `NULL` through, since an absent value resolves to the default `"definitions"`.

## Call sites

- `.readJson()` in `R/project.R`, before the value is stored, so an escaping value makes the project unloadable rather than dangerous on save.
- The `paths$definitionsFolder` setter in `R/project.R`, which also gives the field the type check it previously lacked.
- `.clearProjectArtifacts()` in `R/project-lifecycle.R`. This site needs its own guard: it reads the container straight off disk with `.readContainerField()` and never builds a `Project`, so the read guard does not cover it. Aborting the overwrite is the safe outcome, since the alternative is unlinking a directory outside the destination.

## Tests

The escaping value is now rejected on load, on set, and on `initProject(overwrite = TRUE)`; the overwrite test also asserts that a bystander file in the parent folder survives. A legitimate rename (`"defs"`) still works, and the existing coverage for relocated definition trees stays green.

There is no `## Example` section: the only way to show the previous behaviour is to run the destructive path, which is not something to put in a reprex.


R/project-lifecycle.R
```diff
@@ -651,7 +651,14 @@ initProject <- function(
...
       containers,
       function(path) {
-        .readContainerField(path, "definitionsFolder") %||% "definitions"
+        folder <- .readContainerField(path, "definitionsFolder")
+        # The container is read straight off disk here, without building a
+        # `Project`, so this is the only guard standing between an untrusted
+        # `definitionsFolder` and the recursive `unlink()` below. Refusing to
+        # clear is the safe outcome: better to abort the overwrite than to
+        # delete a directory outside `destination`.
+        .validateDefinitionsFolder(folder)
```
Felixmil: The guard is in the right place: `vapply()` resolves every container before the delete loop below runs, so an escaping value aborts before any `unlink()` — no half-cleared destination.

One diagnostics detail. `.validateDefinitionsFolder()` defaults `call` to `rlang::caller_env()`, and here the caller is this anonymous `vapply` function, whose `sys.call()` is `FUN(X[[i]], ...)`. So the abort reads ``Error in `FUN()`:``, which names nothing the user can act on. That is out of step with both neighbours: the two sibling aborts in this function (the `failedToClearProjectArtifacts` calls below) attribute to `.clearProjectArtifacts()`, and at the `.readJson()` call site the `local_error_call(NULL)` set at the top of that method is picked up through the environment, so the abort correctly reads as a property of the file with no call at all. The new test can't catch this, since `expect_error(regexp = "single folder name")` only matches the message body.

Hoisting the environment out of the lambda is enough:

```r
  callerEnv <- rlang::caller_env()
  definitionsFolders <- if (length(containers) == 0L) {
    "definitions"
  } else {
    unique(vapply(
      containers,
      function(path) {
        folder <- .readContainerField(path, "definitionsFolder")
        .validateDefinitionsFolder(folder, call = callerEnv)
        folder %||% "definitions"
      },
      character(1)
    ))
  }
```

That attributes the abort to `initProject()`. `call = NULL` is the other defensible choice, and arguably the closer analogue of `.readJson()`: like there, this is a property of the file on disk rather than of the call that opened it.

Minor, same site: when `destination` holds more than one container, the message names the offending value but not which container carried it. The sibling `projectPathEscapesRoot` names the root for the same reason, so passing the container path through to the message entry would keep them consistent — worth it only if you think a multi-container destination is realistic here.

---

R/project.R
```diff
@@ -952,6 +952,10 @@ Project <- R6::R6Class(
           definitionsFolder = list(
             get = function() private$.definitionsFolder %||% "definitions",
             set = function(value) {
+              # Unlike its four siblings, this folder is one plain segment under
+              # the project directory, so it is guarded on write rather than
+              # contained on read; see `.validateDefinitionsFolder()`.
+              .validateDefinitionsFolder(value)
```
Felixmil, line 959: This narrows the field's contract, and the three places that document it still describe the old, wider one. A value like `"config/defs"` used to work (`.definitionsDir()` just `file.path()`s it) and is now a load-time abort, so a reader following the current docs can write a project file the package refuses to load:

- the `@field paths` roxygen a few lines up (`R/project.R:140-141`): "`definitionsFolder` (the folder holding the definition files, default `"definitions"`)" — plus the `man/Project.Rd` regen.
- `vignettes/projects.Rmd:81`: "`definitionsFolder`: where the `definitions/` tree lives (`definitions` by default)".
- `NEWS.md`, in the container bullet: "an optional `definitionsFolder` ... that makes the definition-tree location configurable".

"a single folder name directly under the project folder" in each keeps the docs and the guard saying the same thing. The compat risk of the narrowing itself is nil — `schemaVersion` 2.0 is unreleased, so there are no JSON projects in the wild carrying a nested value.

Also worth one line of coverage here: nothing exercises `project$paths$definitionsFolder <- NULL`, and that `is.null()` passthrough in `.validateDefinitionsFolder()` is what keeps the default reachable after a rename. It sits naturally next to the `"defs"` assertion in the new setter test:

```r
  project$paths$definitionsFolder <- NULL
  expect_identical(project$paths$definitionsFolder, "definitions")
```

---

Felixmil (commented):
## Review

The fix is correct and the guard is complete. Nothing here blocks merging; the two inline comments are a diagnostics bug and a docs/coverage gap.

**What it does.** `definitionsFolder` was the one project-folder field read out of `Project.json` with no containment check, and it is the field that feeds the package's only two delete sites. A new `.validateDefinitionsFolder()` rejects anything that is not a single plain path segment, wired in at the three places the value can enter: the `paths$definitionsFolder` setter, `Project$.readJson()`, and `.clearProjectArtifacts()`.

**Completeness — the guard has no bypass.** `private$.definitionsFolder` is written in exactly three places (`R/project.R:771` NULL init, `:959` the setter, `:1436` `.readJson()`), and the latter two are now guarded, so every in-memory value is checked once. Everything downstream reads the guarded getter: `.writeDefinitionTree()` (`R/definition-files.R:486`) takes `project$paths$definitionsFolder`, and `.projectJsonData()` writes it back from the same getter. I checked the paths that could plausibly slip past:

- `project$paths <- <group>` cannot bypass the setter — `.acceptGroupWriteback()` only accepts this instance's own proxy back, so the per-field setter has already run.
- `restoreProject()` materializes a canonical file and loads it through `Project$new(projectFilePath = canonFile)`, i.e. through `.readJson()`, so a snapshot carrying an escaping value is covered before `.writeProjectTree()`.
- `unlink()` / `file.remove()` across `R/` confirms the PR body's claim: `project-lifecycle.R:670` and `definition-files.R:509` are the only delete sites keyed to this field. The others are `on.exit()` cleanups of temp dirs and the sensitivity output dir.

**Two hazards closed, not one.** Besides `../..`, `""` is now rejected — it would have made `file.path(destination, "")` resolve to `destination/`, so `initProject(overwrite = TRUE)` on a project file with `"definitionsFolder": ""` would have recursively unlinked the destination itself. Might be worth a clause in the log entry.

**Reuse and style.** Reusing `.validateFilenameSegment()` is the right call — a segment check is strictly tighter than the siblings' containment check, and the field has no `${VAR}` opt-out to honour, so `.resolveWorkingFolder()`'s machinery would have been the wrong shape. `messages$invalidDefinitionsFolder` follows the established two-argument `(name, given)` contract exactly, leaves both placeholders unglued, and correctly distinguishes "is no name at all" from "is a name that would escape". Guarding before the assignment means an aborted set leaves neither the field nor the dirty bit touched.

**Tests.** Load, set, and `initProject(overwrite = TRUE)` are each covered, and the overwrite test earns its keep: with `".."` the old code would have unlinked `parent`, so asserting the bystander file *and* the original `definitions/` tree both survive is a real regression test rather than just an error-message assertion. `test-coverage` is green on the head commit, so the suite passes.

**Nits, take or leave:**

- The `REFACTORING-LOG.md` entry is the only recent one without a trailing `(#N)`. `REFACTORING.md` says to omit it when there is no issue, so this is fine as-is, but the work traces to a review comment on #1017 and every neighbour carries a reference.
- `messages$invalidDefinitionsFolder` opens with `{.arg definitionsFolder}`. That reads right for the setter, but on the `.readJson()` path this is a project-file field, not an argument — the sibling load-path message `projectPathEscapesRoot` uses `{.field {fieldName}}` for exactly that case.
- Out of scope, noting only so it is not mistaken for something this PR should have caught: a segment check cannot see a symlink. A shared project folder shipping `definitions -> /somewhere/else` aims the same two delete sites outside the project with the *default* value, so the field is not the vector there. Pre-existing and orthogonal.

---

R/project-lifecycle.R
Felixmil: Fixed in cd0e312a. `caller_env()` is now bound outside the `vapply()` lambda, so the abort reads `Error in initProject():`:

```
Error in initProject(destination = d, type = "minimal", createExcel = FALSE,  :
  definitionsFolder must be a single folder name directly under the
project folder.
x The name ".." contains a path separator or is "." / "..", so it could point
  outside the project folder.
i Pass a plain folder name, for example "definitions" (the default) or "defs".
```

I took `caller_env()` over `call = NULL`. The `.readJson()` analogy holds for where the bad value came from, but not for what the user is left with: that path's message names the file, and this one names only the offending value, so dropping the call too would leave the abort with no location at all.

Not taken: naming which container carried the value. It would mean threading a file path into a message entry the setter path shares, where there is no file, and a destination with more than one container is not a shape worth shaping the message around.

R/project.R
Felixmil: Fixed in cd0e312a. All three spots now say "a single folder name directly under the project folder": the `@field paths` roxygen (plus the `man/Project.Rd` regen), `vignettes/projects.Rmd`, and the `NEWS.md` container bullet.

Added the `NULL` assertion next to the `"defs"` one, so the passthrough that keeps the default reachable after a rename is covered.

Also took the message nit from the review body: `{.arg definitionsFolder}` -> `{.field definitionsFolder}`, which reads right on both paths since the setter target is a field too. And the log entry now mentions the `""` hazard.

sorinvoicu (approved):

Issue #950, opened 2026-02-11
Refer to `ospsuite.plots` documentation in the vignette


PavelBal:
There are some other types of plots implemented in `ospsuite.plots` that are not directly in `{ospsuite}`. Mention this in the vignette.

Issue #771, opened 2025-02-19
Detailed documentation of Plot configuration


PavelBal:
The current documentation on setting up the plots (https://esqlabs.github.io/esqlabsR/articles/esqlabsR-plot-results.html) is very limited. It should be expanded with better examples and explaining the different options (e.g., how to plot population profiles with different quantiles).

PR #1224, opened 2026-08-04
Reorganize the package articles by reader need

### LINKED ISSUE #734: food of thought: potential improvements of documentation to get started with esqlabsR
This is a collection of questions I had when doing the get-started and using the package for the first time for my project
- For all the Excel files, there are so many columns. Maybe one could have documentation (as there is for functions) that states which columns must be defined, which are optional, and which will take the information from the PKSim/MoBi file if one does not add information.
- in 2. Design Scenarios: " A scenario is defined by the simulation file containing the model structure, parametrization of the model, application protocol, and (optionally) the physiology of the simulated individual or population." -> I probably would add PKSim/MoBi to specify the simulation file 
- in 2. Design Scenarios: "If you want to run the simulation with the settings as it has been exported from PK-Sim or MoBi, you can proceed to [vignette("esqlabsR-run-simulations")](https://esqlabs.github.io/esqlabsR/articles/esqlabsR-run-simulations.html)." -> This links a bit to my first point that it was not clear to me what from this number of files and options is needed in the simplest case. This sentence is  bit hidden, but for me, it was super helpful to see how easily I can run my model (from my own project where the excel sheets are not pre-filled) and get the plots in the correct format. 
- the whole DataCombine stuff is a bit unclear to me, which makes it a bit harder to understand how the plotting works. and therefore, my following comment might be more related to that than to the documentation itself
-> How do I get the right "dataSet" entry for the plotting file? (I did figure it out eventually, but it was not obvious to me)

I hope it's helpful

Felixmil:
Reorganizes the package articles so that each one serves a single reader need, following Diátaxis. Every vignette was previously a hybrid: it taught, instructed, listed fields, and argued rationale on the same page, which meant a reader with a task in hand had to wade through reasoning to reach a step, and a reader trying to understand the design had nowhere to look. Two of the gaps this created were already reported as issues: there was no page answering "which fields are required, and which are taken from the PK-Sim file if I omit them" (#734), and plot configuration had nowhere to be documented properly (#771).

Nothing under `R/`, `man/`, `NAMESPACE` or `tests/` is touched, and the pkgdown `reference:` index is unchanged. This is a documentation-only change.

## Article structure

Sixteen articles, up from thirteen, grouped by what the reader needs from them:

| Type | Articles |
| --- | --- |
| Tutorial | `esqlabsR` (the get-started lesson, keeping its name so pkgdown treats it as the top-level Get started entry) |
| How-to | ten guides, each named for the task it performs |
| Explanation | `about-the-project-model`, `about-the-plotting-model` (new) |
| Reference | `reference-project-file`, `reference-definition-fields` (new) |

The four new articles are filled entirely with prose that already existed and was misplaced, plus the field tables described below. Nothing is scaffolded empty.

## Explanation articles

`about-the-project-model` gathers the reasoning that was spread across four guides: why an analysis is a Project rather than a script, why the `definitions/` tree is the authoritative copy, why saving is explicit rather than write-through, why ids are canonicalized, why a reference is checked at author time but only reported when a project is read, why validation has two levels and not three, and why Excel is now an interchange format.

`about-the-plotting-model` covers why a figure takes three definitions rather than one, why `createPlots()` is keyed by plot grid, why observed data is stored as a reference rather than copied in, why a source carries both a source id and a DataSet name, and the trust boundary a `script` source creates.

## Reference articles

`reference-project-file` describes the project directory, the twelve definition sections and their folders, every `Project.json` field, the `${VAR}` path form, the in-memory field groups, the snapshot file, and the Scenario Result record.

`reference-definition-fields` gives one field table per definition kind. Each carries an **If omitted** column saying whether a field falls back to a default, is taken from the model file as exported from PK-Sim or MoBi, or does not apply, which is what #734 asked for. It also holds the `simulationTime` grammar, the scenario stored-field name mapping, the plot type catalog with its per-type fields, the vectorized-authoring rules, and the sensitivity plot configuration matrix.

These tables were derived from the field constants and function signatures in the domain files (`.scenarioFieldNames`, `.settableScenarioFields`, `.validPlotTypes`, `fieldsByPlotType`, and their siblings) rather than transcribed from the existing prose, which is how the corrections below surfaced.

## How-to guides

Each guide keeps its procedure and loses the rationale and field listings, replaced by a sentence and a link. Section headings now name the task they perform. The `script`-source security callout is deliberately kept in the observed-data guide as well as the explanation, since a reader following that procedure must see it there rather than behind a link.

`how-to-plot-results` gains a runnable population example showing how `quantiles` sets the displayed percentile bands, documents the three population-only plot fields, and points at `{ospsuite.plots}` for chart types not reachable through a plot definition (#771, #950).

## Corrections

Four factual errors surfaced while checking the documentation against the code:

- The README described project edits as write-through with no separate save step. They are held in memory until `saveProject()`.
- The README named `saveSnapshot()` / `loadSnapshot()`, which do not exist. The functions are `snapshotProject()` / `restoreProject()`.
- The scenario design guide listed `gender` as required on an individual alongside `species`. Only `species` is required; `gender` defaults to `UNKNOWN`.
- The project folder diagram omitted `definitions/initial-conditions/`.

## Breaking change for readers

The ten task-oriented vignettes are renamed, so the old `vignette()` handles and article URLs no longer resolve. The mapping is recorded in `NEWS.md` under breaking changes. The get-started tutorial keeps its `vignette("esqlabsR")` name.

## Notes for review

- All sixteen articles render on both build paths: `pkgdown::build_site()` and the `R CMD check` route. `devtools::check()` reports one NOTE, the pre-existing long snapshot paths under `tests/testthat/_snaps/sensitivity-*`.
- Every `vignette()` cross-reference resolves to a file, no `_pkgdown.yml` entry names a missing article, and no article is missing from the index.
- `vignette("x")` references render as plain text rather than links in a local build, because `downlit` needs an installed vignette index to autolink. This is pre-existing rather than introduced here, and should resolve on CI, but it is worth confirming on the next site deploy since the reorganization leans on readers following those links.
- No `## Example` reprex section: the change alters no exported function's behaviour.

Closes #734, #771, #950

## Summary by CodeRabbit

* **Documentation**
  * Reorganized documentation into Get Started, How-to, Explanation, and Reference sections.
  * Added guides for project setup, validation, plotting, observed data, and sensitivity-plot customization.
  * Added detailed references for project files and definition fields.
  * Clarified in-memory editing, explicit saving, snapshots, restoration, scenario design, simulations, and plotting workflows.
  * Updated terminology, links, examples, and migration guidance throughout the documentation.


vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+| `readPopulationFromCSV` | no | logical | `FALSE`. |
+
+Two behaviours are not visible from the table:
+
+- **Every reference is checked as the scenario is added.** If `individual`, `application`, or any entry in `parameterSets` does not resolve, `addScenario()` aborts and nothing is created.
+- **`parameterSets` is applied in the order you list them.** Later sets override earlier ones where they touch the same parameter, so `c("global", "aciclovir", "ckd")` applies `global` first and lets `ckd` win.
+
+### Stored field names
+
+The on-disk field names differ from the authoring argument names for the reference fields, which matters when reading a `definitions/scenarios/*.json` file or a record pulled off the project.
```
Felixmil: **The "Stored field names" table gives in-memory record names, not the on-disk JSON keys.**

The section is introduced as what you need "when reading a `definitions/scenarios/*.json` file", but the right-hand column lists the parsed `Scenario` record's names rather than the keys actually written to disk. `inst/extdata/projects/Example/definitions/scenarios/aciclovir_iv.json`:

```json
{
  "name": "aciclovir_iv",
  "individual": "adult_male",
  "population": null,
  "parameterSets": ["global", "aciclovir"],
  "application": "aciclovir_iv_250mg",
  "steadyState": false
}
```

So the disk keys are `name` / `individual` / `population` / `application` / `parameterSets` / `steadyState` — the authoring argument names, unchanged. A reader who hand-authors a scenario file per this table writes `individualId` and `populationId`, which aborts on the missing `name` and silently drops the individual and the parameter sets.

Either retitle the section as the in-memory record mapping, or replace the right-hand column with the real JSON keys.

---

vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+Two functions. `addParameterSet(project, id)` creates an empty set; `addParameterEntry(project, id, containerPath, parameterName, value, units = NULL)` fills it.
+
+| Field | Required | Type | Notes |
+| --- | --- | --- | --- |
+| `containerPath` | **yes** | non-empty string | The container in OSPS path notation. |
+| `parameterName` | **yes** | non-empty string | |
+| `value` | **yes** | numeric scalar | |
+| `units` | no | string | Use `""` for a dimensionless parameter. |
+
+`addParameterEntry()` creates the set on demand if it does not yet exist, and is last-write-wins on a repeated `(containerPath, parameterName)` pair. `removeParameterEntry()` takes an entry back out and auto-removes the set if it becomes empty; `removeParameterSet()` drops a whole set.
```
Felixmil: **`addParameterEntry()` is not last-write-wins — it aborts on a repeated pair.**

`R/parameters.R` documents the internal helper precisely:

> A duplicate `(containerPath, parameterName)` aborts unless `overwrite = TRUE`, in which case it replaces the existing entry (last-write-wins).

So the default is refusal; last-write-wins is what `overwrite = TRUE` buys you. As written, a reader building a set incrementally will expect a silent replace and instead get an error.

The same claim appears at `vignettes/how-to-design-scenarios.Rmd:90` ("`addParameterEntry()` is last-write-wins on a repeated `(containerPath, parameterName)` pair") and needs the same qualification.

---

vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+| Argument shape | Behaviour |
+| --- | --- |
+| The `id` vector | Sets the count, N. |
+| A scalar field | Recycled to all N. |
+| A length-N vector | Aligned to the ids by position, one value per definition. |
+| A field that is itself a vector for one definition (such as `parameterSets`) | Applied whole to each definition. |
+
+The call is **all-or-nothing**: if any one definition fails validation, none is applied.
+
+`addPlot()` and `addPlotGrid()` carry a deliberate asymmetry worth knowing, because it is the one place the rules above do not read the way you would guess. A length-N vector passed to a **positional** argument (`dataCombined`, `plotType`, `plots`) aligns by position as usual. A length-N **atomic vector** passed through `...` is treated as one multi-value field and applied whole, collapsed to a comma-separated string. So `title = c("A", "B")` gives every plot the single title `"A, B"`. To vary a `...` field per definition, pass a length-N **list**: `title = list("A", "B")`.
```
Felixmil: **`plots` in `addPlotGrid()` does not align by position when it is a vector.**

The paragraph groups `plots` with `dataCombined` and `plotType` as positional arguments where "a length-N vector … aligns by position as usual". `R/plots.R:946` does the opposite for `plots` specifically:

```r
# `plots` is whole-per-grid: a bare character vector applies to every grid.
perGridPlotIds <- if (n > 1L && is.list(plots) && length(plots) == n) {
  plots
} else if (is.list(plots)) {
  rep(list(unlist(plots, use.names = FALSE)), n)
} else {
  rep(list(plots), n)          # <- bare vector: applied whole to every grid
}
```

Only a length-N **list** aligns per grid. So `addPlotGrid(project, c("g1", "g2"), plots = c("p1", "p2"))` puts both plots in both grids, not one each — the reverse of what this sentence promises, and the paragraph is specifically the one telling readers where the rules "do not read the way you would guess".

Worth noting the sentence also contradicts `addPlotGrid()`'s own roxygen.

---

vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+
+`addDataCombined(project, id, simulated = list(), observed = list())`. Both arguments are lists of entry lists.
+
+A `simulated` entry:
+
+| Field | Required | Type | If omitted |
+| --- | --- | --- | --- |
+| `label` | **yes** | string | Critical error. Names the curve. |
+| `scenario` | **yes** | id | Critical error. |
+| `path` | **yes** | string | Critical error. The full output path, not an output-path id. |
```
Felixmil: **An output-path id *is* accepted for a DataCombined `path`.**

The table says "The full output path, not an output-path id." `R/data-combined.R:179` resolves either:

```r
# Resolve an output-path id to its literal path; a value that is not a known
# id is used verbatim as a literal model path.
path <- outputPaths[[entry$path]] %||% entry$path
```

`addDataCombined()`'s roxygen documents the id form too. The reference article is the page a reader consults precisely to settle this, so the "not an output-path id" clause steers them away from the shorter, more maintainable form. Suggest: "An output-path id, or a full model path used verbatim when it matches no id."

---

vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+| `weightMin`, `weightMax` | no | numeric | Unconstrained. |
+| `heightMin`, `heightMax` | no | numeric | Unconstrained. |
+| `BMIMin`, `BMIMax` | no | numeric | Unconstrained. |
+| `ageUnit`, `weightUnit`, `heightUnit`, `BMIUnit` | no | string | PK-Sim defaults. |
+| `gender` | no | string | Not set. |
+| `population` | no | string | PK-Sim default. |
+| `diseaseState` | no | string | Healthy. |
+| `proteinOntogenies` | no | character vector | No ontogenies applied. |
+
+For a `csv` population that names no `file`, the CSV is looked for as `<id>.csv` in the populations folder. The lookup is case-insensitive against the folder listing, with an exact match winning, because an id is lowercase while a CSV keeps the case its author gave it.
```
Felixmil: **A `type = "csv"` population with no `file` is a validation error, not an `<id>.csv` fallback.**

This paragraph says the CSV "is looked for as `<id>.csv` in the populations folder" when a `csv` population names no `file`. `R/populations.R` makes `file` mandatory for that source type:

```r
.populationSourceTypes <- c("programmatic", "csv")
# a `csv` source needs `file`
"csv" = "file",
```

and the `<id>.csv` derivation is the *legacy scenario-flag* path (`readPopulationFromCSV`), not the `csv` source type — the comment at `R/populations.R:109` says so explicitly.

This also contradicts the table immediately above it, which marks `file` required for a `csv` source. The case-insensitive-lookup detail is accurate but belongs with the `readPopulationFromCSV` path.

---

vignettes/how-to-identify-parameters.Rmd
```diff
@@ -81,25 +73,27 @@ Print the result for a human-readable optimization summary:
...
+| The estimate sitting at one of its bounds | Whether `minValue` and `maxValue` are too tight. Widen them and refit. |
+| An objective function insensitive to the parameter | Whether the data constrain the parameter at all; the output mapping may not see its effect. |
 
-When you see it, treat the estimate with caution and revisit the bounds, the output mapping, or whether the data constrain the parameter at all.
+Until one of those is resolved, treat the estimate as a point value with no confidence attached.
 
-## Building a task programmatically
+## Build a task in R
 
 Instead of relying on a task stored in the project, you can author one in memory with the `add*` family. `addPITask()` creates a task and requires at least one PI Parameter and one PI Output Mapping at creation, so you seed it with existing records and grow it afterwards. The records are the same plain-data values the project already holds, so the simplest seed is to reuse the ones from the example task.
```
Felixmil: **Stale since #1221: `addPITask()` no longer requires a parameter and a mapping at creation.**

> `addPITask()` creates a task and requires at least one PI Parameter and one PI Output Mapping at creation, so you seed it with existing records and grow it afterwards.

The previous commit on this branch made empty tasks creatable — `scenarios`, `parameters`, and `outputMappings` all default to empty, and the requirement is enforced before `runPI()` rather than at creation. The "so you seed it … and grow it afterwards" rationale falls away with it; seeding from a template is still a fine thing to show, just not a requirement.

Worth flagging that `reference-definition-fields.Rmd` — added by this same PR — already documents these three as not-required-to-create, so the two articles currently contradict each other. (CodeRabbit raised the same underlying issue on the reference table in #discussion_r3712642034; this is the sentence that actually needs the edit.)

---

vignettes/reference-project-file.Rmd
```diff
@@ -0,0 +1,177 @@
...
+| --- | --- |
+| Extension | `.esqlabsR`, always added; any extension you supply is replaced |
+| Content | JSON |
+| Default name | `<projectName>-YYYY-MM-DD-HHMMSS`, falling back to `project` when the project has no name |
+| Includes unsaved edits | yes, the snapshot is of the in-memory state |
+| Needs a folder on disk | no, a project that exists only in the session can be snapshotted |
+| Return value | invisibly, the path written |
+| Existing file | aborts unless `overwrite = TRUE` |
+
+`restoreProject()` materializes a snapshot back into a full project folder (a `Project.json` and the `definitions/` tree) and returns a freshly loaded `Project` bound to that directory. It refuses a directory that already holds a project unless `overwrite = TRUE`. The round trip is lossless: a snapshot reloads into a structurally identical project.
```
Felixmil: **`restoreProject()` refuses *any* non-empty directory, not only one holding a project.**

> It refuses a directory that already holds a project unless `overwrite = TRUE`.

The check in `R/project-snapshot.R` is deliberately broader:

```r
# The non-empty check is broader than `isProjectInitialized()` on purpose, because
# a `dir` holding unrelated files, or a partial `definitions/` tree with no
# `Project.json`, is still not a safe place to unpack into without consent.
dirNotEmpty <- dir.exists(dir) &&
  length(list.files(dir, all.files = TRUE, no.. = TRUE)) > 0L
```

Restoring into a directory holding a README or a stray CSV aborts too. Since this is the reference page, the precise rule is worth stating: it refuses any directory that is not empty.

Same understatement at `vignettes/how-to-set-up-a-project.Rmd:121` ("refuses a directory that already holds a project").

---

vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+```
+
+Field-by-field description of every definition kind, the argument grammars the authoring functions accept, and the plot-configuration support matrices.
+
+Each table's **If omitted** column is the one to read when you are deciding what you actually have to supply: it says whether a field falls back to a default, is taken from the model file as exported from PK-Sim or MoBi, or simply does not apply.
+
+Two conventions run through all of it:
+
+- Every authoring function takes `project` first and the definition's `id` second. Only the fields after those are listed below.
+- Every authoring function accepts `overwrite`, a logical defaulting to `FALSE`. An id that already exists aborts unless you pass `overwrite = TRUE`, which replaces the definition (last write wins). It is omitted from the tables.
```
Felixmil: **The blanket `overwrite = TRUE` rule does not hold for `addObservedData()`.**

> Every authoring function accepts `overwrite` … An id that already exists aborts unless you pass `overwrite = TRUE`, which replaces the definition (last write wins).

`addObservedData()` refuses a cross-kind replacement regardless of `overwrite` — you cannot swap an entry for one of a different source kind by passing the flag. Stating the rule as universal here, and then omitting `overwrite` from every table on the strength of it, means the exception has nowhere to be documented.

A one-clause carve-out in this paragraph would cover it. (Related: the same section notes `addObservedData()` is not vectorized either, so it is already the standing exception to the general rules.)

---

vignettes/how-to-plot-results.qmd
```diff
@@ -202,6 +156,63 @@ reportGrids <- createPlots(
...
+populationPlots <- createPlots(
+  project,
+  plots = "aciclovir_population_range",
+  scenarioResults = populationResults
+)
+
+populationPlots$aciclovir_population_range
+```
+
+To show a different range, change the fractions: `quantiles = c(0.25, 0.5, 0.75)` draws the interquartile band, and a single value such as `quantiles = 0.5` draws the median alone.
```
Felixmil: **`quantiles = 0.5` is not a supported shape.**

> a single value such as `quantiles = 0.5` draws the median alone

The field is passed through to `ospsuite.plots::plotPopulationTimeProfile()`, which consumes three probabilities (lower band / middle / upper band). A scalar does not degrade to "median only" — the sentence invites a call that will not produce the described figure.

The `c(0.25, 0.5, 0.75)` half of the sentence is correct and worth keeping; I'd just drop the scalar clause, or replace it with a narrow band such as `c(0.45, 0.5, 0.55)` if the intent is to show how to de-emphasise the spread.

Flagging as plausible rather than confirmed — I could not execute R here to check the failure mode, so it is worth a quick run before editing.

---

vignettes/.gitignore
```diff
@@ -2,4 +2,4 @@
 **/*.quarto_ipynb
 # Quarto writes rendered figures and JS/CSS libs here when the article is
 # rendered in place (e.g. a local `quarto render` or `pkgdown::build_article()`).
-/plot-results_files/
+/how-to-plot-results_files/
```
Felixmil: **The new `about-the-plotting-model.qmd` also emits an `_files/` directory, and nothing ignores it.**

This entry was correctly renamed to track `how-to-plot-results`, but the PR adds a second `.qmd` article, and it carries a mermaid block — so a local `quarto render` / `pkgdown::build_article()` drops `vignettes/about-the-plotting-model_files/` (mermaid JS/CSS libs) as untracked output.

Either add the sibling entry:

```gitignore
/how-to-plot-results_files/
/about-the-plotting-model_files/
```

or collapse both to a pattern that survives the next `.qmd` article:

```gitignore
/*_files/
```

---

Felixmil (commented):
Reviewed the documentation-only diff (`44769f0..bffcf56`): the vignette reorganization into tutorial / how-to / explanation / reference tiers, the `_pkgdown.yml` article index, and the README/NEWS updates. Every code chunk and factual claim in the new and trimmed articles was checked against `R/`, the bundled example project under `inst/extdata/projects/Example/`, and NAMESPACE.

**What holds up.** The structural work is sound. All vignette cross-references resolve to a file, the pkgdown index covers all 16 articles with none orphaned, and both corrections the PR description claims are genuine — `gender` really is optional (only `species` is required) and the README really did describe edits as write-through. Deriving the field tables from the field constants rather than the existing prose was the right call; it is what surfaced most of what follows.

**What needs fixing.** Ten findings, all wrong factual claims rather than structural problems, and concentrated in the two new reference articles — which is the worst place for them, since a reference page is consulted precisely to settle a question the reader could not answer from the guides. Three would actively break a reader's work if followed:

- The scenario **stored field names** table gives in-memory record names as the on-disk JSON keys, so hand-authoring a scenario file per the table produces one that aborts and silently drops definitions.
- `addParameterEntry()` is documented as last-write-wins when it aborts without `overwrite = TRUE` (claimed in two articles).
- `plots` in `addPlotGrid()` is documented as aligning by position when a bare vector is applied whole to every grid — in the very paragraph warning readers where the rules are counterintuitive, and contradicting the function's own roxygen.

The rest are narrower: the DataCombined `path` id form is documented as unsupported when it works, the `csv`-population fallback claim contradicts both the code and the table above it, the `overwrite` rule is stated as universal when `addObservedData()` is an exception, and `restoreProject()`'s directory check is understated in two places.

Two are worth separate mention. `how-to-identify-parameters.Rmd:98` went stale under the previous commit on this branch (#1221 made empty PI tasks creatable), and it now contradicts the reference article this same PR adds — so the fix is a one-sentence edit, but leaving it means the two tiers disagree on their first day. And `vignettes/.gitignore` was updated for the renamed `how-to-plot-results` but not for the newly added `about-the-plotting-model.qmd`, which also emits an `_files/` directory.

One caveat on my own verification: I could not execute R in this environment, so the `quantiles = 0.5` finding is marked plausible rather than confirmed and deserves a quick run before you act on it. Everything else is confirmed against the source, with the file and line cited inline.

Nothing here argues against the reorganization — the tiering is a clear improvement and the two gaps it closes (#734, #771) are real gaps. The corrections are all local edits to prose and tables.

---

vignettes/about-the-plotting-model.qmd
```diff
@@ -0,0 +1,126 @@
...
+
+The four source types (`excel`, `pkml`, `script`, `programmatic`) look like four separate choices and reduce to one question: **is the data fixed, or is it derived?**
+
+For fixed data you already have, the answer is a file-backed source: `excel` for a workbook read through an OSPS importer configuration, `pkml` for a saved DataSet, `programmatic` for a DataSet you built in the session (which `saveProject()` writes to a PKML file, so it becomes an ordinary file-backed source from then on). Nothing runs when the data is resolved; the values are read back. This is the default, and it should stay the default.
+
+For derived data, the answer is `script`: an R file that computes the DataSet each time it is resolved. Reach for it when the data is not a fixed set of numbers but a recipe (digitized points that need transforming, a join across several inputs, a unit conversion applied to a raw file). The recipe then travels with the project and stays in step with its inputs, which is the whole point.
+
+The `script` and `programmatic` choice is often presented as being about persistence. It is not: both survive a reload. It is about whether the numbers are a given or a computation.
```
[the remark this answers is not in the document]
Felixmil: Confirmed. Fixed in c8427e48.

The paragraph now separates `programmatic` from the other file-backed types: the DataSet lives only on the in-memory project, reloading before a save discards it, and `saveProject()` is what writes it to PKML and rewrites the entry as a `pkml` source (after which it is addressed by the file basename, not the DataSet name). The closing sentence now says the `script` versus `programmatic` choice is not about persistence *once that save has happened*, keeping the fixed-values versus computed-recipe distinction intact.

vignettes/about-the-project-model.Rmd
```diff
@@ -0,0 +1,117 @@
...
+
+`esqlabsR` separates the decisions from the machinery. A **Project** holds the decisions as data: named definitions, each a small declarative record. Functions such as `runScenarios()` and `createPlots()` are the machinery, and they read the project rather than being configured by hand at the call site.
+
+The practical consequences are what make this worth the indirection:
+
+- **An analysis becomes reviewable.** A colleague can read what a scenario does without reading the code that runs it.
+- **An analysis becomes comparable.** Two scenarios that differ only in dose differ in one field, visibly.
+- **An analysis becomes portable.** The whole thing travels as a directory, or as a single snapshot file, and behaves the same on another machine.
```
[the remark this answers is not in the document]
Felixmil: Confirmed — the snapshot inlines the project sections; the referenced working files are not carried.

Fixed in c8427e48 across all three places you listed:

- `about-the-project-model.Rmd:34` — the portability bullet now says the definitions travel and that models, observed data and population CSVs must be provided or resolved separately.
- `how-to-set-up-a-project.Rmd` — "self-contained" dropped, with a paragraph naming what does not travel and pointing at the `${VAR}` path form as one way to resolve it on the far side.
- `reference-project-file.Rmd` — the property table gains explicit **Covers** and **Does not cover** rows, and the lossless round-trip claim is now scoped to metadata and definitions.

vignettes/reference-definition-fields.Rmd
Felixmil: Confirmed against `inst/extdata/projects/Example/definitions/scenarios/aciclovir_iv.json` — the on-disk keys are the authoring argument names, with the id written as `name`. The table was the parsed in-memory record mapping mislabelled as the disk format.

Fixed in c8427e48: the section is now "On-disk keys and in-memory field names". It leads with a real scenario file showing `name` / `individual` / `application` / `parameterSets` / `steadyState`, then keeps the rename table explicitly scoped to a record read from `project$definitions$scenarios`.

vignettes/reference-definition-fields.Rmd
Felixmil: Confirmed at `R/parameters.R:1129` — a duplicate pair aborts, and `overwrite = TRUE` is what buys last-write-wins.

Fixed in c8427e48 in both places: the reference now states the abort and the `overwrite = TRUE` opt-in explicitly, and `how-to-design-scenarios.Rmd:90` carries the same correction.

vignettes/reference-definition-fields.Rmd
Felixmil: Confirmed at `R/plots.R:946` — a bare vector is wrapped with `rep(list(plots), n)`, so it goes to every grid, and only a length-N list aligns per grid.

Fixed in c8427e48. The paragraph is split into two named asymmetries, and the `plots` one now shows the contrast directly:

```r
addPlotGrid(project, c("g1", "g2"), plots = c("p1", "p2"))   # both grids get both plots
addPlotGrid(project, c("g1", "g2"), plots = list("p1", "p2")) # g1 gets p1, g2 gets p2
```

vignettes/reference-definition-fields.Rmd
Felixmil: Confirmed at `R/data-combined.R:179` — `outputPaths[[entry$path]] %||% entry$path` resolves an id first and falls back to the literal path.

Fixed in c8427e48 using your suggested wording: "An output-path id, or a full model path used verbatim when it matches no id."

vignettes/reference-definition-fields.Rmd
Felixmil: Confirmed — `.populationEntryRequiredFields()` makes `file` mandatory for `"csv"`, and the comment at `R/populations.R:109` attributes the `<id>.csv` derivation to the legacy scenario-flag path.

Fixed in c8427e48: the `csv` paragraph now says omitting `file` is a validation error, and the `<id>.csv` lookup (with its case-insensitive matching) is attributed to `readPopulationFromCSV = TRUE` instead. The table above it is no longer contradicted.

vignettes/how-to-identify-parameters.Rmd
Felixmil: Confirmed — `addPITask()` defaults `scenarios = character(0)`, `parameters = list()`, `outputMappings = list()`.

Fixed in c8427e48. The sentence now says a task can be created empty and grown, with the requirement enforced before `runPI()` rather than at creation, and reframes seeding from a template as the quickest route to a runnable task rather than a requirement.

You were right that `reference-definition-fields.Rmd` already had this correct, so no change was needed there — the two articles now agree.

vignettes/reference-project-file.Rmd
Felixmil: Confirmed — the `dirNotEmpty` check in `R/project-snapshot.R:175` is a `list.files()` emptiness test, deliberately broader than `isProjectInitialized()`.

Fixed in c8427e48 in both places: the reference states it refuses any directory that is not empty and notes the check is broader than "already holds a project" on purpose, and `how-to-set-up-a-project.Rmd` adds that a stray README or CSV is enough to stop it.

vignettes/reference-definition-fields.Rmd
Felixmil: Confirmed at `R/observed-data.R:428` — a cross-kind replacement is refused in either direction whatever `overwrite` is set to.

Fixed in c8427e48: the blanket paragraph now carries the carve-out, saying `overwrite = TRUE` replaces only a same-kind source and pointing at `removeObservedData()` for the cross-kind case. It sits alongside the existing note that `addObservedData()` is not vectorized, so the standing exception is stated once in the place the general rule is made.

vignettes/how-to-plot-results.qmd
Felixmil: You were right to ask for a run. I executed all three shapes against the example population scenario:

```
quantiles = 0.05,0.5,0.95 -> OK ggplot2::ggplot
quantiles = 0.5           -> ERROR: non-character argument
quantiles = 0.25,0.5,0.75 -> OK ggplot2::ggplot
```

So the scalar does not degrade to a median, it fails when the figure is built.

Fixed in c8427e48: the scalar clause is gone, replaced with your `c(0.45, 0.5, 0.55)` suggestion for de-emphasising the spread, plus an explicit note that the field takes three probabilities and a single value fails.

vignettes/.gitignore
Felixmil: Confirmed — `about-the-plotting-model.qmd` carries a mermaid block, so it drops its own `_files/` directory on a local render.

Fixed in c8427e48 with the pattern form you suggested, `/*_files/`, so it covers both current `.qmd` articles and any added later rather than needing a new entry each time.

vignettes/about-the-project-model.Rmd
```diff
@@ -0,0 +1,117 @@
...
+
+**Reading a project defers.** `loadProject()` and `importProjectFromExcel()` accept a project whose references do not all resolve. Each section is read on its own, and the gaps are reported rather than fatal. That is the right behaviour because the alternative is a project you cannot open. A legacy Excel project with one real gap, or a work in progress a colleague is halfway through, would become unloadable, and the tools you would use to inspect and repair it all need it loaded first. Reading also tells you about every gap at once, instead of stopping at the first.
+
+So a loaded project can hold a dangling reference that `addScenario()` would have refused. That is the deferred half doing its job, not an inconsistency in the rules. The gate that catches it is further along: `runScenarios()` and `createPlots()` validate before they run, so a gap in the part of the project they are about still stops you before any simulation happens.
```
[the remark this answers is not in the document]
Felixmil: Confirmed — `createPlots()` has `validate = TRUE` as a default parameter, so it is an opt-out.

Fixed in c8427e48: the existing sentence is qualified with "on their default settings", and a following paragraph states that the gate is an opt-out rather than a guarantee, that both `runScenarios()` and `createPlots()` accept `validate = FALSE`, and that a project is therefore only as validated as the calls you made.

vignettes/esqlabsR.Rmd
```diff
@@ -89,9 +84,9 @@ addIndividual(
 )
 ```
 
-The new definition is staged in memory; it lands in `definitions/individuals/adult_female.json` once you call `saveProject(project)`.
+Nothing has been written to disk yet: the new individual is held in memory until you save, which we do in a moment.
 
-With the individual in place, add a **Scenario** that simulates aciclovir in her. A Scenario pins a model to a subject, dosing, parameters, a time grid, and the outputs to record; the `id` comes first and `modelFile` is the only other required field, with everything else an optional reference into the project's other definitions. References are checked as the scenario is added, so a typo in an id is caught here rather than at run time. See `vignette("design-scenarios")` for the full set of fields.
+With the individual in place, add a **Scenario** that simulates aciclovir in her. A Scenario pins a model to a subject, dosing, parameters, a time grid, and the outputs to record. Only `modelFile` is required; everything else is an optional reference to another definition in the project. References are checked as the scenario is added, so a typo in an id stops you here rather than at run time.
```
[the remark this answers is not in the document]
Felixmil: Confirmed. Fixed in c8427e48.

The sentence now separates the two kinds of optional field: `individual`, `application`, `parameterSets` and `outputPaths` reference other definitions, while `simulationTime` and `simulationTimeUnit` are plain values that set the time grid. The `modelFile` requirement and the reference-checking explanation are unchanged.

vignettes/esqlabsR.Rmd
```diff
@@ -159,21 +154,23 @@ plots <- createPlots(
...
-- `vignette("design-scenarios")` covers the full authoring API: every supporting definition, editing in place with the `set*` family, vectorized authoring (adding several definitions in one call), and seeding scenarios from model files.
-- `vignette("validate-project")` shows how to check that everything hangs together before you run it.
-- `vignette("run-simulations")` covers running several scenarios at once, populations, steady-state, and saving and reloading results.
-- `vignette("plot-results")` covers richer figures, attaching observed data, and the esqLABS house style.
-- `vignette("observed-data")` covers the ways measured data enters a project.
+Two other kinds of article are worth knowing about. `vignette("about-the-project-model")` and `vignette("about-the-plotting-model")` explain why the package works the way it does, which is the fastest way to stop being surprised by it. `vignette("reference-definition-fields")` and `vignette("reference-project-file")` are for looking things up: every field of every definition, whether it is required, and what happens when you leave it out.
```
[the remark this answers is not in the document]
Felixmil: Confirmed. Fixed in c8427e48 — the closing list now covers all ten guides, adding parameter identification, sensitivity analysis, sensitivity-plot customization, and Excel migration, each with a one-line description of when to reach for it.

vignettes/how-to-set-up-a-project.Rmd
```diff
@@ -0,0 +1,127 @@
...
+project$definitions$scenarios
+```
+
+The sections are read-only. To change a definition, use an authoring function (see `vignette("how-to-design-scenarios")`) or edit its JSON file directly.
+
+If `loadProject()` spots an obvious cross-reference mistake, such as a scenario pointing at an individual that is not defined, it warns immediately but still returns the project, so you can inspect and repair it. For the full picture run `validateProject()`.
+
+## Scaffold a new project
+
+`initProject()` creates a project in a directory that already exists. Pass `type = "minimal"` for an empty scaffold, or `type = "example"` to copy the complete worked example so you have something runnable to start from.
```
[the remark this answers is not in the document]
Felixmil: Confirmed at `R/project-lifecycle.R:494` — `fs::dir_create(destination)` runs before the existing-project check.

Fixed in c8427e48: the sentence now reads that `initProject()` creates a project at `destination`, creating that directory if it is not there yet.

vignettes/reference-definition-fields.Rmd
```diff
@@ -0,0 +1,383 @@
...
+`addPITask(project, id, scenarios, parameters, outputMappings, configuration)`.
+
+| Field | Required to create | Required to run | Type |
+| --- | --- | --- | --- |
+| `scenarios` | no | **yes** | character vector of scenario ids |
+| `parameters` | no | **yes** | list of PI Parameter records |
+| `outputMappings` | no | **yes** | list of PI Output Mapping records |
+| `configuration` | no | no | list (algorithm, confidence-interval method, and similar) |
+
+A task can be created empty and seeded afterwards, but an empty task cannot run: validation reports a missing `scenarios`, `parameters`, or `outputMappings` as a critical error, and the run gates on validation.
```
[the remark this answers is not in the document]
Felixmil: Verified but no change needed in this file: the table here already documents `scenarios`, `parameters` and `outputMappings` as not required to create and required to run, which matches the empty defaults in `addPITask()`.

The stale claim was in `how-to-identify-parameters.Rmd` ("requires at least one PI Parameter and one PI Output Mapping at creation"), and that is what c8427e48 corrects, so the two articles now agree. Tracked on the sibling thread for that line.

PR #1225, opened 2026-08-05
Share duplicated logic, split the Excel importer, and make the test suite assert its warnings

Felixmil:
This branch is a code-quality pass over the 6.0.0 JSON-project code, aimed at the three things that make an agent-written codebase hard for a human to keep: logic written out more than once, functions too long to read in one sitting, and a test suite whose green run hides things. Two user-facing behaviour changes, both noted below; everything else preserves behaviour. The suite goes from 3344 to 3401 passing tests with no failures.

## Duplicated logic

The rule for "this argument must be a single non-empty string" existed as 26 hand-written copies across eight files, while a function doing exactly that already sat in `R/scenarios.R` under a scenario-specific name. It moved to `R/utils.R` as `.assertNonEmptyString()` with the predicate `.isNonEmptyString()` behind it, and all 26 sites now call it.

Other rules that were written more than once and are now written once:

- the individual field lists, as `.individualFields` / `.individualNumericFields`, matching the `.populationNumericFields` constant `R/populations.R` already had
- the numeric-field validity check shared by individuals and populations, as `.isInvalidNumericField()` beside its two existing siblings
- the parameter-identification task lookup, as `.requirePITask()`, replacing the same four lines in four authoring impls
- the whole task member-removal body, as `.removePIMember()`, including the rule that a task left with neither parameters nor output mappings is deleted rather than stored empty
- the parameter-set merge loop in `.mergeScenarioParameters()`, written three times in a row
- the individuals and applications parameter-set reference checks in `.validateCrossReferences()`, which were byte-identical apart from the kind they name

Two candidates were deliberately left alone, because the function that would remove them only forwards arguments and hides no decision: the shared opening of `.runScenariosFromProject()` and `.buildSimulationsFromProject()`, and the `.appendParameterSets()` call block in the Excel importer.

Adding a field to a scenario now means editing two places instead of three: `Project$addScenario()` forwards through `...` like the other authoring methods already did.

## Long functions

`importProjectFromExcel()` was 629 lines. The 260-line table describing what each workbook contributes now lives in `.excelImportSections()`, leaving the importer at 366. The section closures reached the enclosing frame with `<<-` for one variable; that is now an explicit environment the caller owns, so the extraction does not change where the write lands.

`.createScenariosFromPKML_impl()` (447 lines) and the remaining phases of `.validateCrossReferences()` (292 lines) are untouched and worth their own changes.

## A dependency reach that fails R CMD check

`messages$notNamedList()` called `ospsuite.utils:::.getCallingFunctionName()`, reaching into another package's internal namespace, which `R CMD check` rejects outright. It has been there since #685. The message no longer spells out the calling function; `.validateIsNamedList()` takes a `call` and hands it to the abort, so the error still names the public function the user called. The only visible change is that the message text no longer carries a `sensitivityCalculation():` prefix.

`R CMD check` now reports `Status: 1 NOTE`, and that NOTE is pre-existing: 14 snapshot and fixture paths exceed the portable-length limit.

## A NaN that read as "field not set"

`is.na(NaN)` is `TRUE`, so `NaN` was classified as an unset numeric field: `addIndividual(weight = NaN)` silently omitted the field and `setIndividual(weight = NaN)` silently cleared whatever the field already held, while `weight = "80kg"` aborted as it should. The same held for populations. No workbook cell reads as `NaN`; it arrives from a calculation that went wrong, so `.isUnsetNumericField()` now excludes it and the existing guards abort. `NA` still means "not set" and still clears a field, which is what an empty Excel cell relies on.

This is pre-existing rather than introduced here, and it was found by review of this branch. Behaviour change: code passing `NaN` now gets an error instead of a silently dropped or cleared field.

## Test suite

The suite raised 49 warnings that nothing asserted, so a regression that stopped warning would have passed silently. It now raises 11, and all 11 come from dependencies rather than from this package. Where a test used an id in a spelling the project immediately rewrites, it now uses the stored spelling, so no warning fires and no coverage is lost (canonicalization is pinned separately in `test-definition-id.R`). Where the rewriting is the point of the test, the warning is asserted.

Clearing that noise exposed a test passing for the wrong reason. `"removeOutputPath() warns when the path is referenced only by a PI mapping"` asserted a string that appears only in the canonicalization warning, never in the "still referenced by 1 parameter identification task" warning it is named for, so the check it existed to make was never made. It now asserts the real message.

The two warnings from `plotIndividualTimeProfile()`, which ospsuite deprecated in 12.4.2 and will remove in 14.0, are left unsuppressed on purpose: the replacement is not a drop-in, and the warning is the only thing carrying that deadline.

Fixture duplication is down as well. The parameter-identification tests built the same fitting task by hand a dozen times and now call one shared builder. The Excel tests set the same fixtures up 45 times over, including 11 sites reimplementing the `exampleProject()` helper that already existed; that setup now lives in `localExcelProjectDir()`, `excelRoundTrip()` and `scenarioSheetRow()`, and the file is 167 lines shorter. `excelRoundTrip()` muffles only the one cross-reference warning the example project always raises, not every warning, so anything new still surfaces.

`R/definition-list.R` and `R/project-field-group.R` had no test file at all and now have one each, covering the aborts on every assignment form into a project section and the per-field closure capture the source comments warn about.

## Naming

`test-snapshot.R` did not mirror `R/project-snapshot.R`, so `devtools::test_active_file()` could not find it; it and its `_snaps/` file are renamed. Six comments that recounted how the code used to work were rewritten to describe what it does, per the comment rule in `REFACTORING.md`.


R/authoring-vectorize.R
```diff
@@ -353,3 +353,21 @@ NULL
...
+#
+# @keywords internal
+# @noRd
+.isInvalidNumericField <- function(value) {
+  if (is.null(value) || .isUnsetNumericField(value)) {
+    return(FALSE)
+  }
+  coerced <- suppressWarnings(as.double(value))
+  length(value) != 1L || is.na(coerced) || !is.finite(coerced)
+}
```
Felixmil, line 378: Re: CodeRabbit's suggestion to reject `NaN` here — checked it against the code this consolidates. `.isUnsetNumericField()` (line 335) is `length(value) == 1L && is.na(value)`, and `is.na(NaN)` is `TRUE` in R, so a literal `NaN` already hit the "unset" branch and skipped validation before this PR, in all three call sites this replaces (`addIndividual`/`setIndividual` weight/height/age, `addPopulation`/`setPopulation`'s numeric range fields). This function reproduces that behavior unchanged.

So it's a real gap, but not one this PR introduces — the PR is explicitly a no-behavior-change dedup ("No user-facing behaviour changes except one error message"). Fixing it here would be a scope creep into the exact validation-tightening this branch is deliberately not doing. Worth its own follow-up issue (tighten `.isUnsetNumericField`/`.isInvalidNumericField` to distinguish `NaN` from `NA`) rather than folding into this PR — happy to file it if useful.

---

R/project-excel.R
```diff
@@ -485,6 +273,234 @@ importProjectFromExcel <- function(
...
+  # a valid Excel workbook, so `readxl` raises a raw "zip file cannot be opened"
+  # error that names nothing useful; wrap it in a clear message naming the path.
+  pcExcel <- tryCatch(
+    readExcel(projectConfigPath),
+    error = function(e) {
+      cli::cli_abort(
+        c(
+          "{.path {projectConfigPath}} is not a readable Excel project file.",
+          "i" = "It must be a valid {.field .xlsx} workbook \\
+          (the project's {.file Project.xlsx})."
```
Felixmil, line 356: Re: CodeRabbit's flag on the backslash-newline continuation here — this is an established convention already used throughout the codebase (e.g. `R/definition-files.R`, `R/definition-id.R`, and `R/authoring-vectorize.R:117`, all pre-existing), relying on glue's line-continuation trim (backslash + newline + leading whitespace collapses to nothing) so a long message can wrap in source without a hard line break in the rendered text. Not a bug introduced here, and changing it would be inconsistent with the rest of the file/codebase, so I'd leave it as-is.

---

Felixmil (commented):
Self-review pass (medium depth) over this diff against `json-based-project` (base `44ffa45`, head `b5a527b`).

**Overview:** Pure code-quality pass as described — dedup of 26 hand-written "non-empty string" checks into `.assertNonEmptyString()`/`.isNonEmptyString()`, shared `.isInvalidNumericField()` for individuals/populations, `.requirePITask()`/`.removePIMember()` collapsing four PI authoring impls, `.extendWithParameterSets()` collapsing three copies of the scenario-parameter merge loop, `.checkParameterSetRefs()` deduping the individuals/applications cross-reference check, `Project$addScenario()` now forwarding `...`, the Excel importer split into `.excelImportSections()` (629 → 366 lines, `<<-` replaced by an explicit `rlang::env()` the caller owns), and the `ospsuite.utils:::` internal-namespace reach removed from `messages$notNamedList()`.

**Correctness spot-checks (traced each dedup against what it replaced):**
- `.isNonEmptyString`/`.assertNonEmptyString` — message text and predicate logic match the 8+ inline sites checked (individuals, populations, scenarios, parameter-identification, parameters, output-paths).
- `.isInvalidNumericField()` — behaviorally identical to the inline code in `addIndividual`/`setIndividual`/`addPopulation`/`setPopulation` (see inline comment re: the one pre-existing edge case CodeRabbit flagged).
- `.requirePITask()`/`.removePIMember(field=, label=)` — the empty-task deletion check reads `piTask$parameters`/`piTask$outputMappings` directly after `piTask[[field]] <-` mutates the right one; correct either way `field` is called with.
- `.extendWithParameterSets()` — the individual/application null-guards it drops (`if (!is.null(indivData))`, etc.) are redundant given `unlist(NULL$x)` → `NULL` → zero-iteration loop, not behavior changes.
- `Project$addScenario(...)` forwarding — verified `.addScenario_impl()`'s defaults (`simulationTimeUnit = "h"`, `steadyState = FALSE`, etc.) exactly match what was removed from the R6 method's own signature, and the free `addScenario()` now calls `project$addScenario()` with all-named arguments, so position-based forwarding isn't a hazard.
- Excel importer's `importState <- rlang::env(...)`, threaded into `.excelImportSections(state)` — correct replacement for the `<<-`; environments mutate by reference so `state$x <- ...` inside the closure is visible to the caller without needing `<<-`.
- The `test-parameter-identification.R` fix for `"removeOutputPath() warns when the path is referenced only by a PI mapping"` is real: the old assertion matched the substring `"PIOnlyPath"`, which the canonicalization-collector warning also contains, so it was never actually exercising the "still referenced by 1 parameter identification task" path it's named for. Confirmed against the base branch's version of the test.

**Re: CodeRabbit's two actionable comments** — left inline replies. Short version: the `NaN`-as-unset one is real but pre-existing (not introduced here, and fixing it would break this PR's declared no-behavior-change scope — worth its own issue); the backslash-continuation one is a false positive, it's an established repo-wide convention.

**CI / coverage:** all checks green (R-CMD-Check × 3 platforms, pkgdown, lychee, test-coverage); patch coverage 96.8%, project coverage up 94.20% → 94.25%. `mergeable_state: clean`. Didn't find anything blocking.

---

R/authoring-vectorize.R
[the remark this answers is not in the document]
Felixmil: Confirmed and fixed in 41a87a4b.

Verified the whole chain before changing anything: `is.na(NaN)` is `TRUE`, so `.isUnsetNumericField(NaN)` returned `TRUE`, `.isInvalidNumericField(NaN)` short-circuited to `FALSE`, and `.coerceNumericField(NaN)` returned `NULL`. End to end that meant `addIndividual(weight = NaN)` stored no weight at all, and `setIndividual(weight = NaN)` on an individual with `weight = 70` left it cleared, while `weight = "80kg"` aborted correctly. Both held for `addPopulation` / `setPopulation` on `weightMin`, and for `numberOfIndividuals`.

Worth recording that this is not a regression from this PR: I ran the same probe against `origin/json-based-project` and got identical results, so the behaviour predates the refactor and the extraction of `.isInvalidNumericField()` preserved it faithfully. It only became visible because the rule now sits in one named place.

Took the proposed one-line fix. Checked first that nothing can now store `NaN` instead: all four `.coerceNumericField()` call sites sit after the validation loop in the same function, so the abort fires before coercion is reached. `NA` still means "not set" and still clears a field, which is what an empty workbook cell depends on.

Tests added for all five sites you listed, plus the helpers directly over `NA`, `NaN`, `Inf`, a bad string, a length-2 vector and `NULL`. Suite is green at 3422 passing, up 21.

R/project-excel.R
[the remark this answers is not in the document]
Felixmil: Skipping this one: the message renders correctly as it stands.

I checked what it actually emits rather than reading the source, and `cli` resolves the backslash-newline on its own pass, so there is no preserved newline and no leaked indentation:

```
Error in `importProjectFromExcel()`:
! '/tmp/.../file.xlsx' is not a readable Excel project file.
i It must be a valid .xlsx workbook (the project's 'Project.xlsx').
Caused by error in `utils::unzip()`:
! zip file '...' cannot be opened
```

`R/messages.R` documents this as the house convention in its header: a long template wraps across physical lines with a trailing `\`, the builder leaves the continuation in the string, and the raising `cli_abort()` / `cli_warn()` resolves it. The pattern is used throughout the catalog, so changing this one site would make it the odd one out.

Separately, this string is not part of this PR. It came in with #1097 and is untouched here; it only shows up in the diff because lifting the section table out of `importProjectFromExcel()` shifted the line numbers.

R/authoring-vectorize.R
Felixmil: Superseded: the gap is fixed in 41a87a4b, which landed a few seconds after the comment above was written.

The analysis above is right that the behaviour predates this branch (I confirmed the same by running the case against `origin/json-based-project`), but it was fixed here rather than deferred. `.isUnsetNumericField()` now excludes `NaN`, so `.isInvalidNumericField()` reaches its finiteness check and the existing guards abort. `NA` still means "not set" and still clears a field, which is what an empty workbook cell depends on.

Behaviour change, recorded in the PR description: code passing `NaN` gets an error instead of a silently dropped or cleared field.

R/project-excel.R
Felixmil: Agreed, no change made. Confirmed by checking the rendered output rather than the source: `cli` resolves the continuation, so the message emits as a single line with no leaked indentation.

Issue #1226, opened 2026-08-05
Architecture audit: recommended actions for the 6.0.0 project model


Felixmil:
Architecture audit of the 6.0.0 JSON project model, covering layer boundaries, bounded contexts, module depth and public surface. The structure came out well: the `REFACTORING.md` conventions are genuinely followed, validation is a clean convention-based dispatcher, the domain files are 1:1 with their tests, and the private read/write seam is real. This issue collects the actions worth taking, in priority order, so they can be split into their own issues or picked off one at a time.

Measured baseline at the time of the audit: 40 files in `R/`, 31,224 lines, 116 exports, 39 test files. The `Project` R6 class carries 8 active bindings, 58 public methods, 20 section fields and 28 private methods. The authoring surface is 50 exported free functions, 58 methods and 47 `_impl` functions.

Part of #1088.

### 1. Write a glossary and name the overloaded domain terms apart

"Scenario" currently means three different things with no boundary marker: a JSON definition (the `scenarios` section, `addScenario()`), a runtime S3 object (`Scenario()` in `R/scenarios.R`, with `as.list.Scenario` and `print.Scenario`), and an execution unit (`.prepareScenario()` in `R/scenario-execution.R`, which becomes an `ospsuite` `SimulationSet`). "Parameters" spans four meanings: the `parameterSets` section, the `initialConditions` section (also parameters), `PIParameter`, and raw `ospsuite` parameter paths. These are legitimately distinct concepts, which is exactly the case for naming them apart rather than unifying them, but reading `scenario` in an unfamiliar file today tells you nothing about which one you have.

The action is a glossary table of term, meaning and code name, not a rename sweep. Renames, if any, follow from it later and one at a time. Cheapest item here and the highest ratio of value to effort. Fits under #1149.

### 2. Collapse the pass-through authoring methods on `Project` into one dispatch member

Every authoring operation exists three times: an exported free function that guards the type and forwards, a public `Project` method whose whole body is `private$.impl(.someOperation_impl, ...)`, and the `_impl` that holds the logic. The middle layer holds no logic at all. It exists only because R6 privacy stops a free function from reaching `private`, so the method is the only door to `private$.impl()`.

[...]

PR #1227, opened 2026-08-05
Extract the project load's I/O out of the constructor; add a domain-term glossary

### LINKED ISSUE #1150: Project lifecycle atomicity (load/save/reload)
Umbrella for the atomicity of the v6 project lifecycle operations. Part of #1088.

`saveProject()` and in-place `reloadProject()` are best-effort today: a failure partway can leave the on-disk tree or the in-memory object in a mixed state. Tracks making both operations all-or-nothing (write-to-temp-then-swap on save; parse-then-commit on reload), decided together since they are the write/read counterparts of the same gap.

Felixmil:
Ships two of the three top-priority actions from the architecture audit (#1226): action 3 (move file loading out of the `Project` constructor) and action 1 (a glossary for the overloaded domain terms). Action 2 (collapse the 58 pass-through authoring methods into one dispatch member) is scoped out — see below.

## Action 3: move the load's I/O out of `Project`

`Project$new()`'s constructor and the private `.reload()` both used to parse-and-assign in one pass (`private$.readJson()`): read the container JSON, split `filePaths`/`excel`, and load every section's `definitions/<kind>/` tree, assigning straight into `private$` field-by-field as it went. A failure partway (an unsupported schema version hit after `outputPaths` already loaded, say) left the object half-populated.

- `R/project-lifecycle.R`: new free function `.loadProjectTree(path)` does all of that I/O and returns a plain list — nothing is assigned anywhere until the whole tree has parsed successfully.
- `R/project.R`: new private method `.applyLoadedSections(sections)` commits that list to `private` in one unconditional pass (plus resetting the runtime-only state: session-added observed data/populations, the two caches, the dirty/validation flags). The constructor and `.reload()` both call `.applyLoadedSections(.loadProjectTree(...))` directly; `.readJson()` and the private `.loadSection()` are deleted (fully superseded).

**`Project$new(path)` keeps working exactly as before** — same signature, same validation, same behavior for every caller. `loadProject()`, `ProjectConfiguration()`, the Excel bridge, and `restoreProject()` are all unaffected.

Side effect worth flagging (not the primary goal, and this doesn't claim to close #1150, which also covers `saveProject()` atomicity): a failed `reloadProject()` now leaves the project's in-memory state completely untouched, instead of partially overwriting it.

## Action 1: glossary

Adds a **Glossary** section to `CLAUDE.md` (right after the `REFACTORING.md` include, so it surfaces automatically with the rest of the project instructions) disambiguating the three things called "Scenario" and the four called "Parameters" that #1226 names — term, meaning, code name. Descriptive only, no renames.

One correction against the issue's own text: `.prepareScenario()` returns `list(simulation, population)` (an `ospsuite::Simulation` + `ospsuite::Population`), not an `ospsuite::SimulationSet` — grepped the whole package, that type isn't used anywhere in `R/`.

## Action 2: scoped out

The issue's design collapses the 58 pass-through methods (`project$addScenario()` etc.) into one dispatch member, which deletes the method form outright. Both entry points (`addScenario(project, ...)` and `project$addScenario(...)`) need to keep working, so that design doesn't apply here. The issue's own stated fallback — generate the 58 methods from a spec table instead of hand-writing them — was also declined. `R/project.R`'s 58 methods are untouched in this PR; worth a follow-up issue if the duplication is still worth trimming under that fallback shape.

## PR #1225

Checked `gh pr diff 1225` before starting (it has since merged into `json-based-project`, squash commit `6448e815`). It restructured `R/project-excel.R` and touched `R/project.R`'s `addScenario` method. Since action 2 is scoped out, this PR never touches either of those, so there's no overlap.

Closes nothing directly; addresses actions 1 and 3 of #1226.


tests/testthat/test-project-lifecycle.R
```diff
@@ -5,6 +5,34 @@ test_that("loadProject() returns a Project from a valid Project.json", {
...
 
+test_that(".loadProjectTree() returns a plain list, not a Project, with every section", {
+  path <- file.path(.copyTestProjectDir(), "Project.json")
+  sections <- .loadProjectTree(path)
+
+  expect_type(sections, "list")
+  expect_false(inherits(sections, "Project"))
+  expect_equal(sections$schemaVersion, "2.0")
+  expect_equal(fs::path_file(sections$projectFilePath), "Project.json")
+  expect_equal(sections$projectDirPath, dirname(sections$projectFilePath))
```
Felixmil: The behaviour this refactor exists to produce — a failed load or reload leaves nothing half-populated — has no test, while the assertions that were added are largely tautological.

`expect_false(inherits(sections, "Project"))` can never fail: `sections` is a bare `list()` literal built in the `return`, so it carries no class attribute at all. Same for `expect_equal(sections$projectDirPath, dirname(sections$projectFilePath))` — the returned list sets `projectDirPath = dirname(jsonPath)` and `projectFilePath = jsonPath`, so this restates the assignment rather than pinning a contract.

The atomicity guarantee is the testable part, and it's cheap to cover: load a project, mutate it, corrupt one definition file on disk (or bump `schemaVersion` to something unsupported), then assert `reloadProject()` errors *and* the in-memory sections are still exactly what they were before the failed reload. The existing `reloadProject()` tests around line 317 cover discard-edits, silence, and the unbound-project abort, but none of them exercise a failing reload. Without that, nothing stops a future edit from reintroducing field-by-field assignment.

---

R/project.R
```diff
@@ -1358,126 +1353,36 @@ Project <- R6::R6Class(
...
-      private$.plots <- private$.loadSection("plots", jsonData)
-      private$.plotGrids <- private$.loadSection("plotGrids", jsonData)
-      private$.parameterIdentification <- private$.loadSection(
-        "parameterIdentification",
-        jsonData
-      )
+    # Commit a `.loadProjectTree()` snapshot to `private` in one
+    # unconditional pass: every field the parse produced is assigned, and the
+    # runtime-only state a tree carries no record of (session-added observed
+    # data and populations, the two caches, the dirty and validation flags)
```
Felixmil: "the two caches" doesn't match the body. The reset block clears five things: `.programmaticDataSets`, `.observedDataNamesCache`, `.programmaticPopulations`, `.validatedSinceMutation`, `.modified`. Of those, exactly one is a cache — `.observedDataNamesCache` (the only `*Cache` private field in the class, declared at line 719). The two programmatic stores are already covered by "session-added observed data and populations" earlier in the same parenthetical, and the flags by "the dirty and validation flags" after it, so counting the stores as caches double-counts them.

Suggest "the observed-data names cache" (singular, named) so the list enumerates the five resets without overlap.

---

R/messages.R
```diff
@@ -731,7 +731,7 @@ messages$legacySnapshotNotLoadable <- function(jsonPath) {
   )
 }
 
-# Raised from `Project$.readJson()` when it is handed a folder that holds no
+# Raised from `.loadProjectTree()` when it is handed a folder that holds no
```
Felixmil: This one isn't raised from `.loadProjectTree()`. `noProjectContainerInFolder` is raised in `.resolveProjectContainerPath()` (`R/project-lifecycle.R:498`), which is where `folder` is bound — and the comment's next-to-last sentence depends on that: "interpolated exactly once, by the raising call, where `folder` is bound" only makes sense if it names the frame that actually binds it. `.loadProjectTree()` has no `folder` in scope.

The rename mechanically swapped `Project$.readJson()` for `.loadProjectTree()`, but the old attribution was already wrong here (`.readJson()` also only *called* the resolver), so the rename carried the error forward. Since the line is being touched anyway, worth pointing it at `.resolveProjectContainerPath()`.

Contrast the two entries above at lines 190 and 715 (`unsupportedSchemaVersion`, `legacySnapshotNotLoadable`), where `.loadProjectTree()` is now correct — those really do abort in that frame.

---

R/messages.R
```diff
@@ -747,7 +747,7 @@ messages$noProjectContainerInFolder <- function(folder) {
   )
 }
 
-# Raised from `Project$.readJson()` when a folder holds several project
+# Raised from `.loadProjectTree()` when a folder holds several project
```
Felixmil: Same mis-attribution as the entry above: `multipleProjectContainersInFolder` is raised in `.resolveProjectContainerPath()` (`R/project-lifecycle.R:501`), which is also where `folder` and `names` are bound — as this comment's own "bound in the raising frame" clause asserts. `.loadProjectTree()` binds neither.

---

R/project.R
```diff
@@ -248,15 +248,10 @@ Project <- R6::R6Class(
...
-        !is.character(projectFilePath) ||
-          length(projectFilePath) != 1L ||
-          is.na(projectFilePath) ||
-          !nzchar(projectFilePath)
-      ) {
-        cli::cli_abort(messages$invalidPathArgument())
-      }
-      private$.readJson(projectFilePath)
+      # `.loadProjectTree()` validates `projectFilePath` itself (the same
+      # checks this method used to run inline) and does all of the load's
```
Felixmil: "the same checks this method used to run inline" explains the call by what the code no longer does. `REFACTORING.md` rules that out explicitly: comments "describe how the code works currently, never its history (no 'previously', 'used to', ...); that belongs in commits and PRs, not the source." A reader arriving fresh has no inline check to compare against, so the parenthetical is dead weight for them and stale the moment the guard moves again.

The load-bearing half is the first clause — that path validation happens inside `.loadProjectTree()` rather than here — which stands on its own:

```r
# `.loadProjectTree()` validates `projectFilePath` and does all of the
# load's I/O; this just commits what it returns.
```

---

R/project-lifecycle.R
```diff
@@ -65,6 +65,162 @@ loadProject <- function(path = ".") {
...
+# @noRd
+.loadProjectTree <- function(path) {
+  rlang::local_error_call(NULL)
+  if (
+    !is.character(path) ||
+      length(path) != 1L ||
+      is.na(path) ||
+      !nzchar(path)
+  ) {
+    cli::cli_abort(messages$invalidPathArgument())
```
Felixmil: Moving the path-argument guard in here puts it under this function's `local_error_call(NULL)`, which changes how it reports. `initialize()` ran this check in its own frame, so `Project$new(NULL)` aborted as `Error in initialize():`; now it lands as a bare `Error:` with no function named at all.

That's at odds with the docblock two lines up, which justifies the `NULL` error call on the grounds that "everything this aborts on is a property of the file, not of the call that opened it." True of the four aborts below — file-not-found, unparseable JSON, legacy snapshot, unsupported schema version — but this one is precisely the opposite: a non-string / length != 1 / `NA` / empty `path` is a property of the *call*, and it's the one abort where naming the caller's function tells the user where to look. It also can't be "a property of the file", since no file has been touched yet at this point.

Cheap fix: hoist `rlang::local_error_call(NULL)` to just after the guard, so the argument check aborts with the inherited call and everything downstream keeps the call-less attribution the comment describes.

Note `initialize()` only reaches here for a non-`character()` argument — zero-length character takes the empty-project early return at `R/project.R:246` — so the reachable trigger is `Project$new(NULL)` / `loadProject(NULL)`, `loadProject(NA)`, `loadProject("")`, and `loadProject(c("a", "b"))`. Nothing in the suite asserts the call prefix on that message, so this is cosmetic rather than a broken test.

---

Felixmil (commented):
Reviewed `6448e815..bb292f06` (8 commits: the `.readJson()` → `.loadProjectTree()` + `.applyLoadedSections()` split, the load/reload rewiring, and the glossary). No R toolchain in this environment, so this is a static review; the three `R-CMD-Check` jobs were still in progress when I looked, with `pkgdown`, `check-changes`, and the link check already green.

**The extraction itself checks out.** Every private field the deleted `.readJson()` assigned is assigned by `.applyLoadedSections()`, and no caller of the deleted `.readJson()` / private `.loadSection()` survives — `initialize()` and `.reload()` are the only two, and both now compose the pair directly. The `list(definitions = list(outputPaths = outputPaths))` stand-in handed to the scenarios parser is safe: `scenarios` is the only definition-tree spec whose `parse` touches its `project` argument at all, so `project = NULL` for the other twelve kinds is fine. The load-order constraint (`outputPaths` before `scenarios`) is preserved and now documented at the point where it's enforced. Glossary entries all resolve to real symbols, and the `.prepareScenario()` correction in the PR body is right — `ospsuite::SimulationSet` appears nowhere in `R/`.

Nothing here is a correctness bug. Six comment/test-quality findings, roughly in order of how much they matter:

1. **`tests/testthat/test-project-lifecycle.R`** — the guarantee this refactor exists to produce (a failed load/reload leaves nothing half-populated) is untested, and two of the added assertions can't fail.
2. **`R/project.R:1359`** — `.applyLoadedSections()`'s header says it resets "the two caches"; there is one cache field in the class.
3. **`R/messages.R:734` and `:750`** — both comments now attribute their abort to `.loadProjectTree()`, but both are raised in `.resolveProjectContainerPath()`; the rename carried an already-wrong attribution forward, and each comment's own "where `folder` is bound" clause depends on naming the right frame.
4. **`R/project.R:251`** — the new comment explains the call by what the method "used to run inline", the historical narration `REFACTORING.md` rules out.
5. **`R/project-lifecycle.R:91`** — the path-argument guard now aborts with a bare `Error:`; it's the one abort in this function that *is* a property of the call, not of the file.

Scoping action 2 out looks right to me: both `addScenario(project, ...)` and `project$addScenario(...)` have to keep working, which the issue's single-dispatch-member design doesn't allow.

---

tests/testthat/test-project-lifecycle.R
Felixmil: Fixed. Dropped the two tautological assertions and added a real test of the atomicity guarantee: `addScenario()` + a corrupted container (unsupported `schemaVersion`) + `reloadProject()`, asserting the abort *and* that the in-memory scenario and dirty bit are untouched. Also fixed `projectDirPath` to compare against the independently-known temp dir instead of another field of the same return value.

R/project.R
Felixmil: Fixed — named the one real cache (`.observedDataNamesCache`) instead of miscounting the two programmatic stores as caches too.

R/messages.R
Felixmil: Fixed — points at `.resolveProjectContainerPath()` now, where `folder` is actually bound.

R/messages.R
Felixmil: Fixed, same as the entry above.

R/project.R
Felixmil: Fixed — dropped the historical aside, kept just the load-bearing clause.

R/project-lifecycle.R
Felixmil: Good catch, but the suggested fix (hoist `local_error_call(NULL)` below the guard) doesn't land where you'd expect — verified empirically:

```r
f <- function(x) if (x < 0) cli::cli_abort("bad")
tryCatch(f(-1), error = function(e) print(conditionCall(e)))
#> f(-1)
```

With no override in effect, `cli_abort()`'s default attributes to *the function currently executing* (via its own `sys.call()`), not to that function's caller. So hoisting the NULL below the guard would make the argument-check abort read `Error in `.loadProjectTree(projectFilePath)`:` — the exact opaque internal name the NULL treatment exists to avoid, just relocated.

Went with moving the validation itself back into `initialize()`, where it lived before this PR: it's the one call site that receives genuinely untrusted input (`.reload()`'s path is always a previously-resolved, already-valid absolute path by construction), so `.loadProjectTree()` can now trust its argument and its docblock's claim ("everything this aborts on is a property of the file") is true without exception. Verified this restores the exact original attribution:

```r
tryCatch(Project$new(NULL), error = function(e) print(conditionCall(e)))
#> initialize(...)
```

PR #1228, opened 2026-08-06
Validate the initialConditions section

Felixmil:
`initialConditions` was the one project section with no entry in `.validationAdapters`, so `validateProject()` never looked at it. The authoring path is guarded and cross-reference validation catches a dangling reference, but the per-definition shape check every other section gets was missing: a set arriving by hand-editing the definitions tree, or from the Excel importer, passed a full validation clean however malformed it was. This is action 5 of the #1226 audit, minus the `R/parameters.R` split, which stays its own step.

## Validation

- `.validateInitialConditions()` in `R/validation.R`, modelled on `.validateParameterSets()` beside it. A blank molecule path or a missing unit is a critical error (`ospsuite::setQuantityValuesByPath()` rejects a blank unit at run time, so the set is unusable); a non-numeric value and a duplicated path are warnings, which is how the parameter-set validator grades the same two problems.
- `.initialConditionsValidatorAdapter()` in `R/parameters.R`, registered in `.validationAdapters`.
- The rules are the ones `.validateInitialConditionEntryArgs()` already applies on the authoring path, restated for the entries that never went through it.

## Drift

Two hand-written lists were derived from the kind registry instead, since that is how the gap opened in the first place:

- `.applyLoadedSections()` assigns the twelve tree-owned sections in a loop over `.definitionKindNames()`, reusing the existing `private$.sectionField()` mapping. The container fields stay explicit.
- Two tests assert that every registered kind has a validator (bar `plotGrids`, which `.plotsValidatorAdapter()` covers, named in the test with the reason) and an entry in `.emptyTreeSectionsJson()`.

## Example

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/architecture-audit-implementation-238366", quiet = TRUE)

# A project whose initial-condition set was hand-written without a unit.
dir.create(dir)
invisible(file.copy(
  dir,
  recursive = TRUE
))
writeLines(
  jsonlite::toJSON(
    list(id = "presystemic", initialConditions = list(
      list(path = "Organism|VenousBlood|Plasma|Aciclovir", value = 0.5, unit = "")
    )),
    auto_unbox = TRUE,
    pretty = TRUE
  ),
  file.path(dir, "definitions", "initial-conditions", "presystemic.json")
)

project <- loadProject(dir)
validateProject(project)
#> Validation report: 1 critical error, 0 warnings.
#> initialConditions
#>   ✖ [Missing Fields] Set 'presystemic' in initialConditions contains entries without a unit
#> 11 sections OK.
```

</details>

Part of #1226.

### DIFF
--- a/R/validation.R
+++ b/R/validation.R
@@ -784,23 +784,35 @@ validationSummary <- function(validationResults) {
     return(result)
   }
 
+  # Read one field as a scalar string, or `NA` when the entry does not hold
+  # exactly one value for it. A hand-edited set can write `"unit": []` or
+  # `"unit": ["mg", "g"]`, either of which would otherwise abort `vapply()`
+  # with an internal length error instead of being reported.
+  scalarField <- function(entries, field) {
+    vapply(
+      entries,
+      function(e) {
+        value <- e[[field]]
+        if (length(value) != 1L) NA_character_ else as.character(value)
+      },
+      character(1)
+    )
+  }
+
   for (setName in names(initialConditions)) {
     set <- initialConditions[[setName]]
     if (length(set) == 0) {
       next
     }
 
-    paths <- vapply(
-      set,
-      function(e) as.character(e$path %||% NA_character_),
-      character(1)
-    )
-    units <- vapply(
-      set,
-      function(e) as.character(e$unit %||% NA_character_),
-      character(1)
-    )
-    values <- lapply(set, function(e) e$value)
+    # An entry that is a bare string rather than a record (`["Organism|Liver|
+    # Aciclovir"]`) has no fields to read; treating it as an empty record folds
+    # it into the missing-field errors below instead of aborting on `$`.
+    entries <- lapply(set, function(e) if (is.list(e)) e else list())
+
+    paths <- scalarField(entries, "path")
+    units <- scalarField(entries, "unit")
+    values <- lapply(entries, function(e) e$value)
 
     if (any(is.na(paths) | paths == "")) {
       result$addCriticalError(
@@ -846,7 +858,11 @@ validationSummary <- function(validationResults) {
       )
     }
 
-    dupes <- paths[duplicated(paths)]
+    # Only real paths can duplicate: the blanks and `NA`s are already reported
+    # by the missing-path error above, and `duplicated()` would otherwise count
+    # two path-less entries as a duplicate of each other.
+    realPaths <- paths[!is.na(paths) & paths != ""]
+    dupes <- realPaths[duplicated(realPaths)]
     if (length(dupes) > 0) {
       result$addWarning(
         "Uniqueness",
--- a/R/validation.R
+++ b/R/validation.R
@@ -367,6 +367,7 @@ validationSummary <- function(validationResults) {
   scenarios = .scenariosValidatorAdapter,
   outputPaths = .outputPathsValidatorAdapter,
   parameterSets = .parameterSetsValidatorAdapter,
+  initialConditions = .initialConditionsValidatorAdapter,
   applications = .applicationsValidatorAdapter,
   plots = .plotsValidatorAdapter,
   dataCombined = .dataCombinedValidatorAdapter,
@@ -747,6 +748,110 @@ validationSummary <- function(validationResults) {
         "Uniqueness",
         paste0(
           "Duplicate parameter paths in set '",
+          setName,
+          "': ",
+          paste(unique(dupes), collapse = ", ")
+        )
+      )
+    }
+  }
+
+  result
+}
+
+#' Validate the initial-condition sets
+#'
+#' Each set is the array-of-records shape the parser and
+#' `addInitialConditionEntry()` produce: a list of
+#' `list(path, value, unit)` entries, one molecule start value each.
+#'
+#' The rules are the ones `.validateInitialConditionEntryArgs()`
+#' (R/parameters.R) already applies on the authoring path, restated here for
+#' the entries that never went through it: a set hand-written into the
+#' definitions tree, or one built by the Excel importer. `path` and `unit` are
+#' critical because a blank either way makes the set unusable
+#' (`ospsuite::setQuantityValuesByPath()` rejects a blank unit at run time);
+#' a non-numeric `value` and a duplicated `path` are warnings, matching how
+#' `.validateParameterSets()` grades the same two problems.
+#'
+#' @keywords internal
+#' @noRd
+.validateInitialConditions <- function(initialConditions, sectionName) {
+  result <- validationResult$new()
+
+  if (is.null(initialConditions) || length(initialConditions) == 0) {
+    result$addWarning("Data", paste0("No ", sectionName, " defined"))
+    return(result)
+  }
+
+  for (setName in names(initialConditions)) {
+    set <- initialConditions[[setName]]
+    if (length(set) == 0) {
+      next
+    }
+
+    paths <- vapply(
+      set,
+      function(e) as.character(e$path %||% NA_character_),
+      character(1)
+    )
+    units <- vapply(
+      set,
+      function(e) as.character(e$unit %||% NA_character_),
+      character(1)
+    )
+    values <- lapply(set, function(e) e$value)
+
+    if (any(is.na(paths) | paths == "")) {
+      result$addCriticalError(
+        "Missing Fields",
+        paste0(
+          "Set '",
+          setName,
+          "' in ",
+          sectionName,
+          " contains empty molecule paths"
+        )
+      )
+    }
+
+    if (any(is.na(units) | units == "")) {
+      result$addCriticalError(
+        "Missing Fields",
+        paste0(
+          "Set '",
+          setName,
+          "' in ",
+          sectionName,
+          " contains entries without a unit"
+        )
+      )
+    }
+
+    nonNumeric <- vapply(
+      values,
+      function(v) is.null(v) || length(v) != 1L || !is.numeric(v) || is.na(v),
+      logical(1)
+    )
+    if (any(nonNumeric)) {
+      result$addWarning(
+        "Data Type",
+        paste0(
+          "Set '",
+          setName,
+          "' in ",
+          sectionName,
+          " contains non-numeric values"
+        )
+      )
+    }
+
+    dupes <- paths[duplicated(paths)]
+    if (length(dupes) > 0) {
+      result$addWarning(
+        "Uniqueness",
+        paste0(
+          "Duplicate molecule paths in set '",
           setName,
           "': ",
           paste(unique(dupes), collapse = ", ")
--- a/R/parameters.R
+++ b/R/parameters.R
@@ -1,13 +1,23 @@
-# Section validation adapter ----
+# Section validation adapters ----
 #
 # Registered in `.validationAdapters` (R/validation.R) and called by
-# `.runProjectValidation()`. The actual shape check lives in
-# `.validateParameterSets()` (in `R/validation.R`).
+# `.runProjectValidation()`. The actual shape checks live in
+# `.validateParameterSets()` / `.validateInitialConditions()` (in
+# `R/validation.R`).
 
 #' @keywords internal
 #' @noRd
 .parameterSetsValidatorAdapter <- function(project) {
   .validateParameterSets(project$definitions$parameterSets, "parameterSets")
+}
+
+#' @keywords internal
+#' @noRd
+.initialConditionsValidatorAdapter <- function(project) {
+  .validateInitialConditions(
+    project$definitions$initialConditions,
+    "initialConditions"
+  )
 }
 
 # Merge the parameter-set sections of a parsed `Project.json` into the single
--- a/R/project.R
+++ b/R/project.R
@@ -1380,18 +1380,14 @@ Project <- R6::R6Class(
       private$.defaultSimulationRunOptions <- sections$defaultSimulationRunOptions
       private$.filePathsData <- sections$filePathsData
       private$.excelData <- sections$excelData
-      private$.outputPaths <- sections$outputPaths
-      private$.parameterSets <- sections$parameterSets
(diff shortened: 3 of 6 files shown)


R/validation.R
```diff
@@ -367,6 +367,7 @@ validationSummary <- function(validationResults) {
   scenarios = .scenariosValidatorAdapter,
   outputPaths = .outputPathsValidatorAdapter,
   parameterSets = .parameterSetsValidatorAdapter,
+  initialConditions = .initialConditionsValidatorAdapter,
```
Felixmil, line 370: The adapter is registered, so `validateProject()` reaches it — but `runScenarios()` / `buildSimulations()` still don't. `.scenarioBuildPreflight()` (`R/scenario-execution.R:530-538`) names the sections the scenario build depends on, and `initialConditions` isn't among them, even though `parameterSets` — its structural twin — is:

```r
sections = c(
  "outputPaths", "scenarios", "individuals", "populations",
  "applications", "parameterSets", "crossReferences"
)
```

Initial conditions *are* a scenario-build dependency: `.mergeScenarioInitialConditions()` (`R/scenario-execution.R:386`) folds each referenced set into the built simulation, and the result reaches `setQuantityValuesByPath()` via `additionalInitialConditions` at line 469. So the blank-unit case this validator's docstring cites as the reason for grading `unit` critical is exactly the one that still gets through on the run path and fails opaquely inside `ospsuite` — the validator catches it only if the user separately calls `validateProject()`.

Adding `"initialConditions"` to that vector looks like the whole fix. Worth doing here rather than as a follow-up, since without it the new critical error doesn't fire where the failure actually happens.

---

R/validation.R
```diff
@@ -758,6 +759,110 @@ validationSummary <- function(validationResults) {
...
+    paths <- vapply(
+      set,
+      function(e) as.character(e$path %||% NA_character_),
+      character(1)
+    )
+    units <- vapply(
+      set,
+      function(e) as.character(e$unit %||% NA_character_),
+      character(1)
+    )
```
Felixmil: `%||%` only substitutes for `NULL`, so these two `vapply(..., character(1))` calls assume every `path` / `unit` is a length-1 scalar. A hand-edited definitions file — the input this validator exists to police — can break that:

- `"unit": []` parses to `list()`, and `as.character(list())` is `character(0)` → *"values must be length 1, but FUN(X[[1]]) result is length 0"*
- `"unit": ["mg", "g"]` parses to a length-2 vector → same abort at length 2

Either way `validateProject()` dies with an internal `vapply` error instead of reporting the malformed set, which inverts the point of the section: the more mangled the file, the less useful the report.

This mirrors `.validateParameterSets()` at lines 694-703 exactly, so the PR is faithful to the model it says it follows and this isn't a regression — but it's the same latent crash in both, and it's cheap to close here:

```r
scalarChr <- function(x) {
  if (length(x) != 1L) NA_character_ else as.character(x)
}
paths <- vapply(set, function(e) scalarChr(e$path), character(1))
```

which folds the malformed case back into the existing `is.na(paths)` critical error. (Same reasoning applies one level up: if a set is an atomic vector rather than a list of records, `e$path` aborts with *"$ operator is invalid for atomic vectors"* before any of this runs.)

---

R/validation.R
```diff
@@ -758,6 +759,110 @@ validationSummary <- function(validationResults) {
...
+          "Set '",
+          setName,
+          "' in ",
+          sectionName,
+          " contains non-numeric values"
+        )
+      )
+    }
+
+    dupes <- paths[duplicated(paths)]
```
Felixmil: `duplicated()` runs over the raw `paths` vector, which still holds the `NA` / `""` placeholders the critical-error check above just flagged. `duplicated()` compares `NA` as an ordinary value (`incomparables = FALSE` by default), so a set with two path-less entries yields `dupes = NA_character_` and reports:

> Duplicate molecule paths in set 'x': NA

on top of the correct *"contains empty molecule paths"* error. The user gets a second, meaningless finding pointing at a path that doesn't exist.

Filtering to real paths first keeps the two checks independent:

```r
realPaths <- paths[!is.na(paths) & paths != ""]
dupes <- realPaths[duplicated(realPaths)]
```

Minor, and `.validateParameterSets()` has the same shape (its `paste()` turns the placeholders into a literal `"NA|NA"` duplicate) — but the noise lands on precisely the malformed files this section is meant to explain.

---

Felixmil (commented):
Reviewed the three commits (`ffe7a49..5155ad2`) at medium depth. The change does what it says and the registry-driven parts are sound; three findings below, one of which I'd want addressed before merge.

## Verified as correct

- **`.applyLoadedSections()` registry loop is faithful.** `.definitionTreeSpecs()` holds exactly the twelve kinds the explicit assignments covered, all twelve private `.<kind>` backing fields exist, `.loadProjectTree()` returns all twelve under those exact names, and both callers pass its full output. `[[` over `$` is right here (a missing name yields `NULL` for lists, matching the existing `.definitionTreeSpec()` idiom), and the assignments carry no ordering dependency.
- **Both new drift tests hold.** `.validationAdapters` (11) equals the kind set minus `plotGrids`, and `.emptyTreeSectionsJson()`'s keys equal the twelve kinds. Naming `plotGrids` in the test *with its reason* is the right call — that's the exception that would otherwise look like the very gap this PR closes.
- **The critical/warning grading is consistent.** Blank unit as critical matches both the authoring path (`.validateInitialConditionEntryArgs()`) and the Excel importer, which already reject it; the bundled fixtures all carry units, so nothing regresses.
- **Scenario → initial-condition-set cross-referencing was already covered** and is gated on `"scenarios"`, so it's untouched by this.

## Findings

1. **`R/validation.R:370` / `R/scenario-execution.R:531` — the new validator never runs on the `runScenarios()` path.** `.scenarioBuildPreflight()`'s section list omits `initialConditions` while including `parameterSets`, so the blank-unit case the docstring names as the reason for grading `unit` critical still reaches `setQuantityValuesByPath()` and fails opaquely at run time. One vector element fixes it; details inline.

2. **`R/validation.R:793` — `vapply(..., character(1))` assumes scalars.** `%||%` catches `NULL` but not zero-length or multi-element values, so `"unit": []` or `"unit": ["mg","g"]` aborts `validateProject()` with an internal `vapply` error instead of being reported. Mirrors `.validateParameterSets()` exactly, so not a regression — but it's the malformed-input case this section exists for.

3. **`R/validation.R:849` — duplicate check runs over the placeholders.** Two path-less entries produce a spurious `Duplicate molecule paths in set 'x': NA` warning alongside the correct critical error. Cosmetic, same shape as the sibling validator.

(2) and (3) are both consistency-with-`.validateParameterSets()` cases rather than things this PR got wrong — fine to defer to a follow-up that fixes both validators together. (1) is the one I'd fix here, since without it the feature doesn't reach the runtime path where the failure it describes actually occurs.

---

R/validation.R
```diff
@@ -758,6 +759,110 @@ validationSummary <- function(validationResults) {
...
+    return(result)
+  }
+
+  for (setName in names(initialConditions)) {
+    set <- initialConditions[[setName]]
+    if (length(set) == 0) {
+      next
+    }
+
+    paths <- vapply(
```
Felixmil: A sufficiently malformed set makes this throw a raw R error instead of reporting the set.

`e$path` needs `e` to be a list, and `vapply(..., character(1))` needs exactly one value per entry. Two shapes a hand-edited or imported set can take break that:

- entries that are bare strings — `"initialConditions": ["Organism|Liver|Aciclovir"]` — where `e$path` raises `$ operator is invalid for atomic vectors`;
- a `path` written as an array — `"path": ["Organism|Liver", "Aciclovir"]` — where `as.character()` returns length 2 and `vapply` raises `values must be length 1, but FUN(X[[1]]) result is length 2`.

Either way `validateProject()` errors out rather than reporting the bad set, and that's precisely the hand-authored-or-imported input this validator was added to check.

Worth stating plainly: this is inherited, not introduced. `.validateParameterSets()` just above has the identical `vapply` shape over `containerPath` / `parameterName`, and as the PR says this function is modelled on it — so the new code is consistent with the house pattern, and a fix probably belongs in a follow-up that hardens both rather than in this PR. Raising it here only because a validator is the one place the pattern actually bites: everywhere else bad input reaching `$` is a bug, whereas here it's the expected input.

---

Felixmil (commented):
Checked the drift-prevention part closely, since that's the load-bearing claim: every one of the 12 `.definitionKindNames()` maps to an existing `private$.<kind>` field, and `.loadProjectTree()` returns all 12 keys, so the `.applyLoadedSections()` registry loop covers exactly what the hand-written list did. The two "nothing left unwired" tests line up with reality too (11 adapters plus `plotGrids` via `.plotsValidatorAdapter()`, and 12 empty JSON shapes), and the new adapter's collation order matches its siblings.

One robustness note inline on the validator itself — inherited from `.validateParameterSets()` rather than introduced here, so quite possibly a follow-up rather than a change to this PR.

---

Felixmil:
Review of `0dcef22` (commit-loop). One finding — latent rather than live, but the comment asserts the opposite of what holds.

### `.applyLoadedSections()` and `.loadProjectTree()` are not actually coupled

`R/project.R:1388` now commits sections by iterating the registry:

```r
for (kind in .definitionKindNames()) {
  private[[private$.sectionField(kind)]] <- sections[[kind]]
}
```

but `.loadProjectTree()` (`R/project-lifecycle.R:186-212`) still enumerates its return keys by hand (`outputPaths = ...`, `scenarios = ...`, … 12 of them). The two lists agree today — I checked, the 12 registry kinds match the 12 removed explicit assignments exactly — so nothing is broken right now. The problem is the comment directly above the loop:

> a kind added to `.definitionTreeSpecs()` is committed here without a second edit

That second edit is exactly what a new kind does need. Add one to `.definitionTreeSpecs()` without also adding a `loadSection("newKind")` line to `.loadProjectTree()`, and `sections[["newKind"]]` is `NULL`, so the loop assigns `NULL` into the private field on **every load and reload**. It fails silently: `.sectionField()` only aborts for a kind *outside* the registry, and the loop iterates the registry itself, so the resolution always succeeds and the blanking goes unreported.

The drift tests in `5155ad2` don't close this. They assert the registry against `.validationAdapters` and against `.emptyTreeSectionsJson()` — the loader is the third list, and it's the one left unasserted.

Two ways out, either fine:

- have `.loadProjectTree()` build its tree-owned keys by iterating `.definitionKindNames()` too, which would make the comment's claim true; or
- keep the hand-written list and add the matching drift test (`expect_true(all(.definitionKindNames() %in% names(<loaded tree>)))`), then reword the comment to say the coupling is test-enforced rather than automatic.

---

R/validation.R
Felixmil: Fixed.

Confirmed the dependency before touching anything: `.mergeScenarioInitialConditions()` reads `project$definitions$initialConditions[[setId]]`, and its result reaches `initializeSimulation(additionalInitialConditions = )`, so the section really is a scenario-build input and the gate was the only thing not looking at it. `initialConditions` now sits next to `parameterSets` in `.scenarioBuildPreflight()`'s section vector.

No new noise on the run path: `.ensureValid()` aborts on critical errors and discards warnings, so the "No initialConditions defined" warning a project without any set emits stays invisible to `runScenarios()`.

The new test drives `.scenarioBuildPreflight()` itself rather than a copy of the vector, so the list cannot drift out from under it again. The two existing run-gate tests that carry a copy of that list for their own purposes were updated to match, since their comments claim to mirror the real thing.

R/validation.R
Felixmil: Fixed, together with the atomic-entry case from the other thread: one guard covers both.

Reproduced both aborts first. `"unit": []` gives *values must be length 1, but FUN(X[[1]]) result is length 0*, and `["mg", "g"]` the length-2 variant.

`path` and `unit` now go through a small local reader that yields `NA_character_` unless the entry holds exactly one value for the field, which folds both shapes back into the existing missing-field critical error instead of aborting the run. That also lines the two paths up: `.validateInitialConditionEntryArgs()` grades through `.isNonEmptyString()`, which is already false for a zero- or multi-length value, so an authored entry and a hand-edited one are now graded the same way.

On the sibling: `.validateParameterSets()` still has the identical shape over `containerPath` / `parameterName`. You are right that it is the same latent crash, but it is not code this PR touches and fixing it honestly needs its own tests, so I left it for the follow-up you proposed rather than widening this one.

R/validation.R
Felixmil: Fixed.

Confirmed the spurious finding: two path-less entries give `paths = c(NA, NA)`, `duplicated()` compares `NA` as an ordinary value, and the report gains *Duplicate molecule paths in set 'x': NA* on top of the correct missing-path error.

The duplicate check now runs over the paths that are actually there (`!is.na(paths) & paths != ""`), so the two checks stay independent: a set with two blank paths yields exactly one finding, and a genuine duplicate is still reported. A test covers the two-blank-paths case by asserting the set produces no warning at all.

Same caveat as the other threads: `.validateParameterSets()` keeps its version of this (its `paste()` turns the placeholders into a literal `"NA|NA"` duplicate) and is left for the follow-up.

R/validation.R
Felixmil: Fixed by the same guard as the `vapply` thread, and both shapes you name are covered: an entry that is a bare string is now read as a record with no fields, so it collapses into the missing-path and missing-unit critical errors instead of aborting on `$`, and a `path` written as an array reads as `NA` rather than aborting `vapply`.

One thing worth recording, because it changes where the guard has to go: the crash reaches past the two `vapply` calls. `values <- lapply(set, function(e) e$value)` on the next line hits `$` on the same atomic entry, so guarding only the `path` / `unit` reads would have moved the abort down a line instead of removing it. Normalizing each entry once, before any field is read, covers all three reads.

Agreed on the inherited-not-introduced framing. `.validateParameterSets()` is untouched here and left for the follow-up that hardens it with its own tests.

PR #1229, opened 2026-08-06
Keep a declared steady-state time through the Excel bridge

Felixmil:
A scenario can carry a steady-state time it is not currently using: `steadyState: false` alongside a `steadyStateTime` and `steadyStateTimeUnit`. The JSON writer keeps that pair, emitting it whenever a unit is declared. The Excel writer emitted it only when the flag was on, so exporting blanked both columns, re-import read the blank cells as the defaults, and 1 h came back as 1000 min. Silent, and reachable by anyone who exports a project and imports it again.

## Excel bridge

- `.scenarioConfigurationsToExcelDf()` emits the time and unit when the flag is on **or** a unit is declared. A scenario that declares no unit still writes blank, so a blank cell keeps meaning "defaulted" and re-import does not materialize a steady-state configuration nobody authored.
- The `%||% "min"` fallback stays, covering a flag-on record with a null unit.
- Review follow-up: the `> 0` guard this PR originally retained has since been dropped. It predates the unit clause, so it was never the authored-vs-defaulted test, and keeping it left `steadyStateTime = 0` with a declared unit going out blank and returning as `1000` / `"min"` — exactly the drift this PR closes for every other value. It only ever fired for `steadyStateTime <= 0`, and every such record already lost its value, so no previously-correct export changes. Both steady-state fixtures export byte-identically.
- Round-trip test for a flag-off scenario declaring 1 h. It fails without the change (`1` / `"h"` comes back as `1000` / `"min"`).

## NEWS

Folded into the existing #1213 round-trip bullet as its fourth gap, rather than added as a sibling: for a user upgrading from 5.x this is one more thing the bridge now preserves, not a separate change.


R/project-excel.R
```diff
@@ -4010,15 +4010,23 @@ projectStatus <- function(project, silent = FALSE) {
...
+    # would invent a steady-state setup nobody asked for, and re-import would
+    # read it back as real.
     ssTime <- NA
     ssTimeUnit <- NA
     if (
-      isTRUE(sc$simulateSteadyState) &&
+      (isTRUE(sc$simulateSteadyState) || !is.null(sc$steadyStateTimeUnit)) &&
         !is.null(sc$steadyStateTime) &&
         !is.na(sc$steadyStateTime) &&
         sc$steadyStateTime > 0
```
Felixmil: The new disjunct makes the declared unit the test for "authored", but this `> 0` guard isn't, and it drops an authored `0` down exactly the path this commit closes.

`steadyStateTime: 0` with a unit is a value the model accepts end to end:

- the parser takes it — `R/scenarios.R:202` branches on `!is.null(entry[["steadyStateTime"]])`, not on the value — and stores `0` with its unit;
- nothing constrains the sign: `.validateScenarios()` has no positivity check, and there is no `> 0` on `steadyStateTime` anywhere in `R/` except this line;
- `.scenarioToJson()` (`R/project-to-json.R:354`) emits the pair whenever a unit is declared, with no value guard of its own.

So `steadyState: false, steadyStateTime: 0, steadyStateTimeUnit: "h"` survives a JSON save/load, then loses the same way the non-zero case did before this commit: the guard fails, both columns go out blank, and `.naToDefault()` on re-import (`R/project-excel.R:2280-2287`) reads the blanks back as `1000` / `"min"`. Same silent substitution, same false `projectStatus()` drift — just narrowed to one value.

The PR body keeps the guard on the grounds that the steady-state path exports exactly the cells it did before, and it does. But the guard predates the unit check and was standing in for it: distinguishing an authored time from the parser's default `1000`. Now that `!is.null(sc$steadyStateTimeUnit)` answers that question directly, `> 0` is left deciding whether an authored value is worth keeping, and it answers wrong for `0`.

Dropping it looks safe:

```r
      (isTRUE(sc$simulateSteadyState) || !is.null(sc$steadyStateTimeUnit)) &&
        !is.null(sc$steadyStateTime) &&
        !is.na(sc$steadyStateTime)
```

No bundled fixture declares a unit with the flag off (`aciclovir_iv`, `aciclovir_iv_population`, and the three `TestProject` non-steady-state scenarios all carry `steadyStateTime: null` / `steadyStateTimeUnit: null`), so the existing "defaults a non-steady-state scenario's steady-state time and unit" test still sees blank cells and the two steady-state fixtures export unchanged. The `isTRUE(simulateSteadyState)` disjunct plus the `%||% "min"` fallback still covers a drifted record whose flag is on with a null unit; that one would export `0` as `0 min` rather than blank, which is what it actually holds.

Worth a round-trip case at `steadyStateTime = 0` beside the `1 h` one, if you take it.

---

Felixmil (commented):
Reviewed the single commit on `felixmil/excel-steady-state-emit` (932d498) against its base `felixmil/initial-conditions-validator-1226`. R is not installed in this environment, so this is a static review — I did not run the suite.

The change itself holds up. `.scenarioConfigurationsToExcelDf()` now gates on the same condition `.scenarioToJson()` uses (`R/project-to-json.R:354`), so the two writers agree on when a steady-state time is authored rather than defaulted, which is the asymmetry that caused the loss. The "still blank when no unit is declared" half is load-bearing and correct: `.readScenarioRows()` turns a blank cell into `1000` / `"min"` (`R/project-excel.R:2280-2287`), so emitting the parser's default there would materialize a steady-state configuration nobody authored. No bundled fixture declares a unit with the flag off, so the adjacent defaults test and both steady-state fixtures export exactly as before, and the new round-trip test does fail on the old condition.

One finding, inline: the retained `sc$steadyStateTime > 0` guard leaves the same round-trip loss open at exactly `0`. Narrow, but it is the same class of bug and the same silent `1000` / `"min"` substitution.

---

R/project-excel.R
Felixmil: The retained `> 0` guard leaves one value behind that `.scenarioToJson()` keeps, so the two writers still disagree on it.

The new `|| !is.null(sc$steadyStateTimeUnit)` makes the Excel gate match the JSON one for a flag-off scenario, but `.scenarioToJson()` gates on the declared unit alone (`R/project-to-json.R:354`) with no `> 0` test. A scenario declaring `steadyStateTime = 0` with a unit therefore keeps the `0` through JSON, exports blank here, and comes back from `.naToDefault()` as `1000` / `"min"` (`R/project-excel.R:2280-2286`) — the same silent round-trip drift this hunk fixes for every other value.

I read the PR body as keeping `> 0` deliberately to leave the steady-state-on path untouched, which it does. The side effect is that the guard also gates the newly added flag-off branch, which is the path being fixed. Whether `0` is a value worth authoring is a fair question; the writers disagreeing about it is what produces the drift.

---

Felixmil (commented):
The export condition now mirrors `.scenarioToJson()`'s for every value but one. Checked that the pre-existing `aciclovir_iv` round-trip test still holds — that fixture declares a null unit, so it keeps writing blank cells and the "blank means defaulted" contract is intact.

One edge case left, noted inline.

---

Felixmil:
Review of `932d498`. One finding: the widened condition still drops the case it was widened for, when the declared time is `0`.

### `steadyStateTime = 0` plus a declared unit still round-trips to `1000`/`"min"`

`R/project-excel.R:4029-4032`:

```r
(isTRUE(sc$simulateSteadyState) || !is.null(sc$steadyStateTimeUnit)) &&
  !is.null(sc$steadyStateTime) &&
  !is.na(sc$steadyStateTime) &&
  sc$steadyStateTime > 0
```

The `||` correctly widened the flag test to "flag on **or** unit declared", but the trailing `sc$steadyStateTime > 0` still gates on the value, so a scenario with `steadyStateTime = 0` and a unit exports both cells blank.

Meanwhile `.scenarioToJson()` (`R/project-to-json.R:354`) emits the time whenever a unit is declared, independent of the value, so JSON keeps the `0`. The two sides disagree, and the Excel side loses. Round trip:

1. scenario holds `steadyStateTime = 0`, `steadyStateTimeUnit = "min"`
2. `exportProjectToExcel()` → `SteadyStateTime` and `SteadyStateTimeUnit` blank
3. re-import → `.naToDefault(NA, 1000)` and `.naToDefault(NA, "min")` (`R/project-excel.R:2280-2287`) → `1000` / `"min"`

A declared `0` comes back as `1000` — the same silent-substitution drift this hunk targets, just at the boundary value.

Dropping `sc$steadyStateTime > 0` looks right: the two `!is.null` / `!is.na` guards above it already cover "nothing declared", which is the case the blank-cell branch is for, and the comment right above says the blanking is for *no unit* rather than for a zero value. Worth a test pinning `0` + unit through export→import, since that's the boundary the current tests miss.

---

R/project-excel.R
Felixmil: Verified and taken — the guard is gone.

Reproduced first, to be sure the value really reaches the bridge rather than dying earlier: a scenario set to `steadyState = FALSE`, `steadyStateTime = 0`, `steadyStateTimeUnit = "h"` comes back from an Excel round trip as `1000` / `"min"`, exactly as you described. It survives the JSON side untouched, so the two writers were disagreeing about that one value.

The gate is now the three clauses you wrote, and the `%||% "min"` fallback stays for the flag-on-with-null-unit record.

On whether the `> 0` was protecting anything: it only ever fired for `steadyStateTime <= 0`, and every such record already lost its value on a round trip, so removing it cannot change a case that was previously correct. The two steady-state fixtures (`aciclovir_iv_steadystate` at `1 h`, `testscenario_steadystate` at `500 min`) export byte-identically, and no fixture declares a unit with the flag off. The full suite is green, with only the two new expectations added.

Your blame reading is right too: the `> 0` came in with the original bridge and the `!is.null(steadyStateTime)` clause was added later, so it had been standing in for a check that `!is.null(steadyStateTimeUnit)` now makes directly.

Added the `steadyStateTime = 0` round-trip case beside the `1 h` one; it fails without the change with `1000` / `"min"`.

Felixmil: Agreed, and fixed — same substance as the other thread on this line, so the two are answered by one change.

You framed it as the two writers disagreeing, which is the useful way to see it: `.scenarioToJson()` gates on the declared unit alone and writes whatever value the record holds, so any test the Excel writer applies on top of that is a place they can diverge. The `> 0` was the only such test, and `0` was the only value it caught. Confirmed by reproduction: `steadyState = FALSE`, `steadyStateTime = 0`, `steadyStateTimeUnit = "h"` came back from an Excel round trip as `1000` / `"min"` while the JSON round trip kept it.

The guard is dropped, so the Excel gate now differs from the JSON one only in the `isTRUE(simulateSteadyState)` disjunct and the `%||% "min"` fallback, both of which cover a flag-on record with a null unit rather than deciding whether a value is worth keeping. Nothing that previously round-tripped correctly changes: the guard only excluded non-positive times, all of which already drifted. Round-trip test added at `0`, full suite green.

PR #1230, opened 2026-08-06
Give setScenario() the same ... signature as its siblings

Felixmil:
`setScenario()` spelled out all 16 settable fields as explicit arguments and reconstructed the supplied set through 14 `missing()` blocks. That whole interface existed for one reason: `overwrite` is a unique prefix of `overwriteFormulasInSS`, so R's partial matching turned the habit picked up from `addScenario()` into a silent change to a steady-state option. With `...` there is nothing after `project` and `id` to partial-match against, so the name lands in the field set and is rejected by name. `setIndividual()` and `setPopulation()` already had this shape.

## Authoring

- `setScenario(project, id, ...)`, forwarding to the method. Around 70 lines and 14 `@param` blocks go; the settable-field list now lives in two places instead of three (`.settableScenarioFields` stays as the impl's whitelist).
- `.assertScenarioRecordAlone()` takes the extra field names rather than a `match.call()`; `addScenario()` reads its own off `match.call()` through the new `.addScenarioExtraFields()`, since its formals still let `modelFile` through positionally.
- `@param ...` replaces the per-field blocks, modelled on `setIndividual()`'s.

## Unnamed fields

Making the fields travel through `...` exposed a silent drop that was already there for `setIndividual()` and `setPopulation()`: an unnamed field has no name to be looked up under, so the whole alignment skipped it and the definition came back unchanged with nothing said. `setIndividual(project, "indiv1", "Rat")` was a no-op.

- `.assertAuthoringFieldsNamed()` in `R/authoring-vectorize.R`, called on `dots` in all three `set*_impl` functions.
- `messages$unnamedAuthoringFields()` for the text.

## Breaking

A positional field on `setScenario()` (`setScenario(project, "sc", "model.pkml")`) is now an error instead of setting `modelFile`. This matches the two sibling functions. No NEWS entry: `setScenario()` itself is new in 6.0.0, so there is no released behaviour to describe a delta from.

### DIFF
--- a/R/messages.R
+++ b/R/messages.R
@@ -1133,10 +1133,25 @@ messages$setScenarioNoOverwrite <- function() {
   )
 }
 
+# Raised when an authoring field reaches `.alignAuthoringArgs()` without a name.
+# Every field there is looked up by name, so an unnamed one would be dropped in
+# silence; the usual way to produce one is passing a `set*()` field positionally.
+messages$unnamedAuthoringFields <- function(positions) {
+  # `cli::qty()` pluralizes on how many positions there are, not on the position
+  # numbers themselves: a lone field at position 2 is still one field. It has to
+  # sit before every plural marker, since interpolating `positions` resets the
+  # quantity cli pluralizes on, which is why the sentence keeps them together.
+  cli::format_message(c(
+    "Every field must be named.",
+    "x" = "{cli::qty(length(positions))}No name on field{?s} {positions}.",
+    "i" = "Name the field, for example \\
+    {.code setIndividual(project, id, species = \"Rat\")}."
+  ))
+}
+
 # Raised when a name reaching `setScenario()`'s field set is not a scenario
-# field. Only the R6 method can be called with one (the exported function's
-# formals reject it), and without the guard the value would be aligned as a
-# per-definition field and then dropped in silence.
+# field. Without the guard the value would be aligned as a per-definition field
+# and then dropped in silence.
 messages$setScenarioUnknownFields <- function(fields, settable) {
   cli::format_message(c(
     "{.fn setScenario} cannot set {.field {fields}}.",
--- a/R/messages.R
+++ b/R/messages.R
@@ -1137,6 +1137,9 @@ messages$setScenarioNoOverwrite <- function() {
 # Raised when an authoring field reaches `.alignAuthoringArgs()` without a name.
 # Every field there is looked up by name, so an unnamed one would be dropped in
 # silence; the usual way to produce one is passing a `set*()` field positionally.
+# The hint names no function and no field: `setScenario()`, `setIndividual()` and
+# `setPopulation()` all raise this, they share no field, and the abort header
+# already names the one the caller used.
 messages$unnamedAuthoringFields <- function(positions) {
   # `cli::qty()` pluralizes on how many positions there are, not on the position
   # numbers themselves: a lone field at position 2 is still one field. It has to
@@ -1145,8 +1148,7 @@ messages$unnamedAuthoringFields <- function(positions) {
   cli::format_message(c(
     "Every field must be named.",
     "x" = "{cli::qty(length(positions))}No name on field{?s} {positions}.",
-    "i" = "Name the field, for example \\
-    {.code setIndividual(project, id, species = \"Rat\")}."
+    "i" = "Pass each field as {.code name = value}."
   ))
 }
 
--- a/R/messages.R
+++ b/R/messages.R
@@ -1120,9 +1120,10 @@ messages$scenarioRecordWithFields <- function(fields) {
   ))
 }
 
-# `setScenario()` declares `overwrite` only to intercept it: without the formal,
-# R's partial matching would resolve the name to `overwriteFormulasInSS` and
-# quietly set a steady-state model option instead.
+# `setScenario()` intercepts `overwrite` rather than letting it reach the impl's
+# generic unknown-field message, because the name is a habit picked up from
+# `addScenario()` and the scenario carries a second argument that starts the
+# same way.
 messages$setScenarioNoOverwrite <- function() {
   c(
     "{.fn setScenario} has no {.arg overwrite} argument.",
--- a/R/scenarios.R
+++ b/R/scenarios.R
@@ -1362,6 +1362,7 @@ setScenario <- function(
   }
 
   dots <- list(...)
+  .assertAuthoringFieldsNamed(dots)
   # A name that is not a scenario field would otherwise be aligned as a
   # per-definition field and then quietly dropped when nothing applies it.
   unknownFields <- setdiff(names(dots), .settableScenarioFields)
--- a/R/scenarios.R
+++ b/R/scenarios.R
@@ -912,7 +912,9 @@ addScenario <- function(
 ) {
   validateIsOfType(project, "Project")
   if (inherits(id, "Scenario")) {
-    .assertScenarioRecordAlone(match.call(), missing(modelFile))
+    .assertScenarioRecordAlone(
+      .addScenarioExtraFields(match.call(), missing(modelFile))
+    )
     return(do.call(
       project$addScenario,
       c(.scenarioRecordArgs(id), list(overwrite = overwrite))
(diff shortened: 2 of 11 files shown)


R/scenarios.R
```diff
@@ -1200,124 +1202,47 @@ removeScenario <- function(project, id) {
...
+  # `overwrite` reaches the impl as an unknown field, which would name it in the
+  # generic "cannot set" message. Say what it is instead: `addScenario()` is the
+  # one with an `overwrite`, and this scenario also has an argument whose name
+  # starts the same way.
+  if ("overwrite" %in% names(dots)) {
     cli::cli_abort(messages$setScenarioNoOverwrite())
   }
   if (inherits(id, "Scenario")) {
-    .assertScenarioRecordAlone(match.call(), missing(modelFile))
+    .assertScenarioRecordAlone(names(dots))
```
Felixmil, line 1245: On the `Scenario`-record path an unnamed extra field is now dropped in silence, where the old signature caught it.

`names(dots)` is `NULL` when every extra argument is unnamed, so `.assertScenarioRecordAlone(NULL)` sees `length(extra) == 0L` and returns clean:

```r
setScenario(project, sc, "Aciclovir.pkml")  # no error; "Aciclovir.pkml" discarded
```

The old code passed `match.call()` alongside `missing(modelFile)`, and R matched that string to the `modelFile` formal, so `extra` gained `"modelFile"` and it aborted.

The PR's new `.assertAuthoringFieldsNamed()` doesn't cover it either: this branch returns through `do.call(project$setScenario, .scenarioRecordArgs(id))`, which passes only the record's own named fields, so `dots` never reaches `.setScenario_impl()` and its guard.

A partially named call does abort, but names an empty field — `setScenario(project, sc, individual = NULL, "Aciclovir.pkml")` hands `c("individual", "")` to `messages$scenarioRecordWithFields()`, and `{.arg {fields}}` renders the blank as empty backticks.

Calling `.assertAuthoringFieldsNamed(dots)` in `setScenario()` before the record branch closes both halves, and keeps the record path consistent with the id path. Alternatively `.assertScenarioRecordAlone()` could key off how many extra arguments there are rather than how many names, using the names only to build the message — though that needs a second thought about `addScenario()`, which feeds it a plain character vector from `.addScenarioExtraFields()`.

---

Felixmil (commented):
Reviewed the `...` reshape and the unnamed-field guard. The mechanical part holds up: `overwrite` interception survives losing the formal (`"overwrite" %in% names(dots)` fires before the record branch and before the impl's unknown-field check), all in-repo `setScenario()` callers already pass fields by name, and `.addScenarioExtraFields()` preserves `addScenario()`'s positional-`modelFile` detection now that the two entrypoints compute `extra` differently.

One finding, on the `Scenario`-record branch — it's the one path where the reshape re-opens the silent drop the PR set out to close.

---

Felixmil:
Review of `b52fb21` + `e9b63b4`. The `...` reshape is a clear win, but dropping the formals removed two checks R was providing for free, and the new named-field guard has a gap on each side.

### 1. A positional field is silently ignored on the record path — `R/scenarios.R:1240`

```r
if (inherits(id, "Scenario")) {
  .assertScenarioRecordAlone(names(dots))
  return(do.call(project$setScenario, .scenarioRecordArgs(id)))
}
```

`.assertScenarioRecordAlone()` aborts on `length(extra) > 0`, but it's handed `names(dots)`, and for a fully unnamed list that is `NULL` — length 0. So `setScenario(project, <Scenario record>, "NewModel.pkml")` passes the guard, and the branch returns before `.setScenario_impl()`, which is the only place `.assertAuthoringFieldsNamed()` runs. The positional field is dropped without a word.

`addScenario()` doesn't have this hole: it passes `.addScenarioExtraFields(match.call(), missing(modelFile))` instead of `names(dots)`, which sees positional arguments. The two entrypoints now disagree about the same mistake. Passing `dots` (or `length(dots)`) rather than `names(dots)` would line them back up.

### 2. `.setPITask_impl()` didn't get the new guard — `R/parameter-identification.R:1536`

`b52fb21` added `.assertAuthoringFieldsNamed()` to the `set*` impls, but the callers are only `individuals.R:464`, `scenarios.R:1290`, `populations.R:864`. `.setPITask_impl()` is the fourth `...`-taking `set*` impl and was missed, so it still has the exact behaviour the commit set out to remove:

```r
project$setPITask("t1", "scenarioA")   # silent no-op
```

`names(dots)` is `NULL`, so `setdiff(NULL, c("scenarios", "configuration"))` is `character(0)` and the unknown-field abort doesn't fire either; `length(dots)` is 1 so the early return doesn't catch it; neither `"scenarios"` nor `"configuration"` is in `names(dots)`, so nothing is written and it returns `invisible(self)` as if it had worked.

### 3. A repeated field name is now silently accepted — `R/scenarios.R:1312`

With 17 explicit formals, `setScenario(p, "sc", individual = "a", individual = "b")` was an R error (*formal argument "individual" matched by multiple actual arguments*). Under `...`, R permits the duplicate, and:

```r
scalarSupplied <- dots[setdiff(names(dots), wholeNames)]
```

`setdiff()` returns unique values, so this selects `dots["individual"]` — the first match — and `"b"` vanishes. Nothing upstream catches it: `.assertAuthoringFieldsNamed()` sees two named entries, and the `unknownFields` check is also a `setdiff`. An `.assertNoDuplicateIds()`-style check on `names(dots)` would restore what the formals used to give you.

### 4. NEWS doesn't mention the signature change — `NEWS.md:134`

Going from 17 formals to `(project, id, ...)` is breaking for anyone calling positionally or relying on argument abbreviation (`setScenario(p, "sc", indiv = "x")` used to partial-match and now errors as an unknown field). There's no bullet for it in **Breaking changes**.

The existing bullet at line 134 also reads oddly now. The behaviour it describes still holds — `overwrite` is rejected, via the explicit guard at `scenarios.R:1237` — but its explanation is R's partial matching of `overwrite` against `overwriteFormulasInSS`, and with no formals left there is nothing to partial-match. Worth rewriting it around the explicit guard and adding a breaking-change bullet for the signature.

---

Items 1-3 are all the same shape: the reshape moved the mistake-catching from R's argument matcher into hand-written guards, and each of the three ways R used to reject a bad call needs its guard to be complete.

---

R/messages.R
```diff
@@ -1133,10 +1134,25 @@ messages$setScenarioNoOverwrite <- function() {
...
+messages$unnamedAuthoringFields <- function(positions) {
+  # `cli::qty()` pluralizes on how many positions there are, not on the position
+  # numbers themselves: a lone field at position 2 is still one field. It has to
+  # sit before every plural marker, since interpolating `positions` resets the
+  # quantity cli pluralizes on, which is why the sentence keeps them together.
+  cli::format_message(c(
+    "Every field must be named.",
+    "x" = "{cli::qty(length(positions))}No name on field{?s} {positions}.",
+    "i" = "Name the field, for example \\
+    {.code setIndividual(project, id, species = \"Rat\")}."
```
Felixmil: The hint hard-codes `setIndividual`, but `.assertAuthoringFieldsNamed()` is called from all three `set*` paths — `R/individuals.R:464`, `R/scenarios.R:1290` and `R/populations.R:864`. A user who trips it from `setScenario()` or `setPopulation()` is pointed at a different function *and* at `species`, which neither of them can set:

```
Every field must be named.
x No name on field 3.
i Name the field, for example `setIndividual(project, id, species = "Rat")`.
```

Every other message in this catalog names the function the caller actually used, and the two neighbours here (`setScenarioNoOverwrite()`, `setScenarioUnknownFields()`) are per-function for exactly that reason.

Simplest fix is to let the raiser supply its own example, since each call site already knows which function it is. Failing that, a field every `set*()` accepts would at least not be actively wrong — though the function name would still be.

---

Felixmil (commented):
Follow-up to my earlier review on this PR — one more finding, in `b52fb21`'s half of the branch rather than the `...` reshape I looked at before. It surfaced while reviewing #1231 on top of this stack, so I'm leaving it here where the code is.

Cosmetic in the sense that nothing computes wrong, but it is user-facing error text that misdirects two of its three callers. Details inline.

---

Felixmil:
Two findings on this branch, both a field going in unnamed and being dropped without a word.

**1. `setScenario()` silently drops a positional extra after a record** — `R/scenarios.R:1241`

The move from named formals to `...` means an unnamed extra now lands in `dots` with no name, and the record-branch guard only inspects names:

```r
.assertScenarioRecordAlone(names(dots))
```

`names(list("Model.pkml"))` is `NULL`, so `length(extra) > 0L` is `FALSE` and nothing aborts:

```r
setScenario(project, myScenarioRecord, "Model.pkml")
#> accepted; "Model.pkml" is dropped
```

The `missing(modelFile)` check this commit replaces caught exactly this case. `addScenario()` is unaffected because it still routes through `.addScenarioExtraFields(match.call(), ...)`, which matches positionals against formals.

Since a record must be alone, any `dots` at all is an error here. Calling the sibling guard first covers both spellings and gives each its own message:

```r
.assertAuthoringFieldsNamed(dots)
.assertScenarioRecordAlone(names(dots))
```

**2. `.setPITask_impl()` was left out of the new guard** — `R/parameter-identification.R:1507`

`b52fb21` added `.assertAuthoringFieldsNamed(dots)` to the `set*` impls in `individuals.R`, `scenarios.R` and `populations.R`, but not to `.setPITask_impl()`, which is on the same authoring path. Its own guard is name-based only (`R/parameter-identification.R:1523`):

```r
unknown <- setdiff(names(dots), c("scenarios", "configuration"))
```

With no names, `setdiff(NULL, ...)` is `character(0)`, so nothing aborts; `length(dots) == 0L` is `FALSE`, and neither `"scenarios"` nor `"configuration"` is found — the call returns invisibly having changed nothing:

```r
setPITask(project, "task1", list(scenarios = "s1"))
#> silent no-op
```

Verified by reading; no R in this environment, so the suite was not run.

---

Felixmil (commented):
Found while code-reviewing the stack from #1238. This one belongs here rather than there, since it is the `...` rewrite in this PR that introduces it.

### `setScenario(project, <Scenario record>, <positional>)` now silently ignores the extra

Before this PR the record path was guarded with `.assertScenarioRecordAlone(match.call(), missing(modelFile))`, and `missing(modelFile)` is what caught an extra passed positionally. The rewrite guards with `.assertScenarioRecordAlone(names(dots))` instead, and for an all-positional `...` `names(dots)` is `NULL`:

```r
setScenario(project, scenarioRecord, "Other.pkml")
# dots        -> list("Other.pkml")
# names(dots) -> NULL
# length(NULL) > 0L is FALSE  =>  no error
```

Control then reaches `do.call(project$setScenario, .scenarioRecordArgs(id))`, which is built from the record alone, so `"Other.pkml"` is dropped without a word. `addScenario()` is unaffected — it still derives the field names from `match.call()` plus `missing(modelFile)`.

A secondary effect on the mixed case: `setScenario(project, rec, "Other.pkml", individual = "ind1")` gives `names(dots) == c("", "individual")`, so the abort fires but names `""` as one of the offending fields.

Both go away if the guard is given the positions rather than only the names — e.g. have `.assertScenarioRecordAlone()` take `dots`, and describe an unnamed entry by its position the way `.assertAuthoringFieldsNamed()` already does, instead of by an empty string.

---

R/scenarios.R
Felixmil: Confirmed, and fixed. Reproduced on the branch first: `setScenario(project, sc, "Other.pkml")` returned invisibly with the string discarded, and the partially named `setScenario(project, sc, individual = NULL, "Other.pkml")` did abort but rendered the blank as ``` `individual` and `` ```, exactly as you described.

Took the first of your two suggestions. `setScenario()`'s record branch now runs `.assertAuthoringFieldsNamed(dots)` before `.assertScenarioRecordAlone()`. That reuses the helper already added earlier on this branch rather than teaching `.assertScenarioRecordAlone()` a second way to describe a field, and it keeps `addScenario()` out of it — `match.call()` names positional arguments, so that entrypoint never had the gap and needs no change.

Both cases now abort as ``Error in `setScenario()`:`` naming the position, so the empty-field rendering goes away with it. Two snapshot tests in `test-scenarios.R` cover them; with the guard removed the first fails outright ("Expected `setScenario(project, sc, \"Other.pkml\")` to throw a error").

R/messages.R
Felixmil: Confirmed. Checked the three raise sites, and a `setPopulation()` caller really is handed ``i Name the field, for example `setIndividual(project, id, species = "Rat")`.``

I went with your fallback rather than the per-call-site example. The caller name is reachable without a new parameter: `rlang::error_call()` on the `call` the helper already takes resolves through the `_impl`'s `local_error_call()` and yields `setScenario(...)`. But it returns `NULL` when the helper is called directly, as the unit tests do, so it needs a guard and a conditional bullet, and there is no other `error_call()` use in the package to lean on. Set against that, the abort header already prints ``Error in `setScenario()`:``, so the example was only ever demonstrating `name = value` syntax — which is the part that is the same for all three.

The hint now reads "Pass each field as `name = value`." Correct wherever it is raised, and shorter. The two `.assertAuthoringFieldsNamed()` snapshots in `_snaps/authoring-vectorize.md` were re-recorded for it (deliberately, not by a blanket accept), and a new test asserts the message a `setPopulation()` caller receives names no `set*()` function at all.

Happy to switch to the explicit example parameter if you would rather each site spelled out its own; it is a one-argument change at three call sites.

PR #1231, opened 2026-08-06
Give the steady-state step of a scenario build its own function

Felixmil:
The steady-state step of `.prepareScenario()` is the scientifically load-bearing one and was the only step with no direct test: every pure merge step around it is unit-tested, but the branch calling `ospsuite::getSteadyState()` was reachable only by driving the whole 150-line orchestration.

## Scenario execution

- `.applyScenarioSteadyState(simulation, scenario, simulationRunOptions)` extracted alongside the other merge helpers; step 7 of `.prepareScenario()` becomes one call. Behaviour is unchanged, including the truthiness test on `simulateSteadyState` (a corrupt `NA` still errors rather than silently skipping steady state).
- Tests cover the no-op, the solve-and-write-back keyed by simulation id, and the `overwriteFormulasInSS` to `ignoreIfFormula` inversion, with the `ospsuite` bindings mocked. No PK-Sim model is loaded.

### DIFF
--- a/tests/testthat/test-scenario-execution.R
+++ b/tests/testthat/test-scenario-execution.R
@@ -952,7 +952,13 @@ test_that(".applyScenarioSteadyState solves at the scenario's time and writes th
         ignoreIfFormula = ignoreIfFormula,
         simulationRunOptions = simulationRunOptions
       )
-      list(sim1 = list(paths = c("Organism|A"), values = 3.5))
+      # `getSteadyState()` answers one entry per simulation it was handed. The
+      # entry for the simulation under test comes second, so reading the result
+      # positionally instead of by id picks up the other simulation's values.
+      list(
+        other = list(paths = "Organism|B", values = 99),
+        sim1 = list(paths = c("Organism|A"), values = 3.5)
+      )
     },
     setQuantityValuesByPath = function(quantityPaths, values, simulation) {
       written <<- list(
--- a/tests/testthat/test-scenario-execution.R
+++ b/tests/testthat/test-scenario-execution.R
@@ -909,6 +909,106 @@ test_that(".buildScenarioSimulations shares one build cache across scenarios in 
   expect_identical(built, 1L)
 })
 
+# Steady state ----
+
+# A stand-in for the one thing `.applyScenarioSteadyState()` reads off the
+# simulation, so the tests below need no PK-Sim model loaded.
+.fakeSimulation <- function(id = "sim1") {
+  list(id = id)
+}
+
+test_that(".applyScenarioSteadyState is a no-op when the scenario does not ask for one", {
+  called <- FALSE
+  local_mocked_bindings(
+    getSteadyState = function(...) {
+      called <<- TRUE
+      list()
+    },
+    .package = "ospsuite"
+  )
+  scenario <- Scenario(modelFile = "m.pkml", simulateSteadyState = FALSE)
+  simulation <- .fakeSimulation()
+
+  expect_identical(
+    .applyScenarioSteadyState(simulation, scenario, NULL),
+    simulation
+  )
+  expect_false(called)
+})
+
+test_that(".applyScenarioSteadyState solves at the scenario's time and writes the values back", {
+  solved <- NULL
+  written <- NULL
+  local_mocked_bindings(
+    getSteadyState = function(
+      simulations,
+      steadyStateTime,
+      ignoreIfFormula,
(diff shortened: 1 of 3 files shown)


tests/testthat/test-scenario-execution.R
```diff
@@ -909,6 +909,106 @@ test_that(".buildScenarioSimulations shares one build cache across scenarios in
...
+      ignoreIfFormula,
+      simulationRunOptions
+    ) {
+      solved <<- list(
+        simulations = simulations,
+        steadyStateTime = steadyStateTime,
+        ignoreIfFormula = ignoreIfFormula,
+        simulationRunOptions = simulationRunOptions
+      )
+      list(sim1 = list(paths = c("Organism|A"), values = 3.5))
```
Felixmil: The mock returns a one-element list, so this test cannot tell `initialValues[[simulation$id]]` from `initialValues[[1]]`. The comment fifteen lines down says the values land "keyed by its own id", and the description lists "the solve-and-write-back keyed by simulation id" as covered, but a regression to positional indexing passes all three assertions unchanged.

A second entry under a different key pins it:

```r
list(
  other = list(paths = "Organism|B", values = 99),
  sim1 = list(paths = c("Organism|A"), values = 3.5)
)
```

With `other` first, positional indexing writes `Organism|B` / `99` and `expect_identical(written$quantityPaths, c("Organism|A"))` fails — which is the point of keying by id, since `getSteadyState()` takes a list of simulations and returns one entry per simulation.

---

Felixmil (commented):
Reviewed the single commit on `felixmil/steady-state-seam` (7714eb7) against its base `felixmil/set-scenario-dots`. R is not installed in this environment, so this is static analysis — I did not run the suite.

The extraction is behaviour-preserving, including the truthiness claim in the description. `if (scenario$simulateSteadyState)` became `if (!scenario$simulateSteadyState) return(...)`, and the degenerate values propagate identically: `!NA` is `NA` and `!NULL` is `logical(0)`, so a corrupt flag still aborts at the `if` rather than skipping steady state in silence. Inlining `!scenario$overwriteFormulasInSS` at the call site changes nothing, and discarding the return value at step 7 of `.prepareScenario()` is fine — `setQuantityValuesByPath()` mutates the `ospsuite::Simulation` by reference, as the helper's comment says.

Also checked the one thing an extracted steady-state helper could plausibly get wrong: passing `scenario$steadyStateTime` straight through to `getSteadyState()` is correct, because the field contract stores that value in the base unit. The parser converts on the way in (`R/scenarios.R:215`) and `.scenarioToJson()` converts back to `steadyStateTimeUnit` on the way out (`R/project-to-json.R:354`), so no unit handling belongs in this helper.

Placement, the `# .applyScenarioSteadyState ----` header and the plain-comment `@keywords internal` / `@noRd` block match the other helpers in the file, and `local_mocked_bindings(.package = "ospsuite")` is the pattern already used in the test immediately above.

One finding, inline, on the strength of the middle test rather than on the change itself.

---

tests/testthat/test-scenario-execution.R
Felixmil: Confirmed and fixed, with your suggested shape: the mock now returns two entries, the simulation under test second, and a short comment says why the decoy is there.

I checked the strengthened test actually catches the regression rather than merely looking like it would. Three runs of `test-scenario-execution.R`, each after confirming which body was loaded:

1. `initialValues[[simulation$id]]` temporarily changed to `initialValues[[1]]`, test as it stood: `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 144 ]`. Your point exactly, the regression passes unnoticed.
2. Same positional lookup, second entry added to the mock: `[ FAIL 2 | ... PASS 142 ]`, failing on the two assertions that matter:

```
Expected `written$quantityPaths` to be identical to `c("Organism|A")`.
`actual`:   "Organism|B"
Expected `written$values` to be identical to 3.5.
  `actual`: 99.0
```

3. Id lookup restored: back to `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 144 ]`.

Full suite unchanged from its pre-change baseline, `[ FAIL 0 | WARN 11 | SKIP 1 | PASS 3467 ]`.

The two sibling tests do not share the weakness, so I left them alone. The no-op test asserts on a flag the mock sets, and the `overwriteFormulasInSS` test asserts the captured `ignoreIfFormula` directly; neither makes a claim about id keying that its mock cannot back.

PR #1232, opened 2026-08-06
Give the PI bounds-ordering hazard its own function

Felixmil:
Writing a PI parameter's two search bounds is order-dependent: each setter validates the new bound against the *other* one, which still holds the model's default until it is overwritten in turn, so one of the two orders passes through an invalid pair and aborts. Which one depends on where the new window sits relative to the defaults. That was a correct-by-construction `if` buried in a 170-line task builder with a comment and no test surface.

## Parameter identification

- `.assignPIBounds(runtime, minValue, maxValue)` extracted, carrying the explanation with it. `.createSinglePITask()` calls it.
- Tests drive it against a stand-in with the same crossing-rejection semantics, covering a new window above the model defaults, below them, and overlapping them. The stand-in was checked to reproduce the hazard: the naive single order aborts in both directions.

### DIFF
--- a/R/parameter-identification.R
+++ b/R/parameter-identification.R
@@ -983,12 +983,13 @@ print.PITask <- function(x, ...) {
 #
 # The two bounds cannot be set at once, and each setter validates the new bound
 # against the *other* one, which still holds the model's default until it is
-# overwritten in turn. So one of the two orders passes through an invalid pair
-# and aborts: raising `minValue` above the stale `maxValue`, or lowering
-# `maxValue` below the stale `minValue`. Which one depends on where the new
-# bounds sit relative to the model defaults, so the order is chosen per
-# parameter. A `PIParameter()` record is already validated as
-# `minValue <= startValue <= maxValue`, so the other order always succeeds.
+# overwritten in turn. The order only matters when the new window sits entirely
+# outside the current one: above it, `minValue` first is blocked by the stale
+# `maxValue`; below it, `maxValue` first is blocked by the stale `minValue`. A
+# window that overlaps the current one goes through in either order. A
+# `PIParameter()` record is already validated as
+# `minValue <= startValue <= maxValue`, so a window cannot sit both above and
+# below: at most one order is ever blocked, and this picks the other.
 #
 # Mutates `runtime` in place (an R6 object) and returns it invisibly.
 #
--- a/R/parameter-identification.R
+++ b/R/parameter-identification.R
@@ -880,13 +880,8 @@ print.PITask <- function(x, ...) {
       parameters = if (length(paramObjs) == 1L) paramObjs[[1]] else paramObjs
     )
     # Apply the declared display unit first so bounds and start value are
-    # interpreted in it, then the start value. The bound setters validate each
-    # new bound against the *other* bound, which still holds its model-default
-    # value until we overwrite it. Assign the bounds in whichever order keeps
-    # every intermediate state valid: if the new minValue would exceed the
-    # stale maxValue, raise maxValue first; otherwise lower minValue first. The
-    # PIParameter() record is already validated as minValue <= startValue <=
-    # maxValue, so one of the two orders always succeeds.
+    # interpreted in it, then the start value, then the bounds
+    # (`.assignPIBounds()` owns their ordering hazard).
     #
     # A unit that is not one of the parameter's own dimension is reported and
     # left unapplied, rather than aborting the task. The legacy 5.x
@@ -909,13 +904,7 @@ print.PITask <- function(x, ...) {
       }
     }
     runtime$startValue <- p$startValue
-    if (p$minValue >= runtime$maxValue) {
-      runtime$maxValue <- p$maxValue
-      runtime$minValue <- p$minValue
-    } else {
-      runtime$minValue <- p$minValue
-      runtime$maxValue <- p$maxValue
-    }
+    .assignPIBounds(runtime, p$minValue, p$maxValue)
     runtime
   })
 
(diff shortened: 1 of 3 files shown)


Felixmil:
Reviewed `7f1c87a`. **No findings** — this one looks good to me.

`.assignPIBounds()` is a faithful extraction: the branch condition, both assignment orders and the invariant it leans on (`PIParameter()` having already validated `minValue <= startValue <= maxValue`, so the other order always succeeds) all come across unchanged, and the explanation moved with the code instead of being left behind at the call site. The remaining comment in `.createSinglePITask()` correctly shrinks to a pointer rather than duplicating it.

The tests are the part that earns the extraction. The `.boundedRuntime()` stand-in reproduces the real constraint — each setter validating against the *other* bound as it currently stands — so the above-defaults and below-defaults cases genuinely exercise the hazard in both directions rather than just asserting the end state.

One note on scope, since this PR sits mid-stack: I reviewed the whole `ffe7a49..HEAD` range, and it turned up six findings, none of which are in this PR's diff. They're filed on the PRs that own the code:

- #1228 — the kind-registry commit loop and `.loadProjectTree()` aren't coupled the way the new comment claims (latent, not a live bug)
- #1229 — `steadyStateTime = 0` with a declared unit still round-trips to `1000`/`"min"`
- #1230 — four items on the `...` reshape: a positional field silently ignored on the record path, `.setPITask_impl()` missing the new named-field guard, a repeated field name now silently accepted, and undocumented breaking signature change

Also worth knowing: CodeRabbit skipped this PR and every other one in the stack — auto review is off for base branches other than the default — so nothing else is reviewing these automatically.

CI was still in progress at review time (`mergeable_state: unstable` reflects that, not a failure); codecov reports all modified lines covered.

---

R/parameter-identification.R
```diff
@@ -989,6 +978,33 @@ print.PITask <- function(x, ...) {
   )
 }
 
+# Write a PI parameter's search bounds onto its runtime object, in an order that
+# keeps every intermediate state valid.
+#
+# The two bounds cannot be set at once, and each setter validates the new bound
+# against the *other* one, which still holds the model's default until it is
+# overwritten in turn. So one of the two orders passes through an invalid pair
```
Felixmil: "So one of the two orders passes through an invalid pair and aborts" overstates it — for a new window that overlaps the model defaults, both orders are valid.

Model defaults `[0, 10]`, new bounds `[1, 5]`:

- min first: `minValue <- 1` against stale `maxValue = 10` ✓, then `maxValue <- 5` against `minValue = 1` ✓
- max first: `maxValue <- 5` against stale `minValue = 0` ✓, then `minValue <- 1` against `maxValue = 5` ✓

The code already reflects this — the `else` branch is exactly the overlapping case, and it isn't a "the other order would have aborted" branch. One of the new tests covers it too.

The hazard is real for a window entirely outside the defaults, which is what the ordering exists for; it's just that only one of the two directions is ever forced. Suggest "So one of the two orders *may* pass through an invalid pair and abort", or naming the condition directly: the order matters only when the new window sits entirely above or entirely below the current one. Worth getting right since the whole justification for the function is that this ordering is non-obvious.

---

Felixmil (commented):
Found while reviewing #1235 over the whole stack (`e9b63b4..HEAD`) — this one is about the comment introduced here, so leaving it on this PR. Static pass only; no R interpreter in this environment.

The extraction itself is behaviour-preserving and the function is clearly better than the comment-inside-a-170-line-builder it replaces. One inline note: the explanatory comment claims more than the code does.

---

R/parameter-identification.R
Felixmil: Confirmed, and the comment is now fixed. You're right that the overlapping case makes both orders valid, so "one of the two orders passes through an invalid pair and aborts" claimed more than the code does.

I checked it against the real setter guards rather than just the test stand-in. `PIParameters` rejects a new `minValue` when `value > startValue || value >= maxValue`, and a new `maxValue` when `value < startValue || value <= minValue`. Since the call site writes `startValue` before the bounds, the `startValue` half never bites, and what's left is exactly the crossing check. Working the three windows the tests cover:

- **above** (stale `[0, 1]`, new `[10, 20]`): min first is blocked by the stale `maxValue`; max first succeeds. Order forced.
- **below** (stale `[100, 200]`, new `[1, 5]`): max first is blocked by the stale `minValue`; min first succeeds. Order forced.
- **overlapping** (stale `[0, 10]`, new `[2, 8]`, and your `[1, 5]`): both orders succeed. Order does not matter.

I also swept every representable combination on a small grid (22308 of them, stale bounds ordered and `min < start < max`): both orders are valid in 82% of them and exactly one is forced in the remaining 18%. Never are both blocked, and the branch never picks the blocked one. So the function itself has no defect and I left its behaviour alone; this is a prose change only.

New wording:

> The two bounds cannot be set at once, and each setter validates the new bound against the *other* one, which still holds the model's default until it is overwritten in turn. The order only matters when the new window sits entirely outside the current one: above it, `minValue` first is blocked by the stale `maxValue`; below it, `maxValue` first is blocked by the stale `minValue`. A window that overlaps the current one goes through in either order. A `PIParameter()` record is already validated as `minValue <= startValue <= maxValue`, so a window cannot sit both above and below: at most one order is ever blocked, and this picks the other.

Suite stays green: `[ FAIL 0 | WARN 11 | SKIP 1 | PASS 3474 ]`.

One thing I left as is: the test stand-in rejects with `>` / `<` where the real setters use `>=` / `<=`, and it doesn't model `startValue`. That makes it marginally more permissive than the real class (it would accept a degenerate `min == max` window), but it preserves the crossing semantics the ordering actually depends on, so all three cases still discriminate correctly. Happy to tighten it if you'd rather it mirror the real guards exactly.

PR #1233, opened 2026-08-06
Compare a project against its Excel files without importing

Felixmil:
`projectStatus()`'s Excel axis answered a read-only question by running the importer for real: a whole `definitions/` tree written to a temporary directory, re-parsed, diffed, and deleted. A status check therefore cost as much as an import and inherited its side effects. This is the natural follow-on to splitting the importer's read half from its write half.

## Excel bridge

- `.excelToProjectJson()` holds the reading: every workbook the property sheet names is parsed into one inlined container with canonical ids, and nothing is created on disk. `importProjectFromExcel()` calls it and keeps the writing, resolving `outputDir` up front because the read half anchors a relative `${VAR}` data folder to it.
- `.compareJsonToExcel()` reads the workbooks through it and parses the result directly. The container is parsed against a directory holding no `definitions/`, so every section takes the inline fallback — pointing it at the real project folder would resolve each section against the tree sitting beside the workbooks, which is the very thing being compared.

## Project loading

- `.parseProjectSections()` split out of `.loadProjectTree()`, so a container that never came off disk can be parsed too. `.loadProjectTree()` keeps the file reading and calls it.
- `.projectFromSections()` in `R/project.R` commits a parsed section list to a fresh `Project`, next to the class whose private seam it reaches through.

## Comparison normalization

Removing the disk round trip removed two normalizations the comparison had been getting for free, so both sides now get them explicitly and symmetrically:

- Definition order within an array-shaped section is no longer counted as drift. A tree project holds one file per definition, so its order is the order the filesystem lists them in, while a workbook's is the order its rows were typed. Records are ordered by their own serialized content, which needs no per-section id knowledge and is exactly as discriminating. Arrays *inside* a definition (a PI task's parameter list) keep their order.
- Both sides are compared as the JSON they would be written as, since `jsonlite` narrows a whole double to an integer on the way back in: a `250` that reached one side through a file and the other straight from a workbook is the same number stored as two types.

## Tests

The comparison had no test asserting it reports real drift, which is what makes a normalization change dangerous. Two were added: a real edit still flips `excel_in_sync` and names the section, and the comparison leaves the Excel project folder byte-for-byte as it found it.

### DIFF
--- a/R/project-excel.R
+++ b/R/project-excel.R
@@ -992,42 +992,48 @@ exportProjectToExcel <- function(
 # in-memory serialization (`.projectToJson()`) of a loaded project:
 #   - memory side: the live `project`, so the comparison reflects the project as
 #     it is now (unsaved edits included), not the on-disk container;
-#   - Excel side: a fresh re-import of the side-car workbook, loaded back into a
-#     `Project` and serialized the same way.
+#   - Excel side: the side-car workbook read into a `Project` and serialized the
+#     same way.
 # Serializing both from an in-memory project (rather than reading a container
 # file raw) is what makes the comparison container-shape-independent: a tree
 # project's on-disk container keeps every section emptied (the tree owns them),
 # so reading either container raw would blind the comparison and report false
 # drift on every section. The volatile `esqlabsRVersion` is ignored.
 #
+# Nothing is written: a read-only question is answered by reading. The workbooks
+# are parsed into an inlined container (`.excelToProjectJson()`) and that
+# container is parsed into a `Project` directly, so a status check costs one
+# workbook read rather than a whole definitions tree written to a temporary
+# directory, re-read, and deleted.
+#
 # @keywords internal
 # @noRd
 .compareJsonToExcel <- function(project, projectConfigPath, silent = FALSE) {
-  # Create temporary snapshot from current Excel files
-  tempDir <- tempfile("config_snapshot")
-  dir.create(tempDir, showWarnings = FALSE, recursive = TRUE)
-  on.exit(unlink(tempDir, recursive = TRUE), add = TRUE)
-
-  # `copyAssets = FALSE`: this snapshot exists only to be serialized and compared,
-  # then deleted with `tempDir`. Copying the referenced models, data, and
-  # population folders into it would turn every `projectStatus()` read into a
-  # recursive copy of the whole asset tree, and a copy failure (a locked file, a
-  # full disk) would surface as "cannot compare the Excel files".
-  tempJsonPath <- importProjectFromExcel(
+  projectDir <- dirname(fs::path_abs(projectConfigPath))
+  excelJson <- .excelToProjectJson(
     projectConfigPath,
-    outputDir = tempDir,
-    silent = TRUE,
-    copyAssets = FALSE
+    projectDir = projectDir
+  )$jsonData
+
+  # The container is parsed against a directory that does not exist, so every
+  # section resolves to the one inlined in it. Pointing this at the real project
+  # folder would instead resolve each section against the `definitions/` tree
+  # sitting beside the workbooks, which is the very thing being compared: the
+  # Excel side would read back the memory side and never report drift.
+  treelessDir <- tempfile("excel_compare")
+  excelProject <- Project$new(
+    sections = .parseProjectSections(
+      excelJson,
+      jsonPath = file.path(treelessDir, "Project.json"),
+      projectDirPath = treelessDir
+    )
   )
 
-  # The memory side: serialize the live in-memory project. The Excel side: load
-  # the just-imported temp tree project and serialize it the same way. Both go
-  # through `.projectToJson()`, so the two sides are compared in one shape
-  # regardless of how either container is stored on disk. `Project$new()` (not
-  # `loadProject()`) skips the cross-reference warning pass, so a project with
-  # dangling refs compares quietly.
+  # Both sides go through `.projectToJson()`, so they are compared in one shape
+  # regardless of how either container is stored on disk. The memory side is the
+  # live `project`, so the comparison reflects unsaved edits too.
   originalJsonObj <- .projectToJson(project)
-  currentJsonObj <- .projectToJson(Project$new(projectFilePath = tempJsonPath))
+  currentJsonObj <- .projectToJson(excelProject)
 
   # Both sides canonicalize every id (via `.canonicalizeProjectJsonIds()`) so id
   # canonicalization is never counted as drift (which would make an otherwise
@@ -1049,6 +1055,24 @@ exportProjectToExcel <- function(
   # task's parameter list) and is left as it is.
   originalJsonObj <- .sortJsonKeys(originalJsonObj)
   currentJsonObj <- .sortJsonKeys(currentJsonObj)
+
+  # And in the same spirit for the sections held as arrays (`scenarios`,
+  # `individuals`, `populations`, `observedData`): which definition comes first
+  # carries no meaning either. A tree project holds one file per definition, so
+  # its order is the order the filesystem lists them in, while a workbook's is
+  # the order its rows were typed. Only these top-level sections are reordered;
+  # an array inside a definition keeps its order.
+  originalJsonObj <- .sortJsonSections(originalJsonObj)
+  currentJsonObj <- .sortJsonSections(currentJsonObj)
+
+  # Finally, both sides are compared as the JSON they would be written as rather
+  # than as the R values behind it, because `jsonlite` narrows a whole double to
+  # an integer on the way back in: a `250` that reached one side through a file
+  # and the other straight from a workbook is the same number stored as two
+  # types. Encoding and decoding both sides the same way settles that, and
+  # neither side touches disk to do it.
+  originalJsonObj <- .asWrittenJson(originalJsonObj)
+  currentJsonObj <- .asWrittenJson(currentJsonObj)
 
   # Remove esqlabsRVersion -- it changes with package updates and would cause
   # false out-of-sync reports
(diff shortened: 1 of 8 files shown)


R/project-excel.R
```diff
@@ -339,6 +339,144 @@ importProjectFromExcel <- function(
...
+  # the container and fully reconciles the `definitions/` tree (deleting any
+  # definition authored only on the JSON side), so re-importing over a JSON
+  # project the user has since edited would erase that work. Abort unless
+  # `overwrite = TRUE` when the target container file already exists, or a
+  # non-empty `definitions/` tree already sits in `outputDir`.
+  definitionsDir <- file.path(outputDir, "definitions")
+  hasDefinitionTree <- dir.exists(definitionsDir) &&
+    any(grepl("\\.json$", list.files(definitionsDir, recursive = TRUE)))
+  if (!overwrite && (file.exists(outputPath) || hasDefinitionTree)) {
+    cli::cli_abort(messages$importWouldOverwriteProject(outputDir))
```
Felixmil, line 372: Extracting the read half moved this guard from mid-read to after the whole read, so two diagnostics that used to come *after* the overwrite refusal now come before it.

Previously the guard sat between `.parseExcelObservedData()` and `.canonicalizeProjectJsonIds()`. Now `.excelToProjectJson()` runs to completion first, which means `.canonicalizeProjectJsonIds()`'s collision abort and `.warnIncompleteObservedCurves()`'s warnings both fire ahead of it. Re-importing a collision-bearing workbook into an occupied folder now reports the collision rather than the overwrite refusal, and an incomplete-curve warning is emitted for an import that then refuses to run.

Nothing is written before the guard either way, so this is diagnostic precedence rather than data loss — flagging it because "would overwrite" is the more actionable of the two messages for that case, and the read half doing all its work before the cheap precondition is checked is also wasted effort on the refusal path. Hoisting the `outputPath` / `definitionsDir` check above the `.excelToProjectJson()` call would restore the old ordering; nothing in the guard depends on the read.

---

Felixmil (commented):
Found while reviewing #1235 over the whole stack (`e9b63b4..HEAD`) — this one belongs to the read/write split in this PR, so leaving it here. Static pass only; no R interpreter in this environment.

One ordering observation inline. The extraction itself reads as behaviour-preserving otherwise, and the `projectDir` threading is an improvement: `projectStatus()` derives the workbook path from the project's own folder, so `.parseExcelObservedData()` now anchors where it should, where the comparison path previously anchored it at a temp dir.

---

Felixmil (commented):
Found while code-reviewing the stack from #1238; the code is this PR's, so filing it here.

### `.sortJsonSections()` builds its ordering key at `digits = 4`

The sort key is serialized without a `digits` argument, so jsonlite's default of 4 decimal places applies:

```r
as.character(jsonlite::toJSON(record, auto_unbox = TRUE, null = "null"))
```

`.asWrittenJson()`, a few lines below in the same file, passes `digits = NA` for what is essentially the same reason — it needs the comparison to turn on the value rather than on how the value was produced.

The consequence is narrow but it defeats the sort in exactly the case the sort exists for. Two records differing only past the 4th decimal collapse to the same key, and `order(keys, method = "radix")` is stable, so they keep their source order. That is the source order the sort is there to normalize away, so the two sides can present them in different orders and `.compareJsonToExcel()` reports drift for records that are equal.

`digits = NA` on the key matches the sibling and makes the key as discriminating as the comparison downstream of it.

---

R/project-excel.R
Felixmil: Confirmed against the diff of the extraction commit: the guard used to sit between `.parseExcelObservedData()` and `.canonicalizeProjectJsonIds()`, and it now runs after `.excelToProjectJson()` returns.

Verdict: wrong, not merely different, so I moved it. Two things decided it. Nothing in the guard reads anything the read half produces, it only needs `outputDir`, `projectFileName` and `overwrite`, all of which are settled before the call. And the sibling refusal, `exportProjectToExcel()`'s `exportWouldOverwriteWorkbooks` guard, is the first thing that function does, before it serializes a single sheet. The two "would overwrite" guards on the import/export pair should behave the same way, and refusing before doing avoidable work is the behaviour worth keeping.

I did not restore the old position. That spot was an artifact of where the `outputDir` resolution used to sit rather than a chosen point, and preserving it would mean splitting the read half again for no gain. The guard is now above the `.excelToProjectJson()` call, as you suggested.

One correction to that suggestion, though. Hoisting above the whole read also puts the guard ahead of `.excelToProjectJson()`'s own source checks, so a mistyped `projectConfigPath` would report the destination refusal instead of "File not found", and the default `outputDir` is the folder that most often already holds a project. That would have been the same class of accidental reorder, so `importProjectFromExcel()` now repeats the `file.exists()` check ahead of the guard. Three lines, with a comment saying why they are repeated.

The order is now: source exists, destination free, then read. Every diagnostic sits where it can be acted on, and the workbook's own diagnostics only appear for an import that is actually going to run.

Both halves are pinned by snapshots in `test-project-excel.R`. I checked they discriminate rather than pass vacuously: against the previous ordering both fail, the first reporting "is not a readable Excel project file" and the second reporting the same "File not found" attributed to `.excelToProjectJson()`.

Neither `messages$fileNotFound` nor the corrupt-workbook abort was covered by any test or snapshot before this, which is why the extraction was free to move them. Worth noting separately: the corrupt-workbook abort still renders as ``Error in `.excelToProjectJson()`:``, naming a private helper in user-facing text, on the import path and through `projectStatus()`. Same extraction, different symptom, and out of scope for this thread.

R/project-excel.R
```diff
@@ -1123,6 +1184,56 @@ exportProjectToExcel <- function(
...
+    isArraySection <- is.list(section) &&
+      length(section) > 1L &&
+      is.null(names(section))
+    if (!isArraySection) {
+      next
+    }
+    keys <- vapply(
+      section,
+      function(record) {
+        as.character(jsonlite::toJSON(record, auto_unbox = TRUE, null = "null"))
```
Felixmil: The sort key rounds, so it is less discriminating than the comparison it feeds.

`jsonlite::toJSON()` defaults to `digits = 4`, so two records in the same section that agree to four decimals and differ only past them serialize to the same key. `order(..., method = "radix")` is stable, so a tie leaves each side in its own incoming order — the tree side in filesystem order, the workbook side in the order its rows were typed. `.asWrittenJson()` immediately below *does* pass `digits = NA`, so the two records stay distinguishable there, and the `identical(originalJsonObj[[file]], currentJsonObj[[file]])` check reports drift on a project nobody edited. That is exactly the false-positive this normalization exists to remove.

`digits = NA` here makes the key as precise as the comparison that follows it, and matches the file's other two serializations (`.asWrittenJson()`, and the import bootstrap in `importProjectFromExcel()`), which both pass it:

```r
as.character(jsonlite::toJSON(
  record,
  auto_unbox = TRUE,
  digits = NA,
  null = "null"
))
```

Narrow — it needs two definitions in one array-shaped section agreeing to four decimals — but it costs one argument and removes the failure mode outright.

---

Felixmil (commented):
Static review of the two commits in this PR (no R runtime in this environment, so nothing was executed; CI is green on all nine checks).

One finding, inline: `.sortJsonSections()`'s sort key omits `digits = NA`, so it is coarser than the `.asWrittenJson()` comparison it feeds and a tie can surface as false drift.

Checked and found sound:

- **`.excelToProjectJson()` extraction.** No write-half variable (`overwrite`, `silent`, `copyAssets`, `projectFileName`, `outputPath`, `importedProject`) survives in the read half, and `conventionWorkbooks` / `filePathProps` / `pcDir` accumulate in the right scope to be returned. The hoisted `outputDir` default is byte-identical to the old `pcDir` — both are `dirname(fs::path_abs(projectConfigPath))` — and the overwrite guard still sits after the read, as it did before.
- **`.parseProjectSections()` extraction.** `projectDirPath` is used only by `loadSection()`, and `.loadDefinitionTree()` returns the inline section verbatim for a directory that does not exist (`.assertDefinitionTreePathIsDir()` only aborts on a path that exists as a *file*), so the treeless parse genuinely takes the inline fallback for every kind rather than relying on an error path. The new function repeats `.loadProjectTree()`'s own `rlang::local_error_call(NULL)`, so error attribution survives the split.
- **Treeless compare.** `.projectToJson()` reads neither `projectDirPath` nor `projectFilePath`, so pointing the Excel-side project at a nonexistent `treelessDir` cannot leak into either side of the diff. Anchoring the `${VAR}` data folder at the real project folder rather than the old temp dir is a strict improvement: a relative expansion used to resolve under a directory where nothing existed, dropping `observedData` from the Excel side and reporting drift for it.
- **`Project$new(sections = )`.** Equivalent to the loaded-project shape (`.applyLoadedSections()` is the same commit step the file path takes), with the dirty bit correctly left clear, as the new lifecycle test asserts.
- **Normalization symmetry.** `.sortJsonKeys()` runs before `.sortJsonSections()`, so identical records serialize to identical keys on both sides; `.asWrittenJson()` is applied to both sides with the same arguments, so the type settling cannot favour one. Only top-level array sections are reordered, leaving a PI task's parameter list alone.

---

Felixmil:
On the `.sortJsonSections()` review (no inline comment on it, so replying here).

Changed, `digits = NA` on the key. But the reasoning splits into two halves, and only the first holds.

**The mechanism is real.** Verified directly rather than by reading. Two records `list(id = "s", value = 1.00001)` and `list(id = "s", value = 1.00002)` both serialize to `{"id":"s","value":1}` under the default, tie, and `.sortJsonSections()` then returns them in whichever order it received them, so the two sides do not normalize. Under `digits = NA` the keys are distinct and both sides come out in the same order. That part is exactly as described.

**The false drift is not reachable in practice.** The sort runs after `.sortJsonKeys()` and before `.asWrittenJson()`, and a tie needs two records in one section to be equal on *every* field after 4-decimal rounding, including their id. Every array section a real project holds carries a unique id per record: `scenarios` has `name`, `individuals` `individualId`, `populations` `populationId`, `dataCombined` `dataCombinedId`, `plots` `plotId`, `plotGrids` `plotGridId`, `parameterIdentification` `id`, `observedData` the data-file basename in `file`. Distinct ids give distinct keys whatever the rounding does to the numeric fields, so no two records tie and the key's precision never reaches the verdict.

Worth being explicit about a second thing the rounding does *not* do, since it looked like a candidate: it cannot reorder two records that do not tie. Each side computes its key from the same record content by the same rule, so for records that are actually equal the keys are equal too, and only a collapse of two distinct records onto one key defeats the normalization.

So this lands as consistency plus a guard, not a bug fix, and I would rather say that than overstate it. The key should be as discriminating as the `identical()` it feeds regardless of an invariant it does not enforce itself, and the sibling `.asWrittenJson()` passes `digits = NA` a few lines below for the same reason.

Test in `test-project-excel.R` covers the sort directly, with the two records given the same id so they differ only past the 4th decimal. It fails against the previous key (the two source orders come back reversed) and passes with it. The comment above the test says outright that it guards the sort rather than a live false-drift report, so nobody later reads it as a regression test for a bug that never happened.

Full suite: `[ FAIL 0 | WARN 11 | SKIP 1 | PASS 3485 ]`, against a `PASS 3482` baseline on the same branch.

R/project-excel.R
Felixmil: Agreed and already changed — the sort key now passes `digits = NA`, matching `.asWrittenJson()` and the import bootstrap.

The mechanism is exactly as you describe: two records agreeing to four decimals tie, and a stable radix sort then leaves each side in its own incoming order, which is the source order the sort exists to erase.

One qualification from checking how far it reaches. I could not construct the false positive against a real project. A tie needs two records in one array-shaped section equal on *every* field after rounding, id included, and every array section keys on a unique id per record — `scenarios` on `name`, `individuals` on `individualId`, `plots` on `plotId`, `observedData` on the data-file basename, and so on. Distinct ids give distinct keys whatever rounding does to the numeric fields. Rounding also cannot reorder two records that do not tie, since both sides compute keys from the same content by the same rule, so a collapse onto one key is the only failure mode and unique ids block it.

So this went in as a guard on the sort rather than a bug fix, and the test comment says so, to stop a later reader taking it for a regression test of something that happened. The test itself is real: reverting the change makes it fail with the two source orders coming back reversed.

PR #1235, opened 2026-08-06
Unquote a quoted single-value cell on Excel import

Felixmil:
A migration test of the bundled legacy fixture `TestProjectExcelLegacy` found that a quoted single-value cell keeps its quotation marks on import. Its `Plots.xlsx` writes `DataCombinedName` as `"AciclovirPVB"`, quotes included, and the import turned that into the id `_aciclovirpvb_` — the quotes canonicalized to underscores rather than recognised as the cell's syntax. Multi-value list cells were already handled: `"Global", "Aciclovir", "Sheet, with comma"` splits correctly with the interior comma preserved. Nothing broke, because the defining and referencing sides agreed, but the ids were ones no user would author and they differed from the hand-authored equivalent.

## Excel bridge

- `.stripWrappingQuotes()` drops only the outermost pair, so a quote inside the text stays part of the value and an unquoted cell comes back as read.
- It is applied inside `.canonicalizeProjectJsonIds()` rather than at the two `DataCombinedName` readers. That function is the point every id and reference already routes through, and unquoting is an id-spelling concern, so this keeps a definition and a reference on one transform — the property the rest of that function rests on. The same four lines fix roughly twenty sibling cells carrying the identical latent defect (`Scenario_name`, `IndividualId`, `PopulationId`, `ApplicationProtocol`, `plotID`, `plotGridId`, the PI task ids, `OutputPathId`, and the map keys of `outputPaths` / `parameterSets` / `applications` / `initialConditions`).

This is safe for the JSON side of `projectStatus()`'s comparison, which also calls this function: `.canonicalizeOneId()` already maps `"` to `_`, so a canonical JSON id can never carry a literal quote and the strip is a no-op there.

## Tests

The existing characterization test asserted the bug in its title (`a quoted single-value id keeps its quotes as underscores`); it is rewritten to assert the fixed behaviour rather than joined by a second test. `.stripWrappingQuotes()` gets its own unit test covering the wrapping pair, an interior quote, a comma inside a single-value cell, a lone quote, and a non-character cell.

Left alone: observed-data `dataSet` references, which are matched verbatim by design and would need the matching side handled too — that is the inconsistent-quoting residue #1207, not this bug.

### DIFF
--- a/R/project-excel.R
+++ b/R/project-excel.R
@@ -1466,6 +1466,30 @@ projectStatus <- function(project, silent = FALSE) {
     }
     section
   }
+  # A section that is an array of records carries no names for `canonNames()` to
+  # check, yet the parse step keys it by each record's own id field
+  # (`result[[id]] <- entry`) just the same. Two records whose ids canonicalize
+  # alike therefore become one definition, built from the last of them, and the
+  # earlier record is gone without a word. Run the section's ids through the same
+  # collision-checking path for the abort alone; the per-record assignments below
+  # still write each canonical id.
+  checkRecordIds <- function(section, idField) {
+    if (is.null(section)) {
+      return(invisible(NULL))
+    }
+    ids <- vapply(
+      section,
+      function(record) {
+        value <- record[[idField]]
+        if (length(value) == 1L) as.character(value) else NA_character_
+      },
+      character(1)
+    )
+    .silentlyCanonicalized(
+      .canonicalizeId(.stripWrappingQuotes(ids[!is.na(ids)]))
+    )
+    invisible(NULL)
+  }
 
   jsonData$outputPaths <- canonNames(jsonData$outputPaths)
   jsonData$parameterSets <- canonNames(jsonData$parameterSets)
@@ -1482,6 +1506,7 @@ projectStatus <- function(project, silent = FALSE) {
   }
 
   if (!is.null(jsonData$scenarios)) {
+    checkRecordIds(jsonData$scenarios, "name")
     jsonData$scenarios <- lapply(jsonData$scenarios, function(sc) {
       sc$name <- canonScalar(sc$name)
       sc$individual <- canonScalar(sc$individual)
@@ -1495,6 +1520,7 @@ projectStatus <- function(project, silent = FALSE) {
   }
 
   if (!is.null(jsonData$individuals)) {
+    checkRecordIds(jsonData$individuals, "individualId")
     jsonData$individuals <- lapply(jsonData$individuals, function(ind) {
       ind$individualId <- canonScalar(ind$individualId)
       if (!is.null(ind$parameterSets)) {
@@ -1505,6 +1531,7 @@ projectStatus <- function(project, silent = FALSE) {
   }
 
   if (!is.null(jsonData$populations)) {
+    checkRecordIds(jsonData$populations, "populationId")
     jsonData$populations <- lapply(jsonData$populations, function(pop) {
       pop$populationId <- canonScalar(pop$populationId)
       pop
@@ -1543,6 +1570,7 @@ projectStatus <- function(project, silent = FALSE) {
   # data, whose ids are file basenames / DataSet names matched verbatim and
   # never canonicalized, so they are deliberately left untouched.
   if (!is.null(jsonData$dataCombined)) {
+    checkRecordIds(jsonData$dataCombined, "dataCombinedId")
     canonEntryScenario <- function(entry) {
       entry$scenario <- canonScalar(entry$scenario)
       entry
@@ -1567,6 +1595,7 @@ projectStatus <- function(project, silent = FALSE) {
     )
   }
   if (!is.null(jsonData$plots)) {
+    checkRecordIds(jsonData$plots, "plotId")
     jsonData$plots <- lapply(
       jsonData$plots,
       function(plot) {
@@ -1577,6 +1606,7 @@ projectStatus <- function(project, silent = FALSE) {
     )
   }
   if (!is.null(jsonData$plotGrids)) {
+    checkRecordIds(jsonData$plotGrids, "plotGridId")
     jsonData$plotGrids <- lapply(
       jsonData$plotGrids,
       function(grid) {
@@ -1609,6 +1639,7 @@ projectStatus <- function(project, silent = FALSE) {
   # ids), and a parameter's / mapping's own `id` is an inner id, not an
   # definition-file id, so those are left untouched.
   if (!is.null(jsonData$parameterIdentification)) {
+    checkRecordIds(jsonData$parameterIdentification, "id")
     jsonData$parameterIdentification <- lapply(
       jsonData$parameterIdentification,
       function(task) {
@@ -2850,9 +2881,10 @@ projectStatus <- function(project, silent = FALSE) {
 # renamed apart, because a grid naming the id would then reach only one of them
 # and the other plot would belong to no grid.
 #
-# Compared on the canonical id, since that is the id the definitions end up keyed
-# by: two rows spelling one id differently are as much of a collision as two
-# spelling it the same.
+# Compared on the unquoted, canonical id, since that is the id the definitions
+# end up keyed by: two rows spelling one id differently are as much of a
+# collision as two spelling it the same, and a row that quotes the id names the
+# same plot as a row that does not.
 #
 # @param records The parsed records of the sheet (an unnamed list).
 # @param idField The record field holding the id (`plotId` / `plotGridId`).
@@ -2866,7 +2898,9 @@ projectStatus <- function(project, silent = FALSE) {
   }
   ids <- vapply(
     records,
-    function(record) .canonicalizeOneId(as.character(record[[idField]])),
+    function(record) {
+      .canonicalizeOneId(.stripWrappingQuotes(as.character(record[[idField]])))
+    },
     character(1)
   )
   duplicated <- unique(ids[duplicated(ids)])
@@ -2938,7 +2972,11 @@ projectStatus <- function(project, silent = FALSE) {
     if (is.null(name)) {
       next
     }
-    name <- as.character(name)
+    # Group on the unquoted name, the name the definition ends up keyed by.
+    # Grouping on the raw cell splits one combination in two whenever a sheet
+    # spells its name both ways, and the two halves then land on one key, so
+    # whichever is built first is overwritten and its curves are lost.
+    name <- .stripWrappingQuotes(as.character(name))
     dataType <- .naToNull(df$dataType[[i]])
     entry <- list()
     for (col in names(df)) {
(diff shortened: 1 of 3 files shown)


R/project-excel.R
```diff
@@ -1407,17 +1407,26 @@ projectStatus <- function(project, silent = FALSE) {
...
+  # instead turns each one into `_`, so the id reads `_aciclovirpvb_` rather
+  # than the name the modeller wrote. Unquoting here rather than at each of the
+  # ~20 sheet readers is what keeps a definition and a reference on the same
+  # transform, which is the property the rest of this function rests on.
   canonScalar <- function(x) {
     if (is.null(x)) {
       return(x)
     }
-    .canonicalizeOneId(as.character(x))
+    .canonicalizeOneId(.stripWrappingQuotes(as.character(x)))
```
Felixmil, line 1440: Unquoting makes `"X"` and `X` collapse to one canonical id, but the record-keyed sections take the `canonScalar` path, which has no collision check — so an inconsistently quoted workbook now silently loses content where it previously produced two oddly-named definitions.

`canonNames()` routes the map-keyed sections through `.canonicalizeId()`, which aborts on a collision. `dataCombined`, `plots`, `plotGrids` and `parameterIdentification` are arrays of records, so their ids go through `canonScalar()` → `.canonicalizeOneId()` instead, with no such guard.

Concretely for `dataCombined`: `.parseExcelDataCombinedSheet()` groups rows by the **raw** cell (`grouped[[name]]`, before any canonicalization) and returns `unname(grouped)`. A `DataCombined` sheet whose rows spell one name both ways —

| DataCombinedName | dataType |
|---|---|
| `"AciclovirPVB"` | simulated |
| `AciclovirPVB` | observed |

— therefore yields two records. Before this change they canonicalized to `_aciclovirpvb_` and `aciclovirpvb`, two distinct definitions; both survived. Now both become `aciclovirpvb`, and `.parseNestedDataCombined()` (`R/plots.R:69`) does `result[[id]] <- entry`, so the second record overwrites the first and the simulated curve is dropped with no warning. `.parsePlotEntries()` has the same `result[[id]] <- cleaned` assignment for `plots` / `plotGrids`.

The PR body scopes inconsistent quoting out as the #1207 residue, which is fair for the *dangling reference* half of that issue. This is a different consequence, though: silent curve loss rather than a reference that fails validation. Worth either running these four sections' ids through the collision-checking path too, or having `.parseNestedDataCombined()` / `.parsePlotEntries()` abort on a duplicate id instead of overwriting.

---

Felixmil (commented):
Reviewed this PR together with the four commits below it in the stack (`e9b63b4..HEAD`). No R interpreter in this environment, so this is a static pass: I read each touched seam plus its surroundings (`.parseProjectSections` / `.applyLoadedSections`, `.projectToJson` and the per-section serializers, `.loadDefinitionTree`, `.parseObservedDataTree`, `.parseNestedDataCombined`, `.projectSyncStatus`).

One finding on this PR's own diff, inline. Two lower-severity ones belong to commits below this PR, so I left them on the PRs that own the code: an ordering change on #1233, a comment inaccuracy on #1232.

The reasoning about why `.canonicalizeProjectJsonIds()` is the right place for the strip holds up, and the no-op claim for the `projectStatus()` JSON side checks out — `.canonicalizeOneId()` maps `"` to `_`, so a canonical JSON id can't carry a literal quote. The inline comment is about the one case that changes behaviour rather than spelling.

Also checked and clear, in case they come up: the memory and Excel sides of `.compareJsonToExcel()` stay symmetric (`.filePathsToJson()` uses `rawFilePaths()`, so `treelessDir` never reaches the serialization; `paths$definitionsFolder` defaults to `"definitions"` on both sides; `.parseObservedDataTree()` `unname()`s, so the Excel side's named `observedData` map doesn't read as drift); `tempfile()` as a non-existent `projectDirPath` correctly takes the inline fallback in `.loadDefinitionTree()`; `outputDir <- dirname(fs::path_abs(projectConfigPath))` is exactly the old `pcDir`; `projectStatus()` derives the workbook path from the project's own folder, so `.parseExcelObservedData()`'s `projectDir` anchoring is now correct where it was previously anchored at a temp dir; `.stripWrappingQuotes()` is right about only the outermost pair (`sub()` is greedy but anchored at both ends); the new `sections =` constructor argument leaves the object in the same state the removed `.projectFromSections()` did.

---

R/project-excel.R
Felixmil: Reproduced end to end before fixing, importing the legacy fixture with one name spelled both ways. You are right that content disappears, and the picture is slightly different per section.

**Observed on this branch** (`aciclovirpvb` shorthand for the loaded definitions):

| case | result | reported? |
|---|---|---|
| `DataCombined` row 1 `"AciclovirPVB"`, row 2 `AciclovirPVB` | `aciclovirpvb` has `simulated=0, observed=1` — the simulated curve is gone | no |
| `plotConfiguration` row 2 takes row 1's id, quoted | 3 rows, 2 plots (`p1`, `p3`) | no |
| `plotGrids` grid 2 takes grid 1's name, quoted | 2 grids, 1 grid | no |
| `PIParameters` quotes `PITaskName`, the other sheets do not | one `aciclovirfit` with `parameters=2, outputMappings=0` | no |

So the loss is real in all four, and your `dataCombined` walkthrough is exactly right.

**New or pre-existing:** both, in different halves. Running the same cases against the parent commit's importer, all four kept everything — two records each (`_aciclovirpvb_` + `aciclovirpvb`, `_p1_` + `p1` + `p3`, `_aciclovir_` + `aciclovir`, `_aciclovirfit_` + `aciclovirfit`). The loss in those four is genuinely new here.

But a pair that already canonicalized together loses content identically on both sides: `Aciclovir PVB` and `Aciclovir,PVB` (both `aciclovir_pvb`, no quoting involved) give `simulated=0, observed=1` before this PR as well. So the missing collision guard on the record-keyed sections is pre-existing; this PR widens which ids fall into it.

The contrast that explains the silence: `P 1` / `P,1` *is* reported, on both sides, by `.warnDuplicatePlotIds()`. That helper canonicalizes without unquoting, so it does not recognise a quoted twin — which is why the new cases lose a plot where the equivalent old case merely warns.

**Verdict:** two defects, and they want different answers.

The regression is that three readers consume the raw id *before* the canonicalization chokepoint, so the single-transform property the function's comment claims does not actually hold for them. Each now gets the same unquoting:

- `.parseExcelDataCombinedSheet()` groups on the unquoted name. The twins merge into one combination carrying both curves. Aborting would be wrong here: the sheet is long-format, one row per curve, and a workbook naming one combination two ways is unambiguous under this PR's own premise that the quotes are cell syntax.
- The 5.x PI path gathers task names and matches task rows unquoted, so a sheet quoting the name contributes to the one task (`parameters=2, outputMappings=2`).
- `.warnDuplicatePlotIds()` compares on the unquoted canonical id. A plots row is one definition, not a group, so a twin is a reused id: last row wins and it is reported, which is the contract that section already has and that the identical-id test pins.

The pre-existing hole is fixed the way you suggested, reusing the existing machinery rather than growing a second guard: each record-keyed section's ids now run through `.canonicalizeId()`, so two *distinct* names that canonicalize alike abort the import exactly as they do for a keyed section. Applied to all seven record sections (`scenarios`, `individuals`, `populations` as well as your four), since it is a property of the array-of-records shape rather than of those four.

One deliberate limit worth naming: that check does not fire on quoted/unquoted twins. After unquoting they are the same string, and `.canonicalizeId()` treats a repeated id as a duplicate rather than an ambiguity. That is what leaves the merging and the reporting to the three readers above, and it is also what keeps the existing "two plot rows sharing an id lose one plot, and it is reported" behaviour intact.

Four tests added in `test-project-excel.R`, each failing without the fix in its own way: 0 simulated curves, 0 output mappings, no warning, no error. Full suite 3489 -> 3499 passing, `FAIL 0`, same `WARN 11 | SKIP 1`.

Felixmil:
Two follow-ups on the unquoting change, both found while reviewing the stack above this PR. Neither is a regression from the parent — but this PR is where the machinery they concern was added, so they belong here rather than on #1240.

### 1. `checkRecordIds()` does not fire on exactly-repeated ids, and three of its seven sections have no reader-level backstop

`checkRecordIds()` (`R/project-excel.R:1476`) defers the abort to `.canonicalizeId()`, which deliberately keeps only canonicals reached from more than one *distinct* pre-image (`R/definition-id.R:61-70`):

```r
clashing <- clashing[vapply(
  clashing,
  function(c) length(unique(ids[canonical == c])) > 1L,
  logical(1)
)]
```

So two `Scenarios` rows both spelled `Aciclovir_iv` — one distinct pre-image — pass the guard untouched.

This is the limit you named in the resolved thread above, and for four of the seven sections your justification holds exactly as written: `.parseExcelDataCombinedSheet()` merges the twins into one combination, the 5.x PI path merges the task rows, and `.warnDuplicatePlotIds()` reports the reuse for `plotConfiguration` / `plotGrids`.

The three sections the reply did not cover are `scenarios`, `individuals` and `populations`. They have no merging reader and no duplicate warning — `.warnDuplicatePlotIds()` is called only at `R/project-excel.R:2939-2940`, for the two plot sections. Each is keyed by its own id downstream (`result[[id]] <- indiv`, `R/individuals.R:58`; `result[[id]] <- app`, `R/applications.R:283`; `result[[id]] <- task`, `R/parameter-identification.R:577`), so a duplicated row lands in precisely the case `checkRecordIds()`'s own comment describes:

> Two records whose ids canonicalize alike therefore become one definition, built from the last of them, and the earlier record is gone without a word.

Copy-pasting a scenario row in Excel and forgetting to rename it is the ordinary way to produce this. A `duplicated()` check on the canonical ids of those three sections — warn, matching `.warnDuplicatePlotIds()`'s contract, rather than abort — would close it without disturbing the quoted-twin behaviour you deliberately preserved.

### 2. Unquoting is scoped to ids and references; the single-value cells carrying *paths* are still read verbatim

`.canonicalizeProjectJsonIds()` unquotes ids and references, and its comment gives the general reason:

> a workbook quotes a single-value cell as readily as a list one (`"AciclovirPVB"`), and the quotes are the cell's syntax, not part of the name

That reasoning is not specific to ids — it is a property of the cell. But the path-carrying cells are read before canonicalization runs and never pass through `.stripWrappingQuotes()`:

- `dataFile <- prop("dataFile")` (`R/project-excel.R:2575`). A `Property` row holding `"Aciclovir.xlsx"` fails the `file.exists(dataFilePath)` check further down, and the whole `observedData` section is dropped with only the `esqlabsR_importSkippedObservedData` warning — which reads as "your data file is missing", not "your cell is quoted".
- `modelFile = as.character(row[["ModelFile"]])` (`R/project-excel.R:2477`). A quoted cell reaches `sc$modelFile` verbatim and survives the import, failing later at `runScenarios()` where the connection to the workbook cell is gone.

`dataFolder` and `dataImporterConfigurationFile` are read the same way. Routing these through `.stripWrappingQuotes()` at their read sites would extend the property this PR establishes to the cells that also depend on it.


---

PR #1236, opened 2026-08-06
Store a PI member id verbatim, and match an observed-data id either way

Felixmil:
Two authoring entrypoints skipped the id canonicalization every other `add*()` applies, so hand-authoring a project and importing the same project from Excel filed the same definition under different ids. Found by a migration test that built one project both ways and diffed the trees.

Review then showed that the two halves needed opposite answers, so this PR now does two different things rather than one uniform one.

## Observed data: canonicalize on add, match either spelling on remove

`addObservedData()` takes an `entry` list rather than an `id` argument, and the id inside it was written verbatim, so a declared `id = "TestProject_TimeValuesData"` produced a definition file carrying capitals — the case-collision the canonicalization rule exists to prevent on a case-insensitive filesystem.

An observed-data id genuinely does name a definition file, so the add side canonicalizes through the warning-emitting `.canonicalizeId()`, and the rewrite is reported.

Review caught the other half: `.removeObservedData_impl()` still matched verbatim, so you could not remove by the id you had just typed. It now keeps the verbatim pass and retries against the canonical spelling only when that misses. Verbatim stays first so a *derived* id — a file basename spelled exactly as its source spells it, or a programmatic `DataSet` name — still wins its own match. A real miss still warns under the caller's spelling.

## PI members: no door canonicalizes them

The original commit canonicalized the `id =` arguments of `addPIParameter()` and `addPIOutputMapping()`, and `.removePIMember()` with them. Review showed that made things worse, not better: `addPITask()` stores an inline `PIParameter(id = ...)` verbatim, and because canonicalization lowercases, **no input string canonicalizes to anything containing an uppercase letter** — so a member stored as `Lipophilicity` became unremovable by any id at all, and `addPIParameter(..., overwrite = TRUE)` appended a duplicate instead of replacing it.

The transform is therefore removed from all three sites. A PI member id is now stored exactly as every door is given it. The reasoning:

- **A member id is not a definition-file id.** It labels a record inside its task's JSON; nothing derives a filename from it, so the hazard that justifies `.canonicalizeId()` elsewhere does not exist here. The codebase already draws this line: a parameter-set *entry* is matched verbatim on `containerPath` + `parameterName`, while only the *set* id, which does name a file, is canonicalized.
- **Canonicalizing the inline ids would not have restored the invariant.** `.parsePIParameters()` and the new-layout Excel reader also store member ids verbatim, so a task loaded from a hand-edited definition file would stay unremovable. Closing that needs canonicalization on load, which is the round-trip change this stack deliberately defers.
- **A lookup fallback would fix removal only**, leaving a mixed canonical/verbatim section and needing the two-pass match at all three lookups to also fix the `overwrite` duplicate.

The PR's original motive did not survive checking: the legacy 5.x importer canonicalizes via `.uniqueImportedId()` because it *coins* an id from a free-text parameter name, not because members are keyed canonically.

## Consequence worth knowing

Because PI member ids are no longer canonicalized on the authoring side, an Excel-imported project and a hand-authored one can spell the same PI member differently — the importer coins `lipophilicity`, `addPIParameter(id = "Lipophilicity")` stores `Lipophilicity`. That is a spelling difference in a label, not a broken reference, and both projects run identically. Making the two agree needs canonicalization on load, which is the deferred change above.

## Tests

Removal by the typed spelling and by the canonical spelling, for both `addPIParameter()`-supplied and `addPITask()`-inline members; for observed data, the declared, derived and programmatic kinds, so the verbatim-first rule is pinned rather than assumed.

`@param id` docs updated on both families to describe the matching rule that now holds.

### DIFF
--- a/R/observed-data.R
+++ b/R/observed-data.R
@@ -636,11 +636,12 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
 #' @param id Character vector of ids. An observed-data id is the declaration's
 #'   `id` field, or, when it declares none, comes from the data source itself
 #'   (the `DataSet` name of an unsaved programmatic source, or a file basename
-#'   for a file-based source). Either way it is matched exactly as the
-#'   declaration stores it: a declared `id` is stored canonicalized (the form
-#'   [addObservedData()] reported back to you), a derived one exactly as its
-#'   source spells it. A saved programmatic source that declares no `id` is
-#'   matched by its `<name>.pkml` basename (see the note above).
+#'   for a file-based source). Each is matched as the declaration stores it and,
+#'   failing that, as [addObservedData()] would have canonicalized it, so a
+#'   derived id is found by the source's own spelling and a declared one by
+#'   either the form you authored or the canonical form you were given back. A
+#'   saved programmatic source that declares no `id` is matched by its
+#'   `<name>.pkml` basename (see the note above).
 #' @returns The `project` object, invisibly.
 #' @export
 #' @family observedData
@@ -666,6 +667,13 @@ removeObservedData <- function(project, id) {
   dropIdx <- integer()
   programmaticNames <- character()
   missingIds <- character()
+  entriesWithId <- function(key) {
+    which(vapply(
+      observedData,
+      function(e) identical(.observedDataEntryIdOrNA(e), key),
+      logical(1)
+    ))
+  }
   for (one in id) {
     if (one %in% names(private$.programmaticDataSets)) {
       programmaticNames <- c(programmaticNames, one)
@@ -679,11 +687,17 @@ removeObservedData <- function(project, id) {
       dropIdx <- c(dropIdx, matchIdx)
       next
     }
-    matchIdx <- which(vapply(
-      observedData,
-      function(e) identical(.observedDataEntryIdOrNA(e), one),
-      logical(1)
-    ))
+    matchIdx <- entriesWithId(one)
+    if (length(matchIdx) == 0L) {
+      # An id this section stores is one of two kinds, so the match takes two
+      # passes rather than one transform. A derived id is the source's own
+      # string (a `file` basename, a `DataSet` name), which only the verbatim
+      # pass above finds; a declared id is stored the way `addObservedData()`
+      # canonicalized it, so the id as the user authored it needs the same
+      # transform to reach it. Verbatim first, so a derived id that another
+      # entry's declared id happens to canonicalize to still wins its own match.
+      matchIdx <- entriesWithId(.silentlyCanonicalized(.canonicalizeId(one)))
+    }
     if (length(matchIdx) == 0L) {
       missingIds <- c(missingIds, one)
       next
--- a/R/observed-data.R
+++ b/R/observed-data.R
@@ -405,12 +405,15 @@ getObservedDataNames <- function(project) {
 #'
 #' `id` names the declaration itself: it becomes the declaration's file under
 #' the project's definitions folder (`definitions/observed-data/<id>.json` in
-#' the default layout) and is the id [removeObservedData()] matches on. Left
-#' out, the `file` basename serves as both. It is not the name the data is
-#' known by: each imported [`ospsuite::DataSet`] carries the name its source
-#' gives it (the data-set name in an Excel sheet, the name inside a PKML file),
-#' and that name, not the declaration's id, is what a `dataCombined` entry
-#' references.
+#' the default layout) and is the id [removeObservedData()] matches on. It is
+#' canonicalized like every other definition id (lowercased and made a safe
+#' filename segment), and you are told when that changes it. Left out, the
+#' `file` basename serves as both, kept exactly as the source spells it.
+#'
+#' An id is not the name the data is known by: each imported
+#' [`ospsuite::DataSet`] carries the name its source gives it (the data-set name
+#' in an Excel sheet, the name inside a PKML file), and that name, not the
+#' declaration's id, is what a `dataCombined` entry references.
 #'
 #' A `DataSet` you pass lives in the R session until you save. On
 #' [saveProject()] it is written to a PKML file named `<DataSet name>.pkml`
@@ -551,6 +554,16 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
         "An observedData entry's {.field id} must be a single non-empty string."
       )
     }
+    # A declared id names a definition file, so it runs through the same
+    # canonicalizer every other section's `add*` runs its id through: two ids
+    # differing only in case would otherwise be two declarations mapping to one
+    # file on a case-insensitive filesystem. Only a declared id is rewritten; an
+    # id derived from the `file` basename or a programmatic `DataSet` name is the
+    # source's own string, which `.validateObservedDataId()` rejects rather than
+    # reshapes.
+    if (!is.null(entry[["id"]])) {
+      entry[["id"]] <- .canonicalizeId(entry[["id"]])
+    }
     # Config entries are keyed by the id `removeObservedData()` matches on: the
     # entry's own `id` when it declares one, else its `file` basename. That id
     # also names the definition file, so reject an unsafe one here rather than
@@ -623,9 +636,11 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
 #' @param id Character vector of ids. An observed-data id is the declaration's
 #'   `id` field, or, when it declares none, comes from the data source itself
 #'   (the `DataSet` name of an unsaved programmatic source, or a file basename
-#'   for a file-based source). Either way it is matched verbatim, not
-#'   canonicalized. A saved programmatic source that declares no `id` is matched
-#'   by its `<name>.pkml` basename (see the note above).
+#'   for a file-based source). Either way it is matched exactly as the
+#'   declaration stores it: a declared `id` is stored canonicalized (the form
+#'   [addObservedData()] reported back to you), a derived one exactly as its
+#'   source spells it. A saved programmatic source that declares no `id` is
+#'   matched by its `<name>.pkml` basename (see the note above).
 #' @returns The `project` object, invisibly.
 #' @export
 #' @family observedData
--- a/R/parameter-identification.R
+++ b/R/parameter-identification.R
@@ -1624,8 +1624,9 @@ removePITask <- function(project, id) {
   invisible(self)
 }
 
-# Canonicalize the scenario references on a PIParameter record (its `id` is a
-# free label the PI run uses, not a definition-file id, so it is left as-is).
+# Canonicalize the scenario references on a PIParameter record. Only the
+# references: a record handed to `addPITask()` keeps its own `id` verbatim,
+# where `addPIParameter()` canonicalizes the `id` it is given.
 #
 # @keywords internal
 # @noRd
@@ -1711,7 +1712,8 @@ removePITask <- function(project, id) {
 #'   exist in `scenarios` definitions.
 #' @param minValue,maxValue,startValue Numeric scalars.
 #' @param units Optional character scalar.
-#' @param id Optional character scalar; auto-generated as
+#' @param id Optional character scalar, canonicalized the same way
+#'   [addPITask()] canonicalizes a task id; auto-generated as
 #'   `<task>_param_<N>` when absent.
 #' @param overwrite Logical scalar. When `FALSE` (default), an explicit `id`
 #'   that already exists in the task aborts. When `TRUE`, the existing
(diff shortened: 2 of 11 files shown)


R/parameter-identification.R
```diff
@@ -1862,6 +1870,9 @@ removePIParameter <- function(project, task, id) {
 # @keywords internal
 # @noRd
 .removePIMember <- function(self, private, task, id, field, label) {
+  # The add side canonicalizes a supplied id, so the removal has to run the same
+  # transform or an id typed the way it was first added would not find it.
+  id <- .canonicalizeId(id)
```
Felixmil: Canonicalizing here makes a member id supplied through `addPITask()` unreachable, because that door still stores ids verbatim (as the corrected comment on `.canonicalizePIParameterRefs()` now says).

Before this line both sides matched verbatim, so a task built as

```r
addPITask(project, id = "t", scenarios = "s", parameters = list(
  PIParameter(id = "Lipophilicity", ...)
))
```

could be undone with `removePIParameter(project, "t", "Lipophilicity")`. Now that call canonicalizes to `"lipophilicity"`, misses the stored `"Lipophilicity"`, and warns not-found — and no other spelling reaches it, so the member cannot be removed at all.

The add side has the mirror of the same mismatch. `.addPIParameter_impl()` canonicalizes the incoming `id` before `existingIdx <- which(existingIds == id)`, so `addPIParameter(project, "t", id = "Lipophilicity", overwrite = TRUE)` does not see the verbatim entry and appends a second one; the task ends up holding both `"Lipophilicity"` and `"lipophilicity"`. `.validatePI()`'s duplicate check won't flag it either, since the two strings differ.

The PR body calls the verbatim inline-record id known residue left for a follow-up, which is a fair call on its own — but this commit turns it from a cosmetic inconsistency into a functional regression on removal and overwrite. Two ways out that don't touch `addPITask()`'s round-trip semantics:

- canonicalize both sides at lookup time — compare `.silentlyCanonicalized(.canonicalizeId(x))` over `existingIds` as well as over the incoming id, here and in the two `add*` impls; or
- try the verbatim match first and fall back to the canonicalized one.

---

R/observed-data.R
```diff
@@ -551,6 +554,16 @@ addObservedData <- function(project, entry, overwrite = FALSE) {
...
     }
+    # A declared id names a definition file, so it runs through the same
+    # canonicalizer every other section's `add*` runs its id through: two ids
+    # differing only in case would otherwise be two declarations mapping to one
+    # file on a case-insensitive filesystem. Only a declared id is rewritten; an
+    # id derived from the `file` basename or a programmatic `DataSet` name is the
+    # source's own string, which `.validateObservedDataId()` rejects rather than
+    # reshapes.
+    if (!is.null(entry[["id"]])) {
+      entry[["id"]] <- .canonicalizeId(entry[["id"]])
```
Felixmil, line 565: This is the inverse of the gap closed in `.removePIMember()` in the same commit: the add side canonicalizes, the remove side does not.

`.removeObservedData_impl()` still resolves each id with `identical(.observedDataEntryIdOrNA(e), one)`, a verbatim comparison. So

```r
addObservedData(project, list(id = "MyObsData", type = "pkml", file = "x.pkml"))
removeObservedData(project, "MyObsData")
#> Warning: observedData entry "MyObsData" not found; no-op.
```

The user is warned at add time that the id was rewritten, and the updated `@param id` docs describe the behaviour, so it isn't silent — but requiring the caller to remember the canonical form for one section while `removePITask()`, `removeOutputPath()`, and now `.removePIMember()` all accept the id as typed is the kind of per-section rule this commit is otherwise removing.

A blanket `.canonicalizeId()` on the remove side would be wrong here: a derived id is stored exactly as its source spells it (`"MyObs.pkml"`, `"MyProgSet"`), and canonicalizing the search key would break those matches — the very case the new test asserts. A fallback covers both without that cost: keep the verbatim pass, and before pushing an id onto `missingIds`, retry the match against `.silentlyCanonicalized(.canonicalizeId(one))`. Derived ids keep matching verbatim on the first pass; a declared id typed the way it was authored matches on the second.

---

Felixmil (commented):
Automated code review (medium depth) of `e232eb4`, plus the surrounding stack range `ffe7a49..HEAD`, since the branches share a base.

The core change is right: the two entrypoints really were the odd ones out, and running a declared id through the warning-emitting `.canonicalizeId()` rather than the silent importer variant is the correct choice for a hand-authoring call. Test coverage is good — the "leaves a derived id and a `DataSet` name as they are" case is exactly the one a blunter fix would have broken.

Two findings, both about the *other* half of each pair — the id lookup that did not move with the id write. They're inline. In short:

- **`.removePIMember()` canonicalizing is a regression for `addPITask()`-supplied member ids.** Those are still stored verbatim, so a member added as `"Lipophilicity"` through an inline `PIParameter()` record can no longer be removed by any spelling, and `addPIParameter(..., overwrite = TRUE)` appends a duplicate rather than replacing it. The PR body treats the verbatim inline id as cosmetic residue; this commit makes it load-bearing.
- **`addObservedData()` canonicalizes a declared id but `.removeObservedData_impl()` matches verbatim,** so `removeObservedData()` with the id you typed warns not-found. Documented, and not silent, but it's the asymmetry this commit sets out to remove.

Both are fixable at the lookup rather than at the write, so neither needs the round-trip change you deliberately deferred.

**Out of this diff, for whoever picks up #1230:** `setScenario()` guards a `Scenario` record with `.assertScenarioRecordAlone(names(dots))`, which sees nothing when the extra argument is positional — `setScenario(project, <record>, "Other.pkml")` silently discards the field instead of aborting, where `addScenario()`'s `match.call()`/`missing(modelFile)` construction catches it. A partially-named call also renders an empty `{.field }` in the abort text. Not raised there since I was triggered on this PR.

Nothing else in the range needed changing: `.assignPIBounds()` preserves the original ordering logic exactly, `.definitionKindNames()` covers all twelve previously-explicit sections, `.compareJsonToExcel()`'s treeless parse and the new normalizations are symmetric across both sides, the Excel steady-state export condition matches `.scenarioToJson()`'s rule, and `outputDir`'s new default equals the old `pcDir`.

---

Felixmil:
Reviewed this commit against the current stack tip. The add side now canonicalizes, but three matching sides still compare verbatim, so an id that canonicalization rewrites goes in under one spelling and is looked up under another.

**1. A record's id is stored verbatim, so its removal can't find it** — `R/parameter-identification.R:1875`

`.removePIMember()` canonicalizes the id it is given, but `addPITask()` stores a `PIParameter` / `PIOutputMapping` record's own `id` verbatim — the comment at `R/parameter-identification.R:1627` states this is deliberate — and `.parseExcelPIRows()` likewise copies the exported `id` column through unchanged.

```r
addPITask(project, "task1", parameters = list(PIParameter(id = "Lipophilicity", ...)))
removePIParameter(project, "task1", "Lipophilicity")
#> looks for "lipophilicity", warns PIMemberNotFound, no-op
```

The parameter is unremovable by the id it was created with.

**2. `overwrite = TRUE` appends instead of replacing** — `R/parameter-identification.R:1800-1802`

Same asymmetry on the add path: `addPIParameter()` / `addPIOutputMapping()` canonicalize the supplied `id`, then match it against `existingIds` harvested from records that may hold a verbatim id. `overwrite = TRUE` finds no match and appends, leaving `MyParam` and `myparam` side by side; the non-overwrite clash check misses for the same reason. `.validatePI()` won't flag it either — `.checkNoDuplicates()` compares the raw strings (`R/parameter-identification.R:690`), and those two differ.

**3. `removeObservedData()` matches verbatim against a canonicalized id** — `R/observed-data.R:565` vs `R/observed-data.R:656`

`.addObservedData_impl()` canonicalizes a *declared* id; `.removeObservedData_impl()` runs `.assertIdVector(id)` but never `.canonicalizeId(id)`, matching with `identical(.observedDataEntryIdOrNA(e), one)`.

```r
addObservedData(project, list(id = "PlasmaData", file = "obs.csv"))
removeObservedData(project, "PlasmaData")
#> "observedData entry 'PlasmaData' not found; no-op."
```

Worth noting this is the reverse of the derived-id case: `.observedDataEntryIdOrNA()` derives an id from `basename(file)` or a programmatic `name` *without* canonicalizing (`R/definition-files.R:989`), so removal works for a derived id and fails for a declared one. `overwrite = TRUE` misses across the same boundary and appends a second declaration.

`.removePITask_impl()` (`R/parameter-identification.R:1611`) does canonicalize on the way in — that's the shape the other three want.

Verified by reading; no R in this environment, so the suite was not run.

---

Felixmil (commented):
Found while code-reviewing the stack from #1238. Two add/remove id asymmetries, both in this PR's code. They lean opposite ways, so they need different fixes.

### 1. Observed data canonicalizes on add but matches verbatim on remove

`.addObservedData_impl()` canonicalizes a declared `id` (`R/observed-data.R:565`), while `.removeObservedData_impl()` matches the id it is given as-is (`R/observed-data.R:684`) — it runs `.assertIdVector(id)` but never `.canonicalizeId(id)`. Since canonicalization lowercases:

```r
addObservedData(project, list(type = "excel", id = "MyData", file = "..."))
# stored under `mydata`
removeObservedData(project, "MyData")
# no match -> warns "not found", no-op
```

The comment directly above the add-side canonicalization says config entries are "keyed by the id `removeObservedData()` matches on", which now holds only for ids that were already canonical.

### 2. PI members canonicalize on remove but are stored verbatim by `addPITask()`

The mirror image. `.removePIMember()` canonicalizes (`R/parameter-identification.R:1875`) on the stated grounds that "the add side canonicalizes a supplied id". That is true of `addPIParameter()` / `addPIOutputMapping()`, but not of `.addPITask_impl()`: `.canonicalizePIParameterRefs()` and `.canonicalizePIOutputMappingRefs()` touch only `scenarios`, and the comment above them records the intent that "a record handed to `addPITask()` keeps its own `id` verbatim".

So a member introduced through `addPITask()` with a non-canonical id is stored with that id, and the remove side canonicalizes before matching:

```r
addPITask(project, "myTask", parameters = list(PIParameter(id = "scalarWeight", ...)))
# stored as `scalarWeight`
removePIParameter(project, "myTask", "scalarWeight")
# canonicalized to `scalarweight` -> no match -> warns "not found"
```

Because canonicalization lowercases, no input string canonicalizes to anything containing an uppercase letter, so that member cannot be removed by id at all — the warning is the only outcome available.

Worth deciding which side is authoritative for the inline-record path: either canonicalize the inner `id` in `.addPITask_impl()` (which would make the verbatim-id comment obsolete), or have `.removePIMember()` fall back to a verbatim match when the canonical one misses.

---

R/parameter-identification.R
Felixmil: Confirmed, and the overwrite half too. Fixed by taking the third option: **no door canonicalizes a PI member id any more**, so the removal matches exactly what was stored. `.canonicalizeId()` is gone from `.addPIParameter_impl()`, `.addPIOutputMapping_impl()` and `.removePIMember()`.

Why that rather than canonicalizing `addPITask()`'s inline ids, or a verbatim fallback at lookup:

- A member id is not a definition-file id. It labels a record *inside* its task's JSON; nothing derives a filename from it, so the case-insensitive-filesystem hazard that justifies `.canonicalizeId()` elsewhere does not exist here. The precedent is already in the codebase: a parameter-set *entry* is identified by `containerPath` + `parameterName` and matched verbatim on both sides, while only the *set* id, which does name a file, is canonicalized.
- Canonicalizing the inline ids would still not restore the invariant. `.parsePIParameters()` and the new-layout Excel reader (`.parseExcelPIRows()`) also store member ids verbatim, so a task loaded from a hand-edited definition file carrying `"Lipophilicity"` would remain unremovable. Closing that would mean canonicalizing on load, which rewrites ids in existing projects and is exactly the round-trip change deferred here.
- A fallback at the lookup fixes removal but leaves the section holding a mix of canonical and verbatim ids, and needs the same two-pass match repeated at all three lookups to fix the `overwrite = TRUE` duplicate. Dropping the transform fixes both with a deletion.

The stated motive for the add-side canonicalization does not survive checking either: the legacy 5.x importer canonicalizes via `.uniqueImportedId()` because it is *coining* an id from a free-text parameter name, not because members are keyed canonically.

The comment on `.canonicalizePIParameterRefs()` now records that no door reshapes a member id, and the four `@param id` entries say the id is stored, and matched, exactly as given.

Tests: a member added through each door is removable by that spelling; the canonical spelling of a stored id correctly misses; `addPIParameter(overwrite = TRUE)` replaces an inline-added mixed-case member instead of appending a second one. The weight round-trip test keeps its `scalarWeight` mapping id, so the round trip now proves a non-canonical id survives it. Those tests fail 8 without the change.

R/observed-data.R
Felixmil: Agreed, and fixed the way you suggest: `.removeObservedData_impl()` keeps its verbatim pass and, only when that misses, retries against `.silentlyCanonicalized(.canonicalizeId(one))`.

Verbatim stays first for the reason you give. A derived id is the source's own string, so it is the only spelling that finds a `file` basename or a `DataSet` name; ordering it first also means a derived id that some other entry's declared id happens to canonicalize to still wins its own match rather than being shadowed.

The add side keeps canonicalizing. Unlike a PI member id, an observed-data id really does name a definition file, so the rewrite there is right; it was only the lookup that had not moved with it.

`missingIds` still collects the id as typed, so a genuine miss warns under the spelling the caller used rather than the canonical form the second pass tried. The `@param id` docs now describe the two-pass rule instead of claiming an exact match.

Tests cover all three id kinds in one section: a declared id removable by the spelling it was authored with (and, as before, by the canonical one), a file-derived id, and a programmatic `DataSet` name, plus the not-found warning naming the typed id. They fail 3 without the change.

PR #1237, opened 2026-08-06
Import protein ontogenies as one entry per ontogeny

Felixmil:
The `proteinOntogenies` field of an individual and of a population had two on-disk encodings depending on which entrypoint wrote it: the Excel importer joined the entries into one comma-separated string, the authoring API stored a character vector. Both load, and the two agreed for a single ontogeny, diverging from two upward.

```
definitions/individuals/child.json
  imported: "CYP3A4:CYP3A4,CYP2D6:CYP2C8"
  authored: ["CYP3A4:CYP3A4", "CYP2D6:CYP2C8"]
```

The consequence was not cosmetic. An Excel round trip returned the string form, so a hand-authored project carrying more than one ontogeny could never report `excel_in_sync = TRUE`: `projectStatus()` listed `individuals` and `populations` as drifted immediately after an export that should by definition leave both sides in sync.

## Excel bridge

`.excelProteinOntogenies()` returns one entry per ontogeny instead of a joined string, matching what `addIndividual()` and `addPopulation()` store. The single-cell `Protein Ontogenies` spelling is split too rather than passed through verbatim, so both workbook layouts land on the same shape. `auto_unbox = TRUE` on every serializer means a single ontogeny still writes as a scalar string on both paths, unchanged.

The writer needed no change: `.formatOntogeniesToCell()` already splits its input before joining, so it handled a vector before this.

## Compatibility

A project already on disk in the string form still loads: `.parseIndividuals()` and `.parsePopulations()` coerce with `as.character(unlist(val))`, which keeps a joined string as a length-1 vector, and every consumer — including the runtime builder `.readOntogeniesFromList()` — routes through `.splitProteinOntogenies()`. Both encodings resolve to the same `MoleculeOntogeny` objects. Tests in `test-individuals.R` and `test-populations.R` rewrite a real definition file to the old form, reload, and assert both ontogenies are still named.

## Tests

The round-trip test sets two ontogenies on the example project, exports, re-imports, and asserts the value comes back as authored and that `projectStatus()` no longer lists `individuals` or `populations` as drifted. It fails four assertions without the fix. The status assertion is per-section rather than on `excel_in_sync`, because the example project has independent round-trip drift measured before this change.


Felixmil:
Reviewed this diff — no findings in it. What I checked:

- `.excelProteinOntogenies()` returning one entry per ontogeny matches what `addIndividual()` / `addPopulation()` store, so the two encodings genuinely converge rather than just agreeing at length 1.
- Backward compatibility holds. `.parseIndividuals()` / `.parsePopulations()` coerce with `as.character(unlist(val))`, which keeps a comma-joined string as a length-1 vector, and every consumer routes through `.splitProteinOntogenies()`. The two new "still loads" tests pin this.
- `auto_unbox = TRUE` keeps a single ontogeny writing as a scalar on both paths, so the single-ontogeny case stays byte-identical.
- Dropping the `.naToNull()` guard is safe: `.splitProteinOntogenies()` filters on `!is.na(entries) & nzchar(entries)` (`R/utils.R:191`), so an `NA` cell yields `character(0)` and the `length(single) > 0L` test falls through to the pair spelling exactly as the old `.naToNull()` path did.
- Asserting per section on `projectStatus()` rather than on `excel_in_sync` is the right call given the example project's independent round-trip drift, and it still fails without the fix.

Findings from this review pass landed on two PRs in this one's base chain, which own the code they concern: #1236 (three id-canonicalization asymmetries) and #1230 (two silently-dropped unnamed fields).

---

PR #1238, opened 2026-08-06
Declare the observed-data and PI workbooks when exporting a JSON project

Felixmil:
A project authored in JSON lost its observed-data declaration when exported to Excel, silently, and re-importing the exported workbooks produced a project failing validation with two critical errors. This was the most serious of four findings from a migration test of the legacy fixture.

`exportProjectToExcel()` wrote the `Project.xlsx` property table from the project's container blocks alone. An Excel-imported project carries `dataFile` / `dataImporterConfigurationFile` / `parameterIdentificationFile` there, supplied by the legacy property sheet. A JSON-first project does not: `addObservedData()` records the source inside the definition, which the exporter never read. Exporting the same content from each origin gave 18 properties from the imported project and 15 from the authored one.

Re-importing the authored workbooks gave `observedData: 1 -> 0` with no warning at all, because `.parseExcelObservedData()` returns early when no `dataFile` is declared — so the missing-file warning path was unreachable. The loss then cascaded into dangling PI output mappings:

```
Validation report: 2 critical errors, 9 warnings.
crossReferences
  x [Invalid Reference] PI task 'aciclovirfit', outputMapping '...' references undefined observed data '...'
observedData
  ! [Data] No observedData defined
```

## Excel bridge

`.withDeclaredWorkbookProps()` fills only the rows the container blocks leave blank, judged with the existing `.isBlankCell()` so an absent, `NA`, or whitespace value counts as unset:

- `dataFile` and `dataImporterConfigurationFile` from the first `excel` observed-data declaration's `file` / `importerConfiguration`.
- `parameterIdentificationFile` as the conventional `ParameterIdentification.xlsx` when the project has PI tasks, so the exported project declares the workbook it actually wrote. Previously this was masked by the importer's filename fallback — PI survived, but only because of that fallback.

Each derived row carries the description a legacy 5.x property sheet writes. A container value always wins, so an Excel-imported project exports exactly the rows it came in with.

## Several declarations

The property table can name one experimental-data workbook. The first `excel` source is written and every declaration it cannot carry — the remaining `excel` sources, and every `pkml` / `script` / `programmatic` one — is named in a warning (`messages$exportUndeclaredObservedData`, class `esqlabsR_exportUndeclaredObservedData`). Warn-and-write-first beats aborting: the export is an interchange operation someone runs to get a workbook, and refusing to produce one because a second data source exists is worse than producing an honest annotated one. Nothing fires when the written `dataFile` is the project's only declaration.

## Behaviour the fix makes visible

Nine existing tests began emitting a real warning the bug had made unreachable: exporting to a scratch directory writes a `dataFile` whose workbook is not beside it, so the re-import reports the data file missing and skips the observed data. That is the intended outcome — silence becomes a visible warning — so the suite asserts it rather than suppressing it broadly. `excelRoundTrip()`'s muffler also drops the classed `esqlabsR_importSkippedObservedData` warning and is renamed `.muffleRoundTripDataWarnings`; the two tests calling the importer directly assert `"was not found"` explicitly.

## NEWS

The existing bullet said "The Excel round-trip covers every project section except observed data", which is wrong after this change. Amended in place (same dev cycle) to say the round trip covers every section and that observed data travels as a declaration only: the measurements still get no sheet, but the workbook and importer-configuration names round-trip, with the several-declarations behaviour stated.

## Known residue

`projectStatus()` on a hand-authored project with observed data still reports out of sync, but the drift has moved from three definition sections to the container: the export now derives three properties the in-memory project does not hold, so `filePaths` and `excel` differ by construction. Closing that means either having the comparison ignore derived properties or having authoring record them, which is a design decision rather than a bug fix.


Felixmil (commented):
Automated code review, triggered by this PR going ready-for-review. The review ran across the stack (against `json-based-project`), so the summary below covers more than this PR's own diff.

### This PR

No new defects found in the one commit this PR carries. The derivation in `.withDeclaredWorkbookProps()` reads correctly: `.isBlankCell()` gives the container-wins rule uniformly over absent / `NA` / whitespace, the unrepresented-declarations filter matches on `type` **and** `file` so it cannot drop the row it just wrote, and `parameterIdentificationFile` is gated on the project actually having PI tasks.

One thing checked and deliberately not raised as a defect: the export derives three properties the in-memory project does not hold, so `.compareJsonToExcel()` sees them on the Excel side only and `projectStatus()` reports `filePaths` / `excel` drift right after a clean export. That is the "Known residue" section of the PR description, described there in the same terms, and the choice between having the comparison ignore derived properties and having authoring record them is a design call rather than something to fix in this PR. The only part not already stated is that the new round-trip test asserts the `individuals` / `populations` verdicts and so would not catch a regression in that container drift.

### Filed elsewhere in the stack

Four confirmed findings landed on the PR that owns the code, so they are not lost when the lower PRs merge first:

| PR | Finding |
|---|---------|
| #1230 | `setScenario(project, <Scenario record>, <positional>)` silently ignores the extra — `names(dots)` is `NULL` for an all-positional `...`, where the old `missing(modelFile)` caught it |
| #1233 | `.sortJsonSections()` builds its ordering key at jsonlite's default `digits = 4`, so records differing past the 4th decimal tie and keep source order — the false drift the sort exists to prevent |
| #1236 | Observed data canonicalizes an id on add but matches verbatim on remove |
| #1236 | PI members are stored verbatim by `addPITask()` but canonicalized by `.removePIMember()`, leaving such a member unremovable by any id string |

CI is green on everything reported so far; the three `R-CMD-Check` jobs were still running at review time, which is what the `unstable` mergeable state reflects.

---

PR #1239, opened 2026-08-07
Report a malformed parameter set instead of aborting on it

Felixmil:
`.validateParameterSets()` is the twin of the `.validateInitialConditions()` that #1228 hardened, sitting immediately above it in the same file, and it still carries all three of the defects that fix removed. The consequence is worse than a bad validation report: a parameter set hand-edited into a malformed shape aborts with an internal R error while the project is being *loaded*, so the project cannot be opened in order to be fixed. This is a follow-up to #1228 under the #1226 audit.

## Validation

- Each entry is normalized to a list once, before any field is read. Guarding only the two `vapply()` calls would have moved the abort rather than removed it, since `values <- lapply(set, function(e) e$value)` on the next line reads a field the same way.
- A field that does not hold exactly one value reads as `NA` instead of aborting `vapply()` with an internal length error. A hand-edited set can write `"containerPath": []` or `"containerPath": ["a", "b"]`, neither of which was reportable before.
- The duplicate-path check now skips the entries the critical-error check just flagged for having no path, so two path-less entries are no longer reported as both duplicating a path literally called `NA|NA`.

## Shared helper

The scalar-field reader that `.validateInitialConditions()` introduced was a local closure. Both validators now use one file-level `.scalarEntryField()` rather than carrying two copies of it forty lines apart. Net less code than duplicating it, and the initial-condition tests already cover the behaviour.

## Tests

Three tests in `tests/testthat/test-validation.R`, mirroring the three the initial-condition fix added: a non-scalar path field, an entry that is a bare string, and path-less entries not counting as duplicates. Each was written first and confirmed to reproduce the defect against the unfixed validator.

## Example

Before:

``` r
dir.create(dir)
invisible(file.copy(
  dir,
  recursive = TRUE
))

# A parameter set hand-edited so its entry is a bare string, not a record.
writeLines(
  '{"id": "aciclovir", "parameters": ["Organism|Liver|Volume"]}',
  file.path(dir, "definitions", "parameter-sets", "aciclovir.json")
)

project <- loadProject(dir)
#> Error in `e$containerPath`:
#> ! $ operator is invalid for atomic vectors
validateProject(project)
#> Error:
#> ! object 'project' not found
```

After:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/followup-a-paramset-validator", quiet = TRUE)
dir.create(dir)
invisible(file.copy(
  dir,
  recursive = TRUE
))

# A parameter set hand-edited so its entry is a bare string, not a record.
writeLines(
  '{"id": "aciclovir", "parameters": ["Organism|Liver|Volume"]}',
  file.path(dir, "definitions", "parameter-sets", "aciclovir.json")
)

project <- loadProject(dir)
validateProject(project)
#> Validation report: 1 critical error, 1 warning.
#> parameterSets
#>   ✖ [Missing Fields] Set 'aciclovir' in parameterSets contains empty parameter paths
#>   ! [Data Type] Set 'aciclovir' in parameterSets contains non-numeric values
#> 11 sections OK.
```

</details>

Part of #1226.


R/validation.R
```diff
@@ -691,17 +718,16 @@ validationSummary <- function(validationResults) {
...
-      function(e) as.character(e$parameterName %||% NA_character_),
-      character(1)
-    )
-    values <- lapply(set, function(e) e$value)
+    # An entry that is a bare string rather than a record
+    # (`["Organism|Liver|Volume"]`) has no fields to read; treating it as an
+    # empty record folds it into the missing-field error below instead of
+    # aborting on `$`. Normalized before any field is read, so the value read
+    # below is covered too.
+    entries <- lapply(set, function(e) if (is.list(e)) e else list())
```
Felixmil, line 733: Nit, non-blocking. This line and its comment are now the *second* thing duplicated verbatim between the two validators — `.validateInitialConditions()` carries an identical `entries <- lapply(set, function(e) if (is.list(e)) e else list())` at L833 with a near-identical four-line comment. The PR makes exactly this argument for extracting the field reader ("rather than carrying two copies of it forty lines apart"), and the normalization is forty lines apart in the other direction.

Folding it *into* `.scalarEntryField()` wouldn't work — `values <- lapply(entries, function(e) e$value)` needs the normalized list too, so it has to stay at the call site — but a sibling one-liner would:

```r
.entryRecords <- function(set) lapply(set, function(e) if (is.list(e)) e else list())
```

That also gives `.scalarEntryField()`'s `@param entries` a name to point at instead of "the normalization the callers apply", which currently only tells you the precondition exists, not where to find it. Worth having, because the precondition is load-bearing: `e[[field]]` on an atomic entry aborts with `subscript out of bounds`, i.e. the same class of failure this PR removes.

---

Felixmil (commented):
Reviewed the full diff (3 files, +88/-29): the `.scalarEntryField()` extraction, the entry normalization and `hasPath` duplicate filter in `.validateParameterSets()`, and the three new tests. No blocking findings — one non-blocking nit inline.

Things I checked and found sound:

- **The abort is real and the fix reaches it.** `loadProject()` → `.warnOnCrossReferenceErrors()` → `.runProjectValidation(project, sections = NULL)` runs *every* adapter, not just `crossReferences`, so a malformed parameter set genuinely bricks the load. The reprex is not a synthetic path.
- **`[[` is safe where `$` was not.** The definitions tree is read with `simplifyVector = FALSE` throughout (`project-lifecycle.R:90`, `definition-files.R:451`), so entries are plain lists and `e[[field]]` returns `NULL` for an absent name rather than erroring. `"x": []` → `list()` (length 0) and `"x": ["a","b"]` → length 2 both land on `NA`, as documented. Switching from `$` to `[[` also drops partial matching, which no producer of these entries relies on.
- **No `NA` leaks into the `hasPath` subscript.** `containerPaths != ""` is `NA` when the path is `NA`, but `!is.na(...) &` evaluates to `FALSE` first, so `hasPath` is never `NA` and `containerPaths[hasPath]` can't silently inject an `NA` element. With every entry filtered out, `paste(character(0), character(0), sep = "|")` is `character(0)`, so the dupe check no-ops rather than warning.
- **The three new tests fail against the unfixed validator and don't pass vacuously.** `expect_match(msgs, ..., all = FALSE)` reduces to `any(...)`, which is `FALSE` on `character(0)` — so an empty `critical_errors` fails rather than passing silently, which is the trap the `all = TRUE` default would have set.
- **No regression in the existing tests.** The four pre-existing `.validateParameterSets` tests all use entries with `""` halves, which the `hasPath` filter excludes from the dupe check but which the critical-error check still catches; none of them assert on a `NA|NA` warning.
- **Test level matches the precedent.** Unit tests on the validator rather than an end-to-end `loadProject()` test, same as the three #1228 added for `.validateInitialConditions()`.
- `@noRd` on the new helper means no `man/` or `NAMESPACE` regeneration is owed. `REFACTORING-LOG.md` entry is present and carries `(#1226)`.

CI: `pkgdown`, `test-coverage`, both codecov checks and the link check are green; the three `R-CMD-Check` jobs were still running at review time.

---

R/validation.R
Felixmil: Agreed, and the argument is fair — the PR body makes exactly this case for the field reader while leaving the normalization duplicated in the other direction.

Deferring rather than fixing: this round was scoped to blocking findings only, and this is a readability improvement over code that is now correct. Leaving the thread open so it stays visible.

One thing changed underneath it since you wrote this. `2b39a6a9` on this branch tightened `.scalarEntryField()` to require an actual length-1 character, because `"containerPath": [null]` parses to `list(NULL)` — length 1 — and `as.character()` rendered it as the literal string `"NULL"`, which is neither `NA` nor `""`, so it passed the missing-path check the function exists to raise. `true`, `42`, `{"a": 1}` and a one-element array behaved the same way. So the `@param entries` precondition you point at is now load-bearing for a second reason, which strengthens rather than weakens the case for giving it a name.

PR #1240, opened 2026-08-07
Attribute the Excel read aborts to the caller that named the workbook

Felixmil:
Importing a corrupt workbook attributed the abort to `.excelToProjectJson()`, a private helper the caller never named, and included an internal source location carrying the developer's own directory name. Neither of the two aborts in that function had any test or snapshot coverage, which is why the read/write extraction in #1233 could change them unnoticed. This is a follow-up under the #1226 audit.

## Error attribution

`.excelToProjectJson()` takes a trailing `.call` and opens with `rlang::local_error_call(.call)`, per the convention the authoring `_impl`s already follow. Both of its aborts are covered by the one argument, and no call site had to change.

`rlang::local_error_call(NULL)` was the alternative, and there are two precedents for it in this file. Both justify `NULL` by the helper being reached from several entrypoints with no meaningful caller to name, and that reasoning does not transfer here: on the only route where the prefix is visible at all it resolves to `importProjectFromExcel()`, a call the user literally typed, and that function's two sibling preconditions already abort under its own name with a snapshot asserting it. `NULL` would have left one of three preconditions rendering a bare `Error:` while the others named the function.

`.compareJsonToExcel()` is deliberately not threaded: its only production caller discards the prefix, so it would change nothing observable.

Two of the three cases originally suspected turned out not to be defects, and were confirmed by running them rather than by reading:

- The `fileNotFound` abort inside `.excelToProjectJson()` is unreachable from either entrypoint — `importProjectFromExcel()` checks the path itself first and already aborts under its own name. It is now correctly attributed anyway, but nothing changed for a user.
- The `projectStatus()` route never leaked. It re-raises the compare failure through `conditionMessage()`, which returns the message body without the ``Error in `fn`:`` prefix, so the helper's name was never visible there. Before and after are byte-identical.

## Messages

The corrupt-workbook text moves into `messages$excelProjectFileUnreadable()`. It is stored as an unglued template rather than built with `cliFormat()` or `format_message()`, because the path interpolated into it can itself contain braces and both helpers glue eagerly, which the raising call would then glue-parse a second time.

## Tests

- A new `expect_snapshot(error = TRUE)` on the confirmed live leak, with the corrupt workbook built inside the test under `withr::local_tempdir()`.
- The existing `projectStatus()` NA-status test moves from an `expect_warning()` pattern match to `expect_snapshot()`. No coverage is lost — the matched string is in the snapshot verbatim — and it now also guards against a doubled `✖` glyph, which naming the first bullet `"x"` reintroduces because `.projectSyncStatus()` re-raises the body as its own `"x"` bullet.
- `tests/testthat/helpers.R` gains `.redactProjectWorkbookPath()`. The existing `.redactTmpDir()` cannot do this job: it anchors on `tempdir()`, while the chained `readxl` error names the `/private`-prefixed symlink target.

## Example

Before:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/tmp-before-6e614308", quiet = TRUE)
src <- file.path(tempdir(), "excel-project")
dir.create(src)

# A Project.xlsx that is not a readable workbook.
writeLines("not a workbook", file.path(src, "Project.xlsx"))

importProjectFromExcel(
  file.path(src, "Project.xlsx"),
  outputDir = file.path(tempdir(), "imported"),
  overwrite = TRUE
)
#> Error in `.excelToProjectJson()` at tmp-before-6e614308/R/project-excel.R:375:3:
#> ! 
#>   '/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T//RtmpITki8U/excel-project/Project.xlsx'
#>   is not a readable Excel project file.
#> ℹ It must be a valid .xlsx workbook (the project's 'Project.xlsx').
#> Caused by error in `utils::unzip()`:
#> ! zip file '/private/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpITki8U/excel-project/Project.xlsx' cannot be opened
```

After:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/followup-b-excel-call-attribution", quiet = TRUE)
src <- file.path(tempdir(), "excel-project")
dir.create(src)

# A Project.xlsx that is not a readable workbook.
writeLines("not a workbook", file.path(src, "Project.xlsx"))

importProjectFromExcel(
  file.path(src, "Project.xlsx"),
  outputDir = file.path(tempdir(), "imported"),
  overwrite = TRUE
)
#> Error in `importProjectFromExcel()`:
#> ! 
#>   '/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T//RtmpdVnWwi/excel-project/Project.xlsx'
#>   is not a readable Excel project file.
#> ℹ It must be a valid .xlsx workbook (the project's 'Project.xlsx').
#> Caused by error in `utils::unzip()`:
#> ! zip file '/private/var/folders/_6/hdp78hfx2qg6415svlx5rb680000gn/T/RtmpdVnWwi/excel-project/Project.xlsx' cannot be opened
```

</details>

Part of #1226.


tests/testthat/helpers.R
```diff
@@ -429,6 +429,19 @@ importLegacyExcelProject <- function(
...
+#' `expect_snapshot()` of the unreadable-workbook error is stable across runs.
+#'
+#' `.redactTmpDir()` cannot do this job: it anchors on `tempdir()`, and the
+#' `readxl` error chained onto that abort names the path as `utils::unzip()`
+#' resolved it, which on macOS is the `/private`-prefixed symlink target and so
+#' does not start with `tempdir()`. Matching on the workbook basename instead
+#' catches both spellings. The `/` before the basename is required, so the bare
+#' `'Project.xlsx'` the advice bullet mentions is left intact.
+.redactProjectWorkbookPath <- function(lines) {
+  gsub("'[^']*/Project\\.xlsx'", "'<tmp-path>/Project.xlsx'", lines)
```
Felixmil: **This breaks Windows CI** — `R-CMD-Check / windows-latest (release)` is red on this head, and both of the PR's new snapshots are the only two failures ([job log](https://github.com/esqLABS/esqlabsR/actions/runs/31161932516/job/92814027422)):

```
── Failure ('test-project-excel.R:673:3'): importProjectFromExcel names itself when the workbook cannot be read ──
-   ! zip file '<tmp-path>/Project.xlsx' cannot be opened
+   ! zip file 'C:\Users\runneradmin\AppData\Local\Temp\Rtmpctzn7i\file12085f1b12b6\Project.xlsx' cannot be opened
```

The separator class is the whole problem: the pattern requires a literal `/` before the basename, but the chained `utils::unzip()` error names the path as Windows spells it, with `\`. So the match fails and the runner's real temp path lands in the snapshot. Both new snapshots fail this way (the second leaks `...\Example\Project.xlsx`).

Note the esqlabsR-authored line above it redacts fine on Windows — `cli`'s `{.path}` gets the R-normalized forward-slash path. It's only the third-party `unzip()` line that carries the native spelling, which is exactly the line this helper exists to reach.

`.redactTmpDir()` right above already handles this, by normalizing first (`tmp <- gsub("\\\\", "/", tempdir())`). Here the cheapest equivalent is to widen the separator:

```r
gsub("'[^']*[/\\\\]Project\\.xlsx'", "'<tmp-path>/Project.xlsx'", lines)
```

`[^']*` already spans backslashes, so the class is the only change needed, and the roxygen claim just above it still holds — a bare `'Project.xlsx'` with no separator is still left intact, which is what keeps the advice bullet readable.

---

tests/testthat/_snaps/project-excel.md
```diff
@@ -27,6 +27,17 @@
...
+# importProjectFromExcel names itself when the workbook cannot be read
+
+    Code
+      importProjectFromExcel(workbook, outputDir = withr::local_tempdir(), silent = TRUE)
+    Condition
+      Error in `importProjectFromExcel()`:
+      ! '<tmp-path>/Project.xlsx' is not a readable Excel project file.
+      i It must be a valid .xlsx workbook (the project's 'Project.xlsx').
+      Caused by error in `utils::unzip()`:
+      ! zip file '<tmp-path>/Project.xlsx' cannot be opened
```
Felixmil, line 39: Separate from the Windows redaction bug, and lower: both new snapshots bake in the `Caused by error in \`utils::unzip()\`: ! zip file '…' cannot be opened` chain, which is `readxl`/base R wording rather than anything this package controls. If readxl changes which reader it probes with, or base changes that string, CI goes red for a reason unrelated to the call attribution actually under test — and the failure will read as an attribution regression.

The attribution assertion is the ``Error in `importProjectFromExcel()`:`` line plus the two esqlabsR-authored lines; the parent chain is incidental. Worth considering a transform that also collapses the `Caused by` block to a stable placeholder, so the snapshot pins the surface you own and stays sensitive to the thing it exists to catch.

Not blocking — the Windows fix is the one that has to land.

---

R/project-excel.R
Felixmil, line 1: File-level, because the line is outside this PR's hunks — `.projectSyncStatus()` at `R/project-excel.R:1493`:

```r
cli::cli_warn(
  c(
    "Cannot compare the Excel configuration files to the project.",
    "x" = conditionMessage(compareError),
    ...
```

`conditionMessage()` returns text that already has `projectConfigPath` interpolated into it, and `cli_warn()` glue-parses each bullet. So a project directory containing a brace — `/home/user/{draft}/Project.xlsx` — gets a second interpolation pass, and a read-only `projectStatus()` call throws instead of warning.

This is pre-existing, and I'd normally leave it. Flagging it because this PR is what makes it relevant: the new `messages.R` comment states the single-interpolation invariant explicitly as the reason `excelProjectFileUnreadable()` is stored unglued, and the new `projectStatus()` snapshot pins this exact re-raise path. The message is now careful to be interpolated once, and then this call site interpolates it again.

Binding it first keeps the value out of the parser:

```r
compareMessage <- conditionMessage(compareError)
cli::cli_warn(
  c(
    "Cannot compare the Excel configuration files to the project.",
    "x" = "{compareMessage}",
    ...
```

I have not run this — no R in the review environment — so treat it as reasoned from `cli`'s documented inline-substitution behavior rather than reproduced. A brace in a project path is also not a common case; fine to defer to a follow-up under #1226.

---

Felixmil (commented):
Automated review round (medium). Scope: this PR's single commit `6d83d5d`, six files.

**Windows CI is red, and this PR's two new snapshots are the only failures.** `.redactProjectWorkbookPath()` requires a literal `/` before the workbook basename, but the chained `utils::unzip()` error spells the path the Windows way, so the runner's temp path leaks into both snapshots. Details and a one-line fix in the inline comment on `tests/testthat/helpers.R`. Worth noting this is not a flake and not a base-branch failure — macOS and Ubuntu are green, and the diff is what introduced it.

The change itself is sound otherwise. `.call` threading follows the `_impl` convention, and both call sites resolve correctly: from `importProjectFromExcel()` `caller_env()` names the entrypoint, and from `.compareJsonToExcel()` the prefix is discarded by `conditionMessage()` anyway, so leaving that one unthreaded is right. The reasoning in the description for preferring a threaded `.call` over `local_error_call(NULL)` holds up against the two precedents in the file — those helpers genuinely have no single meaningful caller, and this one does. Moving the text into `messages$excelProjectFileUnreadable()` unglued, with an unnamed header, is correct on both counts: the brace-safety argument is real, and leaving the header unnamed is what stops `.projectSyncStatus()` rendering a doubled `✖`.

Two lower-severity notes inline: the snapshots pin third-party `readxl`/base wording that this package does not control, and `.projectSyncStatus()` re-glues the already-interpolated message — pre-existing, but it contradicts the invariant this PR's own comment asserts.

One finding from this review round falls outside this PR — `.scalarEntryField()` in `R/validation.R` accepts any length-1 value and `as.character()`s it, so `"containerPath": [null]` reads as the literal string `"NULL"` and passes the missing-path check the function exists to raise. That belongs to `felixmil/parameter-sets-validator-1226`, this PR's base, so I have not commented on it here.

---

tests/testthat/helpers.R
Felixmil: Fixed in `d3d803fa` — the separator class, exactly as suggested. `windows-latest` is green on that head.

Worth recording why no tempdir-anchored redaction could have worked here, since `.redactTmpDir()` right above looks like it should apply: the two ends of the chained error do not agree on the path at all. R's `tempdir()` reports the 8.3 short name with forward slashes (`C:/Users/RUNNER~1/...`) while `utils::unzip()` reports the long name with backslashes (`C:\Users\runneradmin\...`), so normalizing separators the way `.redactTmpDir()` does would still leave the prefix unmatched. Matching on the basename is what makes this reachable at all; widening the separator is what makes the basename match.

Verified before pushing by feeding the helper the exact failing line from the job log: both paths redact to the committed snapshot text byte-for-byte, the bare `'Project.xlsx'` in the advice bullet survives, and the macOS `/private` form still redacts. The snapshot files themselves are unchanged — they were always correct; only the redaction was platform-blind.

Roxygen updated to name both platform spellings rather than just the macOS one.

tests/testthat/_snaps/project-excel.md
Felixmil: Fair, and I agree the parent chain is incidental to what the snapshot exists to assert. Deferring as you suggested, since the Windows fix was the one that had to land — that is in `d3d803fa` and `windows-latest` is green.

Worth adding one thing to the case for it, though: the exposure is a little wider than "readxl might change its wording". The two ends of that chain do not agree on how they spell the same path across platforms, which is precisely what broke Windows here — the esqlabsR-authored line gets `cli`'s normalized forward-slash path while `unzip()` carries the native spelling. Collapsing the `Caused by` block to a stable placeholder would remove that whole class of platform sensitivity, not just the wording sensitivity. Leaving the thread open.

R/project-excel.R
Felixmil: Confirmed as a real bug, and left alone deliberately: it is pre-existing, and this round was scoped to blocking findings on the PRs themselves.

Your point about why it is newly relevant is the strongest part of it. This PR's `messages.R` comment states the single-interpolation invariant explicitly, as the reason `excelProjectFileUnreadable()` is stored unglued — and then this call site interpolates the already-interpolated text a second time. So the PR documents an invariant that the very path its new snapshot pins goes on to violate.

Leaving the thread open rather than resolving it. A brace in a project directory making a read-only `projectStatus()` throw instead of warn deserves its own fix, and the binding you suggest is the whole of it.

PR #1241, opened 2026-08-07
Record that a PI member id is not a definition-file id

Felixmil:
#1236 removed id canonicalization from every parameter-identification member-id door, and the reason was recorded only in that PR's description. The conventions in `REFACTORING.md` still stated the canonicalization rule without its exception, so the next reader has every reason to "fix" the non-bug. This change writes the rule down; no behaviour changes.

A review of the load and lookup paths confirmed there is nothing to fix. Every door that writes a member id stores it exactly as given, and every lookup matches it exactly: the `add*` collision checks, `.removePIMember()`, the duplicate check in `.validatePI()`, and both Excel readers. A member id never reaches a filename — files are named from the serializer's top-level task keys, and members are arrays inside the task object — never a `DefinitionList` name lookup, and never a cross-reference target. The run path does not resolve a member by id at all; it iterates positionally. The one writer that does canonicalize is the legacy 5.x importer, which *invents* an id from a free-text column that no 5.x sheet carried, and compares canonical against canonical.

Canonicalizing one door alone would make a mixed-case member unreachable, since nothing canonicalizes to a string containing an uppercase letter. Two characterization tests already pin that trap.

## Documentation

- `REFACTORING.md`: the id-canonicalization bullet now says that an id is canonicalized because it names a definition file, that an id naming no file is not, and names the PI member id as the one such id in the tree.
- `CLAUDE.md`: a glossary row for "PI member id", contrasting it with a PI *task* id, which does name a file and is canonicalized. `id` is exactly the kind of overlapping term that table exists to disambiguate.
- `PIOutputMapping()`'s `@param id` gains the sentence its sibling `PIParameter()` already carried, so the two constructors describe the field the same way.

No `## Example` section: nothing observable changed.

Part of #1226.


REFACTORING.md
```diff
@@ -30,7 +30,7 @@ Match these when adding or reviewing code, so new work stays consistent with wha
...
 - **All state writes go through the private seam.** Read/write sections via `private$.getSection()` / `.setSection()` (they set the dirty bit and invalidate the validation cache). Don't poke section fields directly.
-- **Ids are canonicalized; shared vectorization helpers.** Run ids through `.canonicalizeId()`, and use the shared `.assertIdVector()` / `.assertNoDuplicateIds()` / `.recycleField()` / `.alignAuthoringArgs()` helpers rather than re-implementing per domain.
+- **Ids are canonicalized; shared vectorization helpers.** Run ids through `.canonicalizeId()`, and use the shared `.assertIdVector()` / `.assertNoDuplicateIds()` / `.recycleField()` / `.alignAuthoringArgs()` helpers rather than re-implementing per domain. An id is canonicalized because it names a definition file; an id that names no file is not. The one such id in the tree is a PI member id (a `PIParameter` / `PIOutputMapping` `id`), which labels a record inside its task's file: every door stores it exactly as given, and every lookup matches it exactly. Canonicalizing one door alone would make a mixed-case member unreachable, since nothing canonicalizes to a string containing an uppercase letter.
```
Felixmil, line 33: "every door stores it exactly as given" is true of every door that is *given* an id, but the legacy 5.x importer is a door that runs one through the canonicalizer, and this sentence is the only place a future reader will look before deciding that is a bug.

`.parseExcelPI5xParams()` (`R/project-excel.R:3545`) and `.parseExcelPI5xMappings()` (`R/project-excel.R:3668`) both coin the member id via `.uniqueImportedId()`, which calls `.canonicalizeOneId()` on its base (`R/project-excel.R:3923`). So a 5.x-imported member lands lowercased, while a new-layout import (`.parseExcelPIRows()`) stores the sheet's `id` column verbatim.

That is correct as written — the 5.x sheets carry no id column, so the importer invents one rather than being handed one, and it compares canonical against canonical within the task, so nothing goes unreachable. The PR description says exactly this; the rule it is documenting does not. Since the point of writing the rule down is to stop the next reader "fixing" the non-bug, the one place where canonicalization legitimately touches a member id is worth naming here too — something like: *the one exception is the legacy 5.x importer, which invents an id rather than being given one, and matches canonical against canonical.*

Everything else in the bullet checks out against the code: the four `.canonicalizeId()` call sites in `R/parameter-identification.R` (1382, 1529, 1612, 1761) are all task ids; `.canonicalizePIParameterRefs()` / `.canonicalizePIOutputMappingRefs()` touch only `scenarios`; `.removePIMember()`, the `add*` collision checks and `.validatePI()`'s `.checkNoDuplicates()` all match verbatim; and `.canonicalizeOneId()` does `tolower()` (`R/definition-id.R:389`), so the mixed-case-unreachable claim holds.

---

Felixmil (commented):
Documentation-only PR (+5/-3 across 5 files), so the review was an accuracy check of the claims against the code rather than a behaviour review. They hold.

Verified:

- **No PI member-id door canonicalizes.** All four `.canonicalizeId()` call sites in `R/parameter-identification.R` (1382 `.addPITask_impl`, 1529 `.setPITask_impl`, 1612 `.removePITask_impl`, 1761 `.requirePITask`) take a *task* id. `.canonicalizePIParameterRefs()` / `.canonicalizePIOutputMappingRefs()` reshape only `scenarios`. The Excel migration loop (`R/project-excel.R:1750-1770`) canonicalizes the task id, task/member `scenarios` and `outputPath`, and deliberately leaves each member's own `id`.
- **Every lookup matches verbatim.** `.removePIMember()` (`R/parameter-identification.R:1872`), the `existingIds == id` overwrite checks in `.addPIParameter_impl` / `.addPIOutputMapping_impl`, and `.validatePI()`'s two `.checkNoDuplicates()` calls all compare the stored string.
- **The mixed-case trap is real and pinned.** `.canonicalizeOneId()` lowercases (`R/definition-id.R:389`), so a canonical id can never contain an uppercase letter; `tests/testthat/test-parameter-identification.R:2513-2524` and the `overwrite = TRUE` test at 2528 pin both halves.
- **Doc symmetry is now correct.** `PIParameter()` already carried the sentence, `PIOutputMapping()` did not; `man/PIOutputMapping.Rd` is the only regenerated `.Rd`, which is exactly right since it is the only roxygen block that changed.
- The `REFACTORING-LOG.md` bullet is well-formed and correctly ordered.

One non-blocking note inline on `REFACTORING.md`: the "every door stores it exactly as given" clause has an exception the PR description states but the rule itself does not — the legacy 5.x importer coins member ids through `.uniqueImportedId()`, which canonicalizes. Worth naming, since preventing a future "fix" is the whole purpose of the bullet.

CI at review time: `pkgdown`, `check-changes` and the link check green; the three `R-CMD-Check` jobs and `test-coverage` still running.

---

REFACTORING.md
Felixmil: Correct, and the reasoning holds — the rule as written invites exactly the "fix" it exists to prevent, and the legacy 5.x importer is the one door that runs a member id through the canonicalizer.

Deferring rather than fixing: this round was scoped to blocking findings only. The gap is narrow in that the PR description does state the exception, so what is missing is the exception in the durable artifact rather than anywhere in the record. Leaving the thread open so it stays visible; it belongs in the same edit as any future pass over that bullet.

PR #1242, opened 2026-08-07
Let initProject() name the project it creates

Felixmil:
A project's `name` could only be set after the fact, through `project$info$name`, so every project `initProject()` created started out named `"Blank"` — the template folder it was copied from, leaking into the container of every new project. `initProject()` now takes the name up front, and the blank template no longer carries one.

## Authoring

`initProject()` gains a trailing `name` argument, written to the `name` field of the new `Project.json`. It is appended rather than inserted so no positional call changes meaning.

The default is `""`, which writes nothing and leaves whatever name the template carries. That is empty for `type = "minimal"`, since the blank template's `"name": "Blank"` is now `""`. For `type = "example"` it keeps the example project's own `"Example"`, which pairs with the description that template also ships; blanking it would have left the description orphaned, and that is not what this change is for.

## Persistence

The name is applied in the same post-copy pass that already restamps `esqlabsRVersion`, rather than adding a second read-modify-write of the container. That helper is now `.stampContainerFields()`, since it no longer only stamps the version. It has one caller, so the rename is contained, and applying the name there means it is in place before the optional Excel export reads the container back.

## Tests

Three tests in `tests/testthat/test-project-lifecycle.R`, beside the existing version-stamp test: the default leaves a minimal project unnamed, a supplied name reaches both the container file and `project$info$name`, and `type = "example"` keeps the template's own name. No existing test asserted the `"Blank"` name — the one occurrence of that string is a template *path*.

## Example

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/followup-e-init-project-name", quiet = TRUE)

# Named as it is created.
named <- file.path(tempdir(), "aciclovir-study")
initProject(named, type = "minimal", createExcel = FALSE, name = "Aciclovir PK")
jsonlite::fromJSON(file.path(named, "Project.json"))$name
#> [1] "Aciclovir PK"
loadProject(named)$info$name
#> [1] "Aciclovir PK"

# The default leaves a minimal project unnamed.
unnamed <- file.path(tempdir(), "scratch")
initProject(unnamed, type = "minimal", createExcel = FALSE)
jsonlite::fromJSON(file.path(unnamed, "Project.json"))$name
#> [1] ""
```

</details>

Part of #1226.


inst/extdata/projects/Blank/Project.json
```diff
@@ -1,7 +1,7 @@
 {
   "schemaVersion": "2.0",
   "esqlabsRVersion": "6.0.0",
-  "name": "Blank",
+  "name": "",
```
Felixmil: Blanking the name to `""` rather than dropping the key makes every default-scaffolded project report `excel_in_sync = FALSE` against the side-cars `initProject()` itself just wrote.

`""` is not `NULL`, so it survives the whole way: `.parseProjectSections()` sets `name = jsonData$name` (`R/project-lifecycle.R:211`), and `.projectToJson()`'s guard is `if (!is.null(project$info$name))` (`R/project-to-json.R:104`), so the memory side emits `"name": ""`. The Excel side collapses it: the export writes an empty `name` cell (`project$info$name %||% ""`, `R/project-excel.R:836`) and the import runs it through `emptyToNull()` (`R/project-excel.R:543-544`), so `jsonData$name` is `NULL` and the key is omitted (`R/project-excel.R:623-625`). `.compareJsonToExcel()` then diffs `{name: ""}` against a container with no `name` key and reports `name` as out-of-sync.

Since `createExcel = TRUE` is the default, `initProject(dir); projectStatus(loadProject(...))` reports drift on a project nobody has touched. No test covers it — every `projectStatus()` test builds from `with_temp_project()` (`type = "example"`, named `"Example"`), and the three new tests here all pass `createExcel = FALSE`.

The same `""`-is-not-`NULL` gap hits `snapshotProject()`: `projectName <- project$info$name %||% "project"` (`R/project-snapshot.R:66`) falls back only on `NULL`, so a default project snapshots to `-2026-08-07-120000.esqlabsR` — leading dash, `"project"` stem never reached.

Suggest omitting the key from the template instead of emptying it. That is the state `.projectToJson()` already documents as "a project with no name", both symptoms disappear at once, and no consumer needs to learn a second spelling of "unnamed". `.stampContainerFields()`'s `!identical(name, "")` guard works unchanged — it just adds the key rather than overwriting it. The assertion below becomes `expect_null(container$name)`, and a `projectStatus()`-after-`initProject()` test would be worth adding to pin the Excel axis.

---

R/project-lifecycle.R
```diff
@@ -651,10 +661,12 @@ initProject <- function(
   destination = ".",
   type = c("minimal", "example"),
   createExcel = TRUE,
-  overwrite = FALSE
+  overwrite = FALSE,
+  name = ""
 ) {
   destination <- fs::path_abs(destination)
   type <- match.arg(type)
+  validateIsString(name)
```
Felixmil, line 672: `validateIsString()` checks type, not length or `NA`-ness — it forwards to `validateIsOfType(object, "character")`, which accepts a character vector of any length. `name` is documented as a character scalar and is written straight into a scalar container field, so `initProject(dir, name = c("a", "b"))` writes `"name": ["a", "b"]` and `name = NA_character_` writes `"name": null`.

Neither aborts here; both surface later and further away — a `null` name reloads as `NA`, and a two-element name reaches `.validateFilenameSegment()` in `snapshotProject()` as a vector. An `isOfLength(name, 1)` check (or `validateIsString(name)` plus a length guard) next to this line keeps the failure at the door.

---

Felixmil (commented):
Reviewed the single commit on this PR (`856611a`) against base `felixmil/pi-member-id-rule-1226`. R is not installed in this environment, so the findings come from static reading of the code paths rather than from a run.

The shape of the change is right: appending `name` keeps every positional call meaning what it did, folding it into the post-copy write avoids a second read-modify-write of the container, and `.stampContainerFields()` earns its rename. Applying it before the Excel export reads the container back is the correct ordering.

Two findings, both inline:

- **`"name": ""` in the Blank template** (`inst/extdata/projects/Blank/Project.json:4`) — empty-string is not the same "unnamed" that the rest of the codebase means by `NULL`, and the two spellings diverge in `.compareJsonToExcel()` and in `snapshotProject()`. Every default-scaffolded project reports `excel_in_sync = FALSE`, and a nameless project snapshots to a leading-dash filename. Dropping the key rather than emptying it fixes both.
- **`validateIsString(name)`** (`R/project-lifecycle.R:669`) — type-only; a length-2 or `NA` name writes a JSON array or `null` into a scalar field and fails somewhere else later.

The NEWS entry, the roxygen, and the three tests all read accurately against the code as written; the first test's expectation would flip to `expect_null()` under the suggested fix.

---

inst/extdata/projects/Blank/Project.json
Felixmil: Fixed in `ed06c8ca` — the key is dropped from the template rather than emptied.

Reproduced both symptoms first. `initProject(dir)` with defaults, then `projectStatus()`:

```
excel_in_sync: FALSE
file_changes:  name = "Section missing in current Excel"
```

and `snapshotProject()` on that same project produced `-2026-08-07-113018.esqlabsR`. Both are gone.

`.stampContainerFields()`'s `!identical(name, "")` guard did work unchanged, as you predicted — it now adds the key rather than overwriting one.

Test changes: the assertion became `expect_null(container$name)`, and I added the `projectStatus()`-after-`initProject()` test you suggested. That one is the actual regression test — it fails on the empty-string spelling, and you were right that nothing covered the path, since every `projectStatus()` test builds from `with_temp_project()` (`type = "example"`, named) and all three tests here passed `createExcel = FALSE`. I also pinned the snapshot filename, so the second symptom cannot come back either.

NEWS and the roxygen now say the default writes no name, rather than describing it as empty.

R/project-lifecycle.R
Felixmil: Fixed in `ed06c8ca`. Confirmed both shapes by running them first: `name = c("a", "b")` wrote `"name": ["a", "b"]` and `name = NA_character_` wrote `"name": null`, neither aborting.

The guard is a length-and-`NA` check next to `validateIsString()`, raising a new `messages$invalidInitProjectName()`. One deliberate difference from `messages$invalidPathArgument()` beside it: the message asks for a single non-`NA` string, not a non-*empty* one, because `""` is the documented default and is the "leave whatever name the template carries" sentinel — borrowing the neighbouring wording would have described the default as invalid.

Both rejected shapes are snapshot-tested, and both attribute to `initProject()`.

PR #1243, opened 2026-08-07
Resolve a PI output mapping's outputPath when a project is loaded

Felixmil:
A PI output mapping's `outputPath` was stored one way and looked up another. The authoring path resolves the reference and canonicalizes it, and the Excel path canonicalizes it, but the JSON parser did neither: it copied the cell verbatim. Cross-reference validation then compares references *canonically*, while the runtime resolves the mapping with an exact `[[` lookup — so a hand-edited definition file spelling the reference non-canonically passed `validateProject()` clean and then produced a wrong run.

Wrong, not failed. `outputPaths[[outputPathId]]` returns `NULL`, `unique(c(NULL, NULL))` is `NULL`, and assigning `NULL` to a list element does not create it — so the per-scenario output map stays empty, `ospsuite::setOutputs()` is never called at all, and the simulation silently keeps whatever outputs its PKML happens to carry. The build then dies further along on `ospsuite::getQuantity(NULL, ...)` with an untyped error naming neither the task, the mapping, nor the reference: `argument "paths" is of type <NULL>, but expected <character>!`.

Hand-editing the definitions tree is a first-class workflow in this design, and `saveProject()` writes canonical ids, so this is reachable exactly the way a user edits a project by hand.

## Load path

`.parsePIOutputMappings()` resolves the reference through `.matchOutputPathRef()`, the same helper `addPIOutputMapping()` already uses. That helper tries the literal-path match first and only then canonicalizes as an id, which fixes a second gap in the same place: `PIOutputMapping()`'s own docs promise that `outputPath` may be either an output-path id or the literal model path it maps to, and the authoring path honours that, but on load a literal model path was reported as a dangling reference and crashed the build.

Threading the reference resolution needs the already-parsed `outputPaths` section, and there was precedent to copy: the scenarios parser already receives a `list(definitions = list(outputPaths = ...))` stand-in, and `outputPaths` is loaded before `parameterIdentification`, so no load ordering changed.

- `R/parameter-identification.R` — `.parsePITasks(piData, project = NULL)` threads through to `.parsePIOutputMappings(rawList, taskId, project = NULL)`
- `R/definition-files.R` — the kind spec passes `project`
- `R/project-lifecycle.R` — the section is loaded with the stand-in

An unresolvable reference still falls through and stays verbatim, preserving the deliberate parse-leniently-then-validate contract: a mapping that names no output loads and is reported by `validateProject()` rather than aborting the load of the whole project. The `project = NULL` default keeps the existing direct-parser tests byte-identical.

The alternatives were both worse. Making the *runtime* lookup canonical patches two call sites and leaves the record itself disagreeing with what every other door stores, so `removeOutputPath()`'s exact-match in-use scan stays wrong, any future consumer of `m$outputPathId` inherits the bug, and the non-canonical spelling round-trips back to disk. Making *validation* exact would move `.refResolves()`, shared by all fourteen reference kinds, whose canonical tolerance is deliberate and correct — the authoring path really does accept a non-canonical spelling.

## Siblings

Both were checked by running them, not by reading, and neither shares the defect:

- A mapping's `observedData` is verbatim on every door, deliberately — the reference is an OSPS-owned dataset name that canonicalization would destroy. The runtime aborts with a typed `PIDatasetNotFound` naming both spellings. Its only asymmetry is that validation is canonically tolerant for a reference nothing canonicalizes, which is mild over-tolerance, not a wrong run. Left alone.
- A non-canonical `scenarios` reference either fails validation (when only a member is non-canonical, since `.validatePI()` compares member against task exactly) or aborts at run time with a typed, attributed `PIScenarioNotFound` listing the available scenarios. That is a loud failure, so it was left alone rather than changing code that already fails honestly. One residual worth noting separately: a task-level non-canonical spelling makes `.validatePI()` blame every member for the task's spelling, which is misleading attribution.

## Tests

Three tests in `tests/testthat/test-parameter-identification.R` driven by a local helper that rewrites one mapping's `outputPath` in a copied fixture: a non-canonical id in both spellings (asserting the stored id, the exact runtime lookup, and that validation is clean), a literal model path, and a dangling reference kept verbatim and reported. Verified red-then-green by stashing only the three `R/` files with the tests in place: 6 failures before, 0 after.

## Example

Before:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/followup-e-init-project-name", quiet = TRUE)

# A project whose PI output mapping names its output path non-canonically,
# the way a hand-edited definition file can.
dir.create(dir)
invisible(file.copy(
  dir,
  recursive = TRUE
))
pi <- file.path(dir, "definitions", "parameter-identification", "aciclovirsimple.json")
writeLines(
  sub("\"aciclovir_pvb\"", "\"Aciclovir_PVB\"", readLines(pi), fixed = TRUE),
  pi
)

project <- loadProject(dir)
mapping <- project$definitions$parameterIdentification[["aciclovirsimple"]]$outputMappings[[1]]

# What the mapping resolved to, and whether the runtime can look it up.
mapping$outputPathId
#> [1] "Aciclovir_PVB"
project$definitions$outputPaths[[mapping$outputPathId]]
#> NULL

# What validateProject() says about it.
isAnyCriticalErrors(validateProject(project))
#> [1] FALSE
```

After:

``` r
devtools::load_all("/Users/felix/Code/esqlabsR/.claude/worktrees/followup-d-pi-output-path", quiet = TRUE)

# A project whose PI output mapping names its output path non-canonically,
# the way a hand-edited definition file can.
dir.create(dir)
invisible(file.copy(
  dir,
  recursive = TRUE
))
pi <- file.path(dir, "definitions", "parameter-identification", "aciclovirsimple.json")
writeLines(
  sub("\"aciclovir_pvb\"", "\"Aciclovir_PVB\"", readLines(pi), fixed = TRUE),
  pi
)

project <- loadProject(dir)
mapping <- project$definitions$parameterIdentification[["aciclovirsimple"]]$outputMappings[[1]]

# What the mapping resolved to, and whether the runtime can look it up.
mapping$outputPathId
#> [1] "aciclovir_pvb"
project$definitions$outputPaths[[mapping$outputPathId]]
#> [1] "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)"

# What validateProject() says about it.
isAnyCriticalErrors(validateProject(project))
#> [1] FALSE
```

</details>

Part of #1226.


NEWS.md
```diff
@@ -144,6 +144,7 @@ Deprecation intensity follows the lifecycle policy: forwarding shims that still
...
+- `loadProject()` and `restoreProject()` resolve a PI output mapping's `outputPath` the same way `addPIOutputMapping()` does, so the two forms it documents work in a hand-edited definition file too: an output-path id in any spelling that canonicalizes onto a defined one, and the literal model path of a defined output path. Previously a definition file's `outputPath` was stored exactly as written while the run looked it up exactly, and `validateProject()` compared it canonically, so a file spelling the reference `Plasma` where the id is `plasma` validated with 0 errors and then ran with that output path silently missing from `ospsuite::setOutputs()`, failing later with a message naming neither the task nor the reference. A reference that names no defined output path is still kept as written, and `validateProject()` still reports it, so a project with a bad reference stays openable and fixable. (#1226)
```
Felixmil: The `restoreProject()` half of this sentence doesn't hold for the literal-model-path form.

`restoreProject()` runs `.canonicalizeProjectJsonIds()` over the raw snapshot JSON *before* it parses it (`R/project-snapshot.R:213`), and that function canonicalizes a mapping's reference unconditionally:

```r
mapping$outputPath <- canonScalar(mapping$outputPath)   # R/project-excel.R:1765
```

`canonScalar()` is `.canonicalizeOneId()`, which turns every `|`, space and paren into `_`. So a snapshot spelling the reference as `Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)` reaches `.parsePIOutputMappings()` already mangled into `organism_peripheralvenousblood_aciclovir_plasma__peripheral_venous_blood_`: the literal-path branch of `.matchOutputPathRef()` can no longer match it (the value is no longer the path), the id branch can't either, and it falls through verbatim as a dangling reference. `.writeProjectTree()` then persists the mangled string into the restored tree, so the original path isn't even recoverable from the restore output.

The id form is unaffected — definition and reference get the same transform, which is the property that function rests on. It's only the literal path that gets destroyed, and only on this entry point: `importProjectFromExcel()` is fine because `.resolvePIOutputPathRefs()` (`R/project-excel.R:3950`) rewrites a full path to its id *before* the canonicalization pass, and the legacy monolithic snapshot goes through that same Excel bridge via `.upgradeLegacySnapshot()`. The exposed case is a v6-shaped hand-authored snapshot — exactly the audience the paragraph is addressed to.

Two ways out: skip canonicalizing `mapping$outputPath` in `.canonicalizeProjectJsonIds()` when it `identical`-matches one of `jsonData$outputPaths`' values (the same carve-out `observedData` already gets, and `outputPaths` names are canonicalized earlier in the same function so the map is available), or reuse `.resolvePIOutputPathRefs()` there. Either way the sentence becomes true; otherwise it's worth narrowing the claim to `loadProject()` for the literal form.

---

R/parameter-identification.R
```diff
@@ -605,11 +611,24 @@ print.PITask <- function(x, ...) {
...
+    # runtime lookup (`.createSinglePITask()`) and `removeOutputPath()`'s in-use
+    # scan match on exactly: an id in any spelling that canonicalizes onto a
+    # defined one, or the literal model path of a defined output path (the form
+    # `PIOutputMapping()` documents). A reference naming no defined output path
+    # is kept verbatim, so the project still loads and `validateProject()`
+    # reports the dangling reference; without a `project` to resolve against,
+    # every reference is left verbatim for the same reason.
+    outputPath <- raw[["outputPath"]]
+    if (!is.null(project)) {
+      outputPath <- .matchOutputPathRef(outputPath, project) %||% outputPath
```
Felixmil, line 630: This call can abort the load, which the comment right above it (and the PR description) says it won't.

The unresolvable branch of `.matchOutputPathRef()` reaches `.silentlyCanonicalized(.canonicalizeIdRef(value))` (`R/parameter-identification.R:1703`), and `.canonicalizeOneId()` starts with a hard length guard:

```r
if (nbytes > limit) {           # .maxDefinitionIdBytes = 250L
  cli::cli_abort(c("Definition id is too long to be a safe filename: ...
```

`.silentlyCanonicalized()` only suppresses the per-id *warning*; the abort propagates. So a mapping whose `outputPath` is longer than 250 bytes and matches no defined output path now takes down the whole `loadProject()` with `Definition id is too long to be a safe filename: N bytes (limit 250).` — a message naming neither the task, nor the mapping, nor the field. Before this change the parser copied the value verbatim and `validateProject()` reported it.

Narrow, but it's the one shape of bad reference that stops being fixable: the project can't be opened, which is the failure mode the leniency contract exists to prevent, and the fix arrived for the same reason ("a project that cannot be opened cannot be fixed"). A defined long path is fine — the literal-match branch runs first and returns before any canonicalization — so the trigger is specifically an unmatched over-long reference: a deep metabolite path in a large model whose output-path entry was renamed or removed, or a paste error in a hand-edited file.

Cheapest fix is to keep the guard out of the resolution path here, e.g. wrap it:

```r
outputPath <- tryCatch(
  .matchOutputPathRef(outputPath, project) %||% outputPath,
  error = function(e) outputPath
)
```

though a targeted length check inside `.matchOutputPathRef()` (return `NULL` rather than canonicalize an over-long value) reads better and fixes `addPIOutputMapping()`'s error text at the same time — there the abort is at least fail-fast, but it still reports a filename-length problem for what the user experiences as an unknown output path, instead of the typed `outputPathRefNotFound` with its did-you-mean hint.

Worth a fourth case in the new test block: an over-long reference kept verbatim and reported, alongside `no_such_output_path`.

---

Felixmil (commented):
Reviewed the single commit on this PR (`0d2b043`) statically — R isn't installed in this environment, so nothing was executed; I read every hunk plus `.matchOutputPathRef()`, `.canonicalizeOneId()`, `.canonicalizeProjectJsonIds()`, `.resolvePIOutputPathRefs()`, `.parseProjectSections()`, `restoreProject()` and both runtime lookup sites.

The core fix is right, and the reasoning in the description holds up against the code: the runtime really does look the id up exactly (`R/parameter-identification.R:872` and `:936`), `removeOutputPath()`'s in-use scan really does compare with `identical()` (`R/validation.R:540`), and fixing the record at the parse seam is the only one of the three options that leaves all of them agreeing. Load ordering is fine — `outputPaths` is bound at `R/project-lifecycle.R:202`, well before the `parameterIdentification` section is loaded — the `project = NULL` default leaves the direct-parser tests untouched, and `.compareJsonToExcel()` gets the same treatment for free since it goes through `.parseProjectSections()` too, so the two sides stay symmetric.

Two findings, both on the edges of the new resolution rather than its centre, left inline:

- **`NEWS.md`** — the `restoreProject()` half of the claim doesn't hold for the literal-model-path form: `.canonicalizeProjectJsonIds()` mangles the path before the new parser code can see it, and the restore persists the mangled string into the tree.
- **`R/parameter-identification.R:630`** — the new call can abort `loadProject()` on an unresolvable reference over 250 bytes, via `.canonicalizeOneId()`'s filename-length guard. That contradicts the leniency contract stated in the comment directly above the call.

Neither blocks the fix; both are worth a follow-up commit or a narrowed NEWS sentence. CI was still running when I looked (three `R-CMD-Check` legs and `test-coverage` in progress, everything completed green), so the `unstable` mergeable state is pending checks rather than a failure.

---

NEWS.md
Felixmil: Fixed in `4b583c3d`, and I took the larger option rather than narrowing the sentence.

Reproduced it first on a v6-shaped snapshot: the restored mapping carried `organism_peripheralvenousblood_aciclovir_plasma_(peripheral_venous_blood)`, `validateProject()` reported a critical error, and the written tree file kept the mangled string. The id form was unaffected, as you said.

Narrowing the claim would have left the restore silently destroying the path — `.writeProjectTree()` persists it, so it is not recoverable from the output. That is data loss rather than a doc inaccuracy, and the fix turned out smaller than the bug: reuse `.resolvePIOutputPathRefs()` in `.canonicalizeProjectJsonIds()`'s PI block, before the canonicalization pass. Three lines, no new logic, and `jsonData$outputPaths` is already keyed by canonical id at that point so the substituted id needs no further transform.

One thing that required care, and is worth knowing since it argues for the carve-out you offered as the alternative: putting `.resolvePIOutputPathRefs()` on the restore path would itself have added a *new* abort door. Its `vapply(outputPaths, as.character, character(1))` dies on a malformed entry (`values must be length 1, but FUN(X[[2]]) result is length 2`) and its `!is.null(path) && path %in% ...` dies on a multi-length reference. Both are now scalar-guarded, which also removes the same latent abort from the existing Excel caller.

New test in `test-project-snapshot.R` asserts the resolved id in memory, no critical errors, and that the written tree file carries `aciclovir_pvb` rather than a mangled path. The NEWS sentence is widened rather than just repaired: it now states `restoreProject()`'s own failure mode, and "hand-edited definition file" became "definition file or snapshot".

R/parameter-identification.R
Felixmil: Fixed in `4b583c3d` — and neither suggested fix was sufficient, so this went a layer deeper. Reproducing it is what exposed the rest, so thanks for the pointer.

The 276-byte unmatched reference aborted the load exactly as you described. But with a length guard local to `.matchOutputPathRef()`, the parse survives and then `validateProject()` aborts from a *second* door:

```
Error in `FUN(X[[i]], ...)`: Definition id is too long to be a safe filename: 264 bytes (limit 250).
 5. └─.warnOnCrossReferenceErrors(project)
 8.   └─.refResolves(m$outputPathId, outputPathKeys)
10.    └─.canonicalizeForCompare(ref)
```

`.canonicalizeForCompare()` is the shared canonicalize-to-compare chokepoint for all fourteen reference kinds, and it pre-dates this PR — so an over-long *dangling* reference of any kind already made a project unopenable, not just a PI output path. Confirmed by swapping a scenario's `individual` for the same over-long string: same abort, and it now loads and validates.

So the guard went there instead of at the call site: `keep <- !is.na(ids) & nchar(ids, type = "bytes") <= .maxDefinitionIdBytes` inside `.refResolves()`'s canonicalization. A comparison must not abort, and nothing can be *filed* above the bound, so an over-long reference matches nothing and is correctly reported dangling. `.matchOutputPathRef()` now routes through `.canonicalizeForCompare()` as well — behaviourally identical for in-range scalars (both silent, both element-wise `.canonicalizeOneId`), and it makes the matcher and the validator's report share one transform, which is the disagreement class this PR exists to remove.

Your second claimed effect is real, and I checked it rather than assuming: `addPIOutputMapping(outputPath = <276 bytes>)` reported `Definition id is too long to be a safe filename: 276 bytes (limit 250).` before, and now reports the typed `outputPathRefNotFound` with its did-you-mean hint. No existing test pinned the old message — the only snapshot of that text (`tests/testthat/_snaps/definition-id.md:78`) covers `.canonicalizeId()`, the definition side, which is untouched.

Added the fourth case you asked for, plus an expectation pinning the typed message for the over-long reference.

Adjacent and not fixed here: `R/observed-data.R:699` still uses `.silentlyCanonicalized(.canonicalizeId(one))` to match an id for removal, so `removeObservedData()` with an over-long id reports the same wrong problem `addPIOutputMapping()` used to. That is an authoring fail-fast path rather than a load, so no project becomes unopenable by it.

PR #1253, opened 2026-08-28
docs: require a leading dot on internal functions, ban "scaffold"

PavelBal:
Two additions to `AGENTS.md`.

## Internal functions carry a leading dot

Added under **Coding conventions**:

> Internal (non-exported) functions are named with a leading dot: `.addInitialConditionEntry()`, `.absoluteAgainstRoot()`. The dot marks a function as internal at every call site, so a reader never has to check `NAMESPACE` to know whether they are looking at part of the package's interface.

This documents what the package already does rather than introducing anything. Counted on `main`:

| | count |
|---|---|
| top-level functions with a leading dot | 473 |
| top-level functions without one | 130 |

The 130 are largely the exported interface, so the convention is close to universal already — it just was not written down, which is exactly the kind of rule an agent cannot infer reliably from a partial view of the sources.

## "scaffold" joins the jargon list

Added to the list of words to avoid in user-facing documentation, alongside "reconcile", "authoring", "orphan", and the rest.

It fails the same test those fail: to a modeling expert "scaffold" is construction equipment, not a verb for creating files. "create", "set up", or "generate the files for" say the same thing in words the reader already owns.

## Notes

- Draft, per the repository's own PR rule.
- No `NEWS.md` bullet: this changes contributor guidance only, nothing a package user sees.

## Summary by CodeRabbit

* **Documentation**
  * Updated coding conventions for naming internal R functions.
  * Expanded guidance on avoiding developer jargon in user-facing documentation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,7 +22,7 @@ esqlabsR builds on the Open Systems Pharmacology ecosystem. When unsure about an
 
 ## Coding conventions
 
-- Internal (non-exported) functions are named with a leading dot: `.addInitialConditionEntry()`, `.absoluteAgainstRoot()`. The dot marks a function as internal at every call site, so a reader never has to check `NAMESPACE` to know whether they are looking at part of the package's interface.
+- Internal (non-exported) functions are named with a leading dot: `.validateParametersStructure()`, `.getEsqlabsColors()`. The dot marks a function as internal at every call site, so a reader never has to check `NAMESPACE` to know whether they are looking at part of the package's interface.
 - Type checking: use `ospsuite.utils::validateIsOfType()` / `ospsuite.utils::isOfType()` instead of native `inherits()` checks.
 - All user-facing messages, warnings, and errors live in `R/messages.R` as dedicated functions using `ospsuite.utils::cliFormat()` with cli markup (`{.val {x}}`, `{.arg {name}}`). Never inline message strings or build them with `paste()`.
 - Use explicit `package::function()` for functions from other packages. Never use `library()` or `require()` in package code. In tests, call esqlabsR's own functions directly, without a namespace prefix.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,6 +22,7 @@ esqlabsR builds on the Open Systems Pharmacology ecosystem. When unsure about an
 
 ## Coding conventions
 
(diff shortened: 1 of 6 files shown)


nautilus69 (changes requested):
1. Both example functions are absent from the tree; one was never dotted (ships into the repo)

  Neither .addInitialConditionEntry() nor .absoluteAgainstRoot() exists anywhere on
  this branch (R/, man/, tests/) — they appear only on the new line. Both were real
  once and were removed in the v6 / "method-first" refactors, and they fail in two
  different ways:

    - .addInitialConditionEntry(): the real function was addInitialConditionEntry()
      (#1089) — with NO leading dot — and was removed in #1121 ("Tidy the Project
      internal API surface: method-first"). So the example is doubly wrong: the
      function is gone, and when it existed it was itself a counterexample to the rule.

    - .absoluteAgainstRoot(): existed in exactly that dotted form (#1134) but was
      removed in the v6 project-layout work. Purely stale.

  Please swap in two real, current internal helpers, e.g.
    .validateParametersStructure()  (R/error-checks.R:10)
    .getEsqlabsColors()             (R/esqlabs-env.R:41)

  2. The 473 / 130 statistics don't reproduce (PR body + commit message)

  No measurement matches them:

    PR's claim ..................................  473 dotted / 130 non-dotted
    Top-level function definitions (bare) .......   79 dotted /  72 non-dotted
    ...including messages$ members ..............   79 dotted / 196 non-dotted
    Dotted call sites, R/ .......................  714
    Dotted call sites, R/ + tests/ ..............  1506

  473 is larger than the total number of function definitions in R/ (275), so it
  cannot be a count of dotted top-level functions. The reproducible figure is actually
  stronger than the PR's framing — see point 3.

  3. Wording collides with the messages pattern and undersells adherence

  Taken literally there are 124 internal non-dotted functions — but every one is a
  member of the messages object (messages$errorX <- function(...)), i.e. exactly the
  pattern documented two bullets down. Setting those and the 3 S3 print.* methods
  aside:
    - 79 free internal functions correctly carry the dot
    - exactly 1 does not: exampleDirectory() (R/example.R:9)
    - 0 exported names carry a dot

  That is 79/80 = 98.75% adherence — near-universal. Suggest narrowing the rule to
  "free-standing internal functions" and noting the messages/S3 exceptions, so it
  doesn't contradict the messages bullet.

  4. "reader never has to check NAMESPACE" is overstated by one case

  dot => internal is exact (0 dotted exports). The converse has one exception,
  exampleDirectory() — which the rule newly flags as the single function to rename if
  it is meant to be internal.


first two items are the blockers for approval.

nautilus69 (approved):

PR #1256, opened 2026-08-28
Apply checkForNegativeValues to the solver instead of the run options

### LINKED ISSUE #1252: `checkForNegativeValues` moved off `SimulationRunOptions` in ospsuite 13 — setting it in run defaults errors
## Problem

ospsuite 13 removed `checkForNegativeValues` from `SimulationRunOptions` (the option moved to the solver settings, `simulation$solver$checkForNegativeValues`). `SimulationRunOptions` now only carries `numberOfCores` and `showProgress`. Assigning the removed field is a hard error, not a silent no-op:

```r
library(ospsuite) # 13.0.0.9003
o <- SimulationRunOptions$new()
o$checkForNegativeValues <- FALSE
#> Error: cannot add bindings to a locked environment
```

## Where esqlabsR still sets it (branch `json-based-project`)

- `R/scenario-execution.R` — `.buildSimulationRunOptions()` assigns `runOpts$checkForNegativeValues` when the project declares it in `defaultSimulationRunOptions`.
- `R/parameter-identification.R` (~line 1099) — the PI run-options builder does the same.
- `R/project.R` — the `defaultSimulationRunOptions` field docs list `checkForNegativeValues` as a recognized field.
- `R/project-excel.R` (~line 3824) — the Excel import maps the column into the stored defaults.

## Effect

Any v6 project that declares `checkForNegativeValues` in `defaultSimulationRunOptions` (directly in `Project.json` or imported from Excel) errors as soon as scenarios or a parameter identification are run with ospsuite 13.

## Proposed fix

Route the stored `checkForNegativeValues` default to `simulation$solver$checkForNegativeValues` on each prepared simulation (scenario preparation and PI), instead of setting it on `SimulationRunOptions`. Keep the JSON/Excel field as-is so existing projects keep working; update the `defaultSimulationRunOptions` docs to state where the setting is applied.

Related: #408 (making solver settings accessible per scenario) — this field is one of the solver settings that work would cover; the fix here is the minimal compatibility repair.

Found while benchmarking the scenario-execution workflow (2026-08); verified against ospsuite 13.0.0.9003.

PavelBal:
Closes #1252.

## Problem

ospsuite 13 removed `checkForNegativeValues` from `SimulationRunOptions` (it is a solver setting now, `simulation$solver$checkForNegativeValues`). esqlabsR still assigned it on the run options in `.buildSimulationRunOptions()` and `.buildPIConfiguration()`, so any project that declares `checkForNegativeValues` in `defaultSimulationRunOptions` or a PI task's `simulationRunOptions` block aborted with `cannot add bindings to a locked environment` as soon as scenarios or a PI were run.

## Change

- The run-options record is resolved once, by `.resolveRunOptions()`, into `list(runOptions, solverSettings)`. `numberOfCores` / `showProgress` become the `SimulationRunOptions`; `checkForNegativeValues` is written to each simulation's solver by `.applySolverSettings()` in `.prepareScenario()`, right after loading and ahead of the steady-state pre-solve, so that solve runs with the resolved setting too.
- An explicit `simulationRunOptions` argument to `runScenarios()` / `buildSimulations()` wins entirely: the project record is not consulted, and the simulations keep the solver settings of their model files.
- A PI task passes the task-over-project merged record (`modifyList(project default, task block)`) into `.prepareScenario()`; the post-hoc per-simulation loop is gone. `.buildPIConfiguration()` now reuses `.buildSimulationRunOptions()` (PI tasks honour `showProgress` as well).
- `DESCRIPTION` requires `ospsuite (>= 13.0.0)`: in OSPSuite.Core v12 the run option overrode the solver setting at run time (`SimModelManager.cs:98`), so the project value would be silently ignored on 12.x.
- The JSON/Excel field stays as-is; `Project$defaultSimulationRunOptions` roxygen, `runScenarios()` / `buildSimulations()` `@param simulationRunOptions`, the project-file reference vignette, and NEWS state where the setting is applied.

## Tests

- `.applySolverSettings()` unit test on `simple.pkml`; `.resolveRunOptions()` unit test (explicit / project default / neither).
- `buildSimulations()` proves the project default reaches the solver and that an explicit argument suppresses it.
- `.createSinglePITask()` covers task-only, project-default-only, and task-over-project; a mock-based test asserts the task value is already on the solver when the steady-state step is reached.
- `.buildPIConfiguration()` tolerates `checkForNegativeValues` in the block and maps `showProgress`.

`devtools::test()` on `test-scenario-execution.R`, `test-scenarios.R`, `test-project.R` (638 pass) and `test-parameter-identification.R` (331 pass), 0 failures, against ospsuite 13.0.0.9005.

## Before marking ready

- [ ] `devtools::document()` to regenerate `man/Project.Rd`, `man/runScenarios.Rd`, `man/buildSimulations.Rd` (Rd is regenerated once at the end, per the repo rule).

## Summary by CodeRabbit

* **Enhancements**
  * Project-level default simulation options are now applied automatically when no explicit options are provided.
  * Explicit simulation options take precedence over project defaults.
  * Negative-value checking is consistently applied through simulation solver settings.
  * Parameter-identification tasks inherit project defaults, with task-level settings taking precedence.
  * Settings are applied before steady-state pre-solving.

* **Documentation**
  * Updated guidance clarifies how simulation and solver options are inherited and overridden.


PavelBal:
CI failed at dependency resolution: `Can't install dependency ospsuite (>= 13.0.0)`. This branch has no `Remotes:`, and the OSP R-universe registry tracks `*release` for ospsuite (serves 12.4.4), so ospsuite 13 dev was not resolvable. Added `Remotes: ospsuite=Open-Systems-Pharmacology/OSPSuite-R` so pak takes it from GitHub main; ospsuite.plots and tlf still come from the R-universe through `extra-repositories`.

PavelBal:
@coderabbit review

nautilus69 (approved):
Approved with minor requests. 

None of the following items are blocking the first two are just recommendations and the rests are minor suggestions. I am not sure if these are fixed in upstream or some other PRs. 



1. 🟡 Design to confirm — an explicit simulationRunOptions silently drops the project's checkForNegativeValues

  The explicit path returns solverSettings = NULL (.resolveRunOptions, scenario-execution.R:597), so a caller who passes an explicit run-options object only to change
  numberOfCores silently loses the project's checkForNegativeValues default — the simulations revert to their model-file solver setting. Because ospsuite 13 removed the
  field from SimulationRunOptions, there is now no single-call way to say "override cores and keep/set the negative-values check"; the two controls live in different
  places (argument vs project default) yet are coupled all-or-nothing.

  It's documented ("an explicit argument replaces the whole block") and consistent with the prior "explicit wins entirely" semantics, so it's defensible. But the split
  location makes the silent drop a plausible footgun. Please confirm the intent. If kept, either (a) sharpen the docs to state the drop pointedly ("passing
  simulationRunOptions also discards the project's checkForNegativeValues, so the model-file setting is used"), or (b) consider treating checkForNegativeValues as
  orthogonal — sourced from the project default independently of whether an explicit run-options argument is given.

  2. 🟡 Release gate — Depends: ospsuite (>= 13.0.0) + Remotes: ospsuite=…/OSPSuite-R

  ospsuite 13 isn't released (tested against 13.0.0.9005; r-universe serves 12.4.x). The Remotes: entry and the companion "take ospsuite from GitHub main on CI" commit
  are temporary scaffolding — the REFACTORING-LOG says as much. They must be removed before any esqlabsR release (CRAN forbids Remotes; normal/r-universe installs can't
  satisfy >= 13.0.0 until ospsuite 13 ships). So this can't merge to a releasable/installable state until ospsuite 13 is published. Not a code defect — a tracked blocker
  and an unwind step to remember.

  3. 🟢 Scope boundary worth one line — sensitivity and the legacy pkml path don't consult the project default

  The setting reaches the JSON-project scenario-build and PI paths (both via .prepareScenario), but a standalone sensitivity run (R/sensitivity-calculation-utils.R:781
  loads its own simulation) and the legacy scenario-from-pkml.R:440 path don't consult defaultSimulationRunOptions, so a project's checkForNegativeValues won't affect
  those. This matches pre-existing behavior and #1252's scope (scenarios + PI), so it's fine — just confirm it's intended rather than an oversight.

  4. ⚪ Nit

  buildSimulations()'s @param simulationRunOptions says the run options are "not applied to the returned simulations (not run here)" while the solver settings are
  applied to them — the text disambiguates, but it's subtle. Acceptable as written.

  Tests — strong, and they follow the repo's own conventions

  - .applySolverSettings test uses simple.pkml, asserts the fixture ships with the check on (independent oracle), then proves FALSE lands, plus the two no-op paths.
  - .resolveRunOptions covers all three branches and asserts the exact structure, including solverSettings = NULL on the explicit path.
  - buildSimulations regression test proves the project default reaches simulation$solver and that an explicit argument suppresses it (reverting to the model-file
    value).
  - PI: task-only / project-default-only / task-over-project, plus a local_mocked_bindings(.applyScenarioSteadyState=…) test that captures
    simulation$solver$checkForNegativeValues at the moment the pre-solve is reached — directly locking in the ordering guarantee. Since .applyScenarioSteadyState is
    called unconditionally, this mock isn't fragile.

DESCRIPTION
```diff
@@ -19,7 +19,7 @@ License: GPL-2
 URL: https://github.com/esqLABS/esqlabsR, https://esqlabs.github.io/esqlabsR/
 BugReports: https://github.com/esqLABS/esqlabsR/issues
 Depends:
-    ospsuite (>= 12.4.2),
+    ospsuite (>= 13.0.0),
```
sorinvoicu, line 22: 1. Workflows/release-pr.yaml:118-121 runs manage_description.sh pin-remotes, skipped only for packages with no Remotes: field. It will pin to @*release → ospsuite 12.4.3 → contradicts Depends: ospsuite (>= 13.0.0), and the release PR fails its own dependency resolution.
2.  The .NET pin is wrong. ospsuite 13 targets net10.0; nightly-check.yaml:37-40 and the reusable R-CMD-check-build.yaml default both pin 8.0.x. Green only because the runner image already had 10.0.11 — incidental, not declared.
3. Silent behaviour change for projects declaring nothing at all. v12's runSimulations() did simulationRunOptions %||% SimulationRunOptions$new(), whose checkForNegativeValues defaulted TRUE and overrode the solver — so esqlabsR forced the check on regardless of the PKML. Under 13 the PKML wins, and a model shipping false that used to abort on negative values now runs silently. Affects users who never touched defaultSimulationRunOptions.

PR #1257, opened 2026-08-28
Apply the PI sheet's checkForNegativeValues to the solver and require ospsuite 13

### LINKED ISSUE #1252: `checkForNegativeValues` moved off `SimulationRunOptions` in ospsuite 13 — setting it in run defaults errors

PavelBal:
Fixes #1252 for the `main` (5.x, Excel-driven) line. The `json-based-project` fix is #1256.

## Problem

ospsuite 13 removed `checkForNegativeValues` from `SimulationRunOptions`; the check is a solver setting now (`simulation$solver$checkForNegativeValues`). `.createPIConfigurationFromConfig()` still assigned the `PIConfiguration` sheet's `checkForNegativeValues` column on the run options, so with ospsuite 13 `createPITasks()` aborted with `cannot add bindings to a locked environment` for any workbook that fills that cell.

## Change

- `DESCRIPTION` requires `ospsuite (>= 13.0.0)`. ospsuite 12.x has no solver-side `checkForNegativeValues` (12.4.2 `SolverSettings` exposes only `useJacobian`, `h0`, `hMin`, `hMax`, `mxStep`, `relTol`, `absTol`) and its .NET runner honours the run option, so the solver route cannot be made to work there; requiring 13 keeps one code path.
- `.createPIConfigurationFromConfig()` builds `SimulationRunOptions` from `numberOfCores` only.
- `.createSinglePITask()` writes the sheet's `checkForNegativeValues` to each task simulation's solver through the new `.applySolverCheckForNegativeValues()`; an empty cell (`NA`) leaves the model's own setting, a `NULL` simulation entry is skipped.
- The `pi-workflow` vignette says where the setting lands; NEWS carries the version bump under "Breaking changes". The sheet column is unchanged.

## Tests

- `createPITasks applies simulationRunOptions from PIConfiguration columns` reads the value back from the task simulations' solver (test-local `.taskCheckForNegativeValues()`), and asserts that a sheet with only `checkForNegativeValues` builds no run options.
- New unit test for `.applySolverCheckForNegativeValues()` on `simple.pkml`: FALSE lands, `NULL` entry skipped, `NA`/`NULL` leave the setting alone.

`devtools::test()` on `test-utilities-pi.R`, `test-utilities-pi-configuration.R`, `test-pi-task-configuration.R`: 207 pass, 0 failures, against ospsuite 13.0.0.9005.

## CI note

ospsuite 13.0.0 is not a released tag yet (latest release is `v12.4.3`, 2026-06-02; 13.0.0.9005 is `develop`). The `check-released-deps` job pins Remotes to `@*release`, so it cannot satisfy `ospsuite (>= 13.0.0)` until ospsuite 13 is released; expect that job to fail on this PR until then. The dev-deps job installs ospsuite `develop` and is the one that exercises this change.

## Summary by CodeRabbit

- **Breaking Changes**
  - Requires `ospsuite` version 13.0.0 or later.
  - `checkForNegativeValues` is now applied directly to each task simulation’s solver settings.
  - Empty values preserve existing solver or model defaults; `NULL` simulations are safely skipped.

- **Documentation**
  - Updated examples and workflow guidance to show solver-level configuration.
  - Clarified how core-count and negative-value settings are applied.

- **Bug Fixes**
  - Prevented errors when `checkForNegativeValues` is populated during task creation.


PavelBal:
Self-review (independent reviewer pass over a3eb30a0..372c474f) applied in 8adc7cf2:

- README example set `checkForNegativeValues` on the run options, which errors on ospsuite 13; it now sets each scenario's `simulation$solver$checkForNegativeValues`.
- `writeIndividualToXLS()` test expectations updated for ospsuite 13 (117 individual parameters, adds a `µmol/l` unit) — exposed by the version bump, not by the fix.
- PI vignette states what an empty cell means per column; NEWS keeps the version bump (Breaking changes) and the fix (bug fixes) as separate bullets; test helper inlined, `loadTestSimulation("simple")` reused.

Full `devtools::test()` under ospsuite 13.0.0.9005: 1407 pass, 12 fail, 1 skip. All 12 failures are vdiffr SVG snapshot drifts in `test-sensitivity-time-profiles.R` (10) and `test-utilities-sensitivity-calculation.R` (1, two expectations) — files this PR does not touch, and the same SVGs already show as modified in a local `main` checkout, so they are pre-existing and left out of scope here.

PR #1260, opened 2026-08-31
Read the sensitivity parameter factor from the dataset label, not the factor level

PavelBal:
Port of #1259 to `json-based-project`. Fixes 12 of the CI test failures on #1256 (the ones in `test-sensitivity-time-profiles.R` and `test-sensitivity-calculation-utils.R`).

## Root cause

ospsuite 13 returns the `name` column of `DataCombined$toDataFrame()` as a factor. `.aggregateSimulationAndObservedData()` in `R/sensitivity-time-profiles.R` built the numeric `ParameterFactor` with `as.numeric(name)`, which on a factor yields the level position — variation factors `0.1, 1, 2, 20` became `1, 2, 3, 4`. The "Parameter factor" colour legend collapsed to a single break and the lines were drawn in a different order, which is what the failing SVG snapshots show.

## Change

Same as #1259 on `main`:

- `as.numeric(as.character(name))`, with a comment on why.
- Regression test `simulated ParameterFactor carries the variation values` (fails on the unpatched code).
- NEWS bullet.

No snapshot files touched: with the fix the plots render exactly as the committed SVGs. `test-sensitivity-time-profiles.R` + `test-sensitivity-calculation-utils.R`: 67 pass, 0 fail against ospsuite 13.0.0.9005 / ggplot2 4.0.3.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected sensitivity time-profile data so parameter variation values are preserved accurately, including baseline values.
  * Restored variation-value labels in sensitivity plot legends.
  * Preserved parameter-factor ordering when using ospsuite 13 data.


PavelBal:
@coderabbit review

Issue #408, opened 2023-04-05
Make solver setting accessible for simulation scenarios.


PavelBal:

PavelBal:
Each scenario has a corresponding simulation, so solver settings are directly accessible. Closing for now.

PavelBal:
Solver properties should be part of scenarios configuration, to define them before creating simulation objects.

PR #1271, opened 2026-09-10
Make solver settings configurable per scenario

Felixmil:
A scenario definition could not override any solver setting: only the project and a parameter-identification task could, and only for `checkForNegativeValues`. Seven of the eight settable properties of `ospsuite::SolverSettings` were unreachable from the project model altogether; the only route was to build the simulations and edit the object, which excludes `runScenarios()` and puts the setting in a script instead of the shared project artifact. This adds a `solverSettings` block to the scenario definition, and separates how a run is executed from how the solver behaves so that each lives in one place.

## Two concerns, two homes

Following review, a configuration holds `SolverSettings` and nothing else, and `SimulationRunOptions` reaches the package only as an argument of the run functions.

- **How the solver behaves** is project data. It lives in the project's `defaultSolverSettings` and in a scenario's `solverSettings`.
- **How a run is executed** (`numberOfCores`, `showProgress`) is not. It is the `simulationRunOptions` argument of `runScenarios()` and `buildSimulations()`.

`defaultSimulationRunOptions` held both, which is what forced a checker that had to widen its known-key set to avoid reporting `numberOfCores` as an unknown solver setting, and the subtlety that an explicit argument replaced one half of a block while leaving the other. Splitting them deletes both.

Parameter identification is out of scope here: issue #408 is about scenarios. A task's `configuration$simulationRunOptions` keeps its name and still supplies both halves for that task, and `runPI()` keeps its signature. The only PI change is the one the issue asks for, that a scenario's own `solverSettings` wins over the task's when a task runs that scenario. Applying the same split to the PI configuration is worth its own issue.

## Precedence

For each solver setting, the most specific place that gives it a value wins. A setting nobody names keeps the value the model file carries.

- the project's `defaultSolverSettings`,
- a scenario's own `solverSettings` (inside a parameter identification, the task's `simulationRunOptions` block sits between the two),
- the `solverSettings` argument of `runScenarios()` / `buildSimulations()`.

`.resolveRunOptions()` hands `.prepareScenario()` the records below and above the scenario's own block as two separate slots, which is what lets the scenario block sit in the middle on the run path and on top on the PI path without either caller knowing the other's ordering. The shared fold is `.mergeRunOptionRecords()`.

## Validation

`.checkSolverSettings()` returns its problems as strings so `addScenario()`, `setScenario()`, the `solverSettings` argument and `.validateScenarios()` share one set of rules. The type rules earn their place because the .NET side coerces in silence: `simulation$solver$mxStep <- 1.5` truncates to `1`, `as.integer()` yields `NA` above `.Machine$integer.max`, and a negative `hMin` is accepted as given. `hMin`'s own default is `0`, so the numeric rule is non-negative rather than positive.

`.solverSettingFields` is written out rather than read from the class, because the tag is the point: `SolverSettings` exposes active bindings that carry no type information (`relTol` and `useJacobian` are byte-identical apart from the property name), and it is not exported. A test compares the names against `SolverSettings$active`, so a setting added upstream fails loudly instead of being rejected as unknown.

## Persistence and compatibility

`.scenarioToJson()` omits the key when the block is empty rather than emitting `null`: a `jsonlite` round-trip turns `"solverSettings": null` into `{}`, which `.assertNoEmptyObjectFields()` rejects on the next load. `setScenario()` replaces the block whole, and `solverSettings = NULL` clears it. Both authoring paths wrap the block per id, because `.wholeField()` reads a list whose length matches the id count as one element per id.

A project file written before the split still loads: the old `defaultSimulationRunOptions` key is accepted, and any run option it carries is reported and dropped.

## Excel

The scenario sheet has no solver-settings columns and gains none. `exportProjectToExcel()` warns and names the scenarios whose block it could not write; import never produces one. A project carrying a block therefore reports `excel_in_sync = FALSE` for as long as it is there, which is honest.

## Example

```r
project <- loadProject("MyProject")

# One stiff scenario, tighter tolerance, stored in the project
addScenario(project, "stiff_iv",
  modelFile = "Aciclovir.pkml",
  solverSettings = list(relTol = 1e-8, hMax = 0.5)
)

built <- buildSimulations(project, scenarios = "stiff_iv")
built$stiff_iv$simulation$solver$relTol
#> [1] 1e-08

# A one-off override for this run; hMax still comes from the scenario
built <- buildSimulations(project, scenarios = "stiff_iv",
                          solverSettings = list(relTol = 1e-4))
built$stiff_iv$simulation$solver$relTol
#> [1] 1e-04
built$stiff_iv$simulation$solver$hMax
#> [1] 0.5

# How the run is executed is a separate argument and touches no solver setting
runScenarios(project,
  simulationRunOptions = ospsuite::SimulationRunOptions$new(numberOfCores = 8)
)

# A misspelled setting is named rather than dropped
setScenario(project, "stiff_iv", solverSettings = list(reltol = 1e-6))
#> Error in `setScenario()`:
#> ! Cannot modify scenario "stiff_iv":
#> x solverSettings has unknown setting 'reltol'; the settings are absTol,
#>   relTol, h0, hMin, hMax, mxStep, useJacobian, checkForNegativeValues
```

Closes #408


R/scenario-execution.R
```diff
@@ -594,6 +742,17 @@
     argumentName = "customParams",
     nullAllowed = TRUE
   )
+  # `simulationRunOptions` takes either shape `.resolveRunOptions()` accepts.
+  # An empty list counts as a record: it names no field, so it resolves to the
+  # project default like a missing argument does.
+  runOptionsOk <- is.null(simulationRunOptions) ||
```
Felixmil: Two of the three levels of the new chain never reach `.checkSolverSettings()`.

The preflight guard here only asks "is this a named list", and `.resolveRunOptions()` then merges the record straight into the solver chain. `.checkSolverSettings()` is called from `R/scenarios.R` only (the section validator plus `addScenario()` / `setScenario()`), so it covers a scenario's own block and nothing else. The consequences are exactly the ones the PR description gives as the reason the check exists:

- `simulationRunOptions = list(reltol = 1e-4)` — silently dropped. `.applySolverSettings()` intersects with `.solverSettingFields`, `.buildSimulationRunOptions()` ignores the rest, so the run proceeds at the model's tolerance with no message. The same slip in a scenario block is named.
- `simulationRunOptions = list(mxStep = 1.5)` — `as.integer()` truncates to 1, silently.
- `simulationRunOptions = list(relTol = -1)` — written to the solver unchecked.
- `project$defaultSimulationRunOptions <- list(relTol = -1)` — same, and this one persists into `Project.json`, so a bad value is authored into the shared artifact and never reported by `validateProject()` either.

`.checkSolverSettings()` can't be applied to a run-options record as it stands, since it would report `numberOfCores` / `showProgress` as unknown settings. A small wrapper would do it — check the run record against `c(names(.solverSettingFields), "numberOfCores", "showProgress")` for the unknown-key half and hand the solver subset to the existing per-field rules — called here for the argument and for `project$defaultSimulationRunOptions`. A project-container validator for `defaultSimulationRunOptions` would close the persisted case at load time.

---

R/scenario-execution.R
```diff
@@ -26,41 +26,177 @@
...
 }
 
+# Fold run-options records into one, left to right: a field set in a later
+# record wins, a field only an earlier one sets falls through, and a NULL or
+# empty record contributes nothing. This is the merge rule every level of the
+# precedence chain shares, so the chain is expressed by the order of the
+# arguments rather than by a rule per caller.
+# @keywords internal
+# @noRd
+.mergeRunOptionRecords <- function(...) {
```
Felixmil, line 197: `utils::modifyList()` *deletes* a key whose value in the later list is `NULL`, so a `NULL` field does not fall through — it suppresses the level below.

Everything else in the PR treats a `NULL` field as "not carried": `.applySolverSettings()` skips it explicitly, `.checkSolverSettings()` skips it, and `tests/testthat/test-scenario-execution.R:84` asserts `list(relTol = NULL)` is a sound block. But here:

```r
.mergeRunOptionRecords(list(relTol = 1e-5), list(relTol = NULL))
#> list()   — the project default is gone, not preserved
```

So a scenario with `solverSettings = list(relTol = NULL)` silently drops the project's `relTol` instead of inheriting it, which contradicts the `@param solverSettings` docs ("one left out keeps whatever the level below sets"). The `is.null(value)` guard in `.applySolverSettings()` can in fact never fire on a merged record for the same reason.

Dropping `NULL` entries before the fold makes the three places agree:

```r
records <- Filter(function(x) length(x) > 0L, list(...))
records <- lapply(records, function(x) x[!vapply(x, is.null, logical(1))])
```

(If deletion is the intended semantics for clearing a project default, then it's the doc and the checker that need to say so — but a `NULL` that means two different things at two levels seems worth avoiding.)

---

R/scenario-execution.R
```diff
@@ -26,41 +26,177 @@
...
+      .solverSettingFields[[field]],
+      number = if (
+        !is.numeric(value) ||
+          length(value) != 1L ||
+          !is.finite(value) ||
+          value < 0
+      ) {
+        "a single non-negative number"
+      },
+      count = if (
```
Felixmil, line 118: The `count` rule accepts any whole number, but `.applySolverSettings()` coerces with `as.integer()`, which returns `NA` (with a warning) above `.Machine$integer.max`. `mxStep = 5e9` therefore passes validation and reaches the solver as `NA_integer_` — the silent-coercion failure mode this rule was added to prevent, one order of magnitude further out.

Adding an upper bound to the rule keeps checker and applier in agreement:

```r
value > .Machine$integer.max
```

with the message extended to "a single whole number of at least 1" → e.g. "…between 1 and 2147483647".

---

Felixmil (commented):
Automated review of the 8-commit series (`4cb538a...0896158`), medium effort.

The wiring checks out: `.scenarioFieldNames` was extended so the unknown-field backstop accepts the new field, `.wholeField()` recycling is pre-wrapped in both authoring doors, `.scenarioToJson()` omits an empty block (avoiding the `{}`-on-reload trap), `loadFromCache = FALSE` keeps solver settings from leaking between scenarios, and every caller of the renamed `.prepareScenario()` / `.buildScenarioSimulations()` arguments was updated — PI included, sensitivity untouched.

Three findings are inline. The first is the one worth acting on before merge: the check that motivates the whole "the .NET side coerces in silence" argument guards a scenario's own block only, so the two run-options levels of the same chain still accept a typo, a truncating `mxStep` and a negative tolerance without a word — including a `defaultSimulationRunOptions` that persists into `Project.json`. The other two are edges of the same coercion story (`NULL` deleting the level below via `modifyList()`, and `as.integer()` overflow above `.Machine$integer.max`).

One doc gap, in a file this PR doesn't touch so it can't be commented inline — `R/project.R:178-191`, the `@field defaultSimulationRunOptions` roxygen:

- "Recognized fields: `numberOfCores`, `checkForNegativeValues`, `showProgress`" — all eight solver settings are now recognized there.
- "an explicit `simulationRunOptions` argument replaces the whole default, and the simulations then keep the solver settings of their model files" — this PR reverses exactly that, and the argument now also accepts a named list.

Worth refreshing in the same series, since it's the reference doc for the bottom of the new precedence chain.

---

R/parameter-identification.R
```diff
@@ -841,13 +841,15 @@ print.PITask <- function(x, ...) {
...
   cache$populations <- list()
 
-  # Solver settings for the task's simulations: the project's default record
-  # is the baseline and the task's own `simulationRunOptions` block overrides
-  # it entry by entry. `.prepareScenario()` writes the result before its
-  # steady-state pre-solve.
-  solverSettings <- utils::modifyList(
-    project$defaultSimulationRunOptions %||% list(),
-    piTask$configuration$simulationRunOptions %||% list()
+  # Solver settings for the task's simulations: the project's default record is
```
PavelBal: Hard to read for me. "Record", "baseline" are not so common terminology.

Public documentation should not have any reference of private functions, like `.prepareScenario()`.

The second half is too technical/detailed. It reads like explanation of the implementation, which is of no use for the user.

R/project.R
```diff
@@ -176,19 +176,27 @@ Project <- R6::R6Class(
     },
 
     #' @field defaultSimulationRunOptions Named list of the project-level
```
PavelBal: Make here a reference to documentation of the two classes:

https://www.open-systems-pharmacology.org/OSPSuite-R/reference/SolverSettings.html
https://www.open-systems-pharmacology.org/OSPSuite-R/reference/SimulationRunOptions.html

R/project.R
```diff
@@ -176,19 +176,27 @@ Project <- R6::R6Class(
...
-    #'   simulations then keep the solver settings of their model files. A
-    #'   parameter identification applies it to its simulations as well, unless
-    #'   the PI task's own `simulationRunOptions` block sets the field.
+    #'   defaults (the `defaultSimulationRunOptions` JSON field), or `NULL`
+    #'   when none are declared. Recognized fields are the two run options
+    #'   `numberOfCores` and `showProgress`, plus any solver setting: `absTol`,
+    #'   `relTol`, `h0`, `hMin`, `hMax`, `mxStep`, `useJacobian`,
+    #'   `checkForNegativeValues`.
+    #'
+    #'   The run options become the `simulationRunOptions` [runScenarios()] and
```
PavelBal: Not sure this is informative rather then confusing.

R/project.R
```diff
@@ -176,19 +176,27 @@ Project <- R6::R6Class(
...
+    #'   [buildSimulations()] use when the caller passes none. The solver
+    #'   settings have no counterpart on [ospsuite::SimulationRunOptions] and
+    #'   are written to each simulation (`simulation$solver`) instead.
+    #'
+    #'   This block is the bottom of a precedence chain, never the last word.
```
PavelBal: Rewrite with "modeler-friendly" language.

Also it reads as a descriptio of the implementation, rather than user-facing "what is done".

R/scenario-execution.R
```diff
@@ -7,9 +7,9 @@
 # `simulationRunOptions` block, as a plain list parsed from JSON), or return
 # NULL when the record is empty (so the caller keeps the package defaults).
 # Only `numberOfCores` and `showProgress` are fields of `SimulationRunOptions`;
-# an unset one keeps its default. The record's `checkForNegativeValues` is a
-# solver setting with no counterpart on `SimulationRunOptions` (assigning one
-# errors) and is written to each simulation by `.applySolverSettings()`.
+# an unset one keeps its default. A record's remaining keys are solver settings
```
PavelBal: Don't document what fails "(assigning one errors)"

R/scenario-execution.R
```diff
@@ -26,41 +26,247 @@
...
+# each one takes. These are exactly the settable fields of
+# `ospsuite::SolverSettings`; every other key in a record (`numberOfCores`,
+# `showProgress`) belongs to `ospsuite::SimulationRunOptions` instead. The
+# `"count"` and `"flag"` tags exist because the .NET side coerces silently:
+# `mxStep <- 1.5` truncates to 1 and a negative `hMin` is accepted as given, so
+# the tag drives both the coercion here and the type rule
+# `.checkSolverSettings()` enforces.
+# @keywords internal
+# @noRd
+.solverSettingFields <- c(
```
PavelBal: Instead of hard-coding, get the supported fields directly from `ospsuite:::SolverSettings`

R/scenarios.R
```diff
@@ -79,6 +80,12 @@
 #'   referencing `parameterSets` definitions.
 #' @param initialConditions Character vector. Initial-condition set ids
 #'   referencing `initialConditions` definitions.
+#' @param solverSettings Named list or `NULL`. Solver settings for this
```
PavelBal: Give here links to the upstream ospsuite classes

R/scenarios.R
```diff
@@ -79,6 +80,12 @@
...
+#'   scenario's simulation, any of `absTol`, `relTol`, `h0`, `hMin`, `hMax`,
+#'   `mxStep`, `useJacobian` and `checkForNegativeValues`. A setting given here
+#'   overrides the project's `defaultSimulationRunOptions`; one left out keeps
```
PavelBal: "one left out keeps
#'   whatever the level below sets, and finally whatever the model file
#'   carries."

Rephrase, be clearer and simpler

R/scenarios.R
```diff
@@ -761,14 +786,16 @@ runScenarios <- function(
...
 #'   parameter layer.
-#' @param simulationRunOptions Optional [ospsuite::SimulationRunOptions].
-#'   Consulted only for a scenario with `simulateSteadyState` set (the
-#'   steady-state pre-solve still runs); it is not applied to the returned
-#'   simulations, since they are not run here. `NULL` (default) falls back
-#'   to the project's `defaultSimulationRunOptions`, including its solver
-#'   setting `checkForNegativeValues`, which is written to each returned
-#'   simulation; an explicit value replaces the project default entirely,
-#'   so the simulations keep the solver settings of their model files.
+#' @param simulationRunOptions Optional run options, either an
```
PavelBal: Duplicated documentation? Same in line 697

PavelBal (changes requested):
Before addressing my comments about the documentation:

- The use of "SimulationRunOptions" and "SolverSettings", and they are sometimes used  in wrong context. I suggest the following - it is `SolverSettings` that lives within the configurations. 
 `SimulationRunOptions` have only `numberOfCores` and `showProgress`, and should be passed as optional arguments to the `run...` functions, but not defined in the configurations.

So configurations only have `solverSettings` and `SimulationRunOptions` disappers from everywhere except for "run" functions arguments.

man/Scenario.Rd
```diff
@@ -20,7 +20,8 @@ Scenario(
   steadyStateTimeUnit = NULL,
   overwriteFormulasInSS = FALSE,
   modelParameterSets = NULL,
-  initialConditions = NULL
+  initialConditions = NULL,
+  solverSettings = NULL
```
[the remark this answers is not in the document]
Felixmil: @coderabbitai This policy does not exist.

R/parameter-identification.R
```diff
@@ -841,13 +841,15 @@ print.PITask <- function(x, ...) {
...
+  # the baseline and the task's own `simulationRunOptions` block overrides it
+  # field by field. The result is handed to `.prepareScenario()` as the record
+  # *below* each scenario's own `solverSettings` block, so a scenario's setting
+  # wins over the task's, and nothing sits above it because `runPI()` takes no
+  # run options of its own.
+  solverDefaults <- .mergeRunOptionRecords(
+    project$defaultSimulationRunOptions,
+    piTask$configuration$simulationRunOptions
```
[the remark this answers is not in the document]
Felixmil: @coderabbitai not in scope

R/scenario-execution.R
```diff
@@ -26,41 +26,177 @@
...
 .resolveRunOptions <- function(project, simulationRunOptions) {
-  if (!is.null(simulationRunOptions)) {
-    return(list(runOptions = simulationRunOptions, solverSettings = NULL))
-  }
   defaults <- project$defaultSimulationRunOptions
+  if (inherits(simulationRunOptions, "SimulationRunOptions")) {
+    return(list(
+      runOptions = simulationRunOptions,
+      solverDefaults = defaults,
+      solverOverrides = NULL
```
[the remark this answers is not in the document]
Felixmil: @coderabbitai The field is now defaultSolverSettings and the docs state the chain.

R/scenarios.R
```diff
@@ -489,6 +502,16 @@ print.Scenario <- function(x, ...) {
...
     }
+
+    # An unsound `solverSettings` block reaches here only from a hand-edited
+    # definition file; both authoring doors reject one. Report it now rather
+    # than letting the solver take a truncated or negative value in silence.
+    for (problem in .checkSolverSettings(sc$solverSettings)) {
+      result$addCriticalError(
+        "Validation",
+        paste0("Scenario '", name, "': ", problem)
+      )
```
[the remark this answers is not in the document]
Felixmil: @coderabbitai That line is inside .validateScenarios(), and R/messages.R:32-39 names the section validators as the documented exception, building validationResult messages with paste0().

R/parameter-identification.R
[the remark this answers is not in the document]
Felixmil: @coderabbitai yes

tests/testthat/test-validation.R
```diff
@@ -250,6 +250,24 @@ test_that(".validateScenarios flags Population scenario without populationId", {
...
+test_that(".validateScenarios flags an unsound solverSettings block", {
+  # Both authoring doors reject one, so this catches a hand-edited definition
+  # file: the solver would otherwise take a truncated value in silence.
+  sc <- Scenario(modelFile = "model.pkml")
+  sc$solverSettings <- list(reltol = 1e-6, mxStep = 1.5)
+  result <- .validateScenarios(list(s1 = sc))
+  msgs <- vapply(result$critical_errors, \(e) e$message, character(1))
+  expect_match(msgs, "unknown setting 'reltol'", all = FALSE)
+  expect_match(msgs, "mxStep must be a single whole number", all = FALSE)
+  expect_match(msgs, "Scenario 's1'", all = FALSE)
```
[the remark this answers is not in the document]
Felixmil: @coderabbitai The test now asserts the complete message vector in one expect_equal.

vignettes/how-to-migrate-from-excel.Rmd
```diff
@@ -107,6 +107,8 @@ The migration is faithful for the common case, but a few things deserve a manual
 
 * **Empty or absent sections are dropped, not preserved as empty.** A configuration workbook that did not exist, or a sheet with no rows, simply does not appear in the JSON project. If a section you expected is missing after import, check that its workbook and sheet actually held data.
 
+* **A scenario's `solverSettings` is JSON only.** The Excel scenario sheet has no solver-settings columns, so a scenario carrying a `solverSettings` block cannot be written to a workbook. `exportProjectToExcel()` warns and names the scenarios it dropped, and a project re-imported from those workbooks solves with its model files' own solver settings instead. `projectStatus()` also reports such a project as out of sync on the Excel axis for as long as the block is there, because the workbooks genuinely no longer describe it. Import is unaffected: it never produces a block.
```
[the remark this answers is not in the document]
Felixmil: @coderabbitai Reworded to say the scenario override is lost and lower-precedence settings apply

R/project.R
Felixmil: done, The field doc now links both pages

R/project.R
Felixmil: Dropped, the two kinds of setting are listed under their own class instead.

R/scenarios.R
Felixmil: buildSimulations() dropped its own `@param` solverSettings and inherits runScenarios() through the existing @inheritParams, so the two are literally one text

R/scenarios.R
Felixmil: Now: "A setting you leave out keeps the project's value, or the value stored in the model file where the project gives none."

R/scenarios.R
Felixmil: param solverSettings links SolverSettings in all four functions

R/scenario-execution.R
Felixmil: Removed

R/project.R
Felixmil: Rewritten

Felixmil:
Needs #1275 to be merged for CI to pass

PR #1272, opened 2026-09-10
Take species constants from the PK-Sim individual building block (v6)

### LINKED ISSUE #666: Make species scaling (applying different species from individuals file) more robust
- [x]  Include all species supported by PK-Sim

- Allow scaling between, not only from human?

PavelBal:
## Summary

Port of the species-scaling change to the JSON project model. For a non-human individual, `applyIndividualParameters()` now applies the species constants of the individual building block that `ospsuite::createIndividualBuildingBlock()` creates for the same characteristics, on top of the distributed and derived parameters of `ospsuite::createIndividual()`. `SpeciesParameters.xlsx` and `.getSpeciesParameters()` are removed. Closes #666.

The Aciclovir model of the example project and of both test projects is replaced by the PK-Sim 13 export from Open-Systems-Pharmacology/OSPSuite-R#2031. Its application path gained a `No formulation` segment; the example parameter set, the two Excel fixtures (`TestProjectExcel` and `TestProjectExcelLegacy`, whose test copies share that model) and the generator script of the legacy one, the sensitivity article, and the scenario-execution test use the new path.

## Why

PK-Sim's `createIndividual()` returns only parameters flagged `IsChangedByCreateIndividual`: for animals 2 distributed plus about 155 derived parameters. Species constants (tissue composition, lumen geometry, transit, pH, bile salt concentrations) are never returned, which is why the sheet existed. PK-Sim 13 added colon bile salt concentrations the sheet did not carry, and the sheet could not follow an individual's body weight.

## Evidence

Reference simulations for all 8 species built with PK-Sim 13.0.140 (4Comp and TwoPores, IV bolus) were compared parameter by parameter with a human model scaled by each approach. The sheet approach left 4 to 16 parameters per species different from the reference; the building-block approach leaves 24, all inert (lumen wall thickness and volume, whose only consumers, the mucosa volumes, are set by the derived parameters). Scaling between two animal species leaves 0 parameters different. Scaling from an animal to human stays unsupported.

## Changes

- `R/simulation.R`: `applyIndividualParameters()` applies the building-block constants for non-human species (formula entries and paths absent from the model skipped, so the build stays silent); the sheet block in `initializeSimulation()` and `.getSpeciesParameters()` are removed.
- `R/simulation.R`: `applyIndividualParameters()` stops with an error when a human individual is applied to a simulation built for another species (before: 106 of 117 human values were written into a rat model, 11 paths were missing, no error). The species is read from `simulation$configuration$individual$species` (PK-Sim exports since OSP version 12); for an older export or a MoBi model the presence of `Organism|Height` and `Organism|Age` decides, and a model without `Organism|Weight` is left alone. Catalog entry `humanIndividualForNonHumanModel()`.
- `inst/extdata/SpeciesParameters.xlsx` deleted.
- The species-scaling code matches `main` after #1273: `applyIndividualParameters()` builds its parameter set through `.individualParameterSet()` (distributed parameters for a human; distributed, derived, and species constants for any other species, each path once), the documentation drops the gut-wall paragraph and the sentence on skipped parameters, and the long lines are wrapped to the Air width. `writeIndividualToXLS()`, which shares that helper on `main`, does not exist on this branch (#908).
- Tests: `test-simulation.R` checks human to Rat and Mouse scaling against values of PK-Sim 13 individuals, weight dependence, an unchanged human model, scaling between two non-human species (the model scaled to a rat and then to a mouse carries the same organism parameters as the human model scaled to a mouse directly), and the refusal of a human individual on the mouse model of ospsuite's `test_snapshot.json`, with and without the stored individual; the species tests in `test-scenario-execution.R` now mock `.speciesParametersFromBuildingBlock()` and check the colon bile salt concentration.
- Fixtures: three `Aciclovir.pkml` copies, the `aciclovir_iv_250mg_default` parameter sets, `TestProjectExcel` (`Applications.xlsx`, `ProjectConfiguration.json`), `TestProjectExcelLegacy/Configurations/Applications.xlsx`, `data-raw/TestProjectExcelLegacy.R`.
- `REFACTORING-LOG.md` and `NEWS.md` entries.

## Verification

`devtools::test()` on the full suite with ospsuite 13.0.0.9005 (PK-Sim core 13.0.140), on this branch rebased onto `json-based-project` at 4cb538a8: 3613 passed, 0 failed, 0 errors, 3 skipped, 11 warnings. The warnings come from `test-create-plots.R`, `test-plots-utils.R`, and `test-sensitivity-time-profiles.R`, which this PR does not change. The third commit adds the species guard, the fourth the parity with `main`; `devtools::test(filter = "simulation|scenario-execution")` after the fourth: all passed, 0 failures (`test-scenario-execution.R` 156 expectations, `test-simulation.R` 34). CI runs the full suite.

The same change for `main` is #1273.

Not using the stacked-PR tooling: the `gh stack` extension is not installed in this environment, so this is a plain PR against `json-based-project`.


PavelBal:
Same as https://github.com/esqLABS/esqlabsR/pull/1273 but for v6 branch

R/simulation.R
```diff
@@ -18,29 +43,223 @@
...
+  individualCharacteristics,
+  simulation = NULL
+) {
+  gestationalAge <- .snapshotParameterValue(
+    individualCharacteristics$gestationalAge
+  )
+  # If no gestational age is supplied, use the default value.
+  if (is.null(gestationalAge)) {
+    gestationalAge <- 40
+  }
```
Felixmil, line 134: Could be kept NULL so we stay aligned with PKSim default value ?


```r
> bb <- ospsuite::createIndividualBuildingBlock(species = ospsuite::Species$Human,population = ospsuite::HumanPopulation$European_ICRP_2002)
> bb$gestationalAge
[1] "40.00 week(s)"


> bb <- ospsuite::createIndividualBuildingBlock(species = ospsuite::Species$Rat)
> bb$gestationalAge
NULL
```

Felixmil (commented):
Just one minor suggestion.

R/simulation.R
PavelBal: `ospsuite::createIndividualBuildingBlock(species = ospsuite::Species$Rat)` fails with `gestationalAge == NULL`... yes a weird interface

Felixmil: It doesn't on my side:

```r
> ospsuite::createIndividualBuildingBlock(species = ospsuite::Species$Rat)
<IndividualBuildingBlock>
  • Name: Rat
  • Type: Individual

── Origin data ──

  • Species: Rat
  • Population: Rat
  • Gender: Unknown
  • Weight: 0.23 kg
```

PavelBal: Yes but it's not `NULL` :)

PR #1273, opened 2026-09-10
Take species constants from the PK-Sim individual building block

PavelBal:
## Summary

Species scaling of a scenario individual no longer depends on the hand-maintained `SpeciesParameters.xlsx`. For a non-human individual, `applyIndividualParameters()` now applies the species constants of the individual building block that `ospsuite::createIndividualBuildingBlock()` creates for the same characteristics, on top of the distributed and derived parameters of `ospsuite::createIndividual()`. The sheet and its reader are removed. Issue #666.

The shipped example model `Aciclovir.pkml` is replaced by the PK-Sim 13 export from Open-Systems-Pharmacology/OSPSuite-R#2031. Its application path gained a `No formulation` segment; the example configuration files use the new path.

## Why

PK-Sim's `createIndividual()` returns only parameters flagged `IsChangedByCreateIndividual`. For animals that is 2 distributed plus about 155 derived parameters; species constants such as tissue composition, lumen geometry, transit, pH, and bile salt concentrations are never returned, which is why the sheet existed. Every PK-Sim release could add or change such constants; PK-Sim 13 added the colon bile salt concentrations, which the sheet did not carry.

## Evidence

Reference simulations were built for all 8 species with PK-Sim 13.0.140 (4Comp and TwoPores models, IV bolus) and compared parameter by parameter with a human model scaled by each approach.

| Approach | Parameters left different from the PK-Sim reference |
|---|---|
| `createIndividual()` + `SpeciesParameters.xlsx` (before) | 4 to 16 per species (colon bile salt, Dog stomach pH, downstream lumen formulas) |
| `createIndividual()` + building block (this PR) | 24, all inert (lumen wall thickness and volume; only consumers are mucosa volumes, which the derived parameters already set) |

Species to species scaling (Rat to Mouse, Dog to Beagle, Mouse to Rat, Monkey to Minipig) leaves 0 parameters different. Values follow the individual's body weight (checked at Rat 0.4 kg and Beagle 15 kg). Scaling from an animal to human stays unsupported: a human individual needs 426 age- and height-dependent parameters that an animal model does not have.

## Changes

- `applyIndividualParameters()`: species constants from the individual building block for non-human species; formula entries and paths absent from the model are skipped, so no "Could not find quantity" warnings.
- `initializeSimulation()`: the sheet block is removed.
- `applyIndividualParameters()` stops with an error when a human individual is applied to a simulation built for another species (before: 106 of 117 human values were written into a rat model, 11 paths were missing, no error). The species is read from `simulation$configuration$individual$species` (PK-Sim exports since OSP version 12); for an older export or a MoBi model the presence of `Organism|Height` and `Organism|Age` decides, and a model without `Organism|Weight` is left alone. Message `errorHumanIndividualForNonHumanModel()` in `R/messages.R`.
- `writeIndividualToXLS()` writes, for a non-human individual, the complete parameter set `applyIndividualParameters()` applies (distributed and derived parameters plus the species constants, 432 rows for a rat) instead of the two ontogeny factors it wrote before. Both functions build that set through one internal helper, `.individualParameterSet()`, so sheet and runtime cannot drift. The sheet has no model to filter against, so it keeps rows a given model may lack (`Default thickness of gut wall`, 17 rows for the Aciclovir model); the roxygen says to remove them or apply the sheet with `stopIfParameterNotFound = FALSE`.
- `inst/extdata/SpeciesParameters.xlsx` deleted.
- Tests: human to Rat and Mouse scaling checked against values of PK-Sim 13 individuals, weight dependence, an unchanged human model, scaling between two non-human species (the example model scaled to a rat and then to a mouse carries the same organism parameters as the human model scaled to a mouse directly), the refusal of a human individual on the mouse model of ospsuite's `test_snapshot.json`, with and without the stored individual, and the rat sheet of `writeIndividualToXLS()` (content, and a round trip through `readParametersFromXLS()` and `initializeSimulation()` that equals the direct scaling on every parameter the sheet sets).
- Example project: `Aciclovir.pkml` (PK-Sim 13), `Applications.xlsx`, `ProjectConfiguration.json`, the `utilities-config-json` snapshot, two tests, and the sensitivity article use the new application path. The function examples keep the old path because they load the Aciclovir model bundled with ospsuite, which is unchanged.

## Verification

`devtools::test()` on the full suite with ospsuite 13.0.0.9005 (PK-Sim core 13.0.140), run on the first two commits: 33 files, 1430 passed, 0 failed, 0 errors, 1 skipped, 46 warnings (all deprecation warnings of the tlf-based ospsuite plot functions, present before this change). The third commit adds one test and a PR number to a NEWS bullet; the fourth adds the species guard, the sixth the sheet extension (the fifth is an Air reformat). `devtools::test(filter = "utilities-individual|utilities-simulation|scenario")` after the sixth commit: all passed, 0 failures (`test-utilities-individual.R` now 40 expectations). CI runs the full suite.

The port of the same change to `json-based-project` is #1272.

## Summary by CodeRabbit

* **New Features**
  * Added species-specific individual scaling for human and supported non-human species, including conversions between non-human species.
  * Individual parameter exports to Excel now include the complete applied parameter set.
  * Added compatibility updates for PK-Sim 13 and improved support for species-specific simulations.

* **Bug Fixes**
  * Applying human characteristics to non-human simulations now reports a clear error.
  * Simulation initialization now applies selected individual parameters consistently, including non-human species values.

* **Documentation**
  * Documented supported species-scaling directions, limitations, Excel output, and updated PK-Sim 13 usage guidance.


inst/extdata/examples/TestProject/Models/Simulations/Aciclovir.pkml
PavelBal, line 1: Replaced with latest from ospsuite-r, reflecting v13 changes

inst/extdata/examples/TestProject/ProjectConfiguration.json
```diff
@@ -347,7 +347,7 @@
       "column_names": ["Container Path", "Parameter Name", "Value", "Units"],
       "rows": [
         {
-          "Container Path": "Events|IV 250mg 10min|Application_1|ProtocolSchemaItem",
+          "Container Path": "Events|IV 250mg 10min|No formulation|Application_1|ProtocolSchemaItem",
```
PavelBal, line 350: v13 change

inst/extdata/SpeciesParameters.xlsx
PavelBal, line 1: Not needed any more, all parameters are generated in code from the database

PavelBal:
@coderabbitai review

nautilus69 (approved):
Approved, all of these are additional concerns but NOT Blockers 

1- Performance: .speciesParametersFromBuildingBlock() calls createIndividualBuildingBlock() on every applyIndividualParameters()
    for a non-human species, on top of the existing createIndividual() — roughly doubling the per-apply PK-Sim individual-creation
    cost. Fine for one-off scaling; if it's ever called per-scenario in a large batch, consider caching the building block per
    (species, characteristics). Worth a note, not a blocker.
 2- Dedup precedence (evidence-validated): !duplicated(paths, fromLast = TRUE) makes the species-constant/derived value win over
    the distributed value for any shared path. The code asserts derived and building-block agree on shared paths, and the PR's
    parameter-by-parameter evidence (24 inert diffs; 0 for species↔species) validates the outcome — so this is deliberate and
    checked, but if createIndividual's distributed and derived ever diverged on a shared path, the sampled distributed value would be
    silently overridden. Low risk given the validation.
  3 Scalar assumption: .simulationSpecies/.storedIndividualSpecies use is.na(species) / species == Species$Human assuming
    length-1; a length(species) != 1 guard would harden against an unexpected multi-value individual$species. Minor.
  -I am not sure if this one can happen or not: Asymmetric guard: only the human-individual→non-human-model direction is guarded; a non-human individual on a
    mismatched/no-organism model still relies on stopIfNotFound = FALSE silently skipping (pre-existing lenient behavior, not a  regression). All supported directions (human→animal, animal↔animal) are correct.

PR #1275, opened 2026-09-15
Follow ospsuite 13.0.0.9010's re-exported example model and population

Felixmil:
`ospsuite` 13.0.0.9010 re-exported the two fixtures it ships in `extdata/`, and the test suite has been red on every branch since, including `main` at an unchanged commit. This restores it. Nothing in the package's behavior changes: the failures were tests and documentation examples naming a model path that no longer exists, plus two snapshots recording a dependency's example population.

## The dose path

PK-Sim 13 puts a `No formulation` container between an application and its protocol schema item, so the dose of the bundled Aciclovir model is now `Events|IV 250mg 10min|No formulation|Application_1|ProtocolSchemaItem|Dose`. Every sensitivity example and test loads that model through `system.file("extdata", "Aciclovir.pkml", package = "ospsuite")` and named the old path, which the .NET side rejected with `Could not find quantity with path ...` across 144 tests.

The 16 occurrences in the four `R/sensitivity-*.R` example blocks and the five `test-sensitivity-*.R` files now use the path the shipped model carries, read out of the model rather than assumed. The sensitivity vignette already used it.

## Snapshots

The longer label re-wraps a printed tibble in `sensitivity-calculation.md` and widens the legends of 19 plot snapshots. Every changed line in those SVGs is a `<text>` element, so no plotted value moved.

`populations.md` is a separate upstream change: the bundled `extdata/pop.csv` gained `Organism|BSA` and reordered its parameters. Those two tests snapshot all 150 parameters of one individual, so they record the dependency's demographics rather than esqlabsR's own behavior. The parameters they are actually about, the `Organism|Kidney|GFR` and `Organism|Kidney|eGFR` that `extendPopulationByUserDefinedParams()` and `extendPopulationFromXLS()` add, are still there.

## No NEWS entry

Nothing changes for a user of the package: no function behaves differently and no argument or field is added. The examples were broken only for readers who had already updated `ospsuite`. Say so if you would rather have a bullet.

## Scope

`REFACTORING.md` fences off sensitivity analysis, with an exception for updating the wiring when something it depends on changes. This is that exception, and it is limited to the path string, the regenerated `man/*.Rd`, and the snapshots.


R/sensitivity-calculation.R
```diff
@@ -57,7 +57,7 @@
 #' outputPaths <- "Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)"
 #' parameterPaths <- c(
 #'   "Aciclovir|Lipophilicity",
-#'   "Events|IV 250mg 10min|Application_1|ProtocolSchemaItem|Dose",
+#'   "Events|IV 250mg 10min|No formulation|Application_1|ProtocolSchemaItem|Dose",
```
Felixmil, line 60: This path now pins the package to the dev ospsuite, but `DESCRIPTION` still says `ospsuite (>= 13.0.0)`.

With the released 13.0.0 installed the declared dependency is satisfied, yet `system.file("extdata", "Aciclovir.pkml", package = "ospsuite")` hands back the PK-Sim 12 export with no `No formulation` container — so `example(sensitivityCalculation)`, and `R CMD check` which runs examples, aborts with `Could not find quantity with path Events|IV 250mg 10min|No formulation|Application_1|ProtocolSchemaItem|Dose`: the mirror image of the failure this PR fixes. Same for the five `test-sensitivity-*.R` files.

Worth bumping `Depends:` to `ospsuite (>= 13.0.0.9010)` so the requirement is declared rather than implicit in a path string.

---

Felixmil (commented):
Reviewed the whole diff. The path substitution is mechanical and complete — no occurrence of the old `Events|IV 250mg 10min|Application_1|...` string survives anywhere in the repo, and the bundled `TestProject` / `TestProjectExcel` / `Example` pkmls plus `aciclovir_iv_250mg_default.json` already carry the `No formulation` container, so nothing on the project-fixture side is left inconsistent.

Snapshots check out: in `sensitivity-calculation.md` every numeric cell is byte-identical, only tibble column wrapping moved; in the 19 SVGs every changed line is a `<text>` element (verified by filtering the diff), so no geometry moved. In `populations.md` both blocks stay internally consistent (119 paths / 119 values, 120 / 120) and `Organism|Kidney|GFR` and `eGFR` — the parameters the two tests are actually about — are still present.

One remark inline, about the dependency bound.

---

PavelBal (approved):

Issue #1279, opened 2026-09-15
`Scenario$print`


PavelBal:
Currently:

```
> project$definitions$scenarios$aciclovir_iv_female
<Scenario>
  • Name: aciclovir_iv_female
  • Model: Aciclovir.pkml
  • Type: Individual
  • Individual: adult_female
  • Population: <empty string>
  • Protocol: aciclovir_iv_250mg
  • Parameter Sets: global, aciclovir
  • Initial Conditions: <empty string>
  • Output Paths: 1
  • Steady State: FALSE
```

- Suggest to print the `Output Paths` - only the paths, not the names
- Rename `Protocol` to `Administration Protocol`.

PR #1284, opened 2026-09-16
feat(scenarios): show which output paths a scenario uses when printing it

Felixmil:
Printing a scenario reported how many output paths it referenced, not which ones, so a modeler checking a scenario before running it could not see whether the outputs were the ones they meant without going back to the definition files. The same print also labeled the application protocol just `Protocol`, which does not match what the field is called anywhere else. This makes `print.Scenario()` list the paths themselves and use the full label.

## Scenario print

- The `Output Paths` count leaves the scalar item list. An `Output Paths:` bullet takes its place at the end of the field list, with one bullet per referenced path nested one level under it, showing the literal path and never the definition id it was looked up under. One entry per line is the shape `print.ParameterSet()` and `print.InitialConditionSet()` already use, chosen because a full PBPK path wraps mid-path when several are joined onto one labeled line.
- `Protocol` becomes `Administration Protocol`. Display only: the record field stays `applicationProtocol` and the JSON key stays `application`, so nothing about the definition files or the round trip changes.

## Reviewer notes

Two cases the change had to decide, since a count hid both:

- **A reference with no matching definition.** Referential integrity is deliberately lazy, so `.parseScenarios()` keeps a dangling id as a name with an `NA` value and lets the cross-reference validator report it later. Printing only the value would render a bare `NA` and quietly hide the broken reference, so the bullet names the id instead: `kidney_urine (unknown output path)`.
- **A scenario with no output paths.** The block is still printed and shows one `<empty vector>` bullet, so an empty list is distinguishable from a section that was left out.

The nesting is why the block is not another `ospPrintItems()` call. That helper's `title` renders as a flush-left heading, which would put the block outside the field list it belongs to; `.printScenarioOutputPaths()` builds the two levels from the same `cli` pieces `ospPrintItems()` uses instead.

`print.Project()` keeps its own `Output Paths` line as a count. That is the number of output-path definitions in the project, a different quantity from one scenario's references.

## Example

Before:

``` r
devtools::load_all("/tmp/esqlabsr-base", quiet = TRUE)
scenario <- exampleProject()$definitions$scenarios[["aciclovir_iv"]]
print(scenario)
#> <Scenario>
#>   • Name: aciclovir_iv
#>   • Model: Aciclovir.pkml
#>   • Type: Individual
#>   • Individual: adult_male
#>   • Population: <empty string>
#>   • Protocol: aciclovir_iv_250mg
#>   • Parameter Sets: global, aciclovir
#>   • Initial Conditions: testinitialset
#>   • Output Paths: 1
#>   • Steady State: FALSE
#>   • Solver Settings: <empty string>
```

After:

``` r
devtools::load_all("/Users/felix/.openducktor/worktrees/esqlabsr/esqlabsr-7f6a", quiet = TRUE)
scenario <- exampleProject()$definitions$scenarios[["aciclovir_iv"]]
print(scenario)
#> <Scenario>
#>   • Name: aciclovir_iv
#>   • Model: Aciclovir.pkml
#>   • Type: Individual
#>   • Individual: adult_male
#>   • Population: <empty string>
#>   • Administration Protocol: aciclovir_iv_250mg
#>   • Parameter Sets: global, aciclovir
#>   • Initial Conditions: testinitialset
#>   • Steady State: FALSE
#>   • Solver Settings: <empty string>
#>   • Output Paths:
#>     • Organism|PeripheralVenousBlood|Aciclovir|Plasma (Peripheral Venous Blood)
```

Closes #1279

## Summary by CodeRabbit

- **Improvements**
  - Scenario details now identify the application protocol as **“Administration Protocol.”**
  - Output paths are displayed individually by name instead of only showing a count.
  - Scenarios without configured output paths now clearly show an empty output-path section.
  - Unresolved output paths are identified as **“unknown output path”** with their corresponding ID.

- **Documentation**
  - Updated release notes and scenario-printing documentation to reflect the revised display format.


NEWS.md
```diff
@@ -74,8 +74,8 @@ Deprecation intensity follows the lifecycle policy: forwarding shims that still
 
 ## Minor improvements and bug fixes
 
-- Printing a scenario now lists its output paths, one per line, instead of reporting how many there are (#1279).
-- Printing a scenario labels the application protocol field `Administration Protocol` instead of `Protocol` (#1279).
+- A scenario's `print()` now lists the output paths the scenario uses, one full path per line, instead of reporting how many there are. An output path with no matching definition in the project is shown by the id it was asked for under, followed by `(unknown output path)` (#1279).
```
PavelBal: This version was never released - so no "does now instead of".

NEWS.md
```diff
@@ -74,6 +74,8 @@ Deprecation intensity follows the lifecycle policy: forwarding shims that still
 
 ## Minor improvements and bug fixes
 
+- A scenario's `print()` now lists the output paths the scenario uses, one full path per line, instead of reporting how many there are. An output path with no matching definition in the project is shown by the id it was asked for under, followed by `(unknown output path)` (#1279).
+- A scenario's `print()` labels the application protocol field `Administration Protocol` instead of `Protocol` (#1279).
```
PavelBal: Same here - and I think these items must not be mentioned in NEWS.

PavelBal (changes requested):

